PR #242, opened 2014-12-24
[TASK] Update jshintrc

clark800:
@wltsmrz Thanks for polishing up the jshintrc file. There are just two edits that I'd like to add: removing the `boss` and `W004` relaxing options. I also moved `laxbreak` to the end, that is one relaxing option that may have some readability benefits. And I removed `camelcase` since it has no effect.

I can't think of any cases where `boss` (assignments where comparisons are expected) or `W004` (redefining parameters) should be allowed, but if there is a rare case where someone really wants to use it, they can disable the jshint warning with the `// jshint ignore:line` comment directive. But I'd imagine that 99% of the time when these warnings fire it will be catching bugs for us.

### DIFF
--- a/jshintrc
+++ b/jshintrc
@@ -2,7 +2,7 @@
   "browser": true,
   "jquery": true,
   "node": true,
-  "camelcase": false,
+
   "eqeqeq": true,
   "eqnull": true,
   "latedef": true,
@@ -13,10 +13,8 @@
   "curly": true,
   "freeze": true,
   "nonbsp": true,
-  "boss": true,
-  "laxbreak": true,
   "maxparams": 4,
   "maxdepth": 3,
-  "esnext": true,
-  "-W004": true
+
+  "laxbreak": true
 }


geertweening:
I agree with @clark800 that removing both `boss` and `W004` would make the jshintrc file a bit better. The goal is to have the hinting as a tool that makes it easier to catch bad practices or coding styles we don't prefer.

The (case)[https://github.com/ripple/ripple-lib/blob/develop/src/js/ripple/transaction.js#L503] @wltsmrz pointed out doesn't look like great coding practice to me. I'm sure you can use it responsibly but it's hard to read and easy to misunderstand. Allowing assignments where a boolean evaluation is expected should not be a best practice.

For `W004`, I'm not sure if it's more readable to keep `var`. Effectively redefining a var or assigning a previously assigned var doesn't change anything logically, but you might accidentally override a var in your scope you didn't intend to. I expect that to be the reason for jshint to have this option.

Right now we do manual inspection of a pull request to enforce stylistic changes and avoid gotchas. I would like to make the jshint checking part of the test process so it will actually fail the test. That's essentially the same thing as blocking a pull request by a comment for a stylistic change, but now it's automated. It does not replace code review, it's part of the automated tests and makes code reviews a little easier. A certain set of stylistic preferences and gotchas shouldn't come up in code reviews anymore since the hinting will have already caught that.

wltsmrz:
> The goal is to have the hinting as a tool that makes it easier to catch bad practices or coding styles we don't prefer.

Agreed

> The (case)[https://github.com/ripple/ripple-lib/blob/develop/src/js/ripple/transaction.js#L503] @wltsmrz pointed out doesn't look like great coding practice to me. I'm sure you can use it responsibly but it's hard to read and easy to misunderstand. Allowing assignments where a boolean evaluation is expected should not be a best practice.

It's easy to read and hard to misunderstand, to me. It's unlikely that anyone is going to use this without knowing what they're doing. Making jshint more strict in this regard won't catch any bugs

> I would like to make the jshint checking part of the test process so it will actually fail the test. That's essentially the same thing as blocking a pull request by a comment for a stylistic change, but now it's automated.

Bad idea, IMO. You'll still have to review code for style, because jshint is quite lacking. I definitely don't agree with failing tests for jshint.

geertweening:
> It's easy to read and hard to misunderstand, to me. It's unlikely that anyone is going to use this without knowing what they're doing. Making jshint more strict in this regard won't catch any bugs

```
 if ((result = cache[hash])) {
    break;
 }
```
If I look at this case, I'm not sure if the assignment of `result` returns a boolean or if the test is just for a truthy value. In this case the assignment returns a transaction, so it's testing for presence. That's not clear from the snippet, you would have to look elsewhere in the code to see what's going on exactly. IMO this would be more readable and clear:

```
result = cache[hash];
if (result !== void(0)) {
...
```
Or, if you only care about truthy values
```
result = cache[hash];
if (result) {
...
```

I don't really like relying on implicit returns, specially when assigning and testing the result in a statement like this. I would rather be a little more verbose for the sake of clarity. 

I would prefer to take out the `boss` options so we don't write this code, but I care most about agreeing what is a better way to write a case like this. I personally feel the implicit return and assignment in an if case is not the best way.

wltsmrz:
Have it your way. Anyway, whether you have `boss` enabled or not, do you realize it wouldn't catch this assignment within the conditional because it's wrapped in parentheses?

Also @clark800 let's remove `esnext` while we're playing with the jshintrc

clark800:
Yeah, I can see why jshint behaves that way, but it would be nice if it did generate a warning for that case. I submitted an issue, so hopefully they will change that: https://github.com/jshint/jshint/issues/2081. But it would be nice to remove those extra parens in the meantime.

As for esnext, we should leave that in because we are planning on moving to Facebook Flow, which does es6 to es5 compilation, which means that we will be able to start using es6 language features, like the "let" keyword for lexical scoping. That will solve the remaining few cases where we are redeclaring variable names in an elegant way.

wltsmrz:
> I submitted an issue, so hopefully they will change that: jshint/jshint#2081. 

This is not an accidental oversight. It's intentional, and it's the reason I used parentheses in the first place http://jshint.com/docs/options/#boss. Maybe the jshint people will have a change of heart

> As for esnext, we should leave that in because we are planning on moving to Facebook Flow, which does es6 to es5 compilation

I haven't been party to any such plans, but you can remove `esnext` until we make a decision on Facebook Flow. It was added to our javascript style guide repo to accommodate use of `const` in ripple-rest, but we don't use it [const] in ripple-lib

geertweening:
@clark800 can we remove `esnext` for now so we can merge this PR. We can decide later if we want to have `esnext` on for Flow

clark800:
Ok, esnext has been removed. We might not need it either way if we transpile to es5 before running jshint, which might make sense since we have to strip the type annotations before running jshint anyways.

geertweening:
LGTM

PR #245, opened 2015-01-06
Wrap account_tx binary parsing in async.setImmediate

wltsmrz:


src/js/ripple/remote.js
```diff
@@ -1424,10 +1425,19 @@ Remote.prototype.requestAccountTx = function(options, callback) {
...
-    if (options.parseBinary) {
-      res.transactions = res.transactions.map(Remote.parseBinaryTransaction);
+    if (!options.parseBinary) {
+      request.emit('transactions', res);
+      return;
     }
-    request.emit('transactions', res);
+
+    async.mapSeries(res.transactions, function(transaction, next) {
+      async.setImmediate(function() {
```
geertweening: Would be better to replace with `async.nextTick`. Internally `async.setImmediate` is assigned to `async.nextTick`, but the API docs only show `nextTick` as an available method (https://github.com/caolan/async#nextTick)

wltsmrz: Ah. I didn't read the docs, I read [the code](https://github.com/caolan/async/blob/master/lib/async.js#L86-L113). Internally async seems to [prefer](https://github.com/caolan/async/blob/master/lib/async.js#L553) setImmediate. I also prefer to use setImmediate. I'll see if I can get it documented on the async readme so we can use it instead. Leave this pull request open until then

geertweening: cool

wltsmrz: https://github.com/caolan/async/pull/692

beaugunderson: (async maintainer here); just merged that PR. :)

wltsmrz: Thanks!

PR #247, opened 2015-01-07
Decrease redundancy in binary account_tx parsing

wltsmrz:


geertweening:
looks good to me, but can you add a testcase for the `parseBinaryTransaction` method?

wltsmrz:
Yep

geertweening:
LGTM

PR #248, opened 2015-01-08
[TASK] Add flow tasks to Gulpfile

clark800:


geertweening:
LGTM

Gulpfile.js
```diff
@@ -183,6 +194,25 @@ gulp.task('watch', function() {
   gulp.watch('src/js/ripple/*', [ 'build-debug' ]);
 });
 
+// To use this, each javascript file must have /* @flow */ on the first line
+gulp.task('typecheck', function() {
```
lumberj: What's the workflow with Flow type-checking? Maybe the README can be updated?

clark800: The full flow workflow isn't integrated yet, but the transitional workflow is:
1. Add /* @flow */ to the top of a file you are working on
2. Run "gulp typecheck" and fix all the warnings/errors for that file
Once we do this for all the files, we can disable "weak mode" and do a second round of error fixing. Once that is done, we can start adding type annotations and start using the "gulp strip" task.

geertweening: I agree with Alan that we should have the flow you just described documented somewhere. A doc in the docs folder makes sense, maybe a BUILD.md or something similar

clark800: BUILD.md added

geertweening:
LGTM

PR #249, opened 2015-01-09
Request broadcast/filter

wltsmrz:
* Changes the behavior of `Request.broadcast()` (which may not have been documented at all)

@shekenahglory 
@miguelportilla


miguelportilla:
LGTM +1

wltsmrz:
Fixed some things that were noted

PR #255, opened 2015-01-15
[TASK] Refactor to use bignumber.js

clark800:


sublimator:
Some experiments to play with and/or build off:
Using sjcl.bn for the uint classes, and some custom base58 code from Java, stolen originally from BitcoinJ.  The performance is still a bit off, but it's much better than with using bignumber.js for everything:  https://github.com/sublimator/ripple-lib/commit/bb795d30bd71d87e70c72a409bb28d0eb4c4b65d

sublimator:
I tried using this as the ripple-lib version for the rippled integration tests, and all is good except for amount.from_human("1XRP"). Note that there is no space. 

re: serializing bignumbers to hex, (which I guess we've always done?), maybe we want to check that the field is a uint64 type, because strings get parsed as rippled parses uint32 json strings as base 10.

clark800:
> I tried using this as the ripple-lib version for the rippled integration tests, and all is good except for 
> amount.from_human("1XRP"). Note that there is no space.

I will add this to the ripple-lib unit tests and look at this now.

> re: serializing bignumbers to hex, (which I guess we've always done?), maybe we want to check that > > the field is a uint64 type, because rippled parses uint32 json strings as base 10.

Probably a good idea, but please add this as an issue for later; we should try to avoid including behavioral changes in a refactoring commit.

src/js/ripple/amount.js
```diff
@@ -1,53 +1,55 @@
...
+  bi_10:             new BigNumber('10'),
+  bi_1e14:           new BigNumber(String(1e14)),
+  bi_1e16:           new BigNumber(String(1e16)),
+  bi_1e17:           new BigNumber(String(1e17)),
+  bi_1e32:           new BigNumber('100000000000000000000000000000000'),
+  bi_man_max_value:  new BigNumber('9999999999999999'),
+  bi_man_min_value:  new BigNumber('1000000000000000'),
+  bi_xns_max:        new BigNumber('9000000000000000000'), // Json wire limit.
+  bi_xns_min:        new BigNumber('-9000000000000000000'),// Json wire limit.
+  bi_xrp_max:        new BigNumber('9000000000000'),  // Json wire limit.
```
sublimator: I think this is actually wrong, and we want this value, and corresponding xrp values for the xns: 
https://github.com/ripple/rippled/blob/9156633bafd3f121ca7fb6f4bf823797265fcc81/src/ripple/protocol/STAmount.h#L61

src/js/ripple/amount.js
```diff
@@ -110,170 +120,47 @@ Amount.prototype.abs = function() {
...
-    result._currency    = this._currency;
-    result._issuer      = this._issuer;
-
-    result.canonicalize();
-  }
-
-  return result;
+Amount.prototype.multiply = function(multiplicand) {
+  var multiplicandAmount = Amount.from_json(multiplicand);
+  // TODO: probably should just multiply by multiplicandAmount._value
```
sublimator: `divide` seems to go to some troubles to check validity of the operands, yet this doesn't ?

src/js/ripple/amount.js
```diff
@@ -798,29 +576,41 @@ Amount.prototype.parse_quality = function(quality, counterCurrency, counterIssue
...
+  var mantissa      = new BigNumber(mantissa_hex, 16);
+  var offset        = parseInt(offset_hex, 16) - 100;
+
+  var value = new BigNumber(mantissa.toString() + 'e' + offset.toString());
+
   this._currency    = Currency.from_json(counterCurrency);
   this._issuer      = UInt160.from_json(counterIssuer);
   this._is_native   = this._currency.is_native();
 
+  var exponent = 0;
```
sublimator: There's no tests for this function, and at a glance this change seems odd.

sublimator: regarding changing the exponent by +/- 1 when is_native/inverse

sublimator: Oh, there's some tests for from_quality it seems

src/js/ripple/amount.js
```diff
@@ -798,29 +576,41 @@ Amount.prototype.parse_quality = function(quality, counterCurrency, counterIssue
...
+  if (this._is_native) {
+    if (opts.inverse) {
+      exponent += 1;
+    } else {
+      exponent -= 1;
```
sublimator: Suggest changing name of exponent to `power` (even though it's used with toPower()) to make it clearer at a glance.

sublimator: Suggest either: 

```
if (opts.inverse) {
  value = value.toPower(-1);
}

// quality is `takerPays / takerGets`
// counterCurrency equates to takerPays
// We don't want to pay a million times too much
// because the quality is in terms of drops.
if (!opts.xrp_as_drops) {
  if (this._is_native) {
    value = value.div(Amount.bi_xns_unit);
  }
  else if ( baseCurrency.is_valid() && baseCurrency.is_native()) {
    value = value.times(Amount.bi_xns_unit);   
  }  
};

this._set_value(value);
```

or 

```  
var adjusted = opts.inverse ? value.toPower(-1) : value;

var native_scale = opts.xrp_as_drops ? adjusted :
                          // quality is `takerPays / takerGets`
                          // counterCurrency equates to takerPays
                          // We don't want to pay a million times too much
                           this._is_native ? adjusted.div(Amount.bi_xns_unit) :
                           baseCurrency.is_valid() && baseCurrency.is_native() ? 
                          // The quality is how much you must `pay` to get ONE of `gets`
                          // If takerGets was native then we need to pay a million 
                          // times more for each (one) unit. 
                           adjusted.times(Amount.bi_xns_unit) :
                           adjusted;

this._set_value(native_scale);
```

src/js/ripple/serializedobject.js
```diff
@@ -209,8 +216,8 @@ SerializedObject.jsonify_structure = function(structure, field_name) {
 
       if (typeof structure.to_json === 'function') {
         output = structure.to_json();
-      } else if (structure instanceof BigInteger) {
-        output = ('0000000000000000' + structure.toString(16).toUpperCase()).slice(-16);
+      } else if (structure instanceof sjcl.bn) {
+        output = ('0000000000000000' + normalize_sjcl_bn_hex(structure.toString()).toUpperCase()).slice(-16);
```
sublimator: The 16 leading zeroes is clearly for ensuring Uint64 types pad correctly.  With .slice(-16) you couldn't get any larger either.  It would be nice to tighten this up at some point, but as you said, in another patch.

src/js/ripple/serializedtypes.js
```diff
@@ -21,59 +22,52 @@ var amount    = require('./amount');
...
-
-function isCurrencyString(val) {
-  return isString(val) && /^[A-Z0-9]{3}$/.test(val);
-};
+}
 
-function isBigInteger(val) {
-  return val instanceof BigInteger;
-};
+function isBigNumber(val) {
```
sublimator: Given BigNumber everywhere else means bignumber.js perhaps an alternative name?

sublimator:
I'm on my new laptop currently, and have no dumps

src/js/ripple/uint.js
```diff
@@ -107,17 +105,16 @@ UInt.prototype.copyTo = function(d) {
...
 
 UInt.prototype.is_valid = function() {
-  return this._value instanceof BigInteger;
+  return this._value instanceof sjcl.bn;
 };
 
 UInt.prototype.is_zero = function() {
-  return this._value.equals(BigInteger.ZERO);
+  var val = this._value;
+  return val.limbs.length == 1 && val.limbs[0] == 0;
```
sublimator: I never checked if there were other legal limbs values for zero

src/js/ripple/serializedtypes.js
```diff
@@ -354,10 +335,13 @@ var STAmount = exports.Amount = new SerializedType({
...
 
         // Next eight bits: offset/exponent
-        hi |= ((97 + amount._offset) & 0xff) << 22;
+        hi |= ((97 + offset) & 0xff) << 22;
+
         // Remaining 54 bits: mantissa
-        hi |= amount._value.shiftRight(32).intValue() & 0x3fffff;
-        lo = amount._value.intValue() & 0xffffffff;
+        var mantissaDecimal = utils.getMantissaDecimalString(value.abs());
+        var mantissaHex = (new BigNumber(mantissaDecimal)).toString(16);
```
sublimator: Suppose could use sjcl instead for the conversion of the mantissa to hex ?

src/js/ripple/serializedtypes.js
```diff
@@ -354,10 +335,13 @@ var STAmount = exports.Amount = new SerializedType({
...
+        hi |= parseInt(mantissaHex.slice(0, -8), 16) & 0x3fffff;
```
sublimator: maybe we should assert the hex is <= 16 characters

src/js/ripple/amount.js
clark800: Can you clarify what is wrong and what we want? This is a new field corresponding to bi_xns_max, but divided by 1000000 so that it is expressed in XRP rather than drops.

src/js/ripple/amount.js
clark800: Yeah, no throws on the left side.

src/js/ripple/amount.js
clark800: I agree this should be refactored, but this is already a huge pull request and the goal is to introduce bignumber.js; we need not refactor all of amount.js in one pass. We want to make it easy to verify that the logic is the same.

src/js/ripple/serializedtypes.js
clark800: Tried that, but it doesn't work because sjcl doesn't parse decimal (or at least I couldn't find anything about it in the docs).

src/js/ripple/amount.js
```diff
@@ -1,53 +1,55 @@
...
-//
+var assert    = require('assert');
+var BigNumber = require('bignumber.js');
+var extend    = require('extend');
+var utils     = require('./utils');
+var UInt160   = require('./uint160').UInt160;
+var Seed      = require('./seed').Seed;
+var Currency  = require('./currency').Currency;
+
+utils.configureBigNumber(BigNumber);
```
geertweening: I don't like this structure. If I understand correctly, anyone requiring the `bignumber.js` package can override the configuration. I'd rather have a separate class that wraps the `bignumer.js` package, configures it and exposes a singleton for usage across the other source files.

src/js/ripple/amount.js
```diff
@@ -659,96 +462,72 @@ Amount.prototype.invert = function() {
...
-  var matches = String(j).match(Amount.human_RE_hex);
-  if (matches && matches.length === 5 && matches[4]) {
-    integer  = matches[2];
-    fraction = matches[3] || '';
-    currency = matches[4];
-    this._is_negative = Boolean(matches[1]);
-  }
+  var hex_RE = /^[a-fA-F0-9]{40}$/;
+  var currency_RE = /^([a-zA-Z]{3}|[0-9]{3})$/;
+  function isNum(x) { return !isNaN(x); }
```
geertweening: I would prefer `isNumber`. Also, `Infinite` is a valid number, which we don't want here. `isNumber` in serializedtypes does a similar thing.

src/js/ripple/amount.js
clark800: Even if we do that, it won't really solve the problem you described because another file can still require('bignumber.js') without using the wrapper and override the configuration. The only way I can think of to make it safer is to make a new npm package that wraps bignumber.js so that bignumber.js is not a direct dependency of ripple-lib and can't be directly required.

src/js/ripple/serializedtypes.js
sublimator: Oh, interesting, so it can only parse strings in base 16?

src/js/ripple/amount.js
sublimator: The comment says `json wire limit`, yet the number doesn't correspond with rippled's network limits. See:

https://github.com/ripple/rippled/blob/9156633bafd3f121ca7fb6f4bf823797265fcc81/src/ripple/app/transactors/CreateOffer.cpp#L124
https://github.com/ripple/rippled/blob/9156633bafd3f121ca7fb6f4bf823797265fcc81/src/ripple/protocol/STAmount.h#L327
https://github.com/ripple/rippled/blob/9156633bafd3f121ca7fb6f4bf823797265fcc81/src/ripple/protocol/STAmount.h#L64

src/js/ripple/amount.js
sublimator: It's precisely why I suggested a change. The logic was hard to verify.  Even more so than the code it was replacing.

src/js/ripple/serializedtypes.js
clark800: I think that's right. The documentation is really lacking though.

src/js/ripple/bignumber.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+extend(BigNumberWrapper, BigNumber);    // copy class static properties
+BigNumberWrapper.prototype = BigNumber.prototype;
+
+BigNumberWrapper.config = function() {
+  throw new Error('BigNumber.config may only be called from bignumber.js');
+};
+
+BigNumberWrapper.setRoundingMode = function(roundingMode) {
```
sublimator: Does this return the existing rounding mode, for the caller to restore?

src/js/ripple/bignumber.js
```diff
@@ -0,0 +1,28 @@
+var BigNumber = require('bignumber.js');
+var extend = require('extend');
+
+function BigNumberWrapper(value, base) {
+  // reset config every time a BigNumber is instantiated so that
```
sublimator: If we're going to go to some troubles, maybe we should go the full monty. This reconfigures before every construction, but not operations. Is this going to be enough ? ( haven't looked myself )

src/js/ripple/bignumber.js
sublimator: Maybe an api that executes a code block (function) with a certain rounding mode, then `finally` restores the existing one would be nice?

src/js/ripple/amount.js
clark800: Currently it is as similar as possible to the original code. The code you wrote isn't equivalent. The first block of setting the power is to correct for the fact that XRP amounts are now being stored internally as XRP rather then drops. So the bi_xns_unit has to be raised to the power of -2,-1,0,1, or 2. You could do it in two toPower calls, but let's save that decision for the next refactor.

src/js/ripple/bignumber.js
clark800: I think that would be overkill. The main risk is that somebody adds a new file to the project that tries to configure BigNumber.js differently at require-time.

sublimator: Makes sense, this all seems a bit naff. Shame bignumber doesn't have a factory to create isolated contexts.

src/js/ripple/amount.js
sublimator: I don't really get it. What case will the power be -2 or 2 ?
We don't have any tests for the cases where the power is -2, 2 it seems.

sublimator: Note that it's illegal to have two native amounts contributing to a quality. It will never happen in the really real world.

src/js/ripple/amount.js
```diff
@@ -453,56 +322,37 @@ Amount.prototype.invert = function() {
...
-      }
-    }
 
-  } else if (this.is_zero()) {
-    this._offset      = Amount.cMinOffset;
-    this._is_negative = false;
+Amount.prototype.canonicalize = function(roundingMode) {
+  if (this._is_native) {
+    var value = this._value.round(6, BigNumber.ROUND_DOWN);
+    this._value = value.absoluteValue().greaterThan(Amount.bi_xrp_max)
```
sublimator: While historically it seems to_text() would return '0' for native amounts that overflow, internally it sets NaN as the value_. Elsewhere we blow up when limits are exceeded. There's no comments here explaining this strange behaviour or what a future course of action to fix it might be. What are the plans for fixing up amount.js beyond using bignumber.js (which by itself doesn't really buy much)? 

In any case, this seems to be a behavioural change.

clark800: Yeah, let's try to keep the scope of the discussion on this pull request limited to things that apply to this pull request so that it's easier to keep track of what needs to be addressed. This feedback is valuable though and it would be good to put it in the issue tracker.

src/js/ripple/amount.js
clark800: > it's illegal to have two native amounts contributing to a quality

Ok good, we can add an assert and test case for this in a future pull request (behavioral change), which will enable us to use that assumption to simplify the code as you suggest.

src/js/ripple/bignumber.js
clark800: Yeah. Issue submitted: https://github.com/MikeMcl/bignumber.js/issues/50

src/js/ripple/amount.js
sublimator: Who would trade 1 XRP for 5 XRP?

I'm not opposed to changing the behaviour of this code. In fact, I believe
it should be changed, to throw errors more liberally and consistently. It
seems senseless squandering efforts on attempting to faithfully reproduce
old quirks ( which this change set can't achieve a perfect score for
anyway, due to rounding changes ) when it could be put into defining
sensible behaviour.

The discussion is very much in scope. This a huge bunch of changes
requested to be merged, including some amounts now overflowing to zero.
This is a bad change, made worse in the context of avoiding making good
ones.

If there was concerns about the size/scope of the request, it could broken
into logical commits like the rippled guys sometimes do. Anyway, bad
changes need to be excised from this request before merging.

src/js/ripple/amount.js
rec: I think what sublimator is saying is that this change introduces _new_ behavior which is both incorrect and very dangerous - numerical overflow is handled by setting the value to 0 instead of either setting it to NaN or throwing an exception.

Correctness is of supreme importance - this is financial software, and people will be putting serious money through this - so it's _always_ a relevant issue worth bringing up if you see code that you even suspect is incorrect.

clark800: > this change introduces new behavior

Thanks for clarifying this. I interpreted sublimator's "this seems to be a behavioural change" to mean that the change he was proposing. I moved the native amount overflow checking to to_text.

src/js/ripple/amount.js
clark800: > I'm not opposed to changing the behaviour of this code.

I think this blog post does a decent job of explaining the principle that I'm trying to obey here:
http://jakegoulding.com/blog/2012/11/04/refactor-and-make-changes-in-different-commits/

With regards to overflowing to zero, I misinterpreted your original comment; it should be fixed now.

sublimator:
>>> I thought you meant for all overflows.

Hrmm, yeah, the context was lost somewhere, and I was vague, but concerns should not be so readily dismissed. It's a problem with native amounts only right? It seems that _check_limits will throw when an IOU value is out of bounds.

>>> I thought it was ready to be merged

Might be worth taking stock and going back and making the tests more detailed.

I just looked at the existing limit tests and was sad to see this test (existing prior to this patch):
```
   it ('from_json less than minimum XRP', function() {
      var amt = Amount.from_json('-9000000000000000001');
      assert.strictEqual(amt.to_json(), '0');
    });
```
(This was added in the same commit that introduced errors thrown for IOUs exceeding limits it seems)

You can add `assert(isNaN(amt._value));` and see for yourself what happens here. (to_json() must call to_text() which has an `allow_nan` argument)

What a can of worms huh?

clark800:
Yeah, I wasn't trying to dismiss anything, but it didn't make sense since the context got lost. This is part of the reason why I'm suggesting that we try to keep the scope of discussion on pull requests as tight as possible: fewer topics -> less ideas to juggle. Any bugs or suggestions about newly introduced code are welcome on the pull request, but cleaning up pre-existing problems (like the messy overflow handling) is outside the scope. But I totally agree with all of your suggestions and made a JIRA ticket so that we get them done soon (probably right after this pull request is merged): https://ripplelabs.atlassian.net/browse/RLJS-207 Also, feel free to add more comments on the JIRA ticket or create more tickets.

sublimator:
>>> (probably right after this pull request is merged):

Sorry was in transit before, and glossed over this part.

>>> What's the plan for addressing

Ok, so you have my vote for merging this as is, provided the enhancements come ASAP. From my perspective it will save having to do two comb overs, and time wasted fixing up things like the `json wire limit` comments I mentioned prior.

PR #258, opened 2015-01-15
Throw on tls error

wltsmrz:
https://github.com/ripple/ripple-lib/issues/68

### DIFF
--- a/src/js/ripple/server.js
+++ b/src/js/ripple/server.js
@@ -34,7 +34,7 @@ function Server(remote, opts) {
     throw new TypeError('Server configuration is not an Object');
   }
 
-  if (!Server.domainRE.test(opts.host)) {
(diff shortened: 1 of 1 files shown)


src/js/ripple/server.js
```diff
@@ -434,6 +449,11 @@ Server.prototype.connect = function() {
         log.info(self.getServerID(), 'onerror:',  e.data || e);
       }
 
+      if (Server.TLS_ERRORS.indexOf(e.message) !== -1) {
```
clark800: _.contains 

would have prevented the bug in the previous version of this pull request and it's more semantic  (easier to read)

wltsmrz: Maybe later (as in, not in this pull request)

geertweening:
LGTM

PR #260, opened 2015-01-16
double check 'tes' code before emitting 'success'

ripplerm:
result of a received transaction could be in 'tes' or 'tec'.


wltsmrz:
Nice catch, looks good

PR #264, opened 2015-01-21
[BUG] `Request.prototype.ledgerSelect` can accept ledger_index as String

lumberj:
This behavior is consistent with `ripple-lib` before https://github.com/ripple/ripple-lib/commit/8af5f9c2#diff-a8d289f91765fc0f634609ad7c15c239L227


wltsmrz:
We can do this, but we should also validate input in ripple-rest

wltsmrz:
LGTM

PR #265, opened 2015-01-22
[FIX] include request account transactions' date for binary flag

boxbag:
query for ledger and use closed date as account transactions' date


wltsmrz:
I don't agree that ripple-lib should do this automatically

lumberj:
Here's an example of it being done manually in rest, I think related, https://github.com/ripple/ripple-rest/blob/develop/api/transactions.js#L327

PR #268, opened 2015-01-27
Remote updates

wltsmrz:


src/js/ripple/remote.js
```diff
@@ -9,117 +10,64 @@
...
+ *
+ * @param {Object} Connection options.
+ */
+
+function Remote(opts) {
   EventEmitter.call(this);
 
   var self = this;
-  var opts = opts || { };
+  var opts = lodash.extend(this, Remote.DEFAULTS, opts);
```
clark800: This is going to set the options directly on the Remote object. I think we want

    this._opts = lodash.extend({}, Remote.DEFAULTS, opts);

src/js/ripple/remote.js
```diff
@@ -148,155 +97,133 @@ function Remote(opts, trace) {
...
-  // This is used to remove Node EventEmitter warnings
-  var maxListeners = opts.maxListeners || opts.max_listeners || 0;
+  this.setMaxListeners(this.max_listeners);
 
-  this._servers.concat(this).forEach(function(emitter) {
-    if (emitter instanceof EventEmitter) {
-      emitter.setMaxListeners(maxListeners);
-    }
+  this.servers.forEach(function(server) {
+    var connection = self.addServer(server);
```
clark800: addServer returns a server so how about
    
    var server = self.addServer(serverOptions)

src/js/ripple/remote.js
```diff
@@ -148,155 +97,133 @@ function Remote(opts, trace) {
...
-
-  function listenerRemoved(type, listener) {
-    if (type === 'transaction_all') {
-      self._transaction_subs -= 1;
-      if (!self._transaction_subs && self.isConnected()) {
-        self.request_unsubscribe('transactions').request();
+  function listenersModified(action, event) {
+    // Automatically subscribe and unsubscribe to orderbook
+    // on the basis of existing event listeners
+    if (~Remote.TRANSACTION_EVENTS.indexOf(event)) {
```
clark800: You're an assembly coder aren't you? :smile:

This isn't readable. I didn't even expect that the binary representation of -1 is all ones because all numbers are treated as floats in javascript.

I'd suggest _.includes

src/js/ripple/remote.js
```diff
@@ -148,155 +97,133 @@ function Remote(opts, trace) {
...
-    }
-  }
+  this.on('removeListener', function(event) {
+    listenersModified('remove', event);
+  });
 };
 
 util.inherits(Remote, EventEmitter);
 
+Remote.DEFAULTS = {
```
clark800: Can we make this private e.g. a local var in the constructor?

src/js/ripple/remote.js
```diff
@@ -1615,10 +1542,17 @@ Remote.parseBinaryLedgerData = function(ledgerData) {
 Remote.prototype.requestTransactionHistory =
 Remote.prototype.requestTxHistory = function(start, callback) {
   // XXX Does this require the server to be trusted?
-  //utils.assert(this.trusted);
+  // utils.assert(this.trusted);
 
   var request = new Request(this, 'tx_history');
 
+  if (typeof start === 'object') {
+    start = start.start;
```
clark800: This line looks unnecessary

wltsmrz: It isn't unnecessary

src/js/ripple/remote.js
wltsmrz: I have a feeling we'll want access to the default configuration for testing or some other purpose at some point. I'd rather dangle this from the constructor

src/js/ripple/remote.js
wltsmrz: Yeah, I think we'll switch to some lodash method at some point. I'm not in love with ~

src/js/ripple/remote.js
wltsmrz: You're suggesting to redefine `server`? Didn't you claim recently that it was unreadable to do so or against Doug Crockford scripture or somethin?. I will probably change to `server`

src/js/ripple/remote.js
wltsmrz: Nope, we want to extend `this`. Classes such as `TransactionManager` access configuration from their `remote` reference. This is the way it currently works. We could add an options container object like you suggest, and I'd be weakly in favor of it

src/js/ripple/remote.js
clark800: I see what's going on now. I think the code should be oriented toward the non-deprecated usage i.e. 

    function(opts, callback)

In the else block

    opts = {start: opts}

and

    request.message.start = opts.start;

src/js/ripple/remote.js
clark800: It should be kept private until it is known that it needs to be public. It is much easier to switch from private to public than the other way around.

src/js/ripple/remote.js
clark800: Why "at some point"?

src/js/ripple/remote.js
clark800: No, that's true it would be unreadable, this is what I meant:

    this._opts.serverOptionsArray.forEach(function(serverOptions) {
      var server = self.addServer(serverOptions);

src/js/ripple/remote.js
clark800: So what happens if I pass an opts object with keys that overlap with the class method names?

> Classes such as TransactionManager access configuration from their remote reference.

I haven't looked at this part of the code yet, but this sounds really bad.

wltsmrz: Good point, I'll add a check for that

src/js/ripple/remote.js
wltsmrz: Because it doesn't really achieve anything. I might go ahead and add it now since I'll be in the neighborhood

src/js/ripple/remote.js
wltsmrz: Since it doesn't cost us anything to dangle from the constructor, I disagree

src/js/ripple/remote.js
wltsmrz: The parameter could be renamed, yeah

src/js/ripple/remote.js
clark800: As soon as you expose something publicly, there is suddenly a risk that somebody somewhere is depending on it, and there is no way to know for sure that nobody is using it. Therefore it does cost us something, which is either the robustness of the library or the ability to refactor it.

wltsmrz: Just because it's attached to the constructor doesn't make it a *public API* that requires extraordinary care. It'd be so-attached for internal use (at least for now). Although, I wouldn't be opposed to documenting it, with some stability guarantee or warning.

wltsmrz: > Therefore it does cost us something, which is either the robustness of the library or the ability to refactor it.

This may cost someone else something, but it's unlikely to cost anyone anything. People are free to misuse anything. That said, I have no intention of removing it [Remote.DEFAULTS] any time soon and don't see why we would feel compelled to. If we did, we follow proper procedure, make sure it's in the release notes, etc. etc.. It's not exactly the unavoidable one-way street you're imagining.

src/js/ripple/remote.js
```diff
@@ -9,117 +10,72 @@
...
- *      'ledger_closed'   : A good indicate of ready to serve.
- *      'transaction'     : Transactions we receive based on current subscriptions.
- *      'transaction_all' : Listening triggers a subscribe to all transactions
- *                          globally in the network.
- *
- *    @param opts      Connection options.
- *    @param trace
- */
-
-function Remote(opts, trace) {
```
kureus: Why has the inline documentation been removed in this revision?

wltsmrz: This isn't the place for such documentation. Large documentation-comment blocks like this are often unmaintained (as this one was) after the original author stops work. Notice that the `server` configuration description has been out of order for nine months https://github.com/ripple/ripple-lib/commit/41ea820ae0f35384a377a620cbe83759e5500658. Proper documentation is the answer

kureus: Sure, I can understand that. Having said that there are many still knocking around ripple-lib, currency.js, account.js for example, and I have found them to provide helpful clues even working with the maxim, never fully trust a code comment. As an engineer working on top of ripple-lib rather than developing it directly I don't have an up to date knowledge of the libraries nuances and future roadmap and quite honestly have so far found the documentation somewhat lacking. Digging into the ripple-lib source and seeing how ripple-client uses ripple-lib have been the most useful insights. I would personally prefer that these comments were only removed when the documentation was clearly reproduced elsewhere. Can you point me to where this removed comment/explanation is or will be found?

wltsmrz: Most of this comment block is documented [here](https://github.com/ripple/ripple-lib/blob/develop/docs/REFERENCE.md#remote-options). I might be convinced to add the old comment block back, but it's very unlikely that someone else's code comments are going to be maintained, as we've seen. IMO it adds more clutter than utility, but I share your concern for documentation.

Issue #100, opened 2014-06-12
Amount.from_json(fractional)  (Native Amounts)


sublimator:
Amount.from_json("1") != Amount.from_json("1.0")

The presence of a `.` means it's parsed as XRP. It's surely a seductive convenience, but it's nasty.

Does rippled accept "1.0" as a native amount string in json? Looks like no: [1]
(https://github.com/ripple/rippled/blob/develop/src/ripple/module/data/protocol/STParsedJSON.cpp#L408) [2](https://github.com/ripple/rippled/blob/develop/src/ripple/module/data/protocol/STAmount.cpp#L115) [3](https://github.com/ripple/rippled/blob/develop/src/ripple/module/data/protocol/STAmount.cpp#L138-L147) [4](https://github.com/ripple/rippled/blob/develop/src/ripple/module/data/protocol/STAmount.cpp#L201) 
Maybe ripple-lib shouldn't either?

Fix steps:
1. deprecation
2. exception

PR #269, opened 2015-01-28
Cleanup amount.js

clark800:
Behavioral changes:

remove allow_nan option (option to convert NaN to 0)
all overflow checking is done in canonicalize
to_text always returns a string ('NaN' instead of NaN)
from_quality disallows XRP/XRP quality

### DIFF
--- a/src/js/ripple/amount.js
+++ b/src/js/ripple/amount.js
@@ -9,6 +9,19 @@ var Seed      = require('./seed').Seed;
 var Currency  = require('./currency').Currency;
 var BigNumber = require('./bignumber');
 
+
+function isInteger(number) {
+  return parseInt(number) === number;
+}
+
+function ensureDecimalPoint(value) {
+  return isInteger(value) ? String(value) + '.0' : value;
+}
+
+function inverse(number) {
+  return (new BigNumber(number)).toPower(-1);
+}
+
 function Amount() {
   // Json format:
   //  integer : XRP
@@ -26,20 +39,19 @@ var consts = {
   xns_precision:     6,
 
   // bi_ prefix refers to "big integer"
-  bi_5:              new BigNumber('5'),
-  bi_7:              new BigNumber('7'),
-  bi_10:             new BigNumber('10'),
-  bi_1e14:           new BigNumber(String(1e14)),
-  bi_1e16:           new BigNumber(String(1e16)),
-  bi_1e17:           new BigNumber(String(1e17)),
-  bi_1e32:           new BigNumber('100000000000000000000000000000000'),
+  // TODO: we shouldn't expose our BigNumber library publicly
+  bi_5:              new BigNumber(5),
+  bi_7:              new BigNumber(7),
+  bi_10:             new BigNumber(10),
+  bi_1e14:           new BigNumber(1e14),
+  bi_1e16:           new BigNumber(1e16),
+  bi_1e17:           new BigNumber(1e17),
+  bi_1e32:           new BigNumber(1e32),
   bi_man_max_value:  new BigNumber('9999999999999999'),
-  bi_man_min_value:  new BigNumber('1000000000000000'),
-  bi_xns_max:        new BigNumber('9000000000000000000'), // Json wire limit.
-  bi_xns_min:        new BigNumber('-9000000000000000000'),// Json wire limit.
-  bi_xrp_max:        new BigNumber('9000000000000'),
-  bi_xrp_min:        new BigNumber('-9000000000000'),
-  bi_xns_unit:       new BigNumber('1000000'),
+  bi_man_min_value:  new BigNumber(1e15),
+  bi_xns_max:        new BigNumber(1e17),
+  bi_xns_min:        new BigNumber(-1e17),
+  bi_xns_unit:       new BigNumber(1e6),
 
   cMinOffset:        -96,
   cMaxOffset:        80,
@@ -51,18 +63,22 @@ var consts = {
   min_value:         '-1000000000000000e-96'
 };
 
+var MAX_XRP_VALUE = new BigNumber(1e11);
+var MAX_IOU_VALUE = new BigNumber(consts.max_value);
+var MIN_IOU_VALUE = (new BigNumber(consts.min_value)).abs();
+
 // Add constants to Amount class
 extend(Amount, consts);
 
 // DEPRECATED: Use Amount instead, e.g. Amount.currency_xns
 exports.consts = consts;
 
-// Given '100/USD/mtgox' return the a string with mtgox remapped.
+// Given '100/USD/ISSUER' return the a string with ISSUER remapped.
 Amount.text_full_rewrite = function(j) {
   return Amount.from_json(j).to_text_full();
 };
 
-// Given '100/USD/mtgox' return the json.
+// Given '100/USD/ISSUER' return the json.
 Amount.json_rewrite = function(j) {
   return Amount.from_json(j).to_json();
 };
@@ -176,13 +192,7 @@ Amount.prototype.ratio_human = function(denominator, opts) {
   opts = extend({ }, opts);
 
   var numerator = this;
-
-  if (typeof denominator === 'number' && parseInt(denominator, 10) === denominator) {
-    // Special handling of integer arguments
-    denominator = Amount.from_json(String(denominator) + '.0');
-  } else {
-    denominator = Amount.from_json(denominator);
-  }
+  denominator = Amount.from_json(ensureDecimalPoint(denominator));
 
   // If either operand is NaN, the result is NaN.
   if (!numerator.is_valid() || !denominator.is_valid()) {
@@ -242,12 +252,7 @@ Amount.prototype.ratio_human = function(denominator, opts) {
 Amount.prototype.product_human = function(factor, opts) {
   opts = opts || {};
 
-  if (typeof factor === 'number' && parseInt(factor, 10) === factor) {
-    // Special handling of integer arguments
-    factor = Amount.from_json(String(factor) + '.0');
-  } else {
-    factor = Amount.from_json(factor);
-  }
+  factor = Amount.from_json(ensureDecimalPoint(factor));
 
   // If either operand is NaN, the result is NaN.
   if (!this.is_valid() || !factor.is_valid()) {
@@ -281,7 +286,7 @@ Amount.prototype.product_human = function(factor, opts) {
  * @private
  */
 Amount.prototype._invert = function() {
-  this._set_value((new BigNumber(1)).dividedBy(this._value));
+  this._set_value(inverse(this._value));
   return this;
 };
 
@@ -342,13 +347,17 @@ Amount.prototype._check_limits = function() {
   if (this._value.isNaN() || this._value.isZero()) {
     return this;
   }
-  if (!this._is_native) {
-    var absval = this._value.absoluteValue();
-    if (absval.lessThan((new BigNumber(Amount.min_value)).absoluteValue())) {
-      throw new Error('Exceeding min value of ' + Amount.min_value);
+  var absval = this._value.absoluteValue();
+  if (this._is_native) {
+    if (absval.greaterThan(MAX_XRP_VALUE)) {
+      throw new Error('Exceeding max value of ' + MAX_XRP_VALUE.toString());
     }
-    if (absval.greaterThan(new BigNumber(Amount.max_value))) {
-      throw new Error('Exceeding max value of ' + Amount.max_value);
+  } else {
+    if (absval.lessThan(MIN_IOU_VALUE)) {
+      throw new Error('Exceeding min value of ' + MIN_IOU_VALUE.toString());
+    }
+    if (absval.greaterThan(MAX_IOU_VALUE)) {
+      throw new Error('Exceeding max value of ' + MAX_IOU_VALUE.toString());
     }
   }
   return this;
@@ -387,7 +396,7 @@ Amount.prototype.currency = function() {
 };
 
 Amount.prototype.equals = function(d, ignore_issuer) {
-  if (typeof d === 'string') {
+  if (!(d instanceof Amount)) {
     return this.equals(Amount.from_json(d));
   }
 
@@ -506,19 +515,7 @@ Amount.prototype.parse_human = function(j, opts) {
   // Apply interest/demurrage
   if (opts.reference_date && this._currency.has_interest()) {
     var interest = this._currency.get_interest_at(opts.reference_date);
-
-    // XXX Because the Amount parsing routines don't support some of the things
-    //     that JavaScript can output when casting a float to a string, the
-    //     following call sometimes does not produce a valid Amount.
-    //
-    //     The correct way to solve this is probably to switch to a proper
-    //     BigDecimal for our internal representation and then use that across
-    //     the board instead of instantiating these dummy Amount objects.
-    var interestTempAmount = Amount.from_json(''+interest+'/1/1');
-
-    if (interestTempAmount.is_valid()) {
-      this._set_value(this.divide(interestTempAmount)._value);
-    }
+    this._set_value(this._value.dividedBy(interest.toString()));
   }
 
   return this;
@@ -578,40 +575,48 @@ Amount.prototype.parse_quality = function(quality, counterCurrency, counterIssue
   this._issuer      = UInt160.from_json(counterIssuer);
   this._is_native   = this._currency.is_native();
 
-  var power = 0;
-  if (this._is_native) {
-    if (opts.inverse) {
-      power += 1;
-    } else {
-      power -= 1;
+  if (this._is_native && baseCurrency.is_native()) {
+    throw new Error('XRP/XRP quality is not allowed');
+  }
+
+  /*
+  The quality, as stored in the last 64 bits of a directory index, is stored as
+  the quotient of TakerPays/TakerGets.
+
+  When `opts.inverse` is true we are looking at a quality used for determining a
+  `bid` price and it must first be inverted, before our declared base/counter
+  currencies are in line with the price.
+
+  For example:
+
+    quality as stored :  5 USD          /  3000000 drops
+    inverted          :  3000000 drops  /          5 USD
+  */
+  var adjusted = opts.inverse ? inverse(value) : value;
+  var nativeAdjusted = adjusted;
+
+  if (!opts.xrp_as_drops) {
+    // `In a currency exchange, the exchange rate is quoted as the units of the
+    //  counter currency in terms of a single unit of a base currency`. A
+    //  quality is how much taker must `pay` to get ONE `gets` unit thus:
+    //    pays ~= counterCurrency
+    //    gets ~= baseCurrency.
+    if (this._is_native) {
+      // pay:$price              drops  get:1 X
+      // pay:($price / 1,000,000)  XRP  get:1 X
+      nativeAdjusted = adjusted.div(Amount.bi_xns_unit);
+    } else if (baseCurrency.is_valid() && baseCurrency.is_native()) {
+      // pay:$price X                   get:1 drop
+      // pay:($price * 1,000,000) X     get:1 XRP
+      nativeAdjusted = adjusted.times(Amount.bi_xns_unit);
     }
   }
 
-  // Correct offset if xrp_as_drops option is not set and base currency is XRP
-  if (!opts.xrp_as_drops &&
-      baseCurrency.is_valid() &&
-      baseCurrency.is_native()) {
-    if (opts.inverse) {
-      power -= 1;
-    } else {
-      power += 1;
-    }
-  }
-
-  var one = new BigNumber(1);
-  var adjusted = value.times(Amount.bi_xns_unit.toPower(power));
-  var newValue = opts.inverse ? one.dividedBy(adjusted) : adjusted;
-  this._set_value(newValue);
+  this._set_value(nativeAdjusted);
 
   if (opts.reference_date && baseCurrency.is_valid() && baseCurrency.has_interest()) {
     var interest = baseCurrency.get_interest_at(opts.reference_date);
-
-    // XXX If we had better math utilities, we wouldn't need this hack.
-    var interestTempAmount = Amount.from_json(''+interest+'/1/1');
-
-    if (interestTempAmount.is_valid()) {
-      this._set_value(this.divide(interestTempAmount)._value);
-    }
+    this._set_value(this._value.dividedBy(interest.toString()));
   }
 
   return this;
@@ -690,10 +695,6 @@ Amount.prototype.parse_native = function(j) {
     } else {
       this._set_value(value.dividedBy(Amount.bi_xns_unit));
     }
-    // TODO: move this overflow check to canonicalize
-    if (this._value.abs().greaterThan(Amount.bi_xrp_max)) {
-      this._set_value(new BigNumber(NaN));
-    }
   } else {
     this._set_value(new BigNumber(NaN));
   }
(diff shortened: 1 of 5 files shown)


sublimator:
Most of this LGTM, but I'd like to see this taken further really, along with some new
documentation/tests.

It's weird that we throw errors for IOU amounts that are out of bounds, while
setting NaN internally for native amounts. I'd rather see errors thrown
consistently.

The `json wire limits` map to the `network limits` in rippled, which are limits
on values as serialised. I think we DO want to honour these limits for any
amount that is serialised, it's just that the constant copied from rippled was
the wrong one.

If you look at the uses of the source in the rippled code, you'll see the
comment is potentially misleading:

```
    // Maximum native value supported by the code
    static const std::uint64_t cMaxNative  = 9000000000000000000ull;
```

And that we actually want this one:

```
    // Max native value on network.
    static const std::uint64_t cMaxNativeN = 100000000000000000ull;
```

Consider this:

```
  inline
  bool
  isLegalNet (STAmount const& value)
  {
      return ! value.native() || (value.mantissa() <= STAmount::cMaxNativeN);
  }
```

and

```

src/ripple/app/transactors/CreateOffer.cpp:
  476          STAmount saTakerGets = mTxn.getFieldAmount (sfTakerGets);
  477  
  478:         if (!isLegalNet (saTakerPays) || !isLegalNet (saTakerGets))
  479              return temBAD_AMOUNT;
  480  

src/ripple/app/transactors/Payment.cpp:
   82              " saDstAmount=" << saDstAmount.getFullText ();
   83  
   84:         if (!isLegalNet (saDstAmount) || !isLegalNet (maxSourceAmount))
   85              return temBAD_AMOUNT;
   86  

src/ripple/app/transactors/SetTrust.cpp:
   78          std::uint32_t uQualityOut (bQualityOut ? mTxn.getFieldU32 (sfQualityOut) : 0);
   79  
   80:         if (!isLegalNet (saLimitAmount))
   81              return temBAD_AMOUNT;
   82  

src/ripple/app/transactors/Transactor.cpp:
  106      STAmount saPaid = mTxn.getTransactionFee ();
  107  
  108:     if (!isLegalNet (saPaid))
  109          return temBAD_AMOUNT;
  110  

src/ripple/protocol/STAmount.h:
  327  inline
  328  bool
  329: isLegalNet (STAmount const& value)
  330  {
  331      return ! value.native() || (value.mantissa() <= STAmount::cMaxNativeN);

```

In any case, we need to make some kind of distinction between intermediate
amounts and final amounts, which will serialised. Do we want to always enforce
those limits as invariants in the constructor? Or in the serialisation code (to
json or bytes).

(While it does seem to do the right thing and throw, we don't seem to have  any
tests for binary serialisation code when an amount _value is NaN or
BigNumber(NaN))

Because amounts aren't arbitrary precision and have a floating point, there's
problems with operations including `insignificant` operands.

What's ripple-lib do for `12345678901234500000000000000  + 1234` ?

Rounding up there doesn't make sense. And silently nooping doesn't either.

```
it ('has no sense of the significance', function() {
  var amt = Amount.from_json('12344567890123450000000/USD/1');
  var amt2 = Amount.from_json('4999999/USD/1'); 
  var amt3 = amt.add(amt2);
  assert.strictEqual(amt.to_text(), amt3.to_text());
});
```

In summary:

0. Use correct native network limit
1. Tests for binary serialisation of invalid amounts
2. Consider uses of intermediate vs final amounts
3. Consider where invariants should be enforced
    1. Is there value in invalid Amount objects?
4. Consider whether to throw when an operation becomes a noop due 
   to an insignificant operand
6. Document/rationalise/justify behaviours, especially regarding default rounding modes

I'm sort of sold on encouraging the use of arbitrary arithmetic for intermediate results, and enforcing limit invariants in the constructors, forcing callers to be explicit about rounding/truncation etc, but I'm not sure if there's a realistic migration path. I haven't really thought it through, but I'd happily read someone else's thoughts.

sublimator:
Do you want me to raise issues as comments here? On comments on RLJS-207, or as new JIRA issues?

src/js/ripple/amount.js
```diff
@@ -574,40 +566,27 @@ Amount.prototype.parse_quality = function(quality, counterCurrency, counterIssue
...
+    throw new Error('XRP/XRP quality is not allowed');
   }
 
-  // Correct offset if xrp_as_drops option is not set and base currency is XRP
-  if (!opts.xrp_as_drops &&
-      baseCurrency.is_valid() &&
+  var adjusted;
+  var inversion = opts.inverse ? -1 : 1;
+  // Correct for XRP to drops factors
+  if (this._is_native) {
```
sublimator: Shouldn't xrp_as_drops be honoured regardless of whether it's the base or counter currency ?

src/js/ripple/amount.js
```diff
@@ -176,13 +189,7 @@ Amount.prototype.ratio_human = function(denominator, opts) {
...
 
   var numerator = this;
-
-  if (typeof denominator === 'number' && parseInt(denominator, 10) === denominator) {
-    // Special handling of integer arguments
-    denominator = Amount.from_json(String(denominator) + '.0');
-  } else {
-    denominator = Amount.from_json(denominator);
-  }
+  denominator = Amount.from_json(noIntegers(denominator));
```
sublimator: The original tacking on of `.0` seems ill considered. 

Perhaps an IOU style amount should be created for these cases where `noIntegers` is used, otherwise the number is parsed as a native amount (which seem to be truncated to 6 decimal places (iirc))

src/js/ripple/amount.js
```diff
@@ -325,7 +327,9 @@ Amount.prototype.invert = function() {
 
 Amount.prototype.canonicalize = function(roundingMode) {
   if (this._is_native) {
-    this._value = this._value.round(6, BigNumber.ROUND_DOWN);
+    var value = this._value.round(6, BigNumber.ROUND_DOWN);
+    var isOverflow = value.abs().greaterThan(Amount.bi_xrp_max);
+    this._value = isOverflow ? new BigNumber(NaN) : value;
```
sublimator: Why not blow up for native amounts in check limits, as with the iou amounts?

src/js/ripple/amount.js
sublimator: I mean, it seems like from_json isn't desired, so it's not really wanting to interpret it as a native amount. But then later, I see code like equals() which will just use from_json. Hrmmm. Not really consistent semantics.

I can't recall off hand what from_json does for Number values.

sublimator: No cleanup of amount.js would be complete without at least considering this:

```

>>> Amount.from_json(1.0).to_text()
'1'

>>> Amount.from_json(1.1).to_text()
'1100000'

>>> Amount.from_json(1).to_text()
'1'

>>> Amount.from_json('1.0').to_text()
'1000000'

>>> Amount.from_json('1.1').to_text()
'1100000'

>>> Amount.from_json('1').to_text()
'1'

```

test/amount-test.js
```diff
@@ -330,8 +330,8 @@ describe('Amount', function() {
   });
   describe('Amount parsing', function() {
     it('Parse invalid string', function() {
-      assert.strictEqual(Amount.from_json('x').to_text(), '0');
-      assert.strictEqual(typeof Amount.from_json('x').to_text(true), 'number');
+      assert.strictEqual(Amount.from_json('x').to_text(), 'NaN');
+      assert.strictEqual(typeof Amount.from_json('x').to_text(true), 'string');
       assert(isNaN(Amount.from_json('x').to_text(true)));
```
sublimator: Could remove the now ignored argument for to_text

clark800: Will do

src/js/ripple/amount.js
clark800: Yeah, this should be fixed. I think it deserves it's own ticket and pull request because it will probably require a lot of changes to the tests: https://ripplelabs.atlassian.net/browse/RLJS-215

We need to be really careful with this though, I'm not familiar with all the things this change could break.

clark800: > Perhaps an IOU style amount should be created for these cases where noIntegers is used, otherwise the number is parsed as a native amount

I think the code is trying to allow the denominator parameter to be a native amount, but forcing it to be interpreted as XRP rather than drops, as it would be without the integer check. This can also be fixed by https://ripplelabs.atlassian.net/browse/RLJS-215

src/js/ripple/amount.js
clark800: I agree, these should be consistent. The question is whether it is better to throw or use NaN. This problem applies throughout the Amount class; some methods throw, others return NaN. Throwing is nice because the client code finds out about the issue right away.

The counter-argument might be as follows. Let's say ripple-trade adds two amounts using the Amount class to display the sum on the page and they forget to check for exceptions. If we use NaN, then presumably the box on the page would show "NaN", whereas if we throw, it could break the whole page. Arguably, breaking the whole page is ok because technically it is a bug that the client is not checking for errors and it should be fixed immediately.

src/js/ripple/amount.js
clark800: That would make sense, but I'm not really sure what is going on here. The original code only had a check for the base currency. The tests work either way.

clark800:
There are a ton of issues with amount.js and I think we should take it one step at a time. So I think we should do new JIRA tasks for any "big" changes. I'd define "big" as something that would require more than a few minutes to change. I made a couple JIRA tasks:
https://ripplelabs.atlassian.net/browse/RLJS-215
https://ripplelabs.atlassian.net/browse/RLJS-217

clark800:
> Use correct native network limit

Why is the network limit preferable to the code limit? For example if you go over the network limit with an intermediate value, but are back under before serializing, then there shouldn't be an error. So I would think we want to leave it as is.

> Tests for binary serialisation of invalid amounts
> Is there value in invalid Amount objects?

Not that I can think of. I think it might be best to always throw. https://ripplelabs.atlassian.net/browse/RLJS-217
The tests won't be necessary if amounts are always valid. 

> Consider uses of intermediate vs final amounts
> I'm sort of sold on encouraging the use of arbitrary arithmetic for intermediate results

I think the idea is to emulate the behavior of rippled's arithmetic, which is not arbitrary precision if I understand correctly. So if that's the goal, then we can't use arbitrary precision for intermediate results.

> Consider where invariants should be enforced

I think amount.js is doing this correctly already, but it looks like we might need to add a check to serializedtype.js. Should we throw in STAmount.serialize if the value is greater than cMaxNativeN = 100000000000000000ull?

> Consider whether to throw when an operation becomes a noop due to an insignificant operand

This would break compatibility with rippled, so probably not what we want.

> Document/rationalise/justify behaviours, especially regarding default rounding modes

Can you be more specific?

src/js/ripple/amount.js
sublimator: Yep, strong trees grow in windy areas.

sublimator:
> The tests won't be necessary if amounts are always valid.

Maybe, but tests are useful for many reasons: documentation, regression proofing, declaring intent twice (like a password confirm), yada, yada

> This (throwing) would break compatibility with rippled,

I would argue rippled is broken in this respect, and if max compat weren't a concern, I'd most definitely throw.

It gets me to wondering why we don't replicate all the rounding quirks that were mentioned at one point if compatibility is a goal? How much compat, to what end?

> Should we throw in STAmount.serialize if the value is greater than cMaxNativeN = 100000000000000000ull?

I would say yes. No good could ever come from such an amount.

Also, see rippled's canonicalize:

```
src/ripple/protocol/impl/STAmount.cpp:
  784          }
  785  
  786:         if (mValue > cMaxNativeN)   <-----  network limit
  787              throw std::runtime_error ("Native currency amount out of range");
  788  
```

>>> Document/rationalise/justify behaviours, especially regarding default rounding modes

Pro-actively, systematically, and for the ages (comments in the code) describe/ justify how all this works.

So if I'm understanding you right, you believe we should honour rippled limits on both intermediate and final values? And you are in favour of removing the ability to construct invalid amounts, but have some concerns about temporary pains this may cause?

clark800:
> tests are useful for many reasons

I mean that if we throw on every error, there won't be any invalid amounts to test. We would only be able to test that Amount throws on each possible error.

> why we don't replicate all the rounding quirks

Because the one rounding quirk that we diverge on is actually a bug in rippled which will presumably be fixed in the future. The rest of the quirks are not thought to be bugs in rippled.

> I would say yes.

Will do.

> justify how all this works.

I'd prefer to focus on fixing the issues first.

> you believe we should honour rippled limits on both intermediate and final values?

Yes, that is required for compatibility.

> And you are in favour of removing the ability to construct invalid amounts, but have some concerns about temporary pains this may cause?

Yes, but not just constructing them, also creating them through overflow etc.

sublimator:
So in the long term, 100% compatibility is a goal?

>>> I'd prefer to focus on fixing the issues first.

Clarity = power.

I know I'd like to see it, as someone reviewing this, to help me make sense
of all the changes/fixes, and it shows me this has been well thought
through.

For example, why is the noop/insignificance thing not considered a bug?

clark800:
> So in the long term, 100% compatibility is a goal?

It seems like some people are saying so. I'm not convinced that it is necessary. David Schwartz is not convinced either. If you want to discuss this, it is probably better to talk in the slack channel.

> I'd like to see it

It doesn't make sense to document when things are changing so fast. After fixing everything there shouldn't be too much left that needs documentation because it will all just make sense.

> why is the noop/insignificance thing not considered a bug?

That is just how fixed precision arithmetic works.

sublimator:
> That is just how fixed precision arithmetic works.

Sure, but many decimal libraries will report an error when there's not enough precision to describe a certain amount, and you have to be explicit about how you round/truncate.

sublimator:
> It doesn't make sense to document when things are changing so fast.

Yet I'm looking at a pull request where IOU limits exceeded are punished with errors and native amounts become `NaN`.  So you can understand why I'd like to see some better explanation.

clark800:
> Yet I'm looking at a pull request where IOU limits exceeded are punished with errors and native amounts become NaN. So you can understand why I'd like to see some better explanation.

That was the pre-existing behavior. It is bad, it will be removed. Why would we document something just to delete that documentation shortly thereafter?

clark800:
> Sure, but many decimal libraries will report an error when there's not enough precision to describe a certain amount, and you have to be explicit about how you round/truncate.

In any event, if you are going to break compatibility with rippled, you might as well go for the arbitrary precision option.

sublimator:
You need to serialise amounts to talk to rippled at some point

clark800:
I don't think throwing on precision errors is a good option. First of all, it is hard to define when to throw. Assume you have 3 digits of precision and you try to add 1.0 + 0.123, the answer is 1.12; should you throw because you lost the last digit? Or should you throw only if you lose all the digits in one term? If so, what about 100.0 + 0.5 = 101? This case is off from the real sum by 0.5, which is equally off as if the 0.5 was dropped, but in the opposite direction due to rounding. So despite being equally off, one case throws and the other doesn't? And what about precision loss during multiplication and division?

And the exception would only be useful if client code has something useful to do in response.

sublimator:
>  it is hard to define when to throw. 
> And what about precision loss during multiplication and division?

Yeah, division especially (repeats) would need rounding modes.

I think in most cases, when adding/subtracting for like currency, most terms will be compatible. But because rippled uses fixed precision, but floating point, when they aren't compatible, I don't believe we should make guesses as to what should be done. Callers should have to respond to errors, or proactively choose what to do with spills prior. 

The 3 digit precision mental experiments aren't really on point, because we have 16, but in both addition cases there, I guess I'm suggesting throwing.

Maybe these spills really are insignificant, and can be safely ignored and rounded away. (After all, it must be done for 1 divide by 3). I'm not convinced yet though.

> The exception would only be useful if client code has something useful to do in response.

Not doing anything can sometimes be useful.

src/js/ripple/amount.js
sublimator: Is this noIntegers used primarily in the _human family? If so, your interpretation would seem correct.

src/js/ripple/amount.js
```diff
@@ -9,6 +9,19 @@ var Seed      = require('./seed').Seed;
 var Currency  = require('./currency').Currency;
 var BigNumber = require('./bignumber');
 
+
+function isInteger(number) {
+  return parseInt(number) === number;
+}
+
+function noIntegers(value) {
```
sublimator: It feels kind of pointless commenting on this, given it's a transitional commit, but if this commit has any value, so does a comment, so here goes:

`noIntegers` ?

The number `1.0` IS an integer, as `isInteger` would tell you. 
You and I happen to know the purpose for this function, but other people I'm sure would benefit from a better name, and some explanatory comments.  Off the top of my head: `forceFloatRepr` or `forceDecimalPoint`

src/js/ripple/amount.js
```diff
@@ -574,40 +567,30 @@ Amount.prototype.parse_quality = function(quality, counterCurrency, counterIssue
...
-  // Correct offset if xrp_as_drops option is not set and base currency is XRP
-  if (!opts.xrp_as_drops &&
-      baseCurrency.is_valid() &&
-      baseCurrency.is_native()) {
-    if (opts.inverse) {
-      power -= 1;
+  var adjusted;
+  var inversion = opts.inverse ? -1 : 1;
+  // Correct for XRP to drops factors
+  if (!opts.xrp_as_drops) {
```
sublimator:
```javascript
  /*
  The quality, as stored in the last 64 bits of a directory index, is stored as
  the quotient of TakerPays/TakerGets. 

  When `opts.inverse` is true we are looking at a quality used for determining a
  `bid` price and it must first be inverted, before our declared base/counter
  currencies are in line with the price.

  For example:

    quality as stored :  5 USD          /  3000000 drops
    inverted          :  3000000 drops  /          5 USD
  */
  var adjusted = opts.inverse ? inverse(value) : value;
  var nativeAdjusted = adjusted;

  if (!opts.xrp_as_drops) {
    // `In a currency exchange, the exchange rate is quoted as the units of the
    //  counter currency in terms of a single unit of a base currency`. A
    //  quality is how much taker must `pay` to get ONE `gets` unit thus:
    //    pays ~= counterCurrency 
    //    gets ~= baseCurrency.
    if (this._is_native) {
      // pay:$price              drops  get:1 X
      // pay:($price / 1,000,000)  XRP  get:1 X
      nativeAdjusted = adjusted.div(Amount.bi_xns_unit);
    } else if (baseCurrency.is_valid() && baseCurrency.is_native()) {
      // pay:$price X                   get:1 drop
      // pay:($price * 1,000,000) X     get:1 XRP
      nativeAdjusted = adjusted.times(Amount.bi_xns_unit); 
    }
  }
  
  this._set_value(nativeAdjusted);
```

I still prefer this form, as I find it easier to read. That's debatable of course, but it uses measurably less steps.

test/amount-test.js
```diff
@@ -1182,11 +1187,11 @@ describe('Amount', function() {
 
   describe('amount limits', function() {
     it ('max JSON wire limite', function() {
-      assert.strictEqual(Amount.bi_xns_max.toString(), '9000000000000000000');
+      assert.strictEqual(Amount.bi_xns_max.toString(), '100000000000000000');
```
sublimator: There are unfortunately, some entries in old ledgers, that exceed this limit :( 

Monkey patching as configuration works I guess.

test/serializedtypes-test.js
```diff
@@ -570,8 +572,10 @@ describe('Serialized types', function() {
       assert.strictEqual(types.Amount.parse(so).to_json(), '270544960');
     });
     it('Parse 1161981756646125568 XRP', function () {
-      var so = new SerializedObject('5020304050607080');
-      assert.strictEqual(types.Amount.parse(so).to_json(), '1161981756646125696');
+      assert.throws(function() {
+        var so = new SerializedObject('1161981756646125568');
```
sublimator: The original line was `var so = new SerializedObject('5020304050607080');`  which doesn't seem to map to the `Parse 1161981756646125568 XRP'`. 

May as well have it make sense?

src/js/ripple/amount.js
```diff
@@ -342,8 +343,12 @@ Amount.prototype._check_limits = function() {
...
-  if (!this._is_native) {
-    var absval = this._value.absoluteValue();
+  var absval = this._value.absoluteValue();
+  if (this._is_native) {
+    if (absval.greaterThan(Amount.bi_xrp_max)) {
+      throw new Error('Exceeding max value of ' + Amount.bi_xrp_max.toString());
+    }
+  } else {
     if (absval.lessThan((new BigNumber(Amount.min_value)).absoluteValue())) {
       throw new Error('Exceeding min value of ' + Amount.min_value);
```
sublimator: It might be worth making some `bi_min_value` and `bi_max_value` BigNumber constants, given we do these checks a lot ?

src/js/ripple/amount.js
sublimator: It does indeed seem a *_human, force XRP scale for native amounts hack as you said.

sublimator:
No real problems popping out, so I guess LGTM. 

It would be good if some other people had a look too though.

src/js/ripple/amount.js
clark800: Yeah, I was thinking the same thing. I was going to wait because I also want to unexpose all the bi_* symbols so that dependencies aren't tied to our internal bignumber library. But I think it's worth making this part of the change now.

test/amount-test.js
clark800: How is that possible? I thought this was the total number of XRP drops created at the beginning (100 billion XRP * 1 million drops/XRP = 1e11 * 1e6 = 1e17).

src/js/ripple/amount.js
clark800: Ok, using `ensureDecimalPoint`

src/js/ripple/amount.js
clark800: Ok, patch applied

test/serializedtypes-test.js
clark800: It looks like that was supposed to be the hex representation, but it was wrong. I'll change it to the actual hex representation and add a comment.

wltsmrz:
Is this issue addressed? https://github.com/ripple/ripple-lib/issues/175

clark800:
No, should it be? I'm not clear on what the desired behavior is.

wltsmrz:
It doesn't have to be, but I noticed that there are some Amount-related issues. This one I believe is addressed in the current pull request https://github.com/ripple/ripple-lib/issues/100

I think the desired behavior is to make `is_valid` more strict

```
> require('./').Amount.is_valid('1/USD/asdf')
true
> require('./').Amount.is_valid_full('1/USD/asdf')
false

```

clark800:
#100 is https://ripplelabs.atlassian.net/browse/RLJS-215

Does that mean renaming is_valid_full to is_valid and removing the original is_valid?

test/amount-test.js
sublimator: There's thing like old offers, with takerPays of some huge amount. Recall that offers need only be partially funded.

sublimator: @shekenahglory

This change might be of concern to you when processing old offers.

src/js/ripple/amount.js
```diff
@@ -975,29 +937,25 @@ Amount.prototype.not_equals_why = function(d, ignore_issuer) {
     return 'Not an Amount';
   }
 
-  var result = false;
-
   if (!this.is_valid() || !d.is_valid()) {
-    result = 'Invalid amount.';
+    return 'Invalid amount.';
```
geertweening: No need for `else` if you're using returns. Personally I like the old approach of using a `result` var that gets returned in the end, but early return is fine with me as well. If you're using returns though, don't use `else`'s. Makes it confusing

geertweening:
LGTM

src/js/ripple/amount.js
clark800: I switched to returns because of the long-term goal of minimizing the use of variables (one of the todo items on the pull request for the javascript style guide). I'll refactor to remove else's.

PR #276, opened 2015-02-05
Bump dependencies versions

wltsmrz:


geertweening:
wow, looks like we were behind a lot on some libs

geertweening:
LGTM

clark800:
Should we switch some of these to carets instead of tildes?

geertweening:
The package.json was updated, but not the shrinkwrap. I just tested locally and we got some `ws` related stuff failing

PR #277, opened 2015-02-05
[FIX] unsymmetric memo serializing

geertweening:
treat memos as unknown binary, with optionally parsing unsynthesized hint fields

Re-calculating hashes for transactions from rippled was failing. This was due to ripple-lib assuming the binary hex in the MemoType and MemoFormat fields it was receiving for memos represented strings in utf8 format. This is true if the transaction and memo were constructed using ripple-lib, but not if it was constructed in a different way and contain another format.

This PR removes this assumption and moves the utf8 serializing to the transaction level and out of the serializer level.

@wltsmrz @sublimator @shekenahglory


sublimator:
LGTM

wltsmrz:
LGTM

PR #279, opened 2015-02-09
Fix IOU value passed to Amount.from_json()

wltsmrz:


src/js/ripple/orderbook.js
```diff
@@ -439,19 +439,22 @@ OrderBook.prototype.setFundedAmount = function(offer, fundedAmount) {
     return offer;
   }
 
-  var isOfferGetsExceeded = Amount.from_json(fundedAmount)
+  var isOfferGetsExceeded = Amount.from_json(
+    this._currencyGets.is_native()
+    ? fundedAmount
+    : fundedAmount + OrderBook.IOU_SUFFIX
```
clark800: Can't we lookup the _currencyGets currency and issuer and use them to construct an object with value, issuer, and currency keys? The IOU_SUFFIX looks like a hack.

clark800: Also, this pattern of checking if native and adjusting non-native form is duplication, it might be good to make a function to perform this operation.

wltsmrz: It's more trouble than it's worth to set the issuer to something other than an IOU placeholder. I'd say that it's actually worthless, as it makes no difference and is never exposed.

Calling a function on the OrderBook prototype won't de-duplicate very much.

clark800: If I understand correctly, shouldn't something like this work:

```
if (this._currencyGets.is_native()) {
    return fundedAmount;
} else {
    return {value: fundedAmount, currency: this._currencyGets.toString(), issuer: this._issuerGets.toString()};
}
```

That doesn't seem like too much trouble to get rid of a really ugly hack. And I point out the duplication because this bug is a result of missing one case. It would be less likely to be missed if there was a pattern of calling a conversion function on all incoming parameters at the top of the function.

wltsmrz: It's not a really ugly hack, and there is at least one case where your solution wouldn't work. This function and the rest of the orderbook code could use some refactoring, but that's not part of this bug fix. Thanks for the comments

clark800: Can you please let me know the case that won't work? I'm going to create a JIRA task for this and I want to document any risks.

geertweening:
LGTM @wltsmrz can you follow-up with @clark800 on the case that won't work?

PR #281, opened 2015-02-11
fix handling of false parameters in requestLedger

shekenahglory:
I passed in {expand:false, transactions:false} and still got transactions with the ledger.


wltsmrz:
LGTM

PR #283, opened 2015-02-17
[TASK] Disable parsing native amounts in foating point format

clark800:


geertweening:
LGTM

geertweening:
please rebase

clark800:
Made small change to orderbook-test.js to fix a test case and rebased. Bo reviewed the change to orderbook-test.js.

boxbag:
LGTM

sublimator:
An opt in would have been nice, so people can do gradual fixes of their code base (eg. rippled test suite)  but other wise stay up to date.  There's always monkey patches, or just biting the bullet and fixing it all up I guess :)

sublimator:
Err, opt out I meant

clark800:
Using drops instead of XRP is a fairly simple change and supposedly it was only used in the test cases for ripple-lib and rippled. Fixing it shouldn't be much harder than setting a flag, either would require changes to the code, so I think it's better to not increase the complexity.

sublimator:
Yeah, not a big deal.

PR #286, opened 2015-02-18
Set User-Agent Header with ripple-lib/{version}

lumberj:
- This would be helpful for us to analyze usage of ripple-lib for anyone
  using the public rippled

See: https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L530

### DIFF
--- a/src/js/ripple/utils.js
+++ b/src/js/ripple/utils.js
@@ -1,3 +1,8 @@
+var packageJson = require('../../../package.json');
+
+function getPackageVersion() {
+  return packageJson.version;
+}
 
 function getMantissaDecimalString(bignum) {
   var mantissa = bignum.toPrecision(16)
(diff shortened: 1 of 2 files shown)


src/js/ripple/utils.js
```diff
@@ -1,3 +1,8 @@
+var pJson = require('../../../package.json');
```
wltsmrz: Wouldn't you rather call this `package`?

lumberj: Copied from https://github.com/ripple/ripple-rest/blob/7142262ef0d1f4b367a37ffa2d5b8db7b56eb11b/api/lib/utils.js#L4

clark800: > Copied from ...

That doesn't mean it's good

lumberj: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar

Future reserved words include `package`

wltsmrz:
LGTM

src/js/ripple/utils.js
clark800: Why not call it packageJson? I think abbreviations are bad for readability.

lumberj: Changed

clark800:
LGTM

clark800:
Can we add a test case?

PR #292, opened 2015-02-26
Revert "Set User-Agent Header with ripple-lib/{version}"

boxbag:
Reverts ripple/ripple-lib#286


clark800:
For future reference and for the open source community, can you add a note about why you are reverting this?

boxbag:
Browser Websocket API is different. This throws an error in the browser when run.

clark800:
So there is currently no way to make it work in the browser? Is this the correct reference for the browser websocket API? https://developer.mozilla.org/en-US/docs/Web/API/WebSocket

boxbag:
Not to my current knowledge

PR #293, opened 2015-02-27
Offer quality

boxbag:
Keep track of offer quality from rippled and compute quality on new offers. Quality is important because once an offer is placed, its quality never changes. We should NOT be calculating quality on the fly.

### DIFF
--- a/src/js/ripple/orderbook.js
+++ b/src/js/ripple/orderbook.js
@@ -8,16 +8,22 @@
 //  - trade
 //  - transaction
 
-var _            = require('lodash');
-var util         = require('util');
-var extend       = require('extend');
-var assert       = require('assert');
-var async        = require('async');
+'use strict';
+
+var _ = require('lodash');
+var util = require('util');
+var extend = require('extend');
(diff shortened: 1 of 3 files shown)


src/js/ripple/orderbook.js
```diff
@@ -8,16 +8,22 @@
...
-var UInt160      = require('./uint160').UInt160;
-var Currency     = require('./currency').Currency;
-var log          = require('./log').internal.sub('orderbook');
+var Amount = require('./amount').Amount;
+var UInt160 = require('./uint160').UInt160;
+var Currency = require('./currency').Currency;
+var log = require('./log').internal.sub('orderbook');
+
+function assertValidNumber(number, message) {
+  assert(!_.isNull(number) && !isNaN(number), message);
```
wltsmrz: Maybe we should move `assert-diff` to dependencies and use `assert.number` https://github.com/mcavage/node-assert-plus/blob/master/assert.js#L126-L137.

wltsmrz:
LGTM

PR #294, opened 2015-02-28
Refactor base conversion

clark800:


geertweening:
Would like to see some tests for the base class and baseconverter, other than that LGTM

wltsmrz:
LGTM

PR #296, opened 2015-03-04
Fix currency parsing of non-alphanumeric and no-currency currencies

wltsmrz:


test/currency-test.js
```diff
@@ -54,6 +56,16 @@ describe('Currency', function() {
       assert(r.is_valid());
       assert.strictEqual('1D2', r.to_json());
     });
+    it('from_json("1").to_human()', function() {
+      var r = currency.from_json('1');
+      assert(r.is_valid());
+      assert.strictEqual(1, r.to_json());
+    });
+    it('from_json("#$%").to_human()', function() {
+      var r = currency.from_json('#$%');
```
geertweening: Rippled is going to fail on a currency like this right? So this is only to support older currencies that are in the ledger somewhere. Do we want to make this behavior explicit with an options parameter? If the option is provided, currencies like this won't trigger an error, otherwise they will. That way you have to be explicit and can't create an invalid currency in ripple-lib unless you know what you're doing

geertweening:
besides the one comment, LGTM

test/currency-test.js
wltsmrz: Nope, turns out these are in the accepted character set https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/UintTypes.cpp#L37-L43

PR #297, opened 2015-03-04
Add filename and line number to log, use log.warn() for deprecations

wltsmrz:
...rn()


src/js/ripple/log.js
```diff
@@ -43,8 +43,9 @@ Log.prototype._setParent = function(parentLogger) {
 
 Log.makeLevel = function(level) {
   return function() {
-    var args = Array.prototype.slice.call(arguments);
+    var args = Array.prototype.slice.apply(arguments);
     args[0] = this._prefix + args[0];
```
clark800: Why is _prefix added to args[0] but level is unshifted?

src/js/ripple/log.web.js
```diff
@@ -8,21 +8,39 @@ var exports = module.exports = require('./log.js');
...
+    var args = [
+      timestamp,
+      message,
+      '--',
+      location,
+      '\n'
+    ];
+
+    Array.prototype.slice.call(arguments, 2).forEach(function(arg) {
+      args.push(/MSIE/.test(navigator.userAgent)
```
clark800: This is a lot of duplication for this one check. Would it be better to just always do this check so that we can use a single function?

src/js/ripple/log.js
```diff
@@ -61,17 +62,36 @@ Log.prototype.error = Log.makeLevel(4);
...
+    switch (level) {
+      case 3:
+        logFn = console.warn;
+        break;
+      case 4:
+        logFn = console.error;
+        break;
+      default:
+        logFn = console.log;
+    }
```
clark800: This could be simplified if makeLevel took the log function itself (console.warn, console.log, or console.error) rather than an integer which then needs to be mapped back to a function here.

vinniefalco: @clark800 That sounds like a very good change

src/js/ripple/remote.js
```diff
@@ -1494,15 +1501,18 @@ Remote.parseBinaryTransaction = function(transaction) {
...
-  if (typeof meta.DeliveredAmount === 'string' || typeof meta.DeliveredAmount === 'object') {
-    tx_result.meta.delivered_amount = meta.DeliveredAmount;
-  } else {
-    switch (typeof tx_obj.Amount) {
-      case 'string':
-      case 'object':
-        tx_result.meta.delivered_amount = tx_obj.Amount;
-        break;
-    }
+  switch (typeof meta.DeliveredAmount) {
```
wltsmrz: This nested switch doesn't look great

src/js/ripple/remote.js
```diff
@@ -1775,13 +1785,15 @@ Remote.prototype._serverPrepareSubscribe = function(server, callback) {
 
 Remote.prototype.ledgerAccept =
 Remote.prototype.requestLedgerAccept = function(callback) {
+  /* eslint-disable consistent-return */
+  var request = new Request(this, 'ledger_accept');
+
   if (!this._stand_alone) {
+    // XXX This should emit error on the request
```
wltsmrz: I'm saving this for a pull request later

geertweening:
LGTM

vinniefalco:
The commit messages look much cleaner, :+1:

PR #299, opened 2015-03-04
[FIX] fix taker pays funded calculation

boxbag:
We were calling `parseInt` on a string that looks like `12345342e-3`, which gave us `12345342` instead of `12345`.

### DIFF
--- a/test/fixtures/orderbook.js
+++ b/test/fixtures/orderbook.js
@@ -372,6 +372,37 @@ module.exports.QUALITY_OFFERS = [
   }
 ];
 
+// This fixture is to exercise a bug where taker_pays_funded = taker_gets_funded * quality
+// has decimal amounts.
+module.exports.DECIMAL_TAKER_PAYS_FUNDED_OFFERS = [
+  {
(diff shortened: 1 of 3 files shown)


test/fixtures/orderbook.js
```diff
@@ -372,6 +372,35 @@ module.exports.QUALITY_OFFERS = [
   }
 ];
 
+module.exports.INCORRECT_TAKER_PAYS_FUNDED_OFFERS = [
```
lumberj: Why is this called `INCORRECT_TAKER_PAYS_FUNDED_OFFERS`? What is incorrect about it?

boxbag: Hmmm, I guess the taker pays funded result used to be incorrect for this offer when it was passed through `setOffers`. Any suggestions on what to call it?

lumberj: Why not just modify an existing test and fixture to use the large amount?

wltsmrz:
LGTM

PR #301, opened 2015-03-06
Add Amount.strict_mode for toggling range validation

wltsmrz:


src/js/ripple/amount.js
```diff
@@ -1,12 +1,16 @@
+ 'use strict';
+
+ /*eslint-disable new-cap*/
```
clark800: Not necessary to disable this

src/js/ripple/amount.js
```diff
@@ -180,15 +190,19 @@ Amount.prototype.divide = function(divisor) {
...
+ * @param opts.reference_date {Date|Number} Date based on which
+ * demurrage/interest should be applied. Can be given as JavaScript Date or int
+ * for Ripple epoch.
  * @return {Amount} The resulting ratio. Unit will be the same as numerator.
  */
 
 Amount.prototype.ratio_human = function(denominator, opts) {
   opts = extend({ }, opts);
 
+  /*eslint-disable consistent-this */
```
clark800: Not necessary to disable this

src/js/ripple/amount.js
```diff
@@ -804,16 +836,19 @@ Amount.prototype.to_human = function(opts) {
     return 'NaN';
   }
 
+  /* eslint-disable consistent-this */
```
clark800: Not necessary to disable this

src/js/ripple/serializedtypes.js
```diff
@@ -505,41 +588,39 @@ var STPathSet = exports.PathSet = new SerializedType({
...
+       each entry has one or more of the following attributes:
+       amount, currency, issuer.
        */
 
-    var path_list    = [];
+    var path_list = [];
     var current_path = [];
     var tag_byte;
 
+    /* eslint-disable no-cond-assign */
```
clark800: Not necessary to disable this

test/serializedobject-test.js
```diff
@@ -1,6 +1,9 @@
-var assert           = require('assert');
+/* eslint-disable */
```
clark800: Disable max-len only

wltsmrz: No

src/js/ripple/amount.js
wltsmrz: Nothing is necessary; I want to disable this

src/js/ripple/serializedtypes.js
wltsmrz: Nothing is necessary; I want to disable this

src/js/ripple/amount.js
wltsmrz: Nothing is necessary; I want to disable this

src/js/ripple/amount.js
wltsmrz: Nothing is necessary; I want to disable this

geertweening: Instead of disabling, can we rename wherever this `new-cap` setting was throwing an error? We reached consensus on the rules, disabling them instead of following throws away that effort. If we want to our code to look like what we agreed on, we need to make some changes. It doesn't seem like it's a big change unless I'm missing something.

src/js/ripple/amount.js
```diff
@@ -180,15 +190,19 @@ Amount.prototype.divide = function(divisor) {
...
+ * demurrage/interest should be applied. Can be given as JavaScript Date or int
+ * for Ripple epoch.
  * @return {Amount} The resulting ratio. Unit will be the same as numerator.
  */
 
 Amount.prototype.ratio_human = function(denominator, opts) {
   opts = extend({ }, opts);
 
+  /*eslint-disable consistent-this */
   var numerator = this;
```
geertweening: if `numerator` is the same as `this`. Can we just use `this` here without creating a new var?

src/js/ripple/amount.js
```diff
@@ -804,16 +836,19 @@ Amount.prototype.to_human = function(opts) {
     return 'NaN';
   }
 
+  /* eslint-disable consistent-this */
   // Apply demurrage/interest
   var ref = this;
```
geertweening: Could `ref` just be `this`? And if not, what is the `consistent-this` setting want? If it's just `self` instead of `ref` that shouldn't be too hard to change.

geertweening:
Even for test files, we still want to keep a consistent way of writing JavaScript. Disabling all eslint rules doesn't really help there. I'd rather be more strict and only disable things that make sense.

geertweening:
Other than the eslint comments, LGTM

src/js/ripple/amount.js
wltsmrz: I think there's good reason to keep the existing `numerator` semantics. There is nothing to gain by changing it to `self`. Same for the `ref` thing below

src/js/ripple/amount.js
wltsmrz: It's ok to disable eslint rules, as we've also discussed. Keeping this eslint rule will mean changing Amount.NaN(), as `NaN` is capitalized. I think it's fine, and I'd rather disable it in the entire file rather than in eight different locations.

clark800: Instead of Amount.NaN(), why don't we do "new Amount(NaN)"?

test/serializedobject-test.js
```diff
@@ -1,6 +1,19 @@
...
+
+/* eslint-disable max-len*/
+/* eslint-disable key-spacing*/
+/* eslint-disable no-spaced-func*/
+/* eslint-disable space-before-blocks*/
+/* eslint-disable space-unary-ops*/
+/* eslint-disable no-void*/
+/* eslint-disable semi*/
+/* eslint-disable no-unused-vars*/
+/* eslint-disable quotes*/
```
clark800: This is a single find and replace operation, it would have been faster to fix it than to type this disabling comment.

src/js/ripple/amount.js
wltsmrz: I think that's equivalent to `new Amount()`. Are you proposing that we get rid of `Amount.NaN()`?

clark800: Yeah, I think it would be better to get rid of it, it doesn't seem like it should be a public thing. It might be nice to add a parameter to the amount constructor to make it explicit that we are setting it to NaN.

test/serializedobject-test.js
wltsmrz: `VypWWcwquotes` is faster for me than `ggVG:s/"/'/g` actually

src/js/ripple/amount.js
wltsmrz: It's unlikely that it's being used externally but let's not forget to add it to the release notes. I'll change Amount.NaN() to new Amount(NaN) in this file

clark800: Should also add an optional argument to the Amount constructor so that the argument isn't ignored.

PR #302, opened 2015-03-06
[FIX] Don't lint deleted files and pull eslintrc from javascript-style-g...

clark800:
...uide repo to ensure that it is up to date


wltsmrz:
Shorten your commit message to something like "Don't lint deleted files and update eslintrc"

clark800:
@wltsmrz Why?

wltsmrz:
Nevermind, maybe "Don't lint deleted files and pull eslintrc from github"

eslintrc
```diff
@@ -1,155 +0,0 @@
-# ESLint documentation can be found at http://eslint.org/docs/
```
wltsmrz: Why have you removed the eslint configuration?

clark800:
I think it's more clear as is.

eslintrc
clark800: It is being pulled from github on demand.

wltsmrz:
I don't think it's more clear. "Pull from javascript-style-guide" doesn't really give any indication of what's happening. "Pull from github" gives some indication that the configuration is being updated remotely, which is the important part, rather than the name of the repository it's being pulled from.

eslintrc
wltsmrz: Yeah, for Travis. What about locally?

clark800: Locally you can put it in ~/.eslintrc

wltsmrz: Why should we break `npm run lint` locally?

clark800: Oh, I haven't been using that because it runs on all files and spews too much. But yeah, we should either fix it or remove it. How about we remove the "-c" option to make it use the ~/.eslintrc file?

wltsmrz: No, we should leave the eslintrc and update it as needed so we can run locally as usual. There are plenty of files that have been untouched for almost a year and probably won't be touched for a long time.

wltsmrz:
Aside from the commit message, LGTM

lumberj:
LGTM

PR #303, opened 2015-03-06
[TASK] Fix issues in base.js and add more tests

clark800:


src/js/ripple/base.js
```diff
@@ -145,16 +133,10 @@ Base.decode_check = function(version, input, alphabet) {
...
-
-    for (var i = 0, l = version.length; i < l; i++) {
-      match |= version[i] === buffer[0];
-    }
-
-    if (!match) {
-      return NaN;
-    }
+  if (Array.isArray(version) && _.every(version, function(v) {
+          return v !== buffer[0];
```
wltsmrz: This is kind of yucky

wltsmrz:
LGTM

src/js/ripple/base.js
clark800: Yes, ideally it would be written as
`if (Array.isArray(version) && _.every(version, v => v !== buffer[0]))`

wltsmrz: I wouldn't call that ideal

PR #311, opened 2015-03-18
[FIX] Amount: clone in ratio_human, product_human

vhpoet:
Amount.ratio_human and Amount.product_human should only change and return the cloned Amount object.

### DIFF
--- a/src/js/ripple/amount.js
+++ b/src/js/ripple/amount.js
@@ -197,9 +197,7 @@ Amount.prototype.divide = function(divisor) {
 Amount.prototype.ratio_human = function(denominator, opts) {
   opts = extend({ }, opts);
 
-  /*eslint-disable consistent-this */
-  var numerator = this;
-  /*eslint-enable consistent-this */
+  var numerator = this.clone();
 
   denominator = Amount.from_json(denominator);
 


clark800:
clone() does not have side effects; this pull request has no effect

vhpoet:
@clark800 yea, just noticed, will commit an updated in a sec.

vhpoet:
@clark800 should be fine now.

vhpoet:
@clark800 can you please make sure this doesn't happen anywhere else?

src/js/ripple/amount.js
```diff
@@ -283,6 +284,7 @@ Amount.prototype.product_human = function(factor, opts) {
   //
   // See also Amount#ratio_human.
   if (factor._is_native) {
+    product = product.clone();
     product._set_value(product._value.dividedBy(Amount.bi_xns_unit));
```
clark800: This could be `return product._value.dividedBy(Amount.bi_xns_unit)` so you could get rid of the clone and mutation. Might be good to do the same for the other one too.

geertweening:
LGTM

clark800:
LGTM

PR #313, opened 2015-03-19
[FEATURE] add offer autobridging

boxbag:


src/js/ripple/orderbook.js
```diff
@@ -20,11 +20,25 @@ var Amount = require('./amount').Amount;
...
+  DECIMAL_PLACES: 40
+});
 
 function assertValidNumber(number, message) {
   assert(!_.isNull(number) && !isNaN(number), message);
 }
 
+function assertValidLegOneOffer(legOneOffer, message) {
+  assert.strictEqual(typeof legOneOffer, 'object', message);
+  assert.strictEqual(typeof legOneOffer.TakerPays, 'object', message);
```
wltsmrz: Might want to assert(legOneOffer) as it could be `null`, I guess

src/js/ripple/orderbook.js
```diff
@@ -62,6 +79,35 @@ function OrderBook(remote, getsC, getsI, paysC, paysI, key) {
...
+
+  if (!this._currencyGets.is_native() && !this._currencyPays.is_native()) {
+    this._legOneBook = remote.createOrderBook({
+      currency_gets: 'XRP',
+      currency_pays: paysC,
+      issuer_pays: paysI
+    });
+
+    this._legOneBook.on('model', function() {
+      self.computeAutobridgedOffers();
```
wltsmrz: Minor duplication in both of the 'model' listeners

wltsmrz:
LGTM

src/js/ripple/orderbook.js
```diff
@@ -62,6 +80,34 @@ function OrderBook(remote, getsC, getsI, paysC, paysI, key) {
...
+
+  function computeAutobridgedOffersWrapper () {
+    self.computeAutobridgedOffers();
+    self.mergeDirectAndAutobridgedBooks();
+  }
+
+  if (!this._currencyGets.is_native() && !this._currencyPays.is_native()) {
+    this._legOneBook = remote.createOrderBook({
+      currency_gets: 'XRP',
+      currency_pays: paysC,
```
clark800: One letter abbreviations are hard to understand

src/js/ripple/orderbook.js
```diff
@@ -532,79 +659,139 @@ OrderBook.prototype.resetOwnerOfferTotal = function(account) {
...
-OrderBook.prototype.getOfferTakerGetsFunded = function(offer) {
-  assertValidNumber(offer.taker_gets_funded, 'Taker gets funded is invalid');
+OrderBook.prototype.getLeftoverOwnerFunds = function(account) {
+  assert(UInt160.is_valid(account), 'Account is invalid');
 
-  return Amount.from_json(offer.taker_gets_funded + OrderBook.IOU_SUFFIX);
+  var amount = this._ownerFundsLeftover[account];
+
+  if (!amount) {
+    amount = Amount.from_json('0' + OrderBook.IOU_SUFFIX);
```
clark800: Somebody has to fix this IOU_SUFFIX thing...

src/js/ripple/orderbook.js
```diff
@@ -532,79 +659,139 @@ OrderBook.prototype.resetOwnerOfferTotal = function(account) {
...
+ * Retrieve offer quality
+ *
+ * @param {Object} offer
+ */
+
+OrderBook.prototype.getOfferQuality = function(offer) {
+  var amount;
+
+  if (this._currencyGets.has_interest()) {
+    // XXX Should use Amount#from_quality
```
clark800: Could you add an explanation of why it isn't using from_quality currently?

src/js/ripple/orderbook.js
wltsmrz: "Fix" may be the wrong word, but yeah, sometime the improvement which was suggested previously ought to be made, doesn't have to be part of this pull request

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1339,347 @@ OrderBook.prototype.is_valid = function() {
...
+/**
+ * Formats an offer quality amount to a hex that can be parsed by
+ * Amount.parse_quality
+ *
+ * @param [Amount] quality
+ *
+ * @return [String]
+ */
+
+OrderBook.prototype.convertOfferQualityToHex = function(quality) {
```
clark800: Most of this is copy-pasted from serializedtypes.js, is there a way to use the STAmount class for this?

boxbag: It's based on STAmount, but I don't think enough is common for code reuse

src/js/ripple/orderbook.js
boxbag: No idea, do you know @wltsmrz?

src/js/ripple/orderbook.js
boxbag: What's the proposed fix?

src/js/ripple/orderbook.js
wltsmrz: I think there is no good reason

src/js/ripple/orderbook.js
clark800: https://ripplelabs.atlassian.net/browse/RLJS-238

boxbag: This is also a patch, the real solution from what I understand is to change Amount.js to support division for native/non native.

src/js/ripple/orderbook.js
clark800: Everything up to "var valueHex" appears to be duplicated. I looked at it and it should be easy to use STAmount, the only complication is that you have to create a SerializedObject to pass to STAmount, then call to_hex on SerializedObject.

boxbag:
```
hi |= ((100 + offset) & 0xff) << 24;
```

vs

```
STAmount
hi |= ((97 + offset) & 0xff) << 22;
```

clark800: Why is that magic number different?

boxbag: Also, we don't want this from STAmount `hi |= 1 << 31;`

boxbag: `parse_quality` in Amount.js. `var offset = parseInt(offset_hex, 16) - 100;` The offset isn't correct with the value `97`

clark800: You mean there are two nearly identical serialization formats that differ by a magic number? Do you know why?

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1339,347 @@ OrderBook.prototype.is_valid = function() {
...
+  legOneOffer.TakerPays = takerGets.multiply(this.getOfferQuality(legOneOffer));
+};
+
+/**
+ * Merge direct and autobridged offers into a combined orderbook
+ *
+ * @return [Array]
+ */
+
+OrderBook.prototype.mergeDirectAndAutobridgedBooks = function() {
```
clark800: I think most of this function can be replaced with
`this._mergedOffers = _.sortBy(this._offers.concat(this._offersAutobridged), this.getOfferQuality);`

src/js/ripple/orderbook.js
```diff
@@ -62,6 +76,34 @@ function OrderBook(remote, getsC, getsI, paysC, paysI, key) {
...
+  this._legTwoBook = null;
+
+  function computeAutobridgedOffersWrapper() {
+    self.computeAutobridgedOffers();
+    self.mergeDirectAndAutobridgedBooks();
+  }
+
+  if (!this._currencyGets.is_native() && !this._currencyPays.is_native()) {
+    this._legOneBook = remote.createOrderBook({
+      currency_gets: 'XRP',
```
boxbag: `taker_gets_currency`

src/js/ripple/orderbook.js
```diff
@@ -318,6 +402,19 @@ OrderBook.prototype.subscribeTransactions = function(callback) {
 };
 
 /**
+ * Handles notifying listeners that direct offers have changed. For autobridged
+ * books, an additional merge step is also performed
+ */
+
+OrderBook.prototype.notifyDirectOffersChanged = function() {
+  if (this._currencyGets.is_native() || this._currencyPays.is_native()) {
```
boxbag: replace with `isAutobridgedBook`

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+
+  var legOneOffers = _.cloneDeep(this._legOneBook.getOffersSync());
+  var legTwoOffers = _.cloneDeep(this._legTwoBook.getOffersSync());
+
+  var legOnePointer = 0;
+  var legTwoPointer = 0;
+
+  this._offersAutobridged = [];
+
+  this.clearOwnerFundsLeftover();
```
boxbag: owner funds object doesn't belong in orderbook instance

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+ */
+
+OrderBook.prototype.computeAutobridgedOffers = function() {
+  assert(!this._currencyGets.is_native() && !this._currencyPays.is_native(),
+    'Autobridging is only for IOU:IOU orderbooks');
+
+  var legOneOffers = _.cloneDeep(this._legOneBook.getOffersSync());
+  var legTwoOffers = _.cloneDeep(this._legTwoBook.getOffersSync());
+
+  var legOnePointer = 0;
```
boxbag: `legOneDepthIndex`

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+  while (legOneOffers[legOnePointer] && legTwoOffers[legTwoPointer]) {
+    var legOneOffer = legOneOffers[legOnePointer];
+    var legTwoOffer = legTwoOffers[legTwoPointer];
+    var leftoverFunds = this.getLeftoverOwnerFunds(legOneOffer.Account);
+    var autobridgedTakerGets, autobridgedTakerPays;
+    var legOneTakerGetsFunded, legTwoTakerPaysFunded;
+
+    if (legOneOffer.Account === legTwoOffer.Account) {
+      this.unclampLegOneOwnerFunds(legOneOffer);
+    } else if (!legOneOffer.is_fully_funded && !leftoverFunds.is_zero()) {
```
boxbag: this is confusing

src/js/ripple/orderbook.js
```diff
@@ -532,79 +655,139 @@ OrderBook.prototype.resetOwnerOfferTotal = function(account) {
 };
 
 /**
- * Casts and returns offer's taker gets funded amount as a default IOU amount
+ * Clear owner funds leftovers
+ */
+
+OrderBook.prototype.clearOwnerFundsLeftover = function() {
```
boxbag: XRPLeftover

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+    var autobridgedTakerGets, autobridgedTakerPays;
+    var legOneTakerGetsFunded, legTwoTakerPaysFunded;
+
+    if (legOneOffer.Account === legTwoOffer.Account) {
+      this.unclampLegOneOwnerFunds(legOneOffer);
+    } else if (!legOneOffer.is_fully_funded && !leftoverFunds.is_zero()) {
+      this.adjustLegOneFundedAmount(legOneOffer);
+    }
+
+    if (legOneOffer.taker_gets_funded === '0') {
```
boxbag: `legOneOffer.taker_gets_funded.is_zero()`

boxbag:
Extract autobridging to its own module

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+        this.setLegOneTakerGets(legOneOffer, updatedTakerGets);
+
+        this.clampLegOneOwnerFunds(legOneOffer);
+      } else {
+        // Update funded amount since leg one offer was not completely consumed
+        updatedTakerGets = legOneTakerGetsFunded
+          .subtract(legTwoTakerPaysFunded);
+
+        this.setLegOneTakerGetsFunded(legOneOffer, updatedTakerGets);
+      }
```
boxbag: extract everything before pointer increment

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+      // on leg one out to have it match leg two in
+      var legOneQuality = this.getOfferQuality(legOneOffer);
+      var updatedTakerGets;
+
+      autobridgedTakerGets = this.getOfferTakerGetsFunded(legTwoOffer);
+      autobridgedTakerPays = legTwoTakerPaysFunded.multiply(legOneQuality);
+
+      if (legOneOffer.Account === legTwoOffer.Account) {
+        var legOneTakerGets = this.getOfferTakerGets(legOneOffer);
+        updatedTakerGets = legOneTakerGets.subtract(legTwoTakerPaysFunded);
```
boxbag: partially consume leg one, up to limit of leg two in

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+
+        this.setLegOneTakerGets(legOneOffer, updatedTakerGets);
+
+        this.clampLegOneOwnerFunds(legOneOffer);
+      } else {
+        // Update funded amount since leg one offer was not completely consumed
+        updatedTakerGets = legOneTakerGetsFunded
```
boxbag: `updatedTakerGetsFunded`

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+
+      legOnePointer++;
+      legTwoPointer++;
+    }
+
+    var autobridgedOffer = this.formatAutobridgedOffer(
+      autobridgedTakerGets,
+      autobridgedTakerPays
+    );
+    autobridgedOffer.index = legOnePointer + legTwoPointer;
```
boxbag: find another solution for this

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+ * Set taker gets funded amount for a IOU:XRP offer. Also calculates taker
+ * pays funded using offer quality and updates is_fully_funded flag
+ *
+ * @param [Object] legOneOffer - IOU:XRP offer
+ * @param [Amount] takerGetsFunded
+ */
+
+OrderBook.prototype.setLegOneTakerGetsFunded = setLegOneTakerGetsFunded;
+
+function setLegOneTakerGetsFunded(legOneOffer, takerGetsFunded) {
```
boxbag: `setLegOneFundedAmounts`

src/js/ripple/orderbook.js
kureus: @boxbag Does amount.js only currently support division for non native?

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+  return so.to_hex();
+};
+
+/**
+ * Remove funds clamp on leg one offer. This is necessary when the two offers
+ * are owned by the same account. In this case, it doesn't matter if offer one
+ * is not fully funded. Leg one out goes to leg two in and since its the same
+ * account, an infinite amount can flow.
+ *
+ * @param [Object] legOneOffer - IOU:XRP offer
```
sublimator: I got the impression we have JSDoc linting turned on: http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

[bracket-syntax] denotes optionality?

src/js/ripple/orderbook.js
sublimator: It's supported, but the amount has at most 6 decimal places ( like XRP ) iirc

src/js/ripple/orderbook.js
sublimator: Ah, so optional takes this form `@param {string} [somebody] `. Elsewhere, we seem to use squigglies though.

src/js/ripple/orderbook.js
```diff
@@ -1176,4 +1335,307 @@ OrderBook.prototype.is_valid = function() {
...
+OrderBook.prototype.formatAutobridgedOffer = function(takerGets, takerPays) {
+  assert(takerGets instanceof Amount, 'Autobridged taker gets is invalid');
+  assert(takerPays instanceof Amount, 'Autobridged taker pays is invalid');
+
+  var autobridgedOffer = {};
+  var quality = takerPays.divide(takerGets);
+
+  autobridgedOffer.TakerGets = {
+    value: takerGets.to_text(),
+    currency: this._currencyGets.to_hex(),
```
sublimator: We do this because to_json() will lose information, serialising as just an iso code? Maybe we should update to_json() to use to_hex() when it's needed for lossless encoding?

geertweening:
LGTM

PR #317, opened 2015-03-27
Check that Error.stack is available, fixes logging in browser

wltsmrz:


src/js/ripple/log.js
```diff
@@ -75,7 +75,7 @@ function getLogInfo(message, args) {
     message,
     '--',
     // Location
-    (new Error()).stack.split('\n')[4].replace(/^\s+/, ''),
+    (typeof window === 'object') ? '' : (new Error()).stack.split('\n')[4].replace(/^\s+/, ''),
```
clark800: How about explicitly checking whether `stack` is defined?

wltsmrz: Not sure if that would be adequate. Can we assume that it will be the same for all browser or browser-like environments? I'd kind of rather not take a risk, but I guess we could just Array.isArray(new Error().stack) or something

wltsmrz: And by that I mean... typeof new Error().stack === 'string'...

clark800: Yeah, sounds good

clark800:
LGTM

PR #318, opened 2015-03-30
Fix SerializedObject.append performance issue

sublimator:
@shekenahglory 

Try with ledger 7501326

### DIFF
--- a/scripts/verify_ledger_json.js
+++ b/scripts/verify_ledger_json.js
@@ -1,31 +1,34 @@
+'use strict';
+
 var fs = require('fs');
+var Amount = require('../src/js/ripple').Amount;
 var Ledger = require('../src/js/ripple/ledger').Ledger;
 
 function parse_options(from, flags) {
   var argv = from.slice(),
-      opts = {argv:argv};
+      opts_ = {argv: argv};
 
   flags.forEach(function(f) {
     // Do we have the flag?
     var flag_index = argv.indexOf('--' + f);
     // normalize the name of the flag
     f = f.replace('-', '_');
-    // opts has Boolean value for normalized flag key
-    opts[f] = !!~flag_index;
-    if (opts[f]) {
+    // opts_ has Boolean value for normalized flag key
+    opts_[f] = flag_index !== -1;
+    if (opts_[f]) {
       // remove the flag from the argv
       argv.splice(flag_index, 1);
     }
   });
-  return opts;
+  return opts_;
 }
 
 var opts = parse_options(process.argv.slice(2), // remove `node` and `this.js`
                         ['sanity-test']);
 
 if (opts.argv.length < 1) {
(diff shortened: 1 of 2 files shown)


src/js/ripple/serializedobject.js
```diff
@@ -115,9 +119,9 @@ SerializedObject.check_no_missing_fields = function(typedef, obj) {
   if (missing_fields.length > 0) {
     var object_name;
 
-    if (obj.TransactionType !== void(0)) {
+    if (obj.TransactionType !== undefined) {
       object_name = SerializedObject.lookup_type_tx(obj.TransactionType);
-    } else if (obj.LedgerEntryType != null){
+    } else if (obj.LedgerEntryType !== undefined) {
```
wltsmrz: `!= null` catches null and undefined, but I'm guessing this is sufficient

wltsmrz:
LGTM

scripts/verify_ledger_json.js
```diff
@@ -1,31 +1,34 @@
...
   flags.forEach(function(f) {
     // Do we have the flag?
     var flag_index = argv.indexOf('--' + f);
     // normalize the name of the flag
     f = f.replace('-', '_');
-    // opts has Boolean value for normalized flag key
-    opts[f] = !!~flag_index;
-    if (opts[f]) {
+    // opts_ has Boolean value for normalized flag key
+    opts_[f] = !!~flag_index;
```
clark800: I know this isn't part of your pull request, but we should try to replace `!!~x` with `x !== -1` or use `_.includes` instead. You don't have to do it now, but I just want to point out these cases.

src/js/ripple/serializedobject.js
```diff
@@ -272,17 +288,17 @@ SerializedObject.prototype.hash = function(prefix) {
...
+  // var Type = stypes[spec[3]];
 
   if (typeof obj[name] !== 'undefined') {
     // ST: Old serialization code
-    //this.append(SerializedObject.get_field_header(Type.id, field_id));
+    // this.append(SerializedObject.get_field_header(Type.id, field_id));
     try {
       // ST: Old serialization code
-      //Type.serialize(this, obj[name]);
+      // Type.serialize(this, obj[name]);
```
clark800: Can we remove all the commented code?

sublimator: k, will lodash it and excise the commented out codes

src/js/ripple/serializedobject.js
sublimator: I always use != null, cause I learned js from looking at compiled coffee code. iirc existence operator `?` uses `!= null`. But the linter was crying. Using `if ( bla )` probably maintain semantics better, but it's not a big deal here eh?

geertweening:
@sublimator please squash

geertweening:
LGTM

PR #319, opened 2015-03-30
eslint fixes

sublimator:
Was going to have a stab at multi-sign , KeyPair caching and ed25519 sigs in the near future, so `prepping the work area` so to speak.

@wltsmrz 
@clark800 

A few thoughts/questions/observations:
* Is there anyway to narrow the scope from the whole file to a specific block of code when using comment directives to downgrade from error to warning?
* Does `no-multi-space` (or whatever) really buy much?
* `new-cap` is a pain in the arse: I couldn't get the exceptions for `sjcl.bn` working so just surrounded in <enable></enable>
* JSDoc demands some kind of return, for void, use `@return {void}`

### DIFF
--- a/src/js/ripple/seed.js
+++ b/src/js/ripple/seed.js
@@ -1,18 +1,19 @@
+'use strict';
+
 //
 // Seed support
 //
 
 var extend = require('extend');
-var utils  = require('./utils');
-var sjcl   = utils.sjcl;
+var utils = require('./utils');
+var sjcl = utils.sjcl;
 
-var Base    = require('./base').Base;
-var UInt    = require('./uint').UInt;
-var UInt256 = require('./uint256').UInt256;
+var Base = require('./base').Base;
+var UInt = require('./uint').UInt;
 var UInt160 = require('./uint160').UInt160;
 var KeyPair = require('./keypair').KeyPair;
 
-var Seed = extend(function () {
+var Seed = extend(function() {
   this._curve = sjcl.ecc.curves.k256;
   this._value = NaN;
 }, UInt);
@@ -23,7 +24,7 @@ Seed.prototype.constructor = Seed;
 
 // value = NaN on error.
 // One day this will support rfc1751 too.
-Seed.prototype.parse_json = function (j) {
+Seed.prototype.parse_json = function(j) {
   if (typeof j === 'string') {
     if (!j.length) {
       this._value = NaN;
@@ -43,7 +44,7 @@ Seed.prototype.parse_json = function (j) {
   return this;
 };
 
-Seed.prototype.parse_passphrase = function (j) {
+Seed.prototype.parse_passphrase = function(j) {
   if (typeof j !== 'string') {
     throw new Error('Passphrase must be a string');
   }
@@ -56,7 +57,7 @@ Seed.prototype.parse_passphrase = function (j) {
   return this;
 };
 
-Seed.prototype.to_json = function () {
+Seed.prototype.to_json = function() {
   if (!(this.is_valid())) {
     return NaN;
   }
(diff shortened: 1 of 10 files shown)


src/js/ripple/uint160.js
```diff
@@ -19,12 +21,12 @@ UInt160.width = 20;
...
 UInt160.prototype.constructor = UInt160;
 
-var ACCOUNT_ZERO = UInt160.ACCOUNT_ZERO = 'rrrrrrrrrrrrrrrrrrrrrhoLvTp';
-var ACCOUNT_ONE  = UInt160.ACCOUNT_ONE  = 'rrrrrrrrrrrrrrrrrrrrBZbvji';
-var HEX_ZERO     = UInt160.HEX_ZERO     = '0000000000000000000000000000000000000000';
-var HEX_ONE      = UInt160.HEX_ONE      = '0000000000000000000000000000000000000001';
-var STR_ZERO     = UInt160.STR_ZERO     = utils.hexToString(HEX_ZERO);
-var STR_ONE      = UInt160.STR_ONE      = utils.hexToString(HEX_ONE);
+/*var ACCOUNT_ZERO = */UInt160.ACCOUNT_ZERO = 'rrrrrrrrrrrrrrrrrrrrrhoLvTp';
+/*var ACCOUNT_ONE  = */UInt160.ACCOUNT_ONE = 'rrrrrrrrrrrrrrrrrrrrBZbvji';
```
clark800: Looks like you were planning on removing these comments?

wltsmrz:
> Does no-multi-space (or whatever) really buy much?

No. Most of the rules are purely stylistic. In some cases (such as binformat) I see an argument for disabling this particular eslint rule

> new-cap is a pain in the arse: I couldn't get the exceptions for sjcl.bn working so just surrounded in

Yeah. Guess you have to do `var SJCL_BN = sjcl.bn` or something

> JSDoc demands some kind of return, for void, use @return {void}

I think it demands a return description if there is a return in the function, yeah

src/js/ripple/uint160.js
sublimator: No, just undecided. Should just remove the unused vars?

wltsmrz:
LGTM

src/js/ripple/uint160.js
```diff
@@ -90,9 +92,9 @@ UInt160.prototype.to_json = function(opts) {
       }
 
       return output;
-    } else {
-      return this.to_hex();
     }
+    // else
```
clark800: This is why I wasn't an advocate of this rule. Comments like this aren't good. We should either remove this comment or remove the eslint rule.

src/js/ripple/seed.js
```diff
@@ -1,18 +1,20 @@
...
-var sjcl   = utils.sjcl;
+var utils = require('./utils');
+var sjcl = utils.sjcl;
 
-var Base    = require('./base').Base;
-var UInt    = require('./uint').UInt;
-var UInt256 = require('./uint256').UInt256;
+var Base = require('./base').Base;
+var UInt = require('./uint').UInt;
+// var UInt256 = require('./uint256').UInt256;
```
clark800: If we don't need it, let's remove it.

src/js/ripple/uint160.js
sublimator: yeah, I'd be happy with removing the else / return ban

sublimator: Will clean up scraggles tomoz then. Unless someone else wants to... ? What
a joyous thing to wake up to :)

src/js/ripple/shamap.js
```diff
@@ -32,18 +21,26 @@ SHAMapTreeNode.TYPE_TRANSACTION_MD = 3;
...
 SHAMapTreeNode.prototype.add_item = function(tag, node) {
-  throw new Error('Called unimplemented virtual method SHAMapTreeNode#add_item.');
+  throw new Error(
+    'Called unimplemented virtual method SHAMapTreeNode#add_item.');
 };
 
 SHAMapTreeNode.prototype.hash = function() {
   throw new Error('Called unimplemented virtual method SHAMapTreeNode#hash.');
 };
+/*eslint-enable no-unused-vars*/
```
clark800: This can be moved up I think.

clark800:
Yeah, for `new-cap` I think the simplest solution is what Brandon suggested so that you only have to disable the rule in one place.

`no-multi-space` just buys more consistency in style. This question of alignment was discussed in "Clean Code" where the author said he used to do horizontal alignment, but eventually decided it wasn't worth it. I believe he also made the point that if you are tempted to do horizontal alignment, your function is probably too big. I tend to agree with that and add that it can be tedious to maintain the alignment when adding new lines that are longer than the previous ones and you have to adjust all the other lines.

sublimator:
lol, so we enforce new-cap, but not strictly snake_case or camelCase ?

> it can be tedious to maintain

Yeah, it can be tedious, but mostly I prefer to read aligned code, and code
tends to generally more often read than tediously tweaked. Not a big deal,
I guess you can sort of get used to reading it all cramped. Mostly habit.

clark800:
Yeah, it would be nice to enforce the casing, but we left that out for now because there would be too many errors. Once everyone has all of the current lint errors handled we may try adding camelCase.

geertweening:
@sublimator squash please

sublimator:
You're ok with the commit contents, and you want this all squashed down to one commit?

geertweening:
LGTM

PR #320, opened 2015-04-04
[WIP] Remove account families support

sublimator:
A sketch, up for comments and to get a feel for what will need lint fixes prior to modification. The account family feature, is not really supported in rippled anymore, and for good reason: no one really used any account but the first.

### DIFF
--- a/src/js/ripple/message.js
+++ b/src/js/ripple/message.js
@@ -1,20 +1,37 @@
-var async              = require('async');
-var crypto             = require('crypto');
-var sjcl               = require('./utils').sjcl;
-var Remote             = require('./remote').Remote;
-var Seed               = require('./seed').Seed;
-var KeyPair            = require('./keypair').KeyPair;
-var Account            = require('./account').Account;
-var UInt160            = require('./uint160').UInt160;
+'use strict';
+
+var async = require('async');
+var sjcl = require('./utils').sjcl;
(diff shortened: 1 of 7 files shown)


sublimator:
@emschwartz: Can you look over the message.js JSDoc changes, made to appease the lint gods?

ripplerm:
Account Family is a great feature and should be retained.

The reason very few people were using it, is because there's no convenient way (in current version of ripple-lib) to sign a transaction with private-keys directly. So what we should do is adding some methods into the library so that people could easily use the private-keys.

IMO, we should also introduce more feature for Account Families, 
e.g. hardened-child, hierarchical usage.

sublimator:
Will take that into consideration

clark800:
Also note that I am currently working on just that: a convenient way to sign a transaction with specified private keys.

sublimator:
I guess we can quite easily support a few things:
* Generating a KeyPair from a uint256 encoded in base58
* Generating nth keypair/account from a seed
  * Allow serialising the private key as base58

What we should definitely drop is the silly `seed.get_key(tx_json.Account)` caper where we search for the KeyPair in the seed matching the account. There's some discussion going on currently for the form keypair specifiers should take, and allowing describing them using encoded private keys rather than seeds isn't much of a stretch.

PR #321, opened 2015-04-06
[FIX] Fail if PRNG has not been seeded with at least 256 bits of entr…

clark800:
…opy before generating ECDSA signatures

### DIFF
--- a/src/js/ripple/keypair.js
+++ b/src/js/ripple/keypair.js
@@ -92,8 +92,9 @@ KeyPair.prototype.get_address = function() {
 };
 
 KeyPair.prototype.sign = function(hash) {
+  var PARANOIA_256_BITS = 6; // sjcl constant for ensuring 256 bits of entropy
   hash = UInt256.from_json(hash);
-  var sig = this._secret.sign(hash.to_bits(), 0);
+  var sig = this._secret.sign(hash.to_bits(), PARANOIA_256_BITS);
   sig = this._secret.canonicalizeSignature(sig);
   return this._secret.encodeDER(sig);
 };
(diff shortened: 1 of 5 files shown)


src/js/ripple/transaction.js
```diff
@@ -466,7 +466,7 @@ Transaction.prototype.sign = function() {
   }
 
   var key = seed.get_key(this.tx_json.Account);
-  var sig = key.sign(hash, 0);
+  var sig = key.sign(hash);
```
geertweening: Why not set it to 6 explicitly here as well?

src/js/ripple/keypair.js
```diff
@@ -93,7 +93,7 @@ KeyPair.prototype.get_address = function() {
 
 KeyPair.prototype.sign = function(hash) {
   hash = UInt256.from_json(hash);
-  var sig = this._secret.sign(hash.to_bits(), 0);
+  var sig = this._secret.sign(hash.to_bits(), 6);
```
geertweening: Instead of a `6` in here, a constant somewhere would feel more appropriate

src/js/ripple/transaction.js
clark800: That function only takes one parameter: https://github.com/clark800/ripple-lib/blob/c0849e50b3db09429f3b1688fd8e0fd340433ff8/src/js/ripple/keypair.js#L94

src/js/ripple/keypair.js
clark800: Yeah, better to avoid magic numbers

geertweening:
LGTM

PR #326, opened 2015-04-14
[FIX] Amount: to_text_full fix

darkdarkdragon:
turn off group separator in to_text_full

### DIFF
--- a/src/js/ripple/amount.js
+++ b/src/js/ripple/amount.js
@@ -961,7 +961,7 @@ Amount.prototype.to_text_full = function(opts) {
     return 'NaN';
   }
   return this._is_native
-      ? this.to_human() + '/XRP'
+      ? this.to_human({group_sep: false}) + '/XRP'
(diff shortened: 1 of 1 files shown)


clark800:
Is this a bug? It looks like a behavioral change and I'm not sure who is using this function.

darkdarkdragon:
I've found this used here
https://github.com/ripple/ripple-client/blob/develop/src/js/tabs/trade.controller.js#L864
this, obviously, throws exception in from_json because to_text_full returns value with separators.

And, looking into to_text_full it has inconsistent behavior - for XRP values it returns number with separators, but for others - without.
So I concluded that this is a bug

clark800:
I agree, I think we should merge this. @darkdarkdragon There are 4 unit tests failing; would you like me to close this pull request and fix them in a new pull request?

darkdarkdragon:
@clark800 Yes, I'll make new pull request

PR #327, opened 2015-04-16
[WIP] Add keypair service and ed25519 support

wltsmrz:
Minor cleanup of https://github.com/sublimator/ripple-lib/tree/ed25519


src/js/ripple/base.js
```diff
@@ -1,13 +1,15 @@
...
+var extend = require('extend');
 var sjcl = require('./utils').sjcl;
 var utils = require('./utils');
-var extend = require('extend');
 var convertBase = require('./baseconverter');
 
 var Base = {};
 
 var alphabets = Base.alphabets = {
+                                      // rrrrrrrrrrrrrrrrrrrrrrrrrrrr
```
clark800: This comment doesn't make sense to me

src/js/ripple/base.js
```diff
@@ -145,10 +193,9 @@ Base.decode_check = function(version, input, alphabet) {
 
   // We'll use the version byte to add a leading zero, this ensures JSBN doesn't
   // intrepret the value as a negative number
-  buffer[0] = 0;
+  // buffer[0] = 0;
```
clark800: Should this be deleted?

src/js/ripple/keypairs.js
```diff
@@ -0,0 +1,283 @@
...
+hasCachedProperty(Secp256k1Pair.prototype, 'pubKeyBytes', function() {
+  return hexToArray(this.keyPair.to_hex_pub());
+});
+
+/*
+@param {Array<Number>} message (bytes)
+ */
+Secp256k1Pair.prototype.sign = function(message) {
+  var hash = this.hashMessage(message);
+  var signatureBits = this.keyPair.sign(hash, 0);
```
clark800: keypair.sign only takes one parameter

src/js/ripple/keypairs.js
```diff
@@ -0,0 +1,283 @@
...
+Secp256k1Pair.prototype.hashMessage = function(message) {
+  var bits = sjcl.codec.bytes.toBits(message);
+  return Crypt.hashSha512Half(bits);
+};
+
+/*
+@param {Array<Number>} message (bytes)
+@param {Array<Number>} signature (bytes)
+ */
+Secp256k1Pair.prototype.verify = function(message, signature) {
```
clark800: What are `r` and `s`? Is all this logic specific to Secp256k1?

src/js/ripple/seed.js
```diff
@@ -1,21 +1,26 @@
 'use strict';
 
+/*eslint new-cap: 1*/
+
 //
 // Seed support
 //
 
 var extend = require('extend');
+var lodash = require('lodash');
```
clark800: `var _ = require('lodash')`

wltsmrz: `var lodash = require('lodash');` is better, as it's clear that the utility library is lodash and not underscore (which started this convention as far as I'm aware)

src/js/ripple/seed.js
```diff
@@ -57,104 +62,105 @@ Seed.prototype.parse_passphrase = function(j) {
...
+  var privateGen, publicGen;
   var curve = this._curve;
   var i = 0;
 
   do {
-    private_gen = sjcl.bn.fromBits(
-            firstHalfOfSHA512(append_int(this.to_bytes(), i)));
-    i++;
-  } while (!curve.r.greaterEquals(private_gen));
+    // We hash the seed to extend from 128 bits to 256, looping until we are
```
clark800: I think I saw Nic Dudfield asking about whether this is a secure option and Stefan said he thought it was. But has anyone verified this?

src/js/ripple/seed.js
```diff
@@ -57,104 +62,105 @@ Seed.prototype.parse_passphrase = function(j) {
...
+    var accountIndex = opts.accountIndex || 0;
     do {
-      sec = sjcl.bn.fromBits(
-              firstHalfOfSHA512(
-                  append_int(
-                    append_int(public_gen.toBytesCompressed(), account_number)
-                    ,
-                    i
-                  )));
+      // We hash the root key-pair's public key bytes, along with the account
```
clark800: This seems like a really low-level operation. It would probably take me a while to check if this loop is guaranteed to terminate. Are there any libraries that do something like this?

src/js/ripple/keypairs.js
```diff
@@ -0,0 +1,283 @@
...
+  if (keyType === KeyType.secp256k1) {
+    return new Secp256k1Pair(seed, specifier);
+  } else if (keyType === KeyType.ed25519) {
+    return new Ed25519Pair(seed, specifier);
+  }
+
+  throw new Error('unknown key_type ' + keyType);
+};
+
+exports.keyPairCache = {};
```
clark800: GLOBAL STATE

src/js/ripple/seed.js
```diff
@@ -57,104 +62,105 @@ Seed.prototype.parse_passphrase = function(j) {
   return this;
 };
 
+Seed.prototype.parse_base58 = function(j) {
+  var result = Base.decode_multi(j, Seed.width);
+
+  if (result.error) {
+    this._value = NaN;
+  } else if (lodash.isEqual(result.version, [Base.VER_FAMILY_SEED]) ||
```
clark800: Should it just be `versions` and always an array?

src/js/ripple/base.js
sublimator: Along with:
 ```
// We'll use the version byte to add a leading zero, this ensures JSBN doesn't
// intrepret the value as a negative number
```

File: src/js/ripple/keypairs.js [hunk 10]
```diff
@@ -0,0 +1,283 @@
...
+  var hash = sjcl.hash.ripemd160.hash(sjcl.hash.sha256.hash(bits));
+  var id = UInt160.from_bits(hash);
+  id.set_version(0);
+  return id;
+}
+
+function stringifyKey(specifier) {
+  if (typeof specifier === 'object') {
+    var sorted = {};
+    Object.keys(specifier).forEach(function(k) {
```
sublimator: Need to sort the Object.keys()?

src/js/ripple/keypairs.js
sublimator: This logic seems to be extracting r and s from a DER encoded signature. I don't think ed25519 uses DER encoding for sigs.

src/js/ripple/seed.js
sublimator: It has 2**32 chances to create a hash that lies on the curve. I'm guessing it's not a problem and `i` could be counted on one hand most of the time.

clark800: That doesn't answer the question. Why do we have to do this?

src/js/ripple/seed.js
sublimator: I talked to a math whiz, who said he had the same curiosity tickled, and he said that if you brute force intelligently, it takes only the square root of the bits used for the private keys. So 2**256 only needs 2**128 operations. I don't really understand it, but I do understand it's what rippled does to get the private key from a seed, and can't be changed easily.

src/js/ripple/seed.js
sublimator: That's a question for Dave Schwartz. All I know is that the number (hash) must be less than the order of the curve (probably assumptions made by some of the math), and that an incrementing `i` is one way to get a new hash.

src/js/ripple/seed.js
```diff
@@ -57,104 +62,105 @@ Seed.prototype.parse_passphrase = function(j) {
...
+  return this;
+};
+
 Seed.prototype.to_json = function() {
   if (!(this.is_valid())) {
     return NaN;
   }
 
-  var output = Base.encode_check(Base.VER_FAMILY_SEED, this.to_bytes());
+  var version = this._version; // Base.VER_FAMILY_SEED;
```
sublimator: Remove:  ` // Base.VER_FAMILY_SEED;`

src/js/ripple/seed.js
```diff
@@ -57,104 +62,105 @@ Seed.prototype.parse_passphrase = function(j) {
...
-/*
-* @param account
-*        {undefined}                 take first, default, KeyPair
-*
-*        {Number}                    specifies the account number of the KeyPair
-*                                    desired.
-*
-*        {Uint160} (from_json able), specifies the address matching the KeyPair
-*                                    that is desired.
+Seed.prototype.get_key = function(opts) {
```
sublimator: This should check that the version isn't ed25519, as get_key will return an secp256k1 old style interface (sign takes the hash, rather than the message) keypair

sublimator: We might also want to deprecate this so we can phase out this API, so eventually the only way to get a keypair is through the keypairs service. That would mean removing the keypair.js file, and updating the Secp256k1Pair to no longer delegate.

Given we're breaking it now, throwing an error unless the argument is an object, you could argue that we may as well go ahead now and remove the api entirely.

src/js/ripple/keypairs.js
sublimator: I think we could probably remove the caching from this pull request, as it needs more consideration, for various reason.

geertweening:
needs rebase

src/js/ripple/seed.js
geertweening: @wltsmrz ^?

src/js/ripple/keypairs.js
geertweening: @sublimator @wltsmrz any thoughts on this?

src/js/ripple/seed.js
geertweening: Depending on the who uses this api I'm fine with deprecating it. Up to you @sublimator  and @wltsmrz

src/js/ripple/keypairs.js
wltsmrz: This was discussed elsewhere. The order of object keys is never guaranteed but I think this will work most of the time as-is

src/js/ripple/index.js
```diff
@@ -33,6 +33,11 @@ exports.sjcl = require('./utils').sjcl;
 exports.Wallet = require('ripple-wallet-generator')({sjcl: exports.sjcl});
 exports.types = require('./serializedtypes');
 
+exports.keypairs = require('./keypairs');
+exports.getKeyPair = exports.keypairs.getKeyPair;
```
sublimator: Do we want to just export the keypairs module?

src/js/ripple/keypairs.js
```diff
@@ -0,0 +1,380 @@
...
+    if (key.greaterEquals(1) && curve.r.greaterEquals(key)) {
+      break;
+    } /*else {
+      throw new Error('omg unicorn ;) ');
+    }*/
+  }
+  return key;
+}
+
+/* eslint-disable valid-jsdoc */
```
sublimator: May as well just add the `seed ` param

src/js/ripple/keypairs.js
```diff
@@ -0,0 +1,380 @@
...
+KeyPair.prototype.signHex = function(message) {
+  return bytesToHex(this.sign(message));
+};
+
+/* ---------------------------- SECP256K1 KEYPAIR --------------------------- */
+
+/*
+* @class
+* @private
+* @param {Object} - obj secretKey/publicKey Uint8Array members
```
sublimator: missing `keyPair`

src/js/ripple/keypairs.js
```diff
@@ -0,0 +1,380 @@
...
+
+util.inherits(Secp256k1Pair, KeyPair);
+
+Secp256k1Pair.fromSeed = function(seed) {
+  return new Secp256k1Pair(derivek256Secret(seed));
+};
+
+hasCachedProperty(Secp256k1Pair, 'pubKey', function() {
+  var exponent = this.secretKey._exponent;
+  var EcdsaPublicKey = sjcl.ecc.ecdsa.publicKey;
```
sublimator: new-cap is such a pita rule

src/js/sjcl-custom/sjcl-validecc.js
```diff
@@ -1,21 +1,130 @@
...
+function arrayFill(n, fill) {
+  var arr = new Array(n);
+  for (var i = 0; i < n; i++) {
+    arr[i] = fill;
+  }
+  return arr;
+}
+
+function isZero(bn) {
+  return bn.limbs.length === 1 && bn.limbs[0] === 0;
```
sublimator: This needs checking out.  Why not just `equals(0)` ?

src/js/sjcl-custom/sjcl-validecc.js
```diff
@@ -1,21 +1,130 @@
...
+  return isZero(jac.z) && !isZero(jac.y);
+}
+
+function sha256Mac(k) {
+  var hmac = new sjcl.misc.hmac(k, sjcl.hash.sha256);
+  Array.prototype.slice.call(arguments, 1).forEach(hmac.update.bind(hmac));
+  return hmac.digest();
+}
+
+// https://tools.ietf.org/html/rfc6979#section-3.2
```
sublimator: This was basically ported to sjcl from bitcoinjs: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/ecdsa.js#L77. In the process of adding 6979 to ripple-lib-java, I checked the outputs of this against bouncy castle 1.52, with the k256 curve ( which happens to fully implement the spec, and all the test vectors ) and they produced the same output.

test/keypairs-test.js
```diff
@@ -0,0 +1,277 @@
...
+    SigningPubKey: 'EDD3993CDC6647896C455F136648B7750' +
+                   '723B011475547AF60691AA3D7438E021D',
+    TransactionType: 'Payment',
+    expected_sig: 'C3646313B08EED6AF4392261A31B961F' +
+                  '10C66CB733DB7F6CD9EAB079857834C8' +
+                  'B0334270A2C037E63CDCCC1932E08328' +
+                  '82B7B7066ECD2FAEDEB4A83DF8AE6303'
+  },
+
+  'rfc6979': [
```
sublimator: Yoinked from here: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/test/ecdsa.js#L24

sublimator:
Needs tests for:

```javascript
// setSecret should accept a KeyPair
remote.setSecret(account, getKeyPair(seed));
```

```javascript
// setSecret should accept a KeyPair
transaction.setSecret(application.getCachedKeyPair(account));
```

src/js/ripple/keypairs.js
```diff
@@ -0,0 +1,380 @@
...
+    >>> all(h256(r, i) < order for r in randoms() for i in range(2**12))
+    True
+
+    We basically always break here, but for the vanishingly rare case that we
+    don't we'll increment i and continue. :)
+
+    */
+
+    /* eslint-enable max-len */
+    if (key.greaterEquals(1) && curve.r.greaterEquals(key)) {
```
sublimator: There's a subtle logic bug here (also in the existing seed.js code). We want a value greater than zero, and less than (but not equal to) the order of the curve.  sjcl.bn doesn't provide us with the expressiveness to simply say `key < order`, so we must use the messier logic of `!(key >= order)`, which amounts to `!key.greaterEquals(curve.r)`

src/js/ripple/base.js
```diff
@@ -21,7 +22,13 @@ extend(Base, {
...
   VER_ACCOUNT_PRIVATE: 34,
   VER_FAMILY_GENERATOR: 41,
-  VER_FAMILY_SEED: 33
+  VER_FAMILY_SEED: 33,
+
+  // 6 byte version to give `sEd` prefix to 16 byte seed encodings.
+  // VER_ED25519_SEED: [ 73, 141, 118, 70, 140, 214]
+
+  // 3 byte version to give `sEd` prefix to 16 byte seed encodings.
+  VER_ED25519_SEED: [1, 225, 75]
```
sublimator: This needs stability tests for say all (2 ** 128) + n, for values of n between 0, 255?

src/js/ripple/keypairs.js
clark800: Just put something like this at the top of the file
`/* eslint new-cap: [2, {newIsCapExceptions: ["secretKey", "publicKey"]}] */`

src/js/ripple/base.js
clark800: If it's just a base conversion to base 58, wouldn't it suffice to just check 0 and 2^128-1?

npm-shrinkwrap.json
```diff
@@ -28,6 +28,16 @@
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/ripple-wallet-generator/-/ripple-wallet-generator-1.0.3.tgz"
     },
+    "tweetnacl": {
+      "version": "0.13.1",
+      "from": "tweetnacl@>=0.13.0 <0.14.0",
+      "resolved": "https://registry.npmjs.org/tweetnacl/-/tweetnacl-0.13.1.tgz"
+    },
+    "tweetnacl": {
```
clark800: How did this get listed twice?

PR #328, opened 2015-04-16
Fix error logging on Remote for object-type messages

wltsmrz:


src/js/ripple/remote.js
```diff
@@ -519,16 +519,16 @@ Remote.prototype._handleMessage = function(message, server) {
     message = JSON.parse(message);
 
   /*eslint-disable no-empty*/
-  } catch (e) {
-    // TODO: why are we suppressing this?
-    console.error(e);
-  }
+  } catch (e) { }
```
geertweening: We don't specifically log that the parsing of the message fails. So we know the resulting message is not a valid message, but we don't know that it failed the JSON parsing. I don't really like silent catches

clark800:
We don't want to revert to the empty catch that Nic Dudfield was surprised by, which caused this issue in the first place. What is the goal here? Does `_handleMessage` have to accept both object and string? It would be simpler if it just accepted one type.

src/js/ripple/remote.js
wltsmrz: I don't think it matters to whether JSON parsing fails if the error message is descriptive. I'm sympathetic to only handling object-type messages on Remote

wltsmrz:
> We don't want to revert to the empty catch that Nic [sic] Dudfield was surprised by, which caused this issue in the first place.

I don't know who "we" is, and this issue (whatever it is) seems to have been caused by non-descriptive error messages, not the empty catch. It shouldn't currently matter whether the argument is JSON or a JS object, but if it's a malformed one, the error is unfortunately non-descriptive

> What is the goal here?

To prevent logging unnecessarily in the event of an object-type message, and provide a more descriptive error message

> Does _handleMessage have to accept both object and string? It would be simpler if it just accepted one type.

No. I'm sympathetic to parsing all messages on Server instead. I'm gonna do that now

src/js/ripple/remote.js
```diff
@@ -515,20 +515,13 @@ Remote.prototype.disconnect = function(callback) {
...
-    message = JSON.parse(message);
-
-  /*eslint-disable no-empty*/
-  } catch (e) {
-    // TODO: why are we suppressing this?
-    console.error(e);
-  }
-  /*eslint-enable no-empty*/
-
   if (!Remote.isValidMessage(message)) {
```
clark800: isValidMessage still allows strings

wltsmrz: You sure?

```
> var message = { type: 'response' }
undefined
> ripple.Remote.isValidMessage(message)
true
> ripple.Remote.isValidMessage(JSON.stringify(message))
```

clark800: Oops, I misread this: https://github.com/ripple/ripple-lib/pull/328/files#diff-d6c2439c9363ff2c9f9f89fe2a857808R274

src/js/ripple/log.js
```diff
@@ -149,7 +149,11 @@ var NullLogEngine = {
 if (typeof window !== 'undefined' && typeof console !== 'undefined') {
   Log.engine = InteractiveLogEngine;
 } else if (typeof console !== 'undefined' && console.log) {
-  Log.engine = BasicLogEngine;
+  if (process.env.NODE_ENV === 'test') {
+    Log.engine = NullLogEngine;
```
clark800: My original comment on this got lost: "This seems dangerous. What log messages are you trying to hide?"

wltsmrz: Test logs

clark800: This is going to suppress even error messages. If they are intentional errors in the tests, then maybe it would be safer to switch to the NullLogEngine only for the duration of those specific tests and have the BasicLogEngine reinstalled before each test case. Otherwise it could hide error messages that we weren't expecting.

wltsmrz: Tests should cover error conditions, and ripple-lib shouldn't log with `log.js` during tests

clark800: I agree with the first part, but why do you think we should suppress all error messages? We wouldn't have even noticed the problem that started this task if we were suppressing everything.

wltsmrz: The only real "problem" was vague error messages on Remote._handleMessage, which I did actually notice, and referenced earlier in this pull request (https://github.com/ripple/ripple-lib/issues/307). Ostensibly this pull request (328) regards unnecessary logging during a test for error condition, which caught someone else's attention because the log was poor, non-descriptive

clark800: That doesn't answer the question; I'd still like to know if there is a good argument against only hiding error messages that are expected.

test/server-test.js
```diff
@@ -1,14 +1,19 @@
+/* eslint-disable space-before-blocks, no-multi-spaces, space-in-brackets */
```
clark800: https://www.npmjs.com/package/ripple-format

test/remote-test.js
```diff
@@ -28,13 +34,13 @@ describe('Remote', function() {
     remote = new Remote(options);
   });
 
+  afterEach(function() {
+    Log.setEngine(initialLogEngine);
```
clark800: :+1:

geertweening:
LGTM

geertweening:
needs rebase

PR #331, opened 2015-04-27
Use Deterministic Signatures (RFC 6979)

harryprotist:
Adds `signDeterministic` function to `sjcl.ecc.ecdsa.secretKey`, which uses a deterministically generated ephemeral key instead of the usual random one. 

Uses HMAC for generating the ephemeral key, as described in [RFC 6979](http://tools.ietf.org/html/rfc6979). It uses the implementation of HMAC from sjcl, and whatever hash is passed to the function (default sjcl.hash.sha256).

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "ripple-lib",
-  "version": "0.12.4",
+  "version": "0.12.5-rc1",
   "npm-shrinkwrap-version": "5.3.0",
   "node-version": "v0.10.38",
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ripple-lib",
-  "version": "0.12.4",
+  "version": "0.12.5-rc1",
   "description": "A JavaScript API for interacting with Ripple in Node.js and the browser",
   "files": [
     "src/js/*",


src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,22 @@
+/** Diffie-Hellmann function
+* @param {bitArray} hash hash to sign. 
+* @param {Object} type of hash used for hmac (default sjcl.hash.sha256)
+* @param {boolean} fakeLegacyVersion use old legacy version
+*/
+sjcl.ecc.ecdsa.secretKey.prototype.signDeterministic = function(hash, hash_object, fakeLegacyVersion) {
```
clark800: Remove fakeLegacyVersion

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,22 @@
...
+
+  var R = this._curve.r,
+      l = R.bitLength();
```
clark800: Remove `l` variable, move `R` before usage

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,22 @@
...
+
+  var hmac = new sjcl.misc.hmac(this.get(), hash_object || sjcl.hash.sha256);
```
clark800: var secretKey = this.get();

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,22 @@
...
+  var k = hmac.encrypt(hash);
+  while (sjcl.bitArray.bitLength(k) < this._curveBitLength) {
```
clark800: Extract method `generateK`

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
clark800: Call secretKey.sign to reduce duplication

clark800: Use camelcase

sublimator:
@justmoon You probably wanna look at this

harryprotist:
I applied all the suggestions except for "Call secretKey.sign to reduce duplication", because the `k_for_testing` parameter in secretKey.sign doesn't work

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+/*global sjcl*/
+/*eslint-disable new-cap*/
+
+sjcl.ecc.ecdsa.secretKey.prototype.generateK =
```
clark800: I don't think this needs to be exposed publicly

PR #333, opened 2015-04-30
Use Deterministic Signatures (RFC 6979)

harryprotist:
Adds `signDeterministic` function to `sjcl.ecc.ecdsa.secretKey`, which uses a deterministically generated ephemeral key instead of the usual random one. 

Uses HMAC for generating the ephemeral key, as described in [RFC 6979](http://tools.ietf.org/html/rfc6979). It uses the implementation of HMAC from sjcl, and whatever hash is passed to the function (default sjcl.hash.sha256).

Adds signDeterministic and tests for signDeterministic, modifies recoverablepublic key and tests for it, removes validecc.js and uses the sjcl sign and verify functions instead.


Gulpfile.js
```diff
@@ -47,6 +47,7 @@ var sjclSrc = [
   'src/js/sjcl-custom/sjcl-ecdsa-canonical.js',
   'src/js/sjcl-custom/sjcl-ecdsa-der.js',
   'src/js/sjcl-custom/sjcl-ecdsa-recoverablepublickey.js',
+  'src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js',
```
clark800: Also remove the sjcl-validecc.js file

src/js/sjcl-custom/sjcl-ecdsa-recoverablepublickey.js
```diff
@@ -26,13 +26,17 @@ sjcl.ecc.ecdsa.secretKey.prototype.signWithRecoverablePublicKey = function(hash,
   // Convert hash to bits and determine encoding for output
   var hash_bits;
   if (typeof hash === 'object' && hash.length > 0 && typeof hash[0] === 'number') {
-    hash_bits = hash;
+    hash_bits = sjcl.bitArray.clamp(hash, this._curveBitLength);
```
clark800: Isn't clamping here redundant?

test/sjcl-ecdsa-recoverablepublickey-test.js
```diff
@@ -43,7 +43,7 @@ describe('ECDSA signing with recoverable public key', function(){
 
         var pub_val_point = secret_key._curve.G.mult(secret_key._exponent);
         var public_key = new sjcl.ecc.ecdsa.publicKey(curve, pub_val_point);
-        var hash = hash_function(message);
+        var hash = sjcl.bitArray.clamp(hash_function(message), secret_key._curveBitLength);
```
clark800: I think clamping here is redundant too

test/sjcl-ecdsa-signdeterministic-test.js
```diff
@@ -0,0 +1,85 @@
+'use strict';
+
+/*eslint-disable new-cap*/
+
+var assert = require('assert');
+var sjcl = require('ripple-lib').sjcl;
+
+describe('Deterministic ECDSA (RFC 6979)', function() {
```
clark800: You still haven't actually tested that it works. Please add the test cases provided in the RFC: https://tools.ietf.org/html/rfc6979#appendix-A.2.5

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+/*global sjcl*/
+/*eslint-disable new-cap*/
+
+function generateK(secretKey, hash, hashObject) {
```
clark800: This function doesn't appear to match the algorithm described in http://tools.ietf.org/html/rfc6979#section-3.2

sublimator: Definitely not. Another reference: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/ecdsa.js#L11-L12

sublimator:
Another reference: https://github.com/bcgit/bc-java/blob/master/core/src/test/java/org/bouncycastle/crypto/test/DeterministicDSATest.java#L26

Interesting to note, that there's no test vectors in the RFC that use the curve we use. Passing the tests is valuable of itself of course but do we need/want to support a whole heap of different curves? Things are going to be easier if we just use sha 256 hashing along with the k256 curve, which has bit length of 256 for its order.  I've noticed bitcoin implementations, at least claiming to implement the spec, just skip the looped hash building and clamping.

Maybe we want to impl the spec test vectors that match 256b/256b hash/curve order?

geertweening:
@harryprotist please note and resolve the comments on this PR

test/sjcl-ecdsa-signdeterministic-test.js
```diff
@@ -0,0 +1,162 @@
+'use strict';
+
+/*eslint-disable new-cap*/
+
+var assert = require('assert');
+var sjcl = require('ripple-lib').sjcl;
+
+describe.only('Deterministic ECDSA (RFC 6979)', function() {
```
sublimator: Left the `.only` in!  I downloaded the branch locally and indeed the 3 test cases, involving the rfc curve aren't working. 

Check another comment. The `c256` has (sadly) been monkey patched with `k256` as a deprecated backwards compat hack.

sublimator:
Another thing to note: https://github.com/ripple/ripple-lib/blob/10e11d0b41a9f417b1864f4e367ee5dbe36b8101/src/js/sjcl-custom/sjcl-secp256k1.js#L8-L14

We are patching over some ecc math, which may also effect the results.

sublimator:
I can confirm that unless this (below) is commented out (from `sjcl-secp256k1.js`), we can't even calculate the correct public key listed in the rfc.

```
// This is a custom doubling algorithm that
// only works for a=-3 Jacobian curve. It's much
// faster than the generic implementation
/*sjcl.ecc.pointJac.prototype.doubl = function () {
```

Also, I think the comment is buggy, meaning to say only works for `a=0` (which the secp256k1 curve has, as listed [here](https://en.bitcoin.it/wiki/Secp256k1) and elsewhere)

sublimator:
It seems we can (though whether we should ...) work around it all as so:
```javascript
var neg3Doubl = sjcl.ecc.pointJac.prototype.doubl;
sjcl.ecc.pointJac.prototype.doubl = function () {
  if (!this.curve.a.equals(0)) {
    return neg3Doubl.call(this);
  };
  ...
}

...
sjcl.ecc.curves.nist_p256 = sjcl.ecc.curves.c256;
sjcl.ecc.curves.c256 = sjcl.ecc.curves.k256;
```

You would need to do checks on all a=0 specific curve math to see if it should actually use the default math, and later reference c256 as nist_p256.

src/js/sjcl-custom/sjcl-ecdsa-recoverablepublickey.js
harryprotist: yeah, removed

test/sjcl-ecdsa-recoverablepublickey-test.js
harryprotist: Here it needs to know that the hash will be shortened in order to recover the public key, so the clamp is necessary

harryprotist:
@sublimator I switched the rfcTest and messages loops to forEach, and I removed that .only that I accidentaly left in. I also added your workaround for the nist_p256 curve, although I don't know how best to verify that it works

harryprotist:
@clark800 I addressed the redundant clamping in recoverablepublickey and updated the Gulpfile

sublimator:
> I removed that .only that I accidentaly left in

It snuck back in again!

> workaround for the nist_p256 curve, although I don't know how best to verify that it works

I think you've only over-ridden the one method. It's not going to be a `temporary workaround` if the test vector tests will persist in the code base, along side the speedup math. By gulp concatenation order, your over-ride over-rides the existing over-ride, so when you stash away the function as neg3Double, I believe you're actually stashing away the wrong math (and curve)! Yikes!

I think you'd be better off modifying the existing overrides in `src/js/sjcl-custom/sjcl-secp256k1.js`.That includes *both* methods, `add` and `doubl`.

From there, you can check the curve is correct, and the override is working by looking in the rfc6979 spec and making note of these values:

* q (order of curve)
* x (secret key)
* Ux (public key point x)
* Uy (public key point y)

I believe you can then verify these values against:

* q == sjcl.ecc.curves.p256.r.toString()
* Ux == sjcl.ecc.curves.p256.G.mult(x).x.toString()
* Uy == sjcl.ecc.curves.p256.G.mult(x).y.toString()

Using p256 over nist_p256 might be more congruent with the existing c256 and k256?

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,116 @@
...
+* @param {bitArray} hash hash to sign.
+* @param {Object} hashObject type of hash used for hmac
+*   (default sjcl.hash.sha256)
+* @param {boolean} fakeLegacyVersion use old legacy version
+* @return {bitArray} signature
+*/
+sjcl.ecc.ecdsa.secretKey.prototype.signDeterministic =
+function(hash, hashObject, q) {
+
+  var default_q = (new sjcl.bn(1)).toBits();
```
sublimator: I think you want `this._curve.r.toBits()` as q?

test/sjcl-ecdsa-signdeterministic-test.js
```diff
@@ -0,0 +1,153 @@
...
+
+    it('Should create the right k and s for RFC test #' + index, function() {
+      
+      var hashed_message = test.hash.hash(test.message);
+      var k = rfcPrivate.generateK(hashed_message, test.hash, rfcQ);
+
+      console.log(sjcl.codec.hex.fromBits(k.toBits()));
+      console.log(test.k);
+      assert(sjcl.codec.hex.fromBits(k.toBits()), test.k);
+      var s = rfcPrivate.signDeterministic(hashed_message, test.hash);
```
sublimator: `s` returned here, I think is a bitArray concat of r and s, so you need to destructure it before comparing against `test.s` and `test.r`

test/sjcl-ecdsa-signdeterministic-test.js
```diff
@@ -0,0 +1,153 @@
...
+  rfcTests.forEach(function (test, index) {
+
+    it('Should create the right k and s for RFC test #' + index, function() {
+      
+      var hashed_message = test.hash.hash(test.message);
+      var k = rfcPrivate.generateK(hashed_message, test.hash, rfcQ);
+
+      console.log(sjcl.codec.hex.fromBits(k.toBits()));
+      console.log(test.k);
+      assert(sjcl.codec.hex.fromBits(k.toBits()), test.k);
```
sublimator: You want `assert.equal` here, as assert just checks that the first arg is `truthy` I believe, and the second arg is the message to report when it's not.

test/sjcl-ecdsa-recoverablepublickey-test.js
clark800: Both `signWithRecoverablePublicKey` and `recoverFromSignature` should do clamping at some point, so it shouldn't have to happen here.

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,116 @@
...
+    );
+    hmac = new sjcl.misc.hmac(k, hashObject);
+    v = hmac.encrypt(v);
+
+  } while (sjcl.bn.fromBits(t).greaterEquals(q));
+
+  return sjcl.bn.fromBits(t);
+}
+
+/** Diffie-Hellmann function
```
clark800: This JSDoc doesn't match

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,116 @@
...
+*/
+sjcl.ecc.ecdsa.secretKey.prototype.signDeterministic =
+function(hash, hashObject, q) {
+
+  var default_q = (new sjcl.bn(1)).toBits();
+  while (sjcl.bitArray.bitLength(default_q) < sjcl.bitArray.bitLength(hash) + 1) {
+    default_q = sjcl.bitArray.concat(default_q, (new sjcl.bn(0x00)).toBits());
+  }
+  default_q = sjcl.bn.fromBits(default_q);
+  q = q || default_q;
```
clark800: Would be nice to skip the `default_q` computation if `q` is specified

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,116 @@
...
+sjcl.ecc.ecdsa.secretKey.prototype.generateK =
+function (hash, hashObject, q) {
+
+  hashObject = hashObject || sjcl.hash.sha256;
+  var secret = this.get();
+
+  /* k = 0x00 0x00 0x00 ... 0x00 */
+  var k = [];
+  while (sjcl.bitArray.bitLength(k) < sjcl.bitArray.bitLength(hash)) {
+    k = sjcl.bitArray._shiftRight(k, 8);
```
clark800: We should avoid using private methods starting with `_`. I think you can do this with `sjcl.bn`.

harryprotist:
@sublimator I made the changes that you described and added them as a test, and it's now using the right curve

sublimator:
The tests are showing that the implementation is still broken however. I
haven't looked into why exactly that is yet.

sublimator:
There's some talks of using [elliptic](https://github.com/indutny/elliptic) so I guess this is on hold and not worth commenting on? The tests still seem to fail. Anyway, this needs a rebase, due to source files being moved around, and sjcl now using npm.

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
sublimator: You can probably do it by just doing an array fill of `(qlen+7)/8` bytes, then using `sjcl.codec.bytes.toBits`. Currently this routine (`generateK`) doesn't seem to work. You might have more luck decomposing it into various steps/routines that are testable. You could possibly even implement a class.

bitArray's are a bit funny, with the logical unit being a word, rather than byte. So you need to be careful. If you log out `sjcl.codec.bytes.fromBits(v)` you'll see that the first 3 octets are actually `0x0`, even though there's 32 octets. It's probably the `bitLength(...)` method biting you.

test/sjcl-ecdsa-signdeterministic-test.js
```diff
@@ -0,0 +1,184 @@
...
+  ];
+
+  function testRFC(test, index) {
+    it('Should create the right k and s for RFC test #' + index, function() {
+
+      var hashed_message = test.hash.hash(test.message);
+      //var k = rfcPrivate.generateK(hashed_message, test.hash, rfcQ);
+      var k = rfcPrivate.deterministicGenerateK(rfcPrivate._curve, hashed_message, rfcPrivate, test.hash);
+
+      console.log('generated k: ' + sjcl.codec.hex.fromBits(k.toBits()));
```
sublimator: We usually excise `console.log` before submitting pull requests. Recall also that `assert.equal($actual, $expected)` displays the operands when they aren't equal, so we don't need to explicitly log them :)

test/sjcl-ecdsa-signdeterministic-test.js
```diff
@@ -0,0 +1,184 @@
...
+    var k256_curve = sjcl.ecc.curves.k256;
+
+    var old_signature;
+
+    for (var i = 0; i < 4; i++) {
+
+      var secret_bn = sjcl.bn.fromBits(sjcl.codec.hex.toBits(secret_hex));
+      var secret_key = new sjcl.ecc.ecdsa.secretKey(k256_curve, secret_bn);
+      var hashed_message = hash_function.hash(message);
+
```
sublimator: I'd personally find it easier to read if you just move this block:
```
+      var secret_bn = sjcl.bn.fromBits(sjcl.codec.hex.toBits(secret_hex));
+      var secret_key = new sjcl.ecc.ecdsa.secretKey(k256_curve, secret_bn);
+      var hashed_message = hash_function.hash(message);
```
outside of the loop and just did:
```
var sig1 = secret_key.signDeterministic(hashed_message,  hash_function);
var sig2 = secret_key.signDeterministic(hashed_message,  hash_function);
assert.equal(sig1, sig2);
```

instead of a loop.

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
sublimator: Err, _shiftRight(...) I mean :)

sublimator: Oh right, you want the hash length rather than qlen eh? Rest of comment still applies

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,87 @@
+'use strict';
+/*global sjcl*/
+/*eslint-disable new-cap*/
```
clark800: Preferred disabling method is `/* eslint new-cap: [2, {newIsCapExceptions: ["bn"]}] */`

harryprotist:
All the tests that I put in pass now

src/js/sjcl-custom/sjcl-secp256k1.js
```diff
@@ -53,7 +53,13 @@ sjcl.ecc.pointJac.prototype.add = function(T) {
 // This is a custom doubling algorithm that
 // only works for a=-3 Jacobian curve. It's much
 // faster than the generic implementation
+var neg3Doubl = sjcl.ecc.pointJac.prototype.doubl;
 sjcl.ecc.pointJac.prototype.doubl = function () {
+
+  if (!this.curve.a.equals(0)) {
+    return neg3Doubl.call(this);
```
clark800: What if `a` is neither 0 nor 3?

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,126 @@
...
+      v = hmac(v, k);
+      T = sjcl.bn.fromBits(sjcl.bitArray.concat(T.toBits(), v));
+    };
+    T = bits2int(T.toBits());
+  }
+
+  return T;
+}
+
+/** Diffie-Hellmann function with deterministic k
```
clark800: I don't think this is a Diffie-Hellman function

clark800:
I'm getting a stack trace when I try to test your branch, do you have all your changes committed?

harryprotist:
i started getting that too after I merged with your change from develop that removes linter errors. Was that present in that commit, or did I just mess up the merge?

clark800:
Probably a merge issue, looks like you have to remove the require of sjcl-validecc.js from sjcl-custom/index.js

harryprotist:
looks like I did have to make that change, but that wasn't what was causing the stack trace

harryprotist:
I had to do a new `npm install`, but that fixed it. Should I change anything in the repository for that, if the npm stuff is untracked?

clark800:
Shouldn't have to commit anything for that.

src/js/sjcl-custom/sjcl-secp256k1.js
sublimator: Good point, it's not -3 only math that's stashed away, so `neg3Doubl` is a misnomer. 

You could update the errant comments while you're here too.

sublimator:
There's some failing tests due to `sjcl.ecc.ecdsa.secretKey.prototype.signWithRecoverablePublicKey` passing along a `k_for_testing` that no longer works to  `sjcl.ecc.ecdsa.secretKey.prototype.sign` method, which is no longer being over-ridden, and is now demanding k_for_testing as an sjcl.bn object. Some of the tests pass it in as hex.

There's also a bunch of failing interest rate calculation tests. Possibly the merge went awry?

test/sjcl-ecdsa-recoverablepublickey-test.js
sublimator: See `sjcl-ecdsa-recoverablepublickey-test.js`:
```
        var hash = sjcl.bitArray.clamp(hash_function(message), secret_key._curveBitLength);

        var recoverable_signature = secret_key.signWithRecoverablePublicKey(hash, 0, random_value);
        var recovered_public_key = sjcl.ecc.ecdsa.publicKey.recoverFromSignature(hash, recoverable_signature);
```

The change looks like it was done because it's also done in the sjcl version of sign which we now use: 
```
  sign: function(hash, paranoia, fakeLegacyVersion, fixedKForTesting) {
    if (sjcl.bitArray.bitLength(hash) > this._curveBitLength) {
      hash = sjcl.bitArray.clamp(hash, this._curveBitLength);
    }
```

I'm not sure why the tests were using sha512 before, or what the ramifications of using it without clamping are ? But without it, the tests which use sha512 will fail.

src/js/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,126 @@
...
+        sjcl.codec.hex.toBits(repeat('0', Math.ceil((qlen - blen) / 4))),
+        curve.r.toBits()
+      ));
+    }
+    return sjcl.bn.fromBits(bits);
+  }
+  function int2octets(integer) {
+    var iModQ = integer.mulmod(new sjcl.bn(1), curve.r);
+
+    var qlen = sjcl.bitArray.bitLength(curve.r.toBits());
```
sublimator: Should calculate qlen just once :)

test/sjcl-ecdsa-recoverablepublickey-test.js
clark800: I noticed that the blobvault messaging code uses sha512. The blobvault code (message.js and others) lives in ripple-lib even though it isn't required by anything in ripple-lib.

sublimator: > blobvault messaging code uses sha512.
Ah, I glossed over that fact. 

The custom valid-ecc sign/verify code didn't clamp the hash to qlen, where as the sjcl implementations do. Old signatures created before would no longer verify with this change . Hrmmm.

> lives in ripple-lib 
You're proposing extracting it into a separate module?

clark800: Yeah, I created a ticket for extracting it a few weeks ago: https://ripplelabs.atlassian.net/browse/RLJS-358

harryprotist:
Fixed recoverablepublickey. The failing tests for interest calculations happen in ripple/ripple-lib's develop branch at the point that my fork is at

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,119 @@
...
+  /* Utility functions */
+  /* used to generate k and v */
+  function repeat(str, times) {
+    return (new Array(times + 1)).join(str);
+  }
+  function bits2int(bits) {
+
+    let blen = sjcl.bitArray.bitLength(bits);
+
+    if (blen > qlen) {
```
sublimator: This seems confusing, as `blen > qlen` in the first branch, means pretty much `qlen < blen` in the second.

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,119 @@
...
+    } else if (qlen < blen) {
+      return sjcl.bn.fromBits(sjcl.bitArray.concat(
+        sjcl.codec.hex.toBits(repeat('0', Math.ceil((qlen - blen) / 4))),
+        curve.r.toBits()
+      ));
+    }
+    return sjcl.bn.fromBits(bits);
+  }
+  function int2octets(integer) {
+    let iModQ = integer.mulmod(new sjcl.bn(1), curve.r);
```
sublimator: Is there any reason to use mulmod with a multiplication factor of 1, over just using `.mod(curve.r)` ?

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,119 @@
+'use strict';
+let sjcl = require('sjcl');
+/* eslint-disable new-cap */
+
+sjcl.ecc.ecdsa.secretKey.prototype.generateK =
+function(hash, hashObject) {
+
+  let curve = this._curve;
```
sublimator: You should use `const` if the reference will never point to something other than what it is declared at initially. `curve` here is never reassigned so can be const.

However, this code wants to be pushed to sjcl at some point, which doesn't use es6, so you'll want to replace all let/const with var, and use whatever `eslint-disable` directives are required to pacify the linter.

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,119 @@
...
+    k
+  );
+
+  v = hmac(v, k);
+  v = hmac(v, k);
+
+  let T = sjcl.bn.fromBits(v);
+
+  while (
+    sjcl.bitArray.bitLength(T.toBits()) < qlen
```
sublimator: I think the sjcl.bn objects have a `bitLength` method. i.e. you can replace `sjcl.bitArray.bitLength(T.toBits())` with `T.bitLength()`

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+  function bits2int(bits) {
+
+    let blen = sjcl.bitArray.bitLength(bits);
+
+    if (blen > qlen) {
+      return sjcl.bn.fromBits(sjcl.bitArray.bitSlice(bits, 0, qlen));
+    } else if (qlen < blen) {
+      return sjcl.bn.fromBits(sjcl.bitArray.concat(
+        sjcl.codec.hex.toBits(repeat('0', Math.ceil((qlen - blen) / 4))),
```
sublimator: This branch will never execute, but also doesn't use the `bits` at all ?

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,119 @@
...
+      sjcl.codec.hex.toBits(repeat('0', Math.ceil((rlen - ilen) / 4))),
+      iModQ.toBits()
+    );
+  }
+  function bits2octets(bits) {
+    return int2octets(bits2int(bits).mulmod(new sjcl.bn(1), curve.r));
+  }
+
+  function hmac(bits, key) {
+    let hmacK = new sjcl.misc.hmac(key, hashObject);
```
sublimator: You could make this take the `key` as the first parameter, and then use arguments, and take care of the `sjcl.bitArray.concat` dance which would leave the calling code flatter, and easier to read:

```
k = hmac(
 sjcl.bitArray.concat(
   sjcl.bitArray.concat(v, sjcl.codec.hex.toBits('00')),
   sjcl.bitArray.concat(int2octets(x), bits2octets(hash))
 ),
 k
);
```

becomes 

```
k = hmac(k, 
         v, 
         sjcl.codec.hex.toBits('00'), 
         int2octets(x), 
         bits2octets(hash));
```

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  return T;
+};
+
+/**
+* @param {bitArray} hash hash to sign.
+* @param {Object} hashObject type of hash used for hmac
+*   (default sjcl.hash.sha256)
+* @param {sjcl.bn} q set the curve order
```
sublimator: `@param {sjcl.bn} q set the curve order`

Function no longer takes `q`.  I thought the linter was looking out for these things?

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
sublimator: You may notice that it's hard to directly test these units cause they aren't exposed and get state from a closure.

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
clark800: Probably because `mod` is undocumented: https://github.com/bitwiseshiftleft/sjcl/issues/222

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
clark800: I think it's fine to leave the let/const in and just replace with var when preparing the pull request to sjcl.

sublimator: Jesus said to purge bad thoughts

src/sjcl-custom/sjcl-ecdsa-signdeterministic.js
clark800: I just realized that using sjcl's bitLength might be a problem due to this bug I found in sjcl recently: https://github.com/bitwiseshiftleft/sjcl/pull/221

I tested and confirmed that it's possible that `a.bitLength() != b.bitLength()` even if `a.equals(b)`. And they can be different by an arbitrarily large amount. Just try using the `power` method to create numbers with lots of leading zeros. Also, `toBits` does not remove leading zeros either.

For a hash, it makes sense to keep leading zeros, but for arithmetic it doesn't make much sense.

sublimator:
@harryprotist  

Please open this back up @ https://github.com/ripple/sjcl-extended/ :)

Thanks.

PR #334, opened 2015-05-01
Remove config singleton (global state)

clark800:
This is a breaking change, but we must eliminate all global state.


wltsmrz:
LGTM

PR #338, opened 2015-05-04
Fix transaction summary for transactions that fail format validation

wltsmrz:


geertweening:
@wltsmrz can you add a test for this behavior?

wltsmrz:
Yep

src/transaction.js
```diff
@@ -1221,21 +1222,24 @@ Transaction.prototype.summary = function() {
     initialSubmitIndex: this.initialSubmitIndex,
     lastLedgerSequence: this.lastLedgerSequence,
     state: this.state,
-    server: this._server ? this._server._opts.url : undefined,
```
sublimator: Why did we remove this? Cause there can be more than one server per transaction?

wltsmrz: Yep, transactions are submitted to all servers now. I think this would always be undefined currently, just removed it instead

test/transaction-manager-test.js
```diff
@@ -630,6 +675,68 @@ describe('TransactionManager', function() {
...
+    transaction.once('submitted', function(m) {
+      assert.strictEqual(m.error, 'remoteError');
+      receivedSubmitted = true;
+    });
+
+    rippled.on('request_submit', function(m, req) {
+      assert.strictEqual(transactionManager.getPending().length(), 1);
+      assert.strictEqual(m.tx_blob, SerializedObject.from_json(
+        transaction.tx_json).to_hex());
+      req.sendResponse(SUBMIT_REMOTE_ERROR, {id: m.id});
```
sublimator: Where about it's this usually thrown from?

test/transaction-manager-test.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable max-len */
+/* eslint-disable comma-spacing */
```
clark800: This package should handle comma-spacing automatically: https://www.npmjs.com/package/ripple-format

src/transaction.js
```diff
@@ -1221,21 +1223,24 @@ Transaction.prototype.summary = function() {
...
       engine_result: this.result.engine_result,
       engine_result_message: this.result.engine_result_message,
       ledger_hash: this.result.ledger_hash,
       ledger_index: this.result.ledger_index,
-      transaction_hash: this.result.tx_json.hash
+      transaction_hash: undefined
     };
+
+    if (this.result.tx_json) {
+      txSummary.result.transaction_hash = this.result.tx_json.hash;
```
clark800: Could avoid mutation by declaring a variable for `transaction_hash` before `txSummary.result = ...`, or use a ternary expression in the `txSummary.result = ...` assignment.

geertweening:
Travis fails on PR with bunch of lint errors

geertweening:
besides @clark800 's comment, LGTM

clark800:
LGTM

PR #340, opened 2015-05-11
Remove return value of mergeDirectAndAutobridgedBooks

clark800:
Since the function has a side effect, it's better to not return a value.

http://en.wikipedia.org/wiki/Command%E2%80%93query_separation


geertweening:
LGTM

PR #348, opened 2015-05-20
Prevent crash when listening for "model" events on the OrderBook clas…

clark800:
…s and an issuer creates an order with one of their issues


nbougalis:
I think that this is correct, based on my understand of how `ripple-lib` implements autobridging.

geertweening:
LGTM

PR #354, opened 2015-05-28
Move ripple-rest/api into src/api, exposing RippleAPI

clark800:


src/api/common/schemas/Currency.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Currency",
+  "description": "The three-character code or hex string used to denote currencies",
+  "type": "string",
+  "pattern": "^([a-zA-Z0-9]{3}|[A-Fa-f0-9]{40})$"
```
wltsmrz: https://github.com/ripple/rippled/blob/6675ee7f5cb52a898c63f99b7b232b6d1847f9e0/src/ripple/protocol/impl/UintTypes.cpp#L37-L43

npm-shrinkwrap.json
```diff
@@ -26,6 +26,20 @@
       "version": "1.2.1",
       "resolved": "https://registry.npmjs.org/extend/-/extend-1.2.1.tgz"
     },
+    "jayschema": {
```
wltsmrz: https://github.com/ebdrup/json-schema-benchmark

wltsmrz:
General approach LGTM

npm-shrinkwrap.json
clark800: Yeah, looks like we should switch to `is-my-json-valid`, it has 100K downloads in the last day on npm. I guess I'll bundle that with my next pull request since it's for the schemas.

src/api/common/schemas/Currency.json
wltsmrz: https://github.com/ripple/ripple-lib/blob/16e3541a10350f12336a5d1bf621b63f92cf2737/src/currency.js#L65

PR #355, opened 2015-06-02
Validate UInt* hex input

wltsmrz:


src/core/uint160.js
```diff
@@ -74,23 +74,26 @@ UInt160.prototype.parse_generic = function(j) {
 };
 
 // XXX Json form should allow 0 and 1, C++ doesn't currently allow it.
-UInt160.prototype.to_json = function(opts) {
-  opts = opts || {};
+UInt160.prototype.to_json = function(opts_) {
+  let opts = opts_ || {};
```
clark800: Default param can be used to eliminate this line

src/core/uint.js
```diff
@@ -130,34 +144,40 @@ UInt.prototype._update = function() {
...
       break;
 
     default:
-        if (typeof j !== 'string') {
-          this._value = NaN;
-        } else if (this.constructor.width === j.length) {
-          var hex = utils.arrayToHex(utils.stringToArray(j));
+      switch (String(j).length) {
+        case subclass.width:
+          let hex = utils.arrayToHex(utils.stringToArray(j));
```
clark800: Looks like most of the `let`s should be `const`s

src/core/uint.js
```diff
@@ -130,34 +144,40 @@ UInt.prototype._update = function() {
...
           this._value = new sjcl.bn(hex, 16);
-        } else if ((this.constructor.width * 2) === j.length) {
-          // XXX Check char set!
-          this._value = new sjcl.bn(j, 16);
-        } else {
+          break;
+        case subclass.width * 2:
+          if (this.checkHexInput(j)) {
+            this._value = new sjcl.bn(j, 16);
+          }
```
clark800: Should there be `this._value = NaN` here?

test/uint-test.js
```diff
@@ -1,23 +1,55 @@
...
+function makeTests(uIntType) {
+  describe(uIntType, function() {
+    let rippleType = ripple[uIntType];
+    let tests = fixtures[uIntType];
+
+    assert(rippleType && tests);
+
+    lodash.each(tests, (test, key) => {
+      // Example: from_json().to_bytes()
+      let [, inputMethod, outputMethod] = /^(\w+)\(\)\.(\w+)\(\)$/.exec(key);
```
clark800: You've eliminated a lot of redundancy here, which is good, but it's a bit confusing that the code for the tests lives in the fixture.

wltsmrz: Agreed, addressing this now

src/core/uint160.js
wltsmrz: Thanks for the tip

src/core/uint.js
wltsmrz: Addressing

src/core/uint.js
wltsmrz: Updated with tests for parse_generic, incoming

src/core/uint.js
```diff
@@ -166,7 +186,7 @@ UInt.prototype.parse_generic = function(j) {
 };
 
 UInt.prototype.parse_hex = function(j) {
-  if (typeof j === 'string' && j.length === (this.constructor.width * 2)) {
+  if (new RegExp(`^[0-9A-F]{${this.constructor.width * 2}}$`).test(j)) {
```
sublimator: Shouldn't hex also include lower case `a-f` ? That would be a reasonable expectation?

src/core/uint160.js
```diff
@@ -1,16 +1,16 @@
...
-var Base = require('./base').Base;
+const UInt = require('./uint').UInt;
+const Base = require('./base').Base;
 
 //
 // UInt160 support
 //
 
-var UInt160 = extend(function() {
+let UInt160 = extend(function() {
```
sublimator: This can be const

src/core/uint160.js
```diff
@@ -20,8 +20,8 @@ UInt160.width = 20;
 UInt160.prototype = extend({}, UInt.prototype);
 UInt160.prototype.constructor = UInt160;
 
-var HEX_ZERO = UInt160.HEX_ZERO = '0000000000000000000000000000000000000000';
-var HEX_ONE = UInt160.HEX_ONE = '0000000000000000000000000000000000000001';
+let HEX_ZERO = UInt160.HEX_ZERO = '0000000000000000000000000000000000000000';
```
sublimator: These are `const`ants right?

src/core/uint.js
wltsmrz: The idea here is that one can rely on `uint*.is_valid()` to check that input will be reflected by rippled. I think it's a bad idea to allow setting an `InvoiceID` in lowercase that will be returned in uppercase, e.g., and I think `UInt256.is_valid()` is a good place to validate.

wltsmrz: Well, ideally, rippled would reject lowercase or otherwise invalid InvoiceIDs, or InvoiceIDs that certainly won't match, but ripple-lib should also check, no?

sublimator: > I think UInt256.is_valid() is a good place to validate.

Not sure I agree, as UInt256 != InvoiceID, and InvoiceID is not a VL variable length (string) field. It only takes hex, not arbitrary characters, and that hex defines a number. ripple-lib even pads the hex strings passed along to a full 64 nibble string. 

Hrmmm.

wltsmrz: It does pad now. I figure that should be fixed too

sublimator: Still. They are numbers, not strings.

test/uint-test.js
```diff
@@ -1,23 +1,113 @@
...
+  return `Expected ${type(test.input)}: ${test.input} to yield ${type(test.expected)}: ${test.expected === 'null' ? NaN : test.expected}. Actual: ${type(result)}: ${result}`;
+}
+
+function makeTests(uIntType) {
+  describe(uIntType, function() {
+    let rippleType = ripple[uIntType];
+    let tests = fixtures[uIntType];
+
+    it('from_json().to_json()', function() {
+      tests['from_json().to_json()'].forEach(function(test) {
```
clark800: This is much more clear than before :+1:

test/uint-test.js
```diff
@@ -1,23 +1,113 @@
...
+        assert.strictEqual(result.is_valid(), String(test.expected) !== 'null', `Validity check failed: ${test.input}`);
+        assert.strictEqual(result.to_hex(), test.expected, resultError(test, result));
       });
-      it('should create 000000000000000000000000FFFFFFFF when called with 0xFFFFFFFF', function () {
-        var val = UInt128.from_number(0xFFFFFFFF);
-        assert.strictEqual(val.to_hex(), '000000000000000000000000FFFFFFFF');
+    });
+    it('from_generic().to_*()', function() {
+      [
+      ['to_json', ...tests['from_json().to_json()']],
```
clark800: Looks like the names in the fixtures should be "from_generic().to_*()"? I think it would be a bit simpler if this array was just `['to_json', 'to_bytes', 'to_hex']` and the test cases were looked up inside the loop.

sublimator:
The flow issue is something unrelated to this yeah ?

clark800:
All the errors that mention `gulp-flowtype` are due to the stale cache because the project doesn't use that module anymore. The other errors are because you still haven't rebased.

sublimator:
lgtm

PR #356, opened 2015-06-04
Remove sjcl-custom in favor of sjcl-extended module in separate repo

sublimator:
The next step would be to remove message.js into a separate repo that depended on sjcl-extended. From there, we have no real dependency on sjcl, and can move to elliptic/bn.js/hash.js quite easily. This also makes sense as it's es5 code, because ultimately, if we do maintain the sjcl-extended/custom code, we'd want it pushed into sjcl core.  It's likely as message.js will eventually require ed25519 support, and we'll likely be using elliptic for that, that we'll port the signatures that include the public key recovery factor to elliptic. In the meantime though, this is easy and lowers risk, as we aren't messing with the code.

See: https://github.com/ripple/sjcl-extended
We could move the #333 code into that too.


sublimator:
Please review:
@clark800 
@wltsmrz

package.json
```diff
@@ -25,7 +25,7 @@
     "ripple-lib-transactionparser": "^0.3.2",
     "ripple-wallet-generator": "^1.0.3",
     "simple-asyncify": "^0.1.0",
-    "sjcl": "^1.0.3",
+    "sjcl-extended": "ripple/sjcl-extended#1.0.3",
```
clark800: It's funny that npm has a built-in syntax just for GitHub: https://docs.npmjs.com/files/package.json#github-urls

clark800:
LGTM, seems like a good first step

PR #357, opened 2015-06-04
Update schemas, use schemas for validation, and switch to is-my-json-valid

clark800:


src/api/common/errors.js
```diff
@@ -10,19 +10,13 @@ function RippleError(message) {
...
- */
-function InvalidRequestError(message) {
+function ValidationError(message) {
   this.message = message;
 }
-InvalidRequestError.prototype = new RippleError();
-InvalidRequestError.prototype.name = 'InvalidRequestError';
-InvalidRequestError.prototype.error = 'restINVALID_PARAMETER';
+ValidationError.prototype = new RippleError();
+ValidationError.prototype.name = 'ValidationError';
```
mDuo13: This is potentially confusing, since the term "validation" also refers to the approval of ledgers in the consensus process.

lumberj: I don't think it's confusing, since ripple-lib is not doing the validation. Also, is there a validation error during the consensus process?

lumberj: Also, the rippled errors are clearly labeled (I think)

src/api/common/schemas/amount.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "amount",
+  "description": "An Amount on the Ripple Protocol, used also for XRP in the ripple-rest API",
+  "allOf": [
+    {"$ref": "amountbase"},
```
lumberj: Nice

src/api/common/schemas/amountbase.json
```diff
@@ -0,0 +1,44 @@
...
+      "description": "The quantity of the currency, denoted as a string to retain floating point precision",
+      "$ref": "value"
+    },
+    "currency": {
+      "description": "The three-character code or hex string used to denote currencies",
+      "$ref": "currency"
+    },
+    "counterparty": {
+      "description": "The Ripple account address of the currency's issuer or gateway, or an empty string if the currency is XRP",
+      "$ref": "address"
```
lumberj: The address schema doesn't allow empty strings. It might be better to use:

```json
"anyOf": [
  { "$ref": "address" },
  {"type": "string", "description": "Empty string",  "pattern": "^$" }
]
```

clark800: I think we want to disallow counterparty being empty, it can be not specified though in the case of XRP.

src/api/common/schemas/amountbase.json
```diff
@@ -0,0 +1,44 @@
...
+      "required": ["counterparty"]
+    },
+    {
+      "properties": {
+        "currency": {
+          "enum": ["XRP"]
+        }
+      },
+      "not": {
+        "required": ["counterparty"]
```
lumberj: This works if we're allowing empty string OR undefined

src/api/common/schemas/blob.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "blob",
+  "description": "An uppercase hex string representation of a transaction",
+  "type": "string",
+  "minLength": "1",
+  "pattern": "^[0-9A-F]*$"
```
lumberj: The hex is 256bits right? Shouldn't this always be 64 in length?

src/api/common/schemas/instructions.json
```diff
@@ -0,0 +1,44 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "instructions",
+  "description": "Instructions for executing a transaction",
+  "type": "object",
+  "properties": {
+    "sequence": {
+      "description": "The sequence number, relative to the initiating account, of this transaction.",
+      "type": "integer",
+      "minimum": 0
```
lumberj: Ledgers and sequence numbers all start from 0?

src/api/common/schemas/blob.json
clark800: This schema is for a transcation blob, which is arbitrary length

lumberj: Oh yea, missed that woops

src/api/common/schemas/amountbase.json
lumberj: The description says, "or an empty string if the currency is XRP" -- the schema just says it's not required, and additional properties are false. Doesn't that mean an empty string should fail validation when XRP is the currency? Or does it work anyway, since empty string is falsy in JS?

src/api/common/schemas/instructions.json
lumberj: Confirmed that account sequence starts at 1

src/api/common/schemas/natural.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "natural",
+  "description": "A non-negative integer",
+  "type": "integer",
+  "minimum": 0
```
lumberj: I think the first ledger sequence is 1, so, really we want positive integers right?

src/api/ledger/trustlines.js
```diff
@@ -56,7 +53,7 @@ function getTrustLines(account, options, callback) {
         ledger = prevResult.ledger_index;
       } else {
         marker = options.marker;
-        limit = validator.isValid(options.limit, 'UINT32')
+        limit = parseInt(options.limit, 10) === options.limit
           ? Number(options.limit) : DefaultPageLimit;
```
lumberj: Using both `Number` and `parseInt` to parse the limit string. Maybe stick to one?

lumberj:
Aside from comments, LGTM

src/api/common/errors.js
wltsmrz: This caught my attention too; I agree with @mDuo13.

clark800: @mDuo13 @wltsmrz Any suggestions for a better name?

src/api/common/schemas/amountbase.json
clark800: Yeah, empty string for counterparty should fail validation when XRP is currency. I'll fix the description.

src/api/common/schemas/amountbase.json
clark800: I just tested it and it works as intended: validation fails if currency is XRP and counterparty is empty string.

src/api/ledger/trustlines.js
clark800: I'll get rid of this entirely by using the convention that no limit specified means "all"

src/api/common/schemas/natural.json
clark800: First ledger sequence is 0 according to the wiki https://wiki.ripple.com/Genesis_ledger

src/api/common/schemas/instructions.json
clark800: Fixing

src/api/common/errors.js
clark800: Sticking with ValidationError for lack of a better option

src/api/common/schemas/natural.json
sublimator: I think the wiki is wrong:

```
Ledger::Ledger (RippleAddress const& masterID, std::uint64_t startAmount)
    : mTotCoins (startAmount)
    , mLedgerSeq (1) // First Ledger
```

src/api/common/schemas/cancellation.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "cancellation",
+  "type": "object",
+  "properties": {
+    "address": {"$ref": "address"},
+    "orderSequence": {
+      "type": "integer",
+      "minimum": 0
```
sublimator: Would be 1, as the Sequence in a new AccountRoot is 1

clark800: Ok, will fix in next pull request.

PR #358, opened 2015-06-05
Add flow type checking

clark800:
esprima-fb is deprecated, but it seems to work better than babel-eslint currently. We can switch to babel-eslint later.


hserang:
LGTM! Thank you for walking me through the changes.

PR #359, opened 2015-06-06
[FIX] fix AutobridgeCalculator (RT-3445)

darkdarkdragon:
pass issuers to AutobridgeCalculator so
it can create offers with right issuers


clark800:
Lint checks are failing - we switched to babel, so we have a rule that "var" has to be replaced with "const" or "let". Would you mind making the replacements?

darkdarkdragon:
ok, will do it

darkdarkdragon:
@clark800 done

clark800:
LGTM

PR #365, opened 2015-06-10
[WIP] Delegate Base codec duties to base-encoder module

sublimator:
Would need to find [base-encoder](https://github.com/sublimator/base-encoder) a home somewhere before merging, but throwing this out there for comment.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -9,12 +9,14 @@
       "resolved": "https://registry.npmjs.org/async/-/async-0.9.2.tgz"
     },
     "babel-runtime": {
-      "version": "5.4.7",
-      "resolved": "https://registry.npmjs.org/babel-runtime/-/babel-runtime-5.4.7.tgz",
+      "version": "5.5.6",
+      "from": "babel-runtime@>=5.5.4 <6.0.0",
+      "resolved": "https://registry.npmjs.org/babel-runtime/-/babel-runtime-5.5.6.tgz",
       "dependencies": {
         "core-js": {
-          "version": "0.9.13",
-          "resolved": "https://registry.npmjs.org/core-js/-/core-js-0.9.13.tgz"
+          "version": "0.9.15",
+          "from": "core-js@>=0.9.0 <0.10.0",
+          "resolved": "https://registry.npmjs.org/core-js/-/core-js-0.9.15.tgz"
         }
       }
     },
@@ -83,7 +85,7 @@
     "sjcl-extended": {
       "version": "1.0.3",
       "from": "ripple/sjcl-extended#1.0.3",
-      "resolved": "git://github.com/ripple/sjcl-extended.git#908218e0db4fccd08c4e3f1f7269b72f8d93c270",
+      "resolved": "git://github.com/ripple/sjcl-extended.git#d8cf8b22e7d97193c54e1f65113e3edcf200ca17",
       "dependencies": {
         "sjcl": {
           "version": "1.0.3",
@@ -131,6 +133,17 @@
               "resolved": "https://registry.npmjs.org/nan/-/nan-1.8.4.tgz"
             }
           }
+        }
+      }
+    },
+    "x-address-codec": {
+      "version": "0.1.2",
+      "from": "x-address-codec@*",
+      "dependencies": {
+        "base-x": {
+          "version": "1.0.1",
+          "from": "base-x@>=1.0.1 <2.0.0",
+          "resolved": "https://registry.npmjs.org/base-x/-/base-x-1.0.1.tgz"
         }
       }
     }
--- a/package.json
+++ b/package.json
@@ -26,7 +26,8 @@
     "ripple-wallet-generator": "^1.0.3",
     "simple-asyncify": "^0.1.0",
     "sjcl-extended": "ripple/sjcl-extended#1.0.3",
-    "ws": "~0.7.1"
+    "ws": "~0.7.1",
+    "x-address-codec": "^0.1.2"
   },
   "devDependencies": {
     "assert-diff": "^1.0.1",
--- a/src/core/base.js
+++ b/src/core/base.js
@@ -1,17 +1,19 @@
 'use strict';
-const _ = require('lodash');
+
 const sjcl = require('./utils').sjcl;
-const utils = require('./utils');
+const sha256 = require('./utils').sha256;
 const extend = require('extend');
-const convertBase = require('./baseconverter');
 
 const Base = {};
 
 const alphabets = Base.alphabets = {
+  bitcoin: '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz',
   ripple: 'rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz',
-  tipple: 'RPShNAF39wBUDnEGHJKLM4pQrsT7VWXYZ2bcdeCg65jkm8ofqi1tuvaxyz',
-  bitcoin: '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'
+  tipple: 'RPShNAF39wBUDnEGHJKLM4pQrsT7VWXYZ2bcdeCg65jkm8ofqi1tuvaxyz'
 };
+
+const codecOptions = {sha256, alphabets, defaultAlphabet: 'ripple'};
+const {encode, decode} = require('x-address-codec')(codecOptions);
 
 extend(Base, {
   VER_NONE: 1,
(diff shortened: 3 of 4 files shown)


clark800:
This is not quite an ideal interface since you are passing in sha256 and bnFactory, which are not semantically necessary parameters.

My recommendation would be to change the name to `bitcoin-address-codec` and have the interface be `address = encode(bytes, alphabet)` and `bytes = decode(address, alphabet)` where `alphabet` defaults to the bitcoin alphabet if not specified. This package could have its own dependencies for sha256 and ripemd160. I think it would be simpler to just export two functions rather than a class.

sublimator:
Yeah, I already killed off bnFactory, as it felt super naff. Sha256 on the
other hand is valid DI wiring. You could just as well export an Encoder
factory if constructor injection disturbs peoples sense of semantics.  I
just didn't want it to depend on a particular hash256 implementation.

ripemd160 isn't actually required for encoding. It's used to hash the
public key for creating address byes. But for say a seed?

Your saying it makes more sense to slice it up along natural dependency
lines? Will ponder. Need to catch a plane.

sublimator:
Agree with using bitcoin in the name, as it would be more discoverable. However, for 
bitcoin style  identifier encodings, the essence of which is versions and
checksums, and 0 padding, this seems like the logical unit to me.

Note there already exists bs58/b58checked modules. They do have nice
encode/decode apis, but they aren't very flexible, as they hardcode a certain
alphabet, and assume implementations of sha256, and `Buffer` availability
etc.

See:

https://github.com/cryptocoinjs/bs58/blob/master/lib/bs58.js
https://github.com/bitcoinjs/bs58check

There's no reason modules like the above, with cute apis, couldn't be built upon
a more flexible base.

To be honest, I don't know how much is gained from having that bs58 module,
separated from bs58check which is bitcoin only (all bitcoin `identifier`
decoding requires checksum checking) other than allowing high level libs to
wire together chosen hash lib etc. But why not just formalize it with DI and do
the whole thing?

I agree that constructor injection was probably the wrong choice here, and it
should be at the module or class constructor level (I just made it export
encoderFactory). What does it actually buy us though? It allows anyone to use
the code, without being tied to a particular implementation of sha256.

What does this module do? It encodes byte arrays into (primarily base58) encoded
strings, and vice versa. Your suggestion that we name this a `bitcoin-address-
encoder` implies a lot about the functionality it should export. The essence of the
bitcoin addresses, is the checksummed, versioned encodings.

Anyway, it's not much to wire up a bunch of encoders, pretty much as we were
already doing:

```javascript
Base.encoders = _.zipObject(
  _.pairs(alphabets).map(pair => [pair[0], new Encoder(pair[1])])
);
```

Currently we need to delegate anyway, as I wanted to export methods named with
`camelCase` which is sorta/kinda a JavaScript standard. 

Maybe we could export a factory for an instance of higher level class, similar to `Base` in
which you could addAlphabet('ripple', 'rp.....') and setDefaultAlphabet('bitcoin') etc so later
base.decode('rpQtsr...', 'ripple') etc will work? 

May as well pre-instantiate with ripple/bitcoin too?

clark800:
What's wrong with being tied to one implementation of sha256? They're all basically the same.

Why add stateful complexity like `addAlphabet` when you can just make the alphabet an optional parameter?

I suggested `bitcoin-address-codec` because it does both encode and decode.

sublimator:
Ok?  I'M so.so.confused. so4rry

What's wrong with being tied to one implementation of sha256? They're all
basically the same.

Why add stateful complexity like addAlphabet when you can just make the
alphabet an optional parameter?

I suggested bitcoin-address-codec because it does both encode and decode.

—
Reply to this email directly or view it on GitHub
<https://github.com/ripple/ripple-lib/pull/365#issuecomment-111239811>.

sublimator:
> I'M so.so.confused

haha - so so many beers :)

> because it does both encode and decode.

Duh! Yes, good point re: encoder vs codec.

> They're all basically the same.

They often come as part of a library. The obvious and pertinent example is sjcl
vs elliptic/hash.js/bn.js. If a potential client was already using sjcl, and
this explicitly dependend upon hash.js then using this becomes troublesome if
browser use is in consideration, and it's desired to keep the amount of
dependencies down.

clark800:
> They often come as part of a library.

I agree it would be bad to depend on a full crypto library just for sha256, but there are packages that provide just sha256 that are under 200 lines of code: https://www.npmjs.com/package/sha256

From a usability standpoint, it seems much better to simply provide the interface for doing what the user wants to do without asking them to provide additional resources by dependency injection.

sublimator:
> seems much better to simply

It's really not that big a deal to DI, and you can have a thin wrapper
module that wires it up with whatever sha 256 implementation.

clark800:
What's the benefit though? Saving 2KB in the library size?

sublimator:
> Saving 2KB in the library size?
Sure, and choice of implementation.

sublimator:
Update:

I renamed the module to `coin-address-codec`, and it now actually builds off a fork of  [bs58](https://github.com/cryptocoinjs/bs58) as it's faster than the admittedly prettier and more flexible `baseconverter` in ripple-lib.

The [fork of bs58](https://github.com/sublimator/bs58) merely also exports a new factory to create the encode/decode api it normally hardcodes to the bitcoin alphabet. I opened a [pull request](https://github.com/cryptocoinjs/bs58/pull/11) and hope it will be merged. Stellar uses a hard fork, [vendored in] (https://github.com/stellar/js-stellar-base/blob/master/src/vendor/bs58.js).

The injection of sha256 really isn't that painful:

```javascript
const Base = {};
const {apiFactory} = require('coin-address-codec');
const baseOptions = {sha256, defaultAlphabet: 'ripple'};
const {encode, decode, codecs} = apiFactory(baseOptions);
Base.alphabets = Object.keys(codecs);

// --> input: big-endian array of bytes.
// <-- string at least as long as input.
Base.encode = function(input, alphabet) {
  return encode(input, {alphabet});
};
```
I also forked bs58check module to [build off coin-address-codec](https://github.com/sublimator/bs58check/blob/master/index.js), more just to feel it out rather than with any designs for it to be merged:

```javascript
'use strict';

var createHash = require('create-hash')
var addresses = require('coin-address-codec')

function toBuffer(buf) {
  return Buffer.isBuffer(buf) ? buf : new Buffer(buf)
}

function sha256(array) {
  return createHash('sha256').update(toBuffer(array)).digest()
}

var api = addresses.apiFactory({sha256: sha256})
var codec = api.codecs.bitcoin

// Encode a buffer as a base58-check encoded string
function encode(payload) {
  return codec.encodeChecked(payload)
}

// Decode a base58-check encoded string to a buffer
function decode(string) {
  return toBuffer(codec.decodeChecked(string))
}

module.exports = {
  encode: encode,
  decode: decode
}
```

In the process, I noticed that it's worth exporting on the AddressCodec class, some (encode|decode)Checked methods which don't strip the version from the output. Clients of bs58check seem to get the version themselves from the result.

clark800:
> Sure, and choice of implementation.
I would count this as a negative rather than a positive. The goal of abstraction is to alleviate the user of the need to make low-level decisions. Also, what if the user isn't already using sha256 for something else? Each user redundantly has to go search for a library. The 2KB in size savings is a benefit, but it seems trivial unless we are running on machines with barely any memory. It feels like premature optimization to me.

Compare your 8 line example with the simplicity of this:
```
const bitcoinAddressCodec = require('bitcoin-address-codec');
console.log(bitcoinAddressCodec.encode([1,2,3,4], 'abcdefg');
```
I'd prefer not to ask the user to make a factory and even *mutate* things to do something that is purely functional.

How much faster is the bs58 implementation? I think it would be nice if we could keep the generality of the base converter so that it can be used with bases other than 58, depending on the length of the alphabet string. The baseconverter algorithm is pretty fast already, so again I'm concerned that this might be premature optimization.

I also want to point out that the name `coin-address-codec` doesn't make sense because coins don't have addresses. This is an address format invented by "Bitcoin" the system, not "bitcoin" the asset.

sublimator:
I appreciate that you want conceptual purity and to avoid mutations etc. Others care more about flexibility and performance ( base-x was around 12% faster ). 

You can wire together a little module as you want pretty much as so, using the preexisting base-x module (not sure if your proposed API did checksum calculation, and to be clear, this is not a module that you'd expect to see residing in an application, but rather on NPM):

```javascript
const _ = require('lodash');
const basex = _.memoize(require('base-x'));
const BITCOIN = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz';

module.exports = {
  encode: function(bytes, alphabet=BITCOIN) {
    return basex(alphabet).encode(bytes);
  },
  decode: function(string, alphabet=BITCOIN) {
    return basex(alphabet).decode(string);
  }
};
```

Maybe that is a useful building block, that you could build other modules on top of?

Anyway, there's more to it than that. Some people want an API that decodes, and also recalculates the checksum, returning the version and payload in one array. Others want something higher level, where you pass the version in as a parameter and it throws an exception if the version byte doesn't match. We are going to be using a 3 byte prefix for ed25519 seeds, and the normal 1 byte version for the secp256k1 so we want to be able pass in multiple versions, and the payload length.

People also want an API that simply takes 'bitcoin' which looks up an alphabet rather than having to pass the whole alphabet each use. You could do that pretty easily in the application layer, of course.

You could just as easily wire up a little module, that treats the coin-address-codec (*not stuck on that name*) in the same way as the above uses base-x, for doing encoding with checksums (in much the same way as the [bs58check](https://github.com/bitcoinjs/bs58check/blob/master/index.js) module wires up [create-hash](https://www.npmjs.com/package/create-hash) and [bs58](https://github.com/cryptocoinjs/bs58)) 

```javascript
const _ = require('lodash');
const {codecFactory} = require('coin-address-codec');

const hashjs = require('hashjs');
function sha256(bytes) {
  return hashjs.sha256().update(bytes).digest();
}

const Codec = codecFactory({sha256});
const getCodec = _.memoize(a => new Codec(a));
const BITCOIN = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz';

module.exports = {
  encode: function(bytes, alphabet=BITCOIN) {
    return getCodec(alphabet).encodeChecked(bytes);
  },
  decode: function(string, alphabet=BITCOIN) {
    return getCodec(alphabet).decodeChecked(string);
  }
};
```

Anyway, this is how ripple-lib currently wires things up:
```javascript
const sha256 = require('./utils').sha256;

const Base = {};
const {apiFactory} = require('coin-address-codec');
const baseOptions = {sha256, defaultAlphabet: 'ripple'};
const {encode, decode, codecs} = apiFactory(baseOptions);
Base.alphabets = Object.keys(codecs);
```

The encode/decode functions have keyword arguments for alphabet by symbol, and have version[s] options:

```javascript
// --> input: Array
// <-- String
Base.encode_check = function(version, input, alphabet) {
  return encode(input, {version, alphabet});
};
```

> I think it would be nice if we could keep the generality of the base converter so that it can be used with bases other than 58

The `baseconverter` doesn't really do encoding in and of itself yah? It just interprets an array of digits as a number, and then converts any significant digits to a new base, discarding any leading zeroes.  base-x does all the encoding of arbitrary bases too, but doesn't have discrete code for each step.  I'm not saying that baseconverter isn't useful, just not as a standalone "codec", and that there's already a module that inlines the functionality (probably to the benefit of performance).

The codec WILL work for (somewhat) arbitrary alphabets/bases, but realistically, due to the way leading zeroes are handled (I think to avoid the need for padding?) it's doubtful (but not impossible) anyone is going to use it for much other than base58 / bitcoin style address. `'000f'` using the hex alphabet decodes to the base 256 array of digits `[0, 0, 0, 15]`, whereas a normal hex codec assigns 4 bits uniformly to each character. Besides, if you wanted b16, b32, b64 etc, there's specs/libs for that.

clark800:
Well I don't think you'll be able to convince me or I'll be able to convince you, so maybe someone else can weigh in.

The base-x module looks like it's doing basically the same algorithm as the baseconverter, except it is coupling the math with the string encoding. I'm basically agnostic about using base-x vs. baseconverter, though I would probably choose baseconverter. I suspect that the 12% speed difference might be from the encoding step that you would have to add in order to make them comparable.

But I'm less convinced about the interface and naming.

sublimator:
> Well I don't think you'll be able to convince me or I'll be able to convince you, so maybe someone else can weigh in.

Different values. For example I don't really care about the name at all, and am open to changing it. I
borrowed `coin` from [coinstring](https://github.com/cryptocoinjs/coinstring)

> The base-x module looks like it's doing basically the same algorithm as the baseconverter, except it is coupling the math with the string encoding. 

Exactly. The bs58 maintainer seems like he's open to basing bs58 off of base-x.

> I suspect that the 12% speed difference might be from the encoding step that you would have to add in order to make them comparable.

I'm not sure what it was exactly, but I was comparing math+encoding to math+encoding. I noticed bs58 has commit history with "speedup" in the titles.

> But I'm less convinced about the interface
> what if the user isn't already using sha256 for something else? Each user redundantly has to go search for a library

I just showed you how you could wire together a small, publishable module that does what you want from other building blocks. The existence of both bs58 and bs58check should speak to the possibility/utility of wiring.

PR #366, opened 2015-06-10
[FIX] emit empty offers list (RT-3462)

darkdarkdragon:
For autobridged order books, emit 'model' event,
even if offers is empty, to indicate that
orderbook is loaded

### DIFF
--- a/src/core/orderbook.js
+++ b/src/core/orderbook.js
@@ -255,7 +255,7 @@ OrderBook.prototype.unsubscribe = function() {
  * @param {Function} callback
  */
 
-OrderBook.prototype.requestOffers = function(callback) {
+OrderBook.prototype.requestOffers = function(callback=function() {}) {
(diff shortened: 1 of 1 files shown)


src/core/orderbook.js
```diff
@@ -258,6 +258,8 @@ OrderBook.prototype.unsubscribe = function() {
 OrderBook.prototype.requestOffers = function(callback) {
   const self = this;
 
+  var callback = typeof callback === 'function' ? callback : function(){};
```
clark800: This redefines a symbol, which is forbidden by our lint rules. You can use es6 default parameters instead: `function(callback=function() {})`

clark800:
Can you rebase to HEAD?

darkdarkdragon:
@clark800 done

clark800:
LGTM

PR #368, opened 2015-06-12
Remove IOU_SUFFIX from orderbookutils.js and clean up serializedtypes.js

mwshortt:
* Remove IOU_SUFFIX from orderbookutils.js
* Change serialize function in serializedtypes.js (line 412) to accept values as well as amounts
* Clean up serializedtypes.js so it follows eslint rules

### DIFF
--- a/src/core/orderbookutils.js
+++ b/src/core/orderbookutils.js
@@ -1,17 +1,34 @@
 'use strict';
 
-var _ = require('lodash');
-var assert = require('assert');
-var SerializedObject = require('./serializedobject').SerializedObject;
-var Types = require('./serializedtypes');
-var Amount = require('./amount').Amount;
-
-var IOU_SUFFIX = '/000/rrrrrrrrrrrrrrrrrrrrrhoLvTp';
-var OrderBookUtils = {};
+const _ = require('lodash');
+const assert = require('assert');
+const SerializedObject = require('./serializedobject').SerializedObject;
+const Types = require('./serializedtypes');
+const Amount = require('./amount').Amount;
+const OrderBookUtils = {};
 
 function assertValidNumber(number, message) {
   assert(!_.isNull(number) && !isNaN(number), message);
 }
+
+/**
+* Creates a JSON amount object using
+* passed parameters for value, currency and counterparty
+* if currency or counterparty is undefined, defaults to IOU
+*
+* @param amount of value, currency, counterparty
+* @return JSON amount object
+*/
+
+function createAmount(value, currency, counterparty) {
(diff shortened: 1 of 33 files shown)


src/core/serializedtypes.js
```diff
@@ -432,7 +441,7 @@ exports.Quality = new SerializedType({
       lo = parseInt(mantissaHex.slice(-8), 16);
     }
 
-    var valueBytes = sjcl.codec.bytes.fromBits([hi, lo]);
+    let valueBytes = sjcl.codec.bytes.fromBits([hi, lo]);
```
wltsmrz: Could be `const` huh?

mwshortt:
I am reworking this

src/core/serializedtypes.js
mwshortt: It could - thanks for the catch!

src/core/orderbookutils.js
```diff
@@ -23,7 +40,16 @@ function assertValidNumber(number, message) {
 OrderBookUtils.getOfferTakerGetsFunded = function(offer) {
   assertValidNumber(offer.taker_gets_funded, 'Taker gets funded is invalid');
 
-  return Amount.from_json(offer.taker_gets_funded + IOU_SUFFIX);
+  let currency = offer.TakerPays.currency;
```
sublimator: This seems a little redundant at first glance and repeats the solution in many places. Previously the code was using `IOU_SUFFIX` to create a non native amount from what's essentially just the value.

AFAIK it doesn't matter what currency/issuer the amount has, just that it has a quantity and can interoperate with the Amount class math as an IOU. 

I think the goal here was to remove the need for having to tack on a string to force a value to parse as an IOU, and not native. `Amount.from_number` just happens to work for this case I think, as it sets the currency/issuer to 1/1 respectively, which are neutral.  

What `number` actually means in `from_number` is unclear. Currently, it just happens to work with anything that is valid for the BigNumber constructor. The value here (that previously had IOU_SUFFIX tacked on) would be a legitimate numeral, but as a string not a `Number`. 

Depending on how strong your need for order, you could add an `Amount.from_neutral` or the like :) 

`Amount.from_json(offer.taker_gets_funded + IOU_SUFFIX);` 

->

`Amount.from_number(offer.taker_gets_funded);`

@wltsmrz  ?

sublimator: https://github.com/ripple/ripple-lib/blob/develop/src/core/amount.js#L648-L654

clark800:
This is still "work in progress". Maybe we should close this pull request until it's ready?

PR #369, opened 2015-06-12
Rewrite transaction parser and add unit test for getTransaction

clark800:


test/fixtures/transaction-response.json
```diff
@@ -0,0 +1,89 @@
+{
+  "type": "payment",
+  "address": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
+  "specification": {
+    "source": {
+      "address": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
+      "amount": {
+        "currency": "XRP",
+        "value": "1.112209"
```
sublimator: Why is this missing counterparty ? Elsewhere I see an empty string used ? I'm not all that familiar with ripple-rest, but glancing over this.

test/fixtures/transaction-response.json
```diff
@@ -0,0 +1,89 @@
...
+    },
+    "destination": {
+      "address": "rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM",
+      "amount": {
+        "currency": "USD",
+        "value": "0.001",
+        "counterparty": "rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM"
+      }
+    },
+    "fee": "0.00001",
```
sublimator: We're just using a string here? But elsewhere `{"currency" : "XRP" , ... ` ?

src/api/common/index.js
```diff
@@ -9,5 +9,6 @@ module.exports = {
   server: require('./server'),
   dropsToXrp: utils.dropsToXrp,
   xrpToDrops: utils.xrpToDrops,
-  convertAmount: utils.convertAmount
+  unparseAmount: utils.unparseAmount,
```
sublimator: `toRippledAmount` or something ?  Mentioning only cause I had to read to see what `unparse` meant.

src/api/ledger/parse/trustline.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+const assert = require('assert');
+const utils = require('./utils');
+const flags = utils.core.Transaction.flags.TrustSet;
+
+function parseTrustline(tx) {
```
sublimator: `parseTrustLine` -> `parseTrustSet`

src/api/ledger/transactions.js
```diff
@@ -1,102 +1,74 @@
...
-    const maxLedger = Number(options.max_ledger);
-    for (let i = minLedger; i <= maxLedger; i++) {
-      if (!api.remote.getServer().hasLedger(i)) {
-        return callback(new errors.NotFoundError('Ledger not found'));
-      }
+function hasCompleteLedgerRange(remote, options) {
+  const minLedger = Number(options.minLedger || MIN_LEDGER_VERSION);
+  const maxLedger = Number(options.maxLedger || remote.getLedgerSequence());
+  for (let i = minLedger; i <= maxLedger; i++) {
+    if (!remote.getServer().hasLedger(i)) { // TODO: optimize this
```
sublimator: I'm pretty sure I wrote a RangeSet one day when I was bored, that had contains(range), add(range) etc and using bisect, but I can't find it atm. Was on my old computer.

src/api/ledger/transactions.js
```diff
@@ -1,102 +1,74 @@
...
-    queryTransaction,
-    checkIfRelatedToAccount,
-    attachDate
-  ];
+  function callbackWrapper(error, tx) {
+    if (error instanceof errors.NotFoundError
+        && !hasCompleteLedgerRange(remote, options)) {
+      callback(new errors.MissingLedgerHistoryError('Transaction not found,'
+        + ' but the server\'s ledger history is incomplete'));
+    } else if (!error && !isTransactionInRange(tx, options)) {
```
sublimator: Why does it matter if the transaction is not in the passed in minLedger/maxLedger range?  What's the purpose of the minLedger/maxLedger? Is it primarily for detecting if a server doesn't have the enough history to give an authoritative result? This isn't searching for any transactions in a range, it's searching for a single transaction by hash. 

Maybe the error message could be "Transaction found, but was not in specified ledger range". That would give more of a clue as to what's gone wrong.

src/api/transaction/settings.js
```diff
@@ -5,32 +5,23 @@ const _ = require('lodash');
...
- * @param {Object} flags - Object that maps property names to transaction
- *                integer flag values
- *
- * @returns undefined
- */
-function setTransactionIntFlags(transaction, values, flags) {
-  for (let flagName in flags) {
+function setTransactionFlags(transaction, values) {
+  const nameToIndexMap = AccountFlagIndices;
+  for (let flagName in nameToIndexMap) {
```
sublimator: SetFlag and ClearFlag can only be set once for each transaction? I haven't looked elsewhere at calling code, but is it possible that some flags could be silently ignored?

sublimator: I think I looked at this the other day when looking at Vinine's c++ testing framework, flags helper. The AccountSet transactor honours bits set in Flags for everything bar the `AccountTxnID` setting.

src/api/ledger/parse/cancellation.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+const assert = require('assert');
+
+function parseOrderCancellation(tx) {
+  assert(tx.TransactionType === 'OfferCancel');
+  return {
+    orderSequence: tx.OfferSequence
```
sublimator: The other transactor parsers include `fee`

test/fixtures/transaction-response.json
clark800: I'm trying to use the convention that if something doesn't apply, then it is not present. The places where empty string still appears are because that is coming from ripple-lib-transactionparser. At some point either that module will be changed or an adapter will be used to get rid of the counterparty fields.

sublimator: That sounds good to me

test/fixtures/transaction-response.json
clark800: This comment is pointing at the fee line, are you referring to the paths line below?

src/api/ledger/transactions.js
clark800: Yeah, I think we need one where ranges are merged on add, then a method that takes a range and checks if it is fully contained in one of the existing ranges.

test/fixtures/transaction-response.json
sublimator: Was referring to the fee amount using a string instead of an object. With
all the empty string issuer stuff elsewhere it seemed we were gung ho about
uniformity. But a string here?

The paths just seem like raw rippled style json stringified

src/api/ledger/transactions.js
clark800: > Why does it matter if the transaction is not in the passed in minLedger/maxLedger range?

Because that is the semantics of minLedger/maxLedger here. Actually the names should be minLedgerVersion and maxLedgerVersion - I'll change that.

> What's the purpose of the minLedger/maxLedger? Is it primarily for detecting if a server doesn't have the enough history to give an authoritative result?

Yeah, we need to be able to give an authoritative result. If the user wants to verify a transaction exists, they should pass minLedgerVersion as the ledger version at submission and maxLedgerVersion as the maxLedgerVersion given in the transaction instructions. If the transaction was not found and a range was not specified and there was a ledger gap, we would not be able to provide an authoritative result.

> Maybe the error message could be "Transaction found, but was not in specified ledger range"

This would constrain the implementation. The code should not be required to search outside the specified range.

sublimator: Lemme look again, but from memory it was specifically reporting for the
case the transaction (queried by id) was found but not in the range.

How would it constrain the implementation? I'm not suggesting that it
should search outside the range ,but in that particular case you're
basically saying not found, when it was found. That's lying haha :)

If that's the case, more often that not I think they'd have messed
something up with their ledger range code.

src/api/transaction/settings.js
clark800: > SetFlag and ClearFlag can only be set once for each transaction?

That's what it looks like: https://ripple.com/build/transactions/#accountset-flags

> is it possible that some flags could be silently ignored?

I'm not sure, is there a reason to suspect this?

> The AccountSet transactor honours bits set in Flags for everything bar the AccountTxnID setting.

You mean the SetFlag doesn't work for AccountTxnID? The docs make it look like it should work.

src/api/ledger/parse/cancellation.js
clark800: Nice catch, `fee` isn't part of the `specification`, so I should probably move it to `outcome`.

src/api/transaction/settings.js
sublimator: With the ripple-rest ( for lack of a better term atm) api? Does it purport
to allow setting more than one setting at once?

With that code, you couldn't. You could if you set all the settings you
wanted to do by bit setting with Flags.

Only AccountTxnID doesn't support the use of Flags to set it. It requires
the use of SetFlag.

src/api/ledger/transactions.js
sublimator: Yeah, I had one somewhere, I ported from a python version of RegionSet I
wrote for  sublime, which is much the same deal, but it was totally
overkill with subtract methods etc.

test/fixtures/transaction-response.json
clark800: Oh, I see. Yeah, we want to be uniform, but also concise and non-redundant. Since the fee is always XRP, it would be redundant to always label it as XRP.

src/api/ledger/transactions.js
clark800: > How would it constrain the implementation?

For example, if rippled provided a highly optimized method to search a range of ledger versions for a transaction, we should be able to use that method. If we promise to notify the user that the server had the transaction even though it was out of range, we would tie ourselves to always searching the whole history and prevent us from using the optimized method.

The fact that it was internally found outside the specified range is just an implementation detail that need not be exposed externally.

src/api/transaction/settings.js
clark800: > Does it purport to allow setting more than one setting at once?

The schema has a "maxProperties: 1" constraint, so it only lets you set one a time.

The docs say "The preferred way to enable and disable Account Flags is using the SetFlag and ClearFlag parameters of an AccountSet transaction." I'm not sure why it's preferred, but I figured it's best to stick with the recommendation.

src/api/ledger/transactions.js
sublimator: There's not much need not to expose it. That case explicitly catches the
case where it was found but out of range.  You can still throw an error,
just be more precise, with the reporting, so they get some inkling of
what's going on.

Much like, "but the ledger history is incomplete"

src/api/transaction/settings.js
sublimator: So you can only set one property per transaction anyway.

src/api/ledger/transactions.js
clark800: The error message you suggest exposes the details.

src/api/transaction/settings.js
clark800: Well I think rippled supports setting multiple settings per transaction, but not with the recommended method.

src/api/ledger/transactions.js
sublimator: Purists! rolls eyes

src/api/transaction/settings.js
sublimator: I meant more the api thingy here rather than rippled. If the schema
validator will give you a ticket for violating the law ... I guess that
code only needs to be able to support 1.

Could add an assert I guess for the next confused idiot ...

src/api/common/index.js
clark800: Ok, will rename

src/api/ledger/parse/trustline.js
clark800: If we change this, we would also have to change `parseOrder` -> `parseOfferCreate` and then it starts getting confusing with questions like "what's the difference between an order and an offer?" I think it's better to just stick with consistent naming.

sublimator: Ah, so trustline is a ripple-rest name ?

src/api/transaction/settings.js
clark800: Yeah, actually the function can be simplified by using Object.keys, which would also make it more clear

src/api/ledger/parse/trustline.js
clark800: Let's call it "ripple api" because it's unrelated to rest. There are five transaction types in the ripple api: `payment`, `trustline`, `order`, `orderCancellation`, and `settings`.

sublimator: I probably shouldn't have been "reviewing" this, as I was never familiar
with ripple-rest / ripple-api, it's just no one had yet.

I'll stay away from these in the future.

clark800: Well there is a shortage of reviewers for this, so I appreciate the help and invite you to continue reviewing if you have the time.

It is a little tricky though because it is very much a work in progress with a lot of code that is untested an *known* to be broken. In particular, the `src/api/ledger` directory, except for the `parse` directory is mostly broken.

sublimator: Ok, will endeavour to learn `ripple api`

src/api/common/utils.js
```diff
@@ -20,8 +20,20 @@ function convertAmount(amount) {
   };
 }
 
+function wrapCatch(asyncFunction: any): any {
```
lumberj: Isn't there a way to annotate function types in Flow?

lumberj: Looking at function wrapCatch(asyncFunction: any): any {

clark800: Yeah, but this one isn't worth the effort unless you want to try to figure it out.

src/api/common/validate.js
```diff
@@ -25,9 +25,10 @@ function validateAddressAndSecret(obj) {
 }
 
 function validateLedgerRange(options) {
-  if (_.isUndefined(options.minLedger) && _.isUndefined(options.maxLedger)) {
-    if (Number(options.minLedger) > Number(options.maxLedger)) {
-      throw error('minLedger must not be greater than maxLedger');
+  if (_.isUndefined(options.minLedgerVersion)
```
lumberj: If minLedgerVersion is undefined AND maxLedgerVersion is undefined AND minLedgerVersion > maxLedgerVersion -- am I reading that right?

clark800: Oops, somehow the `not` operators got dropped

lumberj: Looks like it happened before your change... I have a suspicion it may have been me

src/api/ledger/parse/utils.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+const _ = require('lodash');
+const transactionParser = require('ripple-lib-transactionparser');
+const toTimestamp = require('../../../core/utils').toTimestamp;
```
lumberj: I added this https://www.npmjs.com/package/root-require to ripple-connect to avoid the `../../../`

src/api/ledger/transactions.js
```diff
@@ -1,102 +1,77 @@
...
 const _ = require('lodash');
-const assert = require('assert');
 const async = require('async');
 const utils = require('./utils');
+const parseTransaction = require('./parse/transaction');
 const validate = utils.common.validate;
 const errors = utils.common.errors;
 
 const DEFAULT_RESULTS_PER_PAGE = 10;
+const MIN_LEDGER_VERSION = 32570;
```
lumberj: Does this one deserve a comment?

src/api/ledger/transactions.js
```diff
@@ -1,102 +1,77 @@
...
-    for (let i = minLedger; i <= maxLedger; i++) {
-      if (!api.remote.getServer().hasLedger(i)) {
-        return callback(new errors.NotFoundError('Ledger not found'));
-      }
+function hasCompleteLedgerRange(remote, options) {
+  const minLedgerVersion = options.minLedgerVersion || MIN_LEDGER_VERSION;
+  const maxLedgerVersion = options.maxLedgerVersion
+    || remote.getLedgerSequence();
+  for (let i = minLedgerVersion; i <= maxLedgerVersion; i++) {
+    if (!remote.getServer().hasLedger(i)) { // TODO: optimize this
```
lumberj: Make this an async function (you could use `async.parallel` or promises) and execute the getServer().hasLedger() calls in parallel?

src/api/transaction/settings.js
```diff
@@ -2,35 +2,28 @@
...
-    const value = values[flagName];
-    const code = flags[flagName];
-
-    if (value === true) {
-      transaction.tx_json.SetFlag = code;
-    } else if (value === false) {
-      transaction.tx_json.ClearFlag = code;
+function setTransactionFlags(transaction, values) {
+  const keys = Object.keys(values);
+  assert(keys.length === 1);    // can only set one setting per transaction
```
lumberj: Add an error message here and get rid of the comment?

src/api/ledger/transactions.js
clark800: Yes

src/api/ledger/transactions.js
clark800: We can do it the optimal way soon: https://ripplelabs.atlassian.net/browse/RLJS-388

src/api/transaction/settings.js
clark800: Good idea

test/fixtures/transaction-response.json
lumberj: The convention used to be "nothing is ever missing or null". Something about compiled languages using the rest api

clark800: Yeah, that argument never made sense to me. Compiled languages should be able to handle missing keys too. But this is a javascript API, so I don't think it matters much anyway. Also, after talking to Bob he seemed to agree that having schemas alleviates the need for showing everything all the time.

src/api/common/utils.js
lumberj: Isn't it like:

```javascript
function wrapCatch(asyncFunction: (() => void): () => void {
}
```

clark800: No, because the `asyncFunction` parameter can take any number of arguments.

lumberj: > Too Few Arguments
Flow produces an error whenever a call provides too few arguments to a function. On the other hand, too many arguments is fine, which is why Flow does not complain on most uses of builtin primitives like Array, whose methods are often called with closures that take fewer parameters than prescribed.


http://flowtype.org/docs/functions.html#too-few-arguments

clark800: Interesting. Well it doesn't fully constrain the type, but I guess it's would at least check that the parameter is a function with no return value, so I'll make the change.

lumberj:
Just a few comments otherwise lgtm

src/api/ledger/parse/utils.js
clark800: Will consider this after this pull request

PR #378, opened 2015-06-22
Add Transaction.signingData

sublimator:
This is needed for ed25519, which, per the rippled implementation, uses its own hashing routine (sha512) and wants the full data. 

A small milestone pull request ...


src/core/transaction.js
```diff
@@ -435,6 +435,15 @@ Transaction.prototype.signingHash = function(testnet) {
   return this.hash(testnet ? 'HASH_TX_SIGN_TESTNET' : 'HASH_TX_SIGN');
 };
 
+Transaction.prototype.signingData = function() {
+  return this._signingData('HASH_TX_SIGN').buffer;
+};
+
+Transaction.prototype._signingData = function(prefix) {
```
sublimator: Kinda pointless taking the prefix, given test net functionality is removed from rippled ...

src/core/transaction.js
```diff
@@ -435,6 +435,15 @@ Transaction.prototype.signingHash = function(testnet) {
...
+  return SerializedObject.from_json(
+        this.tx_json, hashprefixes[prefix].toString(16));
```
sublimator: I guess theoretically there could be a problem if the hash prefix started with a 0.

wltsmrz:
LGTM

bin/ci.sh
```diff
@@ -10,11 +10,9 @@ typecheck() {
 
 lint() {
   REPO_URL="https://raw.githubusercontent.com/ripple/javascript-style-guide"
-  npm install -g eslint
```
clark800: Good idea to avoid double installing

PR #379, opened 2015-06-23
Abstract api deprecation

wltsmrz:


clark800:
What does "abstract api" refer to?

It looks like a lot of this is refactoring the handling of deprecated usage. Do you think it's been deprecated long enough to remove support for the deprecated usage?

wltsmrz:
"Abstract" is a verb. The commit could have a better message, I'll think about it more. After the next release I hope to remove support

wltsmrz:
Halt

PR #384, opened 2015-06-25
Decouple the Amount class from math performed on values

mwshortt:
* Remove IOU_SUFFIX from orderbookutils.js by using the correct currency and counterparty when available 
* Create new Values class that performs math without using the Amount class 
* Change serialize function in serializedtypes.js to accept values as well as amounts
* Clean up serializedtypes.js so it follows eslint rules


src/core/value_XRP.js
```diff
@@ -0,0 +1,30 @@
...
+  Value.call(this, value);
+  this._bi_xns_unit = bi_xns_unit;
+  this._XRP = true;
+}
+
+Value_XRP.prototype = Object.create(Value.prototype);
+
+Value_XRP.prototype.constructor = Value_XRP;
+
+Value_XRP.prototype._canonicalize = function(value) {
```
clark800: Nice polymorphism :+1:

src/core/value_IOU.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+const Value = require('./value').Value;
+const GlobalBigNumber = require('bignumber.js');
+const BigNumber = GlobalBigNumber.another({
+  ROUNDING_MODE: GlobalBigNumber.ROUND_HALF_UP,
+  DECIMAL_PLACES: 40
+});
+
+function Value_IOU(value) {
```
clark800: `IOUValue` and `XRPValue` sound a little more natural to me

src/core/value_IOU.js
```diff
@@ -0,0 +1,54 @@
...
+  Value.call(this, value);
+}
+
+Value_IOU.prototype = Object.create(Value.prototype);
+
+Value_IOU.prototype.constructor = Value_IOU;
+
+Value_IOU.fromXRPValue = function(XRPvalue, bi_xns_unit) {
+  const newV = new Value_IOU(XRPvalue);
```
clark800: Can call it `newValue`?

src/core/value_IOU.js
```diff
@@ -0,0 +1,54 @@
...
+  newV._XRP = true;
+  newV._bi_xns_unit = bi_xns_unit;
+  return newV;
+};
+
+Value_IOU.prototype.multiply = function(multiplicand) {
+  let mult = multiplicand;
+  if (mult._XRP) {
```
clark800: I know this is more or less a result of the way rippled does math, but it's still really strange to have an `_XRP` flag in an IOU value class. Maybe instead of `fromXRPValue` we can just provide `multiplyByXRP` and `divideByXRP`?

src/core/orderbookutils.js
```diff
@@ -71,7 +109,16 @@ OrderBookUtils.getOfferQuality = function(offer, currencyGets) {
       reference_date: new Date()
     });
   } else {
-    amount = Amount.from_json(offer.quality + IOU_SUFFIX);
+
+    let currency = offer.TakerGets.currency;
```
clark800: This can be done with `const`s only.

src/core/orderbookutils.js
```diff
@@ -50,7 +85,10 @@ OrderBookUtils.getOfferTakerPaysFunded = function(offer) {
 OrderBookUtils.getOfferTakerGets = function(offer) {
   assert(typeof offer, 'object', 'Offer is invalid');
 
-  return Amount.from_json(offer.TakerGets + IOU_SUFFIX);
+  let currency = offer.TakerPays.currency;
+  let issuer = offer.TakerPays.issuer;
+
+  return Amount.from_json(createAmount(offer.TakerGets, currency, issuer));
```
clark800: Isn't `offer.TakerGets` an amount?

src/core/orderbookutils.js
```diff
@@ -36,7 +62,16 @@ OrderBookUtils.getOfferTakerGetsFunded = function(offer) {
 OrderBookUtils.getOfferTakerPaysFunded = function(offer) {
   assertValidNumber(offer.taker_pays_funded, 'Taker gets funded is invalid');
 
-  return Amount.from_json(offer.taker_pays_funded + IOU_SUFFIX);
+  let currency = offer.TakerGets.currency;
+  let issuer = offer.TakerGets.issuer;
+
+  if (currency === undefined || issuer === undefined) {
```
clark800: Looks like this block of coded is somewhat redundant with the above block

src/core/orderbookutils.js
```diff
@@ -1,19 +1,36 @@
...
+* if currency or counterparty is undefined, defaults to IOU
+*
+* @param amount of value, currency, counterparty
+* @return JSON amount object
+*/
+
+function createAmount(value, currency, counterparty) {
+  if (currency === undefined || counterparty === undefined) {
+    return {'value': value,
+      'currency': '000',
```
clark800: Still have to get rid of this hack

src/core/amount.js
```diff
@@ -148,20 +156,22 @@ Amount.prototype.subtract = function(subtrahend) {
...
-    multiplicandAmount._value.times(Amount.bi_xns_unit)
-    : multiplicandAmount._value;
-  return this._copy(this._value.times(multiplyBy));
+
+  const multiplicandAmount = Amount.from_json(multiplicand);
+  const multiplyBy = multiplicandAmount.is_native() ?
+    new Value_XRP(multiplicandAmount._value, Amount.bi_xns_unit)
+    : new Value_IOU(multiplicandAmount._value);
+  const thisValue = new Value_IOU(this._value);
+  return this._copy(new BigNumber(thisValue.multiply(multiplyBy)._value));
```
clark800: The math methods in the Amount class should be simpler now that most of the logic is in the value classes. I don't think the amount class should have to create any new BigNumber or Value objects because the Value classes should return them. In fact, the Amount class should have no mention of BigNumber because the Value class handles all the math.

src/core/orderbookutils.js
```diff
@@ -1,19 +1,36 @@
...
+      'issuer': 'rrrrrrrrrrrrrrrrrrrrrhoLvTp'};
+  }
+  return {'value': value, 'currency': currency, 'issuer': counterparty};
```
wltsmrz: "createAmount" sounds misleading. I'd suggest renaming it but it looks like we could probably just return a real Amount here rather than Amount.from_json()ing throughout the rest of the code

src/core/orderbook.js
```diff
@@ -448,11 +449,10 @@ OrderBook.prototype.applyTransferRate = function(balance) {
...
   assertValidNumber(this._issuerTransferRate, 'Transfer rate is invalid');
 
-  const adjustedBalance = OrderBookUtils.normalizeAmount(balance)
-  .divide(this._issuerTransferRate)
-  .multiply(Amount.from_json(OrderBook.DEFAULT_TRANSFER_RATE))
-  .to_json()
-  .value;
+  const adjustedBalance = (new Value_IOU(balance))
+  .divide(new Value_IOU(this._issuerTransferRate))
+  .multiply(new Value_IOU(OrderBook.DEFAULT_TRANSFER_RATE))
```
wltsmrz: Should probably just make _issuerTransferRate and OrderBook.DEFAULT_TRANSFER_RATE IOValues on assignment

src/core/amount.js
```diff
@@ -249,21 +259,21 @@ Amount.prototype.ratio_human = function(denominator, opts) {
...
  *
- * @param {Amount} factor The second factor of the product.
+ * @param {Amount} fac The second factor of the product.
  * @param {Object} opts Options for the calculation.
  * @param {Date|Number} opts.reference_date Date based on which
  * demurrage/interest should be applied. Can be given as JavaScript Date or int
  * for Ripple epoch.
  * @return {Amount} The product. Unit will be the same as the first factor.
  */
-Amount.prototype.product_human = function(factor, opts) {
```
wltsmrz: `options = {}` in the method signature works, I think

src/core/value.js
```diff
@@ -0,0 +1,55 @@
...
+  this._value = new BigNumber(value);
+}
+
+Value.prototype.abs = function() {
+  let result = this._value.abs();
+  return this._canonicalize(result);
+};
+
+Value.prototype.add = function(addend) {
+  let result = this._value.plus(addend._value);
```
wltsmrz: Should probably `assert(addend instanceof Value)` or `if (!(addend instanceof Value)) myAddend = new Value(addend)`. I also think we should add accessors instead of ._value, but not important

src/core/value_IOU.js
```diff
@@ -0,0 +1,54 @@
...
+const BigNumber = GlobalBigNumber.another({
+  ROUNDING_MODE: GlobalBigNumber.ROUND_HALF_UP,
+  DECIMAL_PLACES: 40
+});
+
+function Value_IOU(value) {
+  Value.call(this, value);
+}
+
+Value_IOU.prototype = Object.create(Value.prototype);
```
wltsmrz: I guess it's ok but I would stick to convention

clark800: Speaking of which, how about we use ES6 classes for this?

wltsmrz: Come to think of it this might be a good place to start using ES6 classes

wltsmrz: Huh looks like we agree

src/core/value_IOU.js
wltsmrz: Agree with this too

sublimator: Yes

src/core/amount.js
sublimator: If you go down this path I agree with what Chris says above. If the amount class `has no mention of` of any BigNumber, it follows that the `._value` member should be one of these new *Value classes that encapsulate them.


Anyway, it just occurred to me, that there's no amount class for the new ripple api, that uses counterparty instead of issuer.

src/core/orderbookutils.js
sublimator: Could be either a string OR an object amount, assuming it's in rippled format, which the capitalisation of the word indicates. It seems this (seeming) bug existed prior to this patch. Note the diff:
`- return Amount.from_json(offer.TakerGets + IOU_SUFFIX);`

It seems like the util is used in `autobridgecalculator`, which is code I'm as yet unacquainted with.

sublimator: Not sure if it's a bug, or just poorly named. Taking a peek: https://github.com/ripple/ripple-lib/blob/5b4deabd90e5237bd0165e65520f25456d32a410/src/core/autobridgecalculator.js#L123

sublimator: Also see here:  https://github.com/ripple/ripple-lib/blob/5b4deabd90e5237bd0165e65520f25456d32a410/src/core/autobridgecalculator.js#L13

The code is used for `leg one` where TakerGets is supposed to be native, which makes sense I guess, where you want to trade some IOU for XRP, to then trade that XRP for the wanted IOU.

I guess that function needs some renaming / comments to make it clearer.

src/core/orderbookutils.js
```diff
@@ -71,7 +109,16 @@ OrderBookUtils.getOfferQuality = function(offer, currencyGets) {
       reference_date: new Date()
     });
   } else {
-    amount = Amount.from_json(offer.quality + IOU_SUFFIX);
+
+    let currency = offer.TakerGets.currency;
+    let issuer = offer.TakerGets.issuer;
+
+    if (currency === undefined || issuer === undefined) {
+      currency = offer.TakerPays.currency;
```
sublimator: As always using IOU_SUFFIX before suggests, the issuer/currency don't matter at all. All this seems redundant when createAmount will set some neutral values anyway.

clark800: Well the initial purpose of this task was to eliminate the arbitrary values in IOU_SUFFIX.

`quality` is a ratio of amount values, so `getOfferQuality` should return an `IOUValue` rather than an `Amount`.

src/core/orderbookutils.js
clark800: Good point, looks like everywhere createAmount is called it is wrapped in Amount.from_json, so we can just make this return an Amount object.

src/core/amount.js
clark800: I don't think we need an amount class. The plan is to factor out `XRPValue` and `IOUValue` into a ripple-lib-extensions module. Client code can do math with these value objects. The only thing the Amount class has besides math is some parsing and formatting, which can be removed from ripple-lib and into a separate ripple-lib-extensions module (only needed by UI code). Of course it also keeps the currency and counterparty, but it doesn't actually track them properly for multiply and divide, so this isn't really useful. Then we could potentially replace the Amount class with just a plain javascript object: `{currency, counterparty, value}`.

src/core/value.js
clark800: Adding a flow annotation would be much better than those options.

wltsmrz: How do you make Flow cast the argument to a Value?

clark800: You shouldn't automatically cast, that means the caller is erroneous in passing the wrong type. Flow can do the equivalent of the assert, but statically.

wltsmrz: The caller is only erroneous if the function only accepts Values : )

clark800: Yeah, it keeps things a lot simpler if each function only accepts one type.

src/core/orderbookutils.js
sublimator: True, these are just as arbitrary and pointless, no matter that they are at least taken from one of TakerPays / TakerGets.

src/core/orderbookutils.js
sublimator: Or use IOUValue as you mentioned elsewhere.

src/core/amount.js
sublimator: > a plain javascript object: {currency, counterparty, value}

Composed entirely of primitives (strings) ? Or could I set the value as an IOUValue?

clark800: Most likely the `value` field would be an instance of IOUValue or XRPValue.

mwshortt:
Thank you for all of your comments! Here is the updated version

src/core/IOUValue.js
```diff
@@ -0,0 +1,43 @@
...
+  DECIMAL_PLACES: 40
+});
+
+class IOUValue extends Value {
+
+  constructor(value: string | BigNumber) {
+    super(value);
+  }
+
+  multiplyByXRP(multiplicand: {_value: BigNumber, _bi_xns_unit: number}) {
```
clark800: I think we should be able to make the type of multiplicand be `XRPValue`.

src/core/IOUValue.js
```diff
@@ -0,0 +1,43 @@
...
+    const constant = new BigNumber((multiplicand._bi_xns_unit).toString());
+    const value = new BigNumber(multiplicand._value);
+    multiplicand._value = (value).times(constant);
```
clark800: Instead of mutating I think we could create a new `IOUValue` instance.

src/core/IOUValue.js
```diff
@@ -0,0 +1,43 @@
...
+    return super.divide(divisor);
+  }
+
+  _canonicalize(value) {
+    return new IOUValue(value.toPrecision(16));
+  }
+
+  equals(comparator) {
+    return (comparator instanceof IOUValue)
+    && this._value.equals(comparator.value);
```
clark800: One is `_value` and one is `value`, is that right?

src/core/XRPValue.js
```diff
@@ -0,0 +1,29 @@
...
+  DECIMAL_PLACES: 40
+});
+
+const Value = require('./value').Value;
+
+class XRPValue extends Value {
+
+  constructor(value, bi_xns_unit) {
+    super(value);
+    this._bi_xns_unit = bi_xns_unit;
```
clark800: Might be worth considering a rename of this property.

src/core/amount.js
```diff
@@ -3,22 +3,19 @@
...
+const Seed = require('./seed').Seed;
+const Currency = require('./currency').Currency;
+const Value = require('./value').Value;
+const IOUValue = require('./IOUValue').IOUValue;
+const XRPValue = require('./XRPValue').XRPValue;
 
 
 function inverse(number) {
-  return (new BigNumber(number)).toPower(-1);
+  return new IOUValue(number).invert()._value;
```
clark800: Property names that start with underscore are supposed to be private.

src/core/amount.js
```diff
@@ -38,25 +35,25 @@ function Amount() {
...
-  bi_1e14: new BigNumber(1e14),
-  bi_1e16: new BigNumber(1e16),
-  bi_1e17: new BigNumber(1e17),
-  bi_1e32: new BigNumber(1e32),
-  bi_man_max_value: new BigNumber('9999999999999999'),
-  bi_man_min_value: new BigNumber(1e15),
-  bi_xns_max: new BigNumber(1e17),
-  bi_xns_min: new BigNumber(-1e17),
-  bi_xns_unit: new BigNumber(1e6),
+  bi_5: new Value(5)._value,
```
clark800: I don't know if anyone is using these, but now might be a good time to delete them. We shouldn't be exposing the bignumber implementation here.

src/core/amount.js
```diff
@@ -113,32 +110,45 @@ Amount.is_valid_full = function(j) {
...
-  var addendAmount = Amount.from_json(addend);
+  const addendAmount = Amount.from_json(addend);
 
   if (!this.is_comparable(addendAmount)) {
     return new Amount(NaN);
   }
 
-  return this._copy(this._value.plus(addendAmount._value));
+  const thisValue = new IOUValue(this._value);
+  const addendValue = new IOUValue(addendAmount._value);
```
clark800: You can just make `this._value` be stored as an instance of `XRPValue` or `IOUValue` so you don't have to do this every time.

src/core/amount.js
```diff
@@ -249,21 +264,20 @@ Amount.prototype.ratio_human = function(denominator, opts) {
...
+ * @param {Amount} fac The second factor of the product.
  * @param {Object} opts Options for the calculation.
  * @param {Date|Number} opts.reference_date Date based on which
  * demurrage/interest should be applied. Can be given as JavaScript Date or int
  * for Ripple epoch.
  * @return {Amount} The product. Unit will be the same as the first factor.
  */
-Amount.prototype.product_human = function(factor, opts) {
-  opts = opts || {};
+Amount.prototype.product_human = function(fac, options = {}) {
```
clark800: Better not to use abbreviations, especially in method signatures.

src/core/amount.js
```diff
@@ -343,12 +358,15 @@ Amount.prototype.invert = function() {
  */
 
 Amount.prototype.canonicalize = function(roundingMode) {
```
clark800: Does Amount even need a canonicalize method anymore? Don't the value classes already ensure canonicalization?

src/core/orderbookutils.js
```diff
@@ -1,19 +1,61 @@
...
+* @return currency
+*/
+
+function getCurrencyFromOffer(offer) {
+  let currency = offer.TakerPays.currency;
+
+  if (!currency) {
+    currency = offer.TakerGets.currency;
+  }
+  return currency;
```
clark800: `return offer.TakerPays.currency || offer.TakerGets.currency`

src/core/amount.js
```diff
@@ -113,32 +110,45 @@ Amount.is_valid_full = function(j) {
...
 Amount.NaN = function() {
-  var result = new Amount();
-  result._value = new BigNumber(NaN); // should have no effect
+  const result = new Amount();
+  result._value = new Value(NaN)._value; // should have no effect
   return result;                      // but let's be careful
 };
 
 // be sure that _is_native is set properly BEFORE calling _set_value
 Amount.prototype._set_value = function(value, roundingMode) {
```
clark800: We should be able to get rid of the `_set_value` method by just requiring that a Value instance is passed into the Amount constructor.

src/core/amount.js
mwshortt: Most of them aren't being used, but a handful of them still are (bi_man_max_value, bi_man_min_value, bi_xns_max, bi_xns_min, bi_xns_unit). I've removed the superfluous ones

clark800: For the constants that we need, I think they should be local variables in the appropriate module. I think the ones you listed probably belong in the `XRPValue` or `IOUValue` classes.

bin/ci.sh
```diff
@@ -11,7 +11,7 @@ typecheck() {
 lint() {
   REPO_URL="https://raw.githubusercontent.com/ripple/javascript-style-guide"
   curl "$REPO_URL/es6/eslintrc" > ./eslintrc
-  echo "parser: esprima-fb" >> ./eslintrc
+  echo "parser: babel-eslint" >> ./eslintrc
```
sublimator: This supports facebook flow (vaguely recall someone telling me that's what esprima-fb was for )  
?

sublimator: I guess so, judging by the rest of the code

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+    return this._value.toFixed(decimalPlaces, roundingMode);
+  }
+
+  getExponent() {
+    return this._value.e;
+  }
+
+  getMantissaDecimalString() {
+    let mantissa = this._value.abs().toPrecision(16)
+    .replace(/\./, '')      // remove decimal point
```
sublimator: Is this non indenting subsequent lines of multi-line expressions in our style-guide or something? I've seen it a bit recently.

mwshortt:
Updated once again:
* In the Amount class, this._value now stores an IOU or XRP Value
* In IOU and XRP Value, this._value is now private
* Out BN implementation is now hidden (I removed the exposed BN constants in the Amount class)

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+
+  subtract(subtrahend: Value) {
+    const result = this._value.minus(subtrahend._value);
+    return this._canonicalize(result);
+  }
+
+  multiply(multiplicand: Value) {
+    const val = this._value;
+    const mult = multiplicand._value;
+    const result = (val).times(mult);
```
clark800: Seems more consistent to inline like the others: `const result = this._value.times(multiplicand._value);`

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+    return this._canonicalize(result);
+  }
+
+  scale(scaleFactor: Value) {
```
clark800: This function can be deleted since it's exactly the same as multiply now.

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+  static getBNRoundDown() {
+    return BigNumber.ROUND_DOWN;
+  }
+
+  abs() {
+    const result = this._value.abs();
+    return this._canonicalize(result);
+  }
+
+  add(addend: Value) {
```
clark800: Throw an exception if types aren't the same in `add` and `subtract`

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+
+  scale(scaleFactor: Value) {
+    const result = this._value.times(scaleFactor._value);
+    return this._canonicalize(result);
+
+  }
+
+  divide(divisor: Value) {
+    if (this._value.isNaN()) {
+      throw new Error('Invalid dividend');
```
clark800: I'm not sure why we are checking for NaN only in division. Probably we should either delete these checks or maybe check for NaN in canonicalize?

src/core/amount.js
```diff
@@ -729,7 +703,8 @@ Amount.prototype.parse_native = function(j) {
 // Requires _currency to be set!
 Amount.prototype.parse_value = function(j) {
   this._is_native = false;
-  this._set_value(new BigNumber(j), BigNumber.ROUND_DOWN);
+  const newValue = new IOUValue(j).toPrecision(16, Value.getBNRoundDown());
```
clark800: `toPrecision` sounds like it should return a string since that's what bignumber.js toPrecision returns. Maybe a better option would be to pass the rounding mode to the IOUValue constructor and call canonicalize from the constructor.

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+    const result = (new BigNumber(this._value)).toPower(-1);
+    return this._canonicalize(result);
+  }
+
+  round(decimalPlaces, roundingMode) {
+    this._value = this._value.round(decimalPlaces, roundingMode);
+    return this;
+  }
+
+  toPrecision(significantDigits, roundingMode) {
```
clark800: Can remove this method.

src/core/value.js
clark800: I prefer indentation.

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+
+  toFixed(decimalPlaces, roundingMode) {
+    return this._value.toFixed(decimalPlaces, roundingMode);
+  }
+
+  getExponent() {
+    return this._value.e;
+  }
+
+  getMantissaDecimalString() {
```
clark800: Can this function be moved to serializedtypes.js by using `value.abs().toString()` ?

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+  lessThan(comparator: Value) {
+    return this._value.lessThan(comparator._value);
+  }
+
+  comparedTo(comparator: Value) {
+    return this._value.comparedTo(comparator._value);
+  }
+
+  equals(comparator) {
+    return (comparator instanceof Value)
```
clark800: I think an XRPValue and IOUValue should be not equal regardless of their numerical value.

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+  isNegative() {
+    return this._value.isNegative();
+  }
+
+  toString() {
+    return this._value.toString();
+  }
+
+  greaterThan(comparator: Value) {
+    return this._value.greaterThan(comparator._value);
```
clark800: Comparisons should probably throw an exception if you try to compare different subtypes.

src/core/IOUValue.js
```diff
@@ -0,0 +1,44 @@
...
+
+  divide(divisor) {
+    if (divisor instanceof XRPValue) {
+      return super.divide(
+        new IOUValue(divisor._value.times(divisor.ripple_units)));
+    }
+    return super.divide(divisor);
+  }
+
+  negated() {
```
clark800: For consistency, we should rename to `negate`

src/core/amount.js
```diff
@@ -113,32 +100,37 @@ Amount.is_valid_full = function(j) {
...
   return result;                      // but let's be careful
 };
 
 // be sure that _is_native is set properly BEFORE calling _set_value
-Amount.prototype._set_value = function(value, roundingMode) {
-  assert(value instanceof BigNumber);
-  this._value = value.isZero() && value.isNegative() ? value.negated() : value;
-  this.canonicalize(roundingMode);
+Amount.prototype._set_value = function(value) {
+  assert(value instanceof Value);
```
clark800: Might be good to use a flow annotation instead of the assert.

src/core/orderbookutils.js
```diff
@@ -1,19 +1,56 @@
...
+* @return issuer
+*/
+
+function getIssuerFromOffer(offer) {
+  let issuer = offer.TakerPays.issuer;
+
+  if (!issuer) {
+    issuer = offer.TakerGets.issuer;
+  }
+  return issuer;
```
clark800: Like above: `offer.TakerPays.issuer || offer.TakerGets.issuer;`

src/core/XRPValue.js
```diff
@@ -0,0 +1,43 @@
...
+  DECIMAL_PLACES: 40
+});
+
+const Value = require('./value').Value;
+
+class XRPValue extends Value {
+
+  constructor(value) {
+    super(value);
+    this.ripple_units = new BigNumber(1e6);
```
clark800: I think we're trying to use camelCase now.

src/core/XRPValue.js
```diff
@@ -0,0 +1,43 @@
...
+const BigNumber = GlobalBigNumber.another({
+  ROUNDING_MODE: GlobalBigNumber.ROUND_HALF_UP,
+  DECIMAL_PLACES: 40
+});
+
+const Value = require('./value').Value;
+
+class XRPValue extends Value {
+
+  constructor(value) {
```
clark800: It might be a good idea to throw an exception in the constructor if `value` has more than 6 digits of precision past the decimal point. This would help catch bugs where an IOUValue is being cast to an XRPValue.

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+'use strict';
+
+const GlobalBigNumber = require('bignumber.js');
+
+const BigNumber = GlobalBigNumber.another({
+  ROUNDING_MODE: GlobalBigNumber.ROUND_HALF_UP,
+  DECIMAL_PLACES: 40
+});
+
+class Value {
```
clark800: We should make this an abstract base class so that there are only two Value classes that appear in the code: in the constructor, throw an exception if `new.target === Value`. Also, in all the places that instantiate Value directly, replace it with either XRPValue or IOUValue.

src/core/value.js
```diff
@@ -0,0 +1,134 @@
...
+
+  constructor(value: string | BigNumber, base: number) {
+    this._value = new BigNumber(value, base);
+  }
+
+  static getBNRoundDown() {
```
clark800: Might be simpler to use class properties like the BigNumber class does. Basically just `Value.ROUND_DOWN = BigNumber.ROUND_DOWN` etc.

src/core/XRPValue.js
mwshortt: That's a good idea, but there are four tests in amount-tests.js using Amount.from_quality that violate this rule even though they are using XRP.

src/core/amount.js
```diff
@@ -614,10 +586,12 @@ function(quality, counterCurrency, counterIssuer, opts) {
     quality as stored :  5 USD          /  3000000 drops
     inverted          :  3000000 drops  /          5 USD
   */
-  var adjusted = opts.inverse ? inverse(value) : value;
-  var nativeAdjusted = adjusted;
+  const valueStr = mantissa.toString() + 'e' + offset.toString();
+  let nativeAdjusted = this._is_native ?
+    new XRPValue(valueStr) : new IOUValue(valueStr);
```
clark800: "quality" is a ratio, so it should always be an `IOUValue`

src/core/XRPValue.js
clark800: Since quality should always be an IOUValue, I think that won't be an issue.

src/core/amount.js
mwshortt: But the tests are expecting canonicalization according to XRPValue's canonicalize function when the currency is native. This results in assertion errors where it is expecting the rounded value: 
```AssertionError: '90,442.19667786521/XRP' === '90,442.196677/XRP'```

Comment (clark800) [inline src/core/amount.js]: Hmmm, this is another quirk of the prior design: it doesn't really make sense to convert a quality to an amount because a quality is a ratio and an amount doesn't support ratio dimensions.

There are a couple options: throw an exception when from_quality's second parameter is XRP (this would be a breaking change). Or we could put in a hack that uses IOUValue.toFixed to cast to an XRPValue without violating the precision constraint. We should try to avoid breaking changes since this is a refactoring pull request though. Maybe later we can do the breaking change in a separate pull request. Even better would be to just delete Amount.from_quality and replace it with a function (not in amount.js) called `parseQuality` that returns an IOUValue.

Comment (mwshortt) [inline src/core/amount.js]: Even if we use this hack we still run into issues with _check_limits (called from _set_value), because it's then trying to compare XRPValues with the IOUValue created in parse_quality. 
What about initially having it be an IOUValue and only at the end, when _set_value is called, turning it into an XRPValue if it's native? This will allow the >6 decimal place check in the XRPValue constructor to work appropriately

Comment (clark800) [inline src/core/amount.js]: Yeah, that's what I was thinking.

File: src/core/value.js (hunk 35 above)
Comment (mwshortt) [inline src/core/value.js]: I changed this to now call utils.js' method of the same name, passing this._value.abs()

Comment (clark800) [inline src/core/value.js]: I'd like if we can delete this method entirely because it seems too low-level to be on the value class.

**Comment by mwshortt** [comment]
Integrated @clark800 's comments to make Value.js an abstract base class and other fixes. Let me know how it looks!

File: src/core/IOUValue.js [hunk 46]
```diff
@@ -0,0 +1,54 @@
...
+const BigNumber = GlobalBigNumber.another({
+  ROUNDING_MODE: GlobalBigNumber.ROUND_HALF_UP,
+  DECIMAL_PLACES: 40
+});
+
+class IOUValue extends Value {
+
+  constructor(value: string | BigNumber, base: ?number) {
+    super(value, base);
+    this._type = IOUValue;
```
clark800: Can use `instance.constructor`

src/core/IOUValue.js
```diff
@@ -0,0 +1,54 @@
...
+
+  negate() {
+    return new IOUValue(this._value.neg(), undefined);
+  }
+
+  _canonicalize(value) {
+    if (value.isNaN()) {
+      throw new Error('Invalid result');
+    }
+    return new IOUValue(value.toPrecision(16), undefined);
```
clark800: Can we remove all the `undefined`s?

src/core/IOUValue.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  equals(comparator) {
+    return (comparator instanceof IOUValue)
+    && this._value.equals(comparator._value);
```
clark800: Indentation

src/core/XRPValue.js
```diff
@@ -0,0 +1,60 @@
...
+  _canonicalize(value) {
+    if (value.isNaN()) {
+      throw new Error('Invalid result');
+    }
+    return new XRPValue(value.round(6, BigNumber.ROUND_DOWN));
+  }
+
+  equals(comparator) {
+    return (comparator instanceof XRPValue)
+    && this._value.equals(comparator._value);
```
clark800: Indentation

src/core/IOUValue.js
mwshortt: I was getting flow errors if I tried to call a function without all the variables, so I added them in. But I can add a default for base in the IOUValue constructor and that solves the flow errors.

clark800: That sounds like a good option

src/core/utils.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
-function getMantissaDecimalString(bignum) {
-  let mantissa = bignum.toPrecision(16)
-    .replace(/\./, '')      // remove decimal point
+// Should only be used directly when the string is from a BigNumber
```
clark800: I don't think it matters if the string is from a BigNumber. Parameter name could be `decimalString` maybe? Any decimal-encoded numeric string should be the same right? Also, I don't think the function requires 16 digits of precision. It might be good to add a case for if `mantissa.length` is greater than 16 though so that it works for all inputs.

src/core/value.js
```diff
@@ -0,0 +1,121 @@
...
+
+const BigNumber = GlobalBigNumber.another({
+  ROUNDING_MODE: GlobalBigNumber.ROUND_HALF_UP,
+  DECIMAL_PLACES: 40
+});
+
+class Value {
+
+  constructor(value: string | BigNumber, base: number) {
+    if (this.constructor.name === 'Value') {
```
clark800: Does `this.constructor === Value` work?

src/core/value.js
```diff
@@ -0,0 +1,121 @@
...
+  add(addend: Value) {
+    if (this._type !== addend._type) {
+      throw new Error('Cannot add two different Value types');
+    }
+    const result = this._value.plus(addend._value);
+    return this._canonicalize(result);
+  }
+
+  subtract(subtrahend: Value) {
+    if (this._type !== subtrahend._type) {
```
clark800: Could save two lines per method if we convert these to asserts.

src/core/amount.js
```diff
@@ -729,7 +705,9 @@ Amount.prototype.parse_native = function(j) {
 // Requires _currency to be set!
 Amount.prototype.parse_value = function(j) {
   this._is_native = false;
-  this._set_value(new BigNumber(j), BigNumber.ROUND_DOWN);
+  const newValue = new IOUValue(
+    new IOUValue(j).toPrecision(16, Value.getBNRoundDown()));
```
clark800: Looks like the suggestion for this got lost.

src/core/value.js
mwshortt: It does, I will simplify that

src/core/value.js
```diff
@@ -0,0 +1,121 @@
...
+    return this._canonicalize(result);
+  }
+
+  invert() {
+    const result = (new BigNumber(this._value)).toPower(-1);
+    return this._canonicalize(result);
+  }
+
+  round(decimalPlaces: number, roundingMode: number) {
+    this._value = this._value.round(decimalPlaces, roundingMode);
```
clark800: Can we just return the new value like the other methods and avoid mutation?

src/core/amount.js
mwshortt: It seems messier to me to be passing another value into the constructor for one specific case versus calling a specific function. But I can change it

clark800: Yeah, the messiness is the fault of the legacy code that uses two different rounding modes.

The advantage is that we don't have to expose the toPrecision method on the interface. So it's a trade-off between an option on the constructor, vs. an additional method on the interface. I was thinking an option on one method counts as less complexity than a new method, but it's debatable. The other factor is that toPrecision seems less of something that a "Value" should do than having options for how to construct one. What do you think?

src/core/utils.js
mwshortt: You're correct. If the ```mantissa.length``` is greater than 16 should it return it or should it add trailing zeros to a different length (maybe adding a parameter to the function?)?

clark800: I think for our use cases we would probably want to truncate the least significant digits. We might want to add "16" to the name of the function as a reminder.

src/core/amount.js
mwshortt: Yeah it's a close call and I see your point. The BigNumber class does have the toPrecision function so it seems plausible that the Value could do it too. I'm going to go with the constructor option as I agree it's better to not expose the toPrecision method.

mwshortt:
Updated once again

src/core/utils.js
```diff
@@ -1,8 +1,9 @@
...
-function getMantissaDecimalString(bignum) {
-  let mantissa = bignum.toPrecision(16)
-    .replace(/\./, '')      // remove decimal point
+// returns the mantissa from the passed in string,
+// adding zeros until it has 16 sd
+function getMantissa16FromString(decimalString) {
+  let mantissa = decimalString.replace(/\./, '')      // remove decimal point
     .replace(/e.*/, '')     // remove scientific notation
     .replace(/^0*/, '');    // remove leading zeroes
   while (mantissa.length < 16) {
```
clark800: What about when length > 16?

mwshortt: Oh I get what you meant earlier, use a toFixed() method to round it to only be of length 16, right?

src/core/XRPValue.js
```diff
@@ -0,0 +1,59 @@
...
+  DECIMAL_PLACES: 40
+});
+
+const Value = require('./value').Value;
+
+class XRPValue extends Value {
+
+  constructor(value: string | BigNumber) {
+    super(value);
+    this.rippleUnits = new BigNumber(1e6);
```
clark800: This can be a global constant in this module and IOUValue.js, which would actually reduce it's exposure. Also I just noticed, we should probably stick to the file naming convention of all lowercase, no separator between words.

mwshortt: Okay. So rename to xrpvalue and iouvalue?

clark800: yeah

mwshortt:
Updated to include the newest round of comments

src/core/amount.js
```diff
@@ -626,22 +598,26 @@ function(quality, counterCurrency, counterIssuer, opts) {
...
     } else if (baseCurrency.is_valid() && baseCurrency.is_native()) {
       // pay:$price X                   get:1 drop
       // pay:($price * 1,000,000) X     get:1 XRP
-      nativeAdjusted = adjusted.times(Amount.bi_xns_unit);
+      nativeAdjusted = nativeAdjusted.multiply(bi_xns_unit);
     }
   }
+  if (this._is_native) {
+    this._set_value(
+      new XRPValue(nativeAdjusted.round(6, Value.getBNRoundDown())));
```
clark800: It looks like this is passing an IOUValue instance to the constructor of XRPValue.

src/core/amount.js
mwshortt: @clark800 Re your most recent comment: this is why I am passing an IOUValue instance to the constructor of XRPValue.

src/core/utils.js
```diff
@@ -1,16 +1,25 @@
...
-  let mantissa = bignum.toPrecision(16)
-    .replace(/\./, '')      // remove decimal point
+// returns the mantissa from the passed in string,
+// adding zeros until it has 16 sd
+function getMantissa16FromString(decimalString) {
+  let mantissa = decimalString.replace(/\./, '')      // remove decimal point
     .replace(/e.*/, '')     // remove scientific notation
     .replace(/^0*/, '');    // remove leading zeroes
+  if (mantissa.length > 16) {
+    const rounded = Number(mantissa).toFixed(16);
```
clark800: This isn't safe because the maximum javascript integer is 2^53-1 = 9007199254740991 < 10^16-1.

src/core/amount.js
clark800: The signature of the XRPValue constructor is `constructor(value: string | BigNumber)`, so this should be a type error, but I guess flow isn't smart enough to catch it? I think if you just add a `.toString()` it should work.

mwshortt: It does, will fix

src/core/utils.js
mwshortt: Right. Instead should I create a new IOUValue and or throw an error if the mantissa is longer than 16 or is there a better solution I'm missing?

clark800: You can truncate by doing `mantissa.substring(0, 16)`

mwshortt: Hmm okay - so it's alright to truncate versus treating it like a number and rounding it somehow?

clark800: Well truncate is equivalent to round down. You could argue that we should use the same rounding mode as the rest of the program, which is round half up. But because this function is basically getting the first 16 digits of the mantissa, round down seems more natural to me. It could go either way though.

mwshortt:
Updated with the two small fixes

src/core/amount.js
```diff
@@ -38,25 +31,17 @@ function Amount() {
...
-  bi_1e14: new BigNumber(1e14),
-  bi_1e16: new BigNumber(1e16),
-  bi_1e17: new BigNumber(1e17),
-  bi_1e32: new BigNumber(1e32),
-  bi_man_max_value: new BigNumber('9999999999999999'),
-  bi_man_min_value: new BigNumber(1e15),
-  bi_xns_max: new BigNumber(1e17),
-  bi_xns_min: new BigNumber(-1e17),
-  bi_xns_unit: new BigNumber(1e6),
+  // man refers to mantissa
```
clark800: This is good

clark800:
LGTM!!

PR #385, opened 2015-06-25
Convert getOrders and add unit test

clark800:


src/api/index.js
```diff
@@ -10,7 +10,7 @@ const getAccountTransactions = require('./ledger/transactions');
 const getTrustlines = require('./ledger/trustlines');
 const getBalances = require('./ledger/balances');
 // const getPathFind = require('./ledger/pathfind');
-// const getOrders = require('./ledger/orders');
+const getOrders = require('./ledger/orders');
```
sublimator: What does it mean `convert` ?  Is some of the code still ripple-rest specific ?

src/api/ledger/orders.js
```diff
@@ -1,119 +1,37 @@
...
-    });
-
-    return promise;
-  }
-
-  getAccountOrders()
-    .then(respondWithOrders)
-    .catch(callback);
+  const defaultLimit = 100;
+  const limit = options.limit || defaultLimit;
```
sublimator: You are probably aware,  but I noticed es6 has some nifty destructuring syntax, that supports defaults:
```javascript
const {limit = defaultLimit, 
       ledgerVersion = this.remote.getLedgerSequence()} = options;
```

src/api/ledger/parse/account-order.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+const parseOrderBase = require('./order-base');
```
sublimator: What is the `base` denoting?  Just the specification part?

sublimator: Ah right, the base common part for all the {account|book|*}orders I guess?

test/fixtures/get-orders-response.json
```diff
@@ -0,0 +1,307 @@
...
+        "value": "18.46856867857617",
+        "counterparty": "rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q"
+      }
+    },
+    "state": {
+      "sequence": 757002
+    }
+  },
+  {
+    "specification": {
```
sublimator: Why the specification and state split? It seems like specification includes fields that can change, as an offer gets partially executed rather than "this is the order as placed/specified".

src/api/ledger/parse/order-base.js
```diff
@@ -0,0 +1,27 @@
...
+  const takerGetsAmount = parseAmount(takerGets);
+  const takerPaysAmount = parseAmount(takerPays);
+  const quantity = (direction === 'buy') ? takerPaysAmount : takerGetsAmount;
+  const totalPrice = (direction === 'buy') ? takerGetsAmount : takerPaysAmount;
+
+  return utils.removeUndefined({
+    direction: direction,
+    quantity: quantity,
+    totalPrice: totalPrice,
+    passive: ((flags & orderFlags.Passive) !== 0) || undefined,
```
sublimator: So the convention will be to have boolean parameters false by omission?

src/api/ledger/orders.js
```diff
@@ -1,119 +1,37 @@
...
+  const ledgerVersion = options.ledgerVersion
+                      || this.remote.getLedgerSequence();
+  const getter = _.partial(requestAccountOffers, this.remote, account,
+                           ledgerVersion, options);
+  utils.getRecursive(getter, limit,
```
sublimator: Nice! Factored out all the marker/limit stuff huh

sublimator:
Just a few questions, but basically LGTM

src/api/index.js
clark800: Yeah, I copied the `api` directory from ripple-rest and now I'm going through converting all the methods to the schemas and conventions of RippleAPI.

src/api/ledger/orders.js
clark800: Yeah, simplifies things quite a bit.

src/api/ledger/parse/account-order.js
clark800: Yeah

src/api/ledger/parse/order-base.js
clark800: Yeah, omission would either mean "equal to the default value" or "equal to false", but these should be equivalent because false should always be default.

test/fixtures/get-orders-response.json
clark800: It can be interpreted as "this is a specification that you could use to create an order equivalent to the current one". I think it's nice to be able to immediately extract the specification and tweak it to create new transactions rather than needing to filter out all the state properties (either by the user or by the code).

sublimator: [k,lgtm]

src/api/ledger/orders.js
clark800: I didn't know that destructuring accepted defaults, that's cool. But I don't think it makes it more readable in this case.

sublimator: Agreed

PR #386, opened 2015-06-26
Remove message.js

sublimator:
Removing message.js which is available for ripple-vault-client/ripple-client in ripple-lib-extensions repository.

See: https://github.com/ripple/ripple-vault-client/issues/26
See: https://github.com/ripple/ripple-lib-extensions/pull/11


clark800:
LGTM

PR #387, opened 2015-06-26
Refactor API deprecation, fix request constructors with positional args

wltsmrz:


wltsmrz:
buomp

clark800:
Looks like eslint didn't run. I'm not sure what happened, do you know?

wltsmrz:
Nope

clark800:
Hmmm, I wonder if it's because you are using branches in the same repo? Maybe FETCH_HEAD doesn't get set in that case? Try running through the commands in the CircleCI log to see if you can reproduce.

clark800:
We could also echo the value of FETCH_HEAD in the ci.sh script. Might be useful enough to keep for future debugging needs.

PR #388, opened 2015-06-26
Convert getOrderBook and add unit test

clark800:


src/api/ledger/parse/account-order.js
```diff
@@ -1,14 +1,27 @@
 'use strict';
-const parseOrderBase = require('./order-base');
+const utils = require('./utils');
+const flags = utils.core.Remote.flags.offer;
+const parseAmount = require('./amount');
 
 // rippled 'account_offers' returns a different format for orders than 'tx'
```
sublimator: You mean a OfferCreate txn vs the Offer ledger entries ?

src/api/ledger/parse/orderbook-order.js
```diff
@@ -0,0 +1,40 @@
...
+  const takerGetsFunded = order.taker_gets_funded ?
+      parseAmount(order.taker_gets_funded) : undefined;
+  const takerPaysFunded = order.taker_pays_funded ?
+      parseAmount(order.taker_pays_funded) : undefined;
+  const available = utils.removeUndefined({
+    availableQuantity: direction === 'buy' ? takerPaysFunded : takerGetsFunded,
+    priceOfAvailableQuantity: direction === 'buy' ?
+      takerGetsFunded : takerPaysFunded
+  });
+  const state = _.isEmpty(available) ? undefined : available;
```
sublimator: Still digging in, but just an observation:  I noticed `sequence` was in `state` in one place and in `properties` elsewhere.

src/core/remote.js
```diff
@@ -1610,7 +1610,7 @@ Remote.prototype.requestBookOffers = function(options_, callback_) {
   const request = new Request(this, 'book_offers');
 
   request.message.taker_gets = {
-    currency: Currency.json_rewrite(gets.currency, {force_hex: true})
```
sublimator: Has json_rewrite and friends been updated so that force_hex is no longer required? (i.e. automatically uses hex for "rich" currency definitions? )

src/api/ledger/parse/orderbook-order.js
sublimator: Hrmm, nup I was confused it seems.

src/api/ledger/parse/orderbook-order.js
```diff
@@ -0,0 +1,40 @@
...
+  const takerGetsAmount = parseAmount(order.TakerGets);
+  const takerPaysAmount = parseAmount(order.TakerPays);
+  const quantity = (direction === 'buy') ? takerPaysAmount : takerGetsAmount;
+  const totalPrice = (direction === 'buy') ? takerGetsAmount : takerPaysAmount;
+
+  const specification = utils.removeUndefined({
+    direction: direction,
+    quantity: quantity,
+    totalPrice: totalPrice,
+    passive: ((order.Flags & flags.Passive) !== 0) || undefined
```
sublimator: Why only just the `passive flag` ?

src/api/ledger/parse/account-order.js
sublimator: Oh right, book_offers vs account_offers.  The former is using rippled' native json format.

test/fixtures/book-offers-response.js
```diff
@@ -0,0 +1,1181 @@
...
+  _.defaults(options, {
+    gets: {
+      currency: 'BTC',
+      issuer: 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B'
+   },
+    pays: {
+      currency: 'USD',
+      issuer: 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B'
+   }
+ });
```
sublimator: indentation ? ^

src/api/ledger/parse/orderbook-order.js
```diff
@@ -0,0 +1,40 @@
...
+  });
+  // "quality" is omitted intentionally as it corresponds to
```
sublimator: Hrmmm.

So there won't be any kind of price per unit of `quantity` property on the order? There may be precision issues but they're possibly inescapable ones, if you need to provide functionality that things like ripple charts need.  I guess you could call it "approx_price" or something to convey the lack of precision. Intentional or not, this seems like a glaring omission.

People would just be left to calculate the price using quantity/totalPrice. How often would be people want a list of orders, and not want to know the quality/price etc? Maybe it should be provided, using the quality where possible.

Interesting related bit of trivia: orders are always executed at the price they were placed at, not what they may end up at after float/rounding issues have potentially caused TakerPays/TakerGets to lose the initial ratio. 

@vhpoet 
@shekenahglory    

Any thoughts? You guys probably have some insight here.

src/api/ledger/parse/orderbook-order.js
sublimator: Oh, so the passive flag is the only flag that can make it into a ledger entry, and the others are execution time flags. FillOrKill and ImmediateOrCancel imply some degree of immediate crossing required.

sublimator:
Other than the questions about json_rewrite and force hex, and omitting the quality, this basically LGTM.

src/api/ledger/parse/account-order.js
clark800: Rippled has three different formats for offers depending on whether you get them from `tx`, `account_offers`, or `book_offers`.

sublimator: Wah? From `tx`? You mean like as in the transaction metadata?

clark800: Yeah. It contains all the same information, but it even uses different values for the flags.

sublimator: That can't be right.  I can see transaction vs ledger entry having different flags, but the metadata format is basically the same as ledger entry, only modified to show deltas.  You're getting Flags from the NewFields/FinalFields etc ?

clark800: Here are the two different values for indicating whether an offer is a sell, depending on whether it is for the transaction or for the ledger entry:

https://github.com/ripple/ripple-lib/blob/develop/src/core/transaction.js#L118
https://github.com/ripple/ripple-lib/blob/develop/src/core/remote.js#L220

sublimator: Right, the TransactionType:OfferCreate transaction Flags are different to the flags set in the LedgerEntryType:Offer Flags.

sublimator: I thought you meant there was 3 different sets of flags?

clark800: The flags are the same for `account_offers` and `book_offers`, but the naming is different: `taker_gets` vs. `TakerGets`. So there are three different parsers needed.

src/api/ledger/parse/orderbook-order.js
clark800: Yeah, it was in `state` in the last pull request.

src/api/ledger/parse/account-order.js
sublimator: Phew, ok, that's all good then :)

src/api/ledger/parse/orderbook-order.js
clark800: Right, `passive` and `sell` technically, but `sell` is encoded in `direction` here.

sublimator: :+1:

src/api/ledger/parse/orderbook-order.js
clark800: So "quality" is a rippled-specific technical term that means "price or inverse price". I don't think it makes sense to expose this implementation detail in a high-level interface.

Adding a "price" field is dangerous because that "price" and "totalPrice" appear on equal footing, whereas really "totalPrice" is the only one that is actually correct.

Adding "approximatePrice" would be a user-interface convenience, which belongs in the user-interface, not baked into the API.

test/fixtures/book-offers-response.js
clark800: Will fix

src/core/remote.js
clark800: The only place `force_hex` is used is here, where it also checks whether it is a 3-character currency code: https://github.com/ripple/ripple-lib/blob/develop/src/core/currency.js#L366

So I think it should be safe to make this change. Does that look right?

sublimator: I don't think so. Out atm tho in droid. Look for where iso_code gets set.
It's only set for simple currencies? Interest bearing (i know they are
deprecated) currencies don't set them too?

clark800: This means that the only case in which the behavior is different from before is when iso_code is set to a three-character code.

sublimator: Not sure. On droid. Try doing
Currency.from_json(interestBearingCur).to_json() should say a lot.

clark800: > Currency.from_json('015841551A748AD2C1F76FF6ECB0CCCD00000000').to_json()
> 'XAU (-0.5%pa)'

sublimator: rippled aint going to understand that eh?

clark800: I don't know, but same behavior on the develop branch

sublimator: I guess a better experiment would be
'Currency.json_rewrite(interestBearing, {force_hex: true})'. With force_hex
both true and false.

You can check but 99% sure rippled only understands hex (likely reason the
force b4) Arguably to_json() should speak rippled. That rate suffix is for
humanoids.

sublimator: rippled can only handle 3 letter codes, or 40 hex characters for the currency string:
https://github.com/ripple/rippled/blob/develop/src/ripple/rpc/handlers/BookOffers.cpp#L79
https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/UintTypes.cpp#L83-L99

sublimator:
```javascript
> Currency.json_rewrite('015841551A748AD2C1F76FF6ECB0CCCD00000000', {force_hex: false})
'XAU (-0.5%pa)'
> Currency.json_rewrite('015841551A748AD2C1F76FF6ECB0CCCD00000000', {force_hex: true})
'015841551A748AD2C1F76FF6ECB0CCCD00000000'
```

test/api-test.js
```diff
@@ -137,4 +139,18 @@ describe('RippleAPI', function() {
       _.partial(checkResult, getOrdersResponse, done));
   });
 
+  it('getOrderBook', function(done) {
+    const orderbook = {
+      base: {
```
sublimator: It looks like these amounts eventually get passed in verbatim, to the book_offers request

sublimator: Yeah:
```json
{
  "error": "remoteError",
  "error_message": "Remote reported an error.",
  "remote": {
    "error": "invalidParams",
    "error_code": 29,
    "error_message": "Invalid field 'taker_pays.issuer', not string.",
    "id": 1,
    "request": {
      "command": "book_offers",
      "id": 1,
      "ledger_index": "validated",
      "taker": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B",
      "taker_gets": {
        "currency": "USD",
        "issuer": null
      },
      "taker_pays": {
        "currency": "BTC",
        "issuer": null
      }
    },
    "status": "error",
    "type": "response"
  },
  "result": "remoteError",
  "engine_result": "remoteError",
  "result_message": "Remote reported an error.",
  "engine_result_message": "Remote reported an error.",
  "message": "Remote reported an error."
}
```

The `null` is there because of the json_rewrite on the issuer of undefined, which returns `NaN`, which serialises to null by JSON.stringify.

```javascript
>>> console.log(JSON.stringify({issuer: NaN}))
{"issuer":null}
```

src/api/ledger/orderbook.js
```diff
@@ -1,171 +1,54 @@
...
+  return isSameIssue(issue, order.specification.quantity)
+    && order.specification.direction === direction;
+}
+
+function formatBidsAndAsks(orderbook, orders) {
+  // the "base" currency is the currency that you are buying or selling
+  // the "counter" is the currency that the "base" is priced in
+  // a "bid"/"ask" is an order to buy/sell the base, respectively
+  const bids = orders.filter(_.partial(orderFilter, orderbook.base, 'buy'));
+  const asks = orders.filter(_.partial(orderFilter, orderbook.base, 'sell'));
```
sublimator: There was something about this is nagging at me about being off. I can't quite articulate it. I can say `bids.length + asks.length !== orders.length` if that alerts you to anything :)

src/api/ledger/parse/orderbook-order.js
sublimator: > really "totalPrice" is the only one that is actually correct

Incorrect. Only `quality` (when it doesn't need to be inverted)  is correct in the sense that that's the rate you'll pay when executing the order. (Offers get executed at the rate they were placed at. The price/quality is static and retrieved from the directory nodes.  "Implementation detail, nanu nanu" ? 

I could almost buy the argument that if you can't be correct in all cases there's no point in trying, but I think you may as well be as accurate as you can be (like shying away from inverted qualities) 

> Adding a "price" field is dangerous because that "price" and "totalPrice" appear on equal footing,

I didn't really understand that.

I don't really feel strongly about this, but It may be throwing away information that may be of interest to someone. 

Questions are:

What is this api going to be used for? 
Who is going to use it? 
What are they all going to need?
Are all the orders sorted by price as they were before? 
Is there expectations that they will be?

Also, before the asks/bids were quite clearly, and simply grouped via virtue of being the responses to two requests. I'm not a trader, but I imagine people looking at orders want to know order depth etc, in the sense of availability bands. (I can't tell whether there's a bug, or you're trying to provide some new semantics I don't really understand but it seems like not all orders returned by rippled make it into the `{asks: [], bids: []}` object returned by getOrderBook).

Meaning, I'm not quite sure where you're going at with these changes.

src/core/remote.js
clark800: I think this is a bug in the currency class, I created this ticket: https://ripplelabs.atlassian.net/browse/RLJS-403

I'm going to revert this change to avoid scope creep on this pull request.

test/api-test.js
clark800: Nice catch, will fix.

src/api/ledger/orderbook.js
clark800: That's actually expected. The `orders` coming in are all bids and asks for both currencies, whereas the result we return is just the bids and asks for the base priced in terms of the counter. In other words, switching `base` and `counter` in the request will give you an entirely different set of orders.

src/api/ledger/parse/orderbook-order.js
clark800: > Incorrect. 

I meant that "totalPrice" is the only one that is correct out of "price" and "totalPrice". "quality" is correct, but it is specific to rippled, unlike "price", which everyone understands. And I'm not sure if it's any more correct to say that orders get executed at the rate of the "quality" than it is to say they get executed at the ratio specified by "TakerGets/TakerPays". "quality" is basically an intermediate result of the calculation for partial executions and you still have to multiply by the quantity traded in order to get the final result, unless the quantity is exactly 1.0. You could argue that including the intermediate result might reduce the total accumulated error, but it won't (I think) if the client code uses the `IOUValue` class to calculate the partial execution price.

> I didn't really understand that.

If a user sees a result that contains "price" and "totalPrice" they are liable to assume that they can use either one safely, whereas this would not be true because "price" would be an approximation.

> What is this api going to be used for? 
> Who is going to use it? 
> What are they all going to need?

The design goal of this API is to be "the one Ripple API", meaning that ideally it should be able to replace all of rippled's APIs, ripple-lib, ripple-rest, and everything else. For REST servers, the API would have to be slightly different, but the schemas should be the same. Of course, for it to replace rippled's APIs it would have to be ported to a compilable language. But also we can't implement everything at once, so we're just trying to do the most important features first.

> Are all the orders sorted by price as they were before? 
> Is there expectations that they will be?

Yes, I forgot to sort the results, will fix.

> not all orders returned by rippled make it into the object returned by getOrderBook

That's correct. Semantically the request should only return "half" of the results. For convenience though, we may as well return the flipped order book since we already have all the results from rippled. I'll try to find a clean way to do that.

clark800:
@sublimator I think all the issues have been addressed now.

src/api/ledger/parse/orderbook-order.js
sublimator: >>> Semantically the request should only return "half" of the results.

You've changed the usual semantics, giving more weight it seems to the Sell
flag to determine buy/sell - > bid/ask.
Does it really matter to a trader whether someone is semantically buying
USD or selling XRP ( for USD) ?
I guess yes, if only to get a truer sense of how much output is available
in each offer (which is dependent on the execution rate)

Hrmmm.

Was there any discussion about this?
Is it actually a useful change, and a useful distinction to make?
This is not something I'm really well suited to judge.

>>> I'm not sure if it's any more correct to say that orders get executed
at the rate of the "quality" than it is to say they get executed at the
ratio specified by "TakerGets/TakerPays"

You have it the wrong way around. The `quality` is actually `TakerPays
divided by TakerGets` and means how much you must pay for 1 unit of
TakerGets.

And it's more correct, in that the `quality` field in the json, is derived
from the quality baked into the directory nodes where the offer is indexed.
Offers are always executed at this rate, determined at the time of
placement, rather than recalculating the ratio each time. With partial
executions, the ratio may slip.

>>> "price" would be an approximation.

As would totalPrice

>>> they can use either one safely

Personally, I'd rather use quality when it's oriented correctly. i.e. where
possible.

sublimator:
I think it's basically LGTM :+1:  

Would be good if someone else could take a look.

src/api/ledger/parse/orderbook-order.js
clark800: > You've changed the usual semantics

"Usual" according to who? This semantics may not be the way rippled does it, but it is the way the rest of the world does it.

> Is it actually a useful change

It's debatable. The way the FX market deals with this is to have a conventional ordering of each pair based on which currency is considered "stronger" so that you don't have two different books for basically the same thing. We probably want to go with that convention unless we are really convinced we have an objectively better option.

> You have it the wrong way around. 

I think I meant `/` as "slash" not division; I wasn't paying attention to the ordering.

> With partial executions, the ratio may slip.

This is a good point, maybe we will have to show the price. Maybe we can drop the totalPrice and show either "price" xor "inversePrice" depending on the case.

sublimator: > You've changed the usual semantics

I don't have any background in trading/finance. I'm only going by what I've
seen in various ripple products.

It's probably something that should be discussed though.

clark800: I'm going to be giving a talk on the API once the I finish the first pass of the conversion so that everyone can critique. That will probably be followed by another architects mailing list thread. This API is based on the ripple-http-api v2 proposal which has already been posted to architects, though I don't think we went into this level of detail in that discussion.

sublimator: Sounds good.

sublimator:
Damn it, I merged before you squashed :( Sorry about that.

PR #389, opened 2015-06-29
[WIP] Ripple keypairs

sublimator:
Work in progress.  

Using x-address-codec and ripple-keypairs, which use bn.js/elliptic/hash.js


src/core/seed.js
```diff
@@ -4,23 +4,34 @@
...
+  this._version = Base.VER_FAMILY_SEED;
 }, UInt);
 
 Seed.width = 16;
 Seed.prototype = extend({}, UInt.prototype);
 Seed.prototype.constructor = Seed;
+Seed.versions = [Base.VER_FAMILY_SEED, Base.VER_ED25519_SEED];
+
+// Backwards compatibility hack
+Secp256k1Pair.prototype.get_address =
```
sublimator: How much should we strive to maintain backwards compatibility?  

Candidates:  get_address /  pub_key_hex   ?

sublimator: This is just a sketch hack

src/core/utils.js
```diff
@@ -138,6 +140,11 @@ function fromTimestamp(timestamp) {
   return Math.round(timestamp_ / 1000) - 0x386D4380;
 }
 
+function sha256(bytes) {
```
sublimator: Could just use hash.js I guess, or ripple-address-codec, which is x-address-codec bound setup with hash.js and ripple as the default alphabet for exported encode/decode. Leads the question:  Do we want to move away from sjcl in a big bang, or over a few pull requests?

src/core/seed.js
```diff
@@ -62,99 +82,19 @@ Seed.prototype.to_json = function() {
...
-*        {Number}                    specifies the amount of attempts taken
-*                                    to generate a matching KeyPair
-*
-*/
-Seed.prototype.get_key = function(account, maxLoops) {
-  let account_number = 0, address;
-  let max_loops = maxLoops || 1;
-
+Seed.prototype.get_key = function() {
   if (!this.is_valid()) {
```
sublimator: Arguably should throw an error here if there's any arguments, to let people know things have changed.

src/api/transaction/sign.js
```diff
@@ -34,17 +32,12 @@ function hashSerialization(serialized, prefix) {
...
-function signingHash(txJSON, isTestNet=false) {
-  return hashJSON(txJSON, isTestNet ? HASH_TX_SIGN_TESTNET : HASH_TX_SIGN);
+function signingData(txJSON) {
+  return ripple.Transaction.from_json(txJSON).signingData().buffer;
 }
 
 function computeSignature(txJSON, keypair) {
-  const signature = keypair.sign(signingHash(txJSON));
-  return ripple.sjcl.codec.hex.fromBits(signature).toUpperCase();
+  return keypair.signHex(signingData(txJSON));
```
sublimator: Seems like converting from json to binary more than needed in this code. Could be tightened up.

src/core/utils.js
clark800: Smaller steps would be easier to review.

src/api/common/validate.js
```diff
@@ -15,12 +15,8 @@ function validateAddressAndSecret(obj) {
...
     throw error('Parameter missing: secret');
   }
-  try {
-    if (!ripple.Seed.from_json(secret).get_key(address)) {
-      throw error('secret does not match address');
-    }
-  } catch (exception) {
-    throw error('secret does not match address');
+  if (!ripple.Seed.from_json(secret).is_valid()) {
+    throw error('secret is in_valid');
```
clark800: in_valid -> invalid

src/core/base.js
```diff
@@ -1,18 +1,20 @@
...
 
 const alphabets = Base.alphabets = {
+  bitcoin: '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz',
   ripple: 'rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz',
-  tipple: 'RPShNAF39wBUDnEGHJKLM4pQrsT7VWXYZ2bcdeCg65jkm8ofqi1tuvaxyz',
-  bitcoin: '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'
+  tipple: 'RPShNAF39wBUDnEGHJKLM4pQrsT7VWXYZ2bcdeCg65jkm8ofqi1tuvaxyz'
 };
 
+const codecOptions = {sha256, alphabets, defaultAlphabet: 'ripple'};
```
clark800: Couldn't you just pass in one alphabet?

clark800: I still think simple interfaces trump micro-optimizations (regarding sha256 dependecy injection).

test/fixtures/sign-output.json
```diff
@@ -1,4 +1,4 @@
 {
-  "signedTransaction": "12000322000000002400000017201B0086955468400000000000000C732102F89EAEC7667B30F33D0687BBA86C3FE2A08CCA40A9186C5BDE2DAA6FA97A37D87446304402207660BDEF67105CE1EBA9AD35DC7156BAB43FF1D47633199EE257D70B6B9AAFBF02207F5517BC8AEF2ADC1325897ECDBA8C673838048BCA62F4E98B252F19BE88796D770A726970706C652E636F6D81144FBFF73DA4ECF9B701940F27341FA8020C313443",
-  "id": "DB44C111583A95AF973A0B0A40348D90512FCBCDDCA3315A286D2BF4FAC100F1"
+  "id": "771B7F94C1E9E0D3B64654799491D6DB06CD02D1C80EE09B2E63B9B3B7AAEF65",
+  "signedTransaction": "12000322000000002400000017201B0086955468400000000000000C732102F89EAEC7667B30F33D0687BBA86C3FE2A08CCA40A9186C5BDE2DAA6FA97A37D874473045022100B3CCDC7DFD929A45B97C55F2E784FFE372A8CE0B2071FC13458F3B2FC0D03DC60220299C1CC39A1A051A9C1C47FAA5210EC6A72D1102CB5F8811CB21DCF60A6319AB770A726970706C652E636F6D81144FBFF73DA4ECF9B701940F27341FA8020C313443"
```
clark800: Why are these changing?

sublimator: elliptic use rfc6979 deterministic sigs

clark800: Will rippled understand this?

sublimator: easier to change than to try and match the existing random k

clark800: Oh, I guess it shouldn't make a difference to rippled because it should work for any arbitrary value of k right?

sublimator: Yeah, it's just using hmac on the message+secret to create the "random" k.  Rippled will be none the wiser.

sublimator: Yeah, exactly

src/api/common/validate.js
sublimator: w00pz

src/core/base.js
sublimator: Agreed it would be a pita to instantiate it like this everywhere wanted to use it (even though most ripple clients would just use `Base`).

Anyway, it's Transitional code. We can use ripple-address-codec, already bound to hash.js, instantiated with a bunch of alphabets and using ripple alphabet by default.  The alphabets have symbolic names like `ripple` for `rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz` and encode/decode functions both honour `alphabet` arg for non default alphabets.

src/core/utils.js
sublimator: Smaller steps it is

src/core/base.js
clark800: I don't think there's much benefit to having symbolic names for alphabets considering that each application is probably only going to use one of them and `ripple-address-codec` will already have it specified. Having symbolic names means that you have to add an additional method to the interface to register alphabets rather than being able to pass them in directly.

sublimator: We could give the `alphabet` keyword argument actual alphabet semantics as you said, and call the symbolic name something else. `codec` shrug?

I think it's fine as it is. It's not like we're talking about an api that is used all the time, and the verbosity would lead to pain and disgrace. Each app would configure it exactly once, if that. 

As you said  `probably only going to use one of them and ripple-address-codec will already have it specified.`

ripple-client uses the symbolics, eg:  https://github.com/ripple/ripple-client/blob/develop/src/js/directives/validators.directive.js#L160

PR #391, opened 2015-07-01
Convert getPathFind and add unit test

clark800:


src/api/ledger/parse/pathfind.js
```diff
@@ -2,13 +2,23 @@
 'use strict';
 const parseAmount = require('./amount');
 
+// test fixture has an example response where source_amount is an IOU without
```
sublimator: Interesting. Checking.

sublimator: I don't think rippled will ever actually respond with that.

sublimator: It must be that the test fixture was hand crafted

src/api/ledger/pathfind.js
```diff
@@ -1,147 +1,99 @@
...
-        pathfindResults.alternatives.unshift({
-          paths_canonical: [],
-          paths_computed: [],
-          source_amount: pathfindResults.destination_amount
-        });
-      }
-      _callback(null, pathfindResults);
-    });
+function conditionallyAddDirectXRPPath(remote, address, paths, callback) {
+  if (typeof paths.destination_amount.currency === 'string' // why not ==='XRP'?
```
sublimator: So `destination_amount` is supposed to be in rippled format here I guess?

src/api/ledger/pathfind.js
```diff
@@ -1,147 +1,99 @@
...
-    addDirectXrpPath,
-    asyncify(formatPath),
-    asyncify(formatResponse)
-  ];
+function getPathFind(pathfind, callback) {
+  validate.pathfind(pathfind);
 
-  async.waterfall(steps, callback);
+  const address = pathfind.source.address;
+  async.waterfall([
```
sublimator: Probably beyond the scope of this changeset, but wrt: direct XRP, it may be useful to check the account roots of the src and destination, for the (available after reserve) balance of the source and whether the destination accepts XRP. There's a flag that disallows it.

test/fixtures/get-pathfind-response.json
```diff
@@ -0,0 +1,65 @@
...
...000001\"}],[{\"account\":\"rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q\",\"type\":1,\"type_hex\":\"0000000000000001\"},{\"currency\":\"USD\",\"issuer\":\"rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM\",\"type\":48,\"type_hex\":\"0000000000000030\"},{\"account\":\"rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM\",\"type\":1,\"type_hex\":\"0000000000000001\"}],[{\"account\":\"rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q\",\"type\":1,\"type_hex\":\"0000000000000001\"},{\"currency\":\"XRP\",\"type\":16,\"type_hex\":\"0000000000000010\"},{\"currency\":\"USD\",\"issuer\":\"rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM\",\"type\":48,\"type_hex\":\"0000000000000030\"},{\"account\":\"rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM\",\"type\":1,\"type_hex\":\"0000000000000001\"}],[{\"account\":\"rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q\",\"type\":1,\"type_hex\":\"0000000000000001\"},{\"currency\":\"XRP\",\"type\":16,\"type_hex\":\"0000000000000010\"},{\"currency\":\"USD\",\"issuer\":\"rpHgehzdpfWRXKvSv6duKvVuo1aZVimdaT\",\"type\":48,\"type_hex\":\"0000000000000030\"},{\"account\":\"rpHgehzdpfWRXKvSv6duKvVuo1aZVimdaT\",\"type\":1,\"type_hex\":\"0000000000000001\"},{\"account\":\"rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM\",\"type\":1,\"type_hex\":\"0000000000000001\"}]]",
```
sublimator: Have we ever considered stripping the type* fields from these?

src/api/ledger/pathfind.js
sublimator: `'100000000'.currency` would have typeof of `undefined` eh.

src/api/ledger/pathfind.js
```diff
@@ -1,147 +1,99 @@
...
+  if (typeof params.dst_amount === 'object' && !params.dst_amount.issuer) {
+    // Convert blank issuer to sender's address
+    // (Ripple convention for 'any issuer')
+    // https://ripple.com/build/transactions/
+    //     #special-issuer-values-for-sendmax-and-amount
+    // https://ripple.com/build/ripple-rest/#counterparties-in-payments
+    params.dst_amount.issuer = params.dst_account;
+  }
+  if (pathfind.source.amounts && pathfind.source.amounts.length > 0) {
+    params.src_currencies = pathfind.source.amounts;
```
sublimator: So what's going on here?  Why is the `src_currencies` array getting set from `source.amounts`?

sublimator: Where else is `source.amounts` used? Is it a good name?

test/fixtures/get-pathfind-response.json
```diff
@@ -0,0 +1,65 @@
...
+    "allowPartialPayment": false,
```
sublimator: You're not omitting the default values here?

test/fixtures/get-pathfind-response.json
clark800: Do you know what they are for? Not in the documentation: https://github.com/ripple/ripple-dev-portal/issues/126

I would assume they are necessary, but if not, that would be something to address in rippled.

src/api/ledger/pathfind.js
clark800: `source.amounts` specifies what amounts can be sent in the payment. Rippled doesn't support setting value limits in pathfinding yet, so currently it is just filtering based on currency and/or counterparty. I forgot to rename counterparty to issuer again though. Also, I'll remove the `value` property here.

test/fixtures/get-pathfind-response.json
sublimator: Consider that they aren't serialized in a blob. Basically safe to strip if
there was any inclination to.

src/api/ledger/pathfind.js
sublimator: 'inputs' or summin might be a better name than 'amounts'

clark800: `source.amounts` and `destination.amount` is most similar to the schema for `preparePayment`, which uses `source.amount` and `destination.amount`

sublimator: If it's not always an amount - seems like a misleading name. Later it can
be either a currency, a currency/issuer pair or an amount?

sublimator: Sorry 'counterparty' or whatever its called now

clark800: You can think of it as an amount that is not fully specified. Yeah.

test/fixtures/get-pathfind-response.json
clark800: Ok, I'll strip them then. David Schwartz said "Really neither is needed as you can determine the type from the other fields. We need to get API versioning to make progress on issues like that though."

src/api/ledger/pathfind.js
sublimator: Ok, seems like a false symmetry to me, but moving on.

sublimator: Yep, type is implied

src/api/ledger/parse/pathfind.js
clark800: Are you sure? I didn't get an answer to that when I asked in the development channel yesterday.

sublimator: Yeah, pretty sure. I looked at the code and its just basically
'amount.getJson(0)'. Ask again :)

clark800: Ok, I'll delete this and fix the fixture then.

test/fixtures/get-pathfind-response.json
clark800: I considered it, not sure. Seems like it might be safer to be explicit here in case the defaults change.

src/api/ledger/pathfind.js
clark800: Right, I'll extract a method to clarify.

src/api/ledger/pathfind.js
clark800: Yeah, I'll make a note about that for later

src/api/ledger/pathfind.js
sublimator: Oh well, I think it's OK to set the correct XRP scale, maybe even sensible

sublimator: I'll take your word for it. That's not immediately obvious to me.

sublimator: Though curious to know why exactly? The mantissa stays the same?

clark800:
@sublimator I think this is ready

src/api/ledger/pathfind.js
```diff
@@ -1,147 +1,106 @@
...
+  if (typeof params.dst_amount === 'object' && !params.dst_amount.issuer) {
+    // Convert blank issuer to sender's address
+    // (Ripple convention for 'any issuer')
+    // https://ripple.com/build/transactions/
+    //     #special-issuer-values-for-sendmax-and-amount
+    // https://ripple.com/build/ripple-rest/#counterparties-in-payments
+    params.dst_amount.issuer = params.dst_account;
+  }
+  if (pathfind.source.amounts && pathfind.source.amounts.length > 0) {
+    params.src_currencies = pathfind.source.amounts.map(amount =>
```
sublimator: If I'm not mistaken, `pathfind.source.amounts` is in ripple-api format. src_currencies can have the `issuer` specified, but `counterparty` would be passed in?

sublimator: Does this want to be `_.omit(toRippledAmount(amount), 'value')`

src/api/ledger/pathfind.js
```diff
@@ -1,147 +1,106 @@
...
-    request.timeout(utils.common.server.CONNECTION_TIMEOUT, function() {
-      request.removeAllListeners();
-      reconnectRippled();
-      _callback(new TimeOutError('Path request timeout'));
+function addDirectXrpPath(paths, xrpBalance) {
+  // Add XRP "path" only if the source acct has enough XRP to make the payment
+  const destinationAmount = paths.destination_amount;
+  if ((new BigNumber(xrpBalance)).greaterThanOrEqualTo(destinationAmount)) {
+    paths.alternatives.unshift({
+      paths_canonical: [],
```
sublimator: `paths_canonical` is basically never used afaik. I'm not sure if it's officially deprecated, but what matters is what fields the ripple-api formatter requires. `parsePathFind` uses only `paths_computed`.

sublimator:
```
➜  rippled git:(develop) ✗ grep -nsHR "paths_canonical" src                    
src/ripple/app/paths/tests/Path_test.cpp:86:        expectEquals(alt[jss::paths_canonical].size(), 0);
src/ripple/app/paths/tests/Path_test.cpp:172:        expectEquals(alt["paths_canonical"].size(), 0);
src/ripple/protocol/JsonFields.h:272:JSS ( paths_canonical );            // out: RipplePathFind
src/ripple/rpc/handlers/RipplePathFind.cpp:334:                jvEntry[jss::paths_canonical] = Json::arrayValue;
```
The path_find (websocket only path find) command doesn't even set it, and it's always just set to an empty array by the rpc command.

src/api/ledger/pathfind.js
```diff
@@ -1,147 +1,106 @@
...
-          paths_canonical: [],
-          paths_computed: [],
-          source_amount: pathfindResults.destination_amount
-        });
-      }
-      _callback(null, pathfindResults);
-    });
-  }
+function isRippledIOUAmount(amount) {
+  // rippled XRP amounts are specified as decimal strings
```
sublimator: Strictly speaking, rippled will never output an XRP amount as anything other than a drops string. It will never output `{"currency" : "XRP", ... }` afaik.

src/api/ledger/pathfind.js
clark800: Yes, I forgot to add that after your last comment, thanks!

src/api/ledger/pathfind.js
clark800: Good point, removing

src/api/ledger/pathfind.js
clark800: Right, but I figured it's better to be defensive in case the format changes or somebody uses the function on a parsed amount.

test/fixtures/get-pathfind-response.json
clark800: I decided to remove the default values here.

hserang:
LGTM!

PR #392, opened 2015-07-01
Convert server methods, add unit tests, and cleanup

clark800:


src/api/server/server.js
```diff
@@ -1,33 +1,45 @@
...
   this.remote.connect(callback);
 }
 
+function disconnect(callback) {
+  this.remote.disconnect(callback);
+}
+
+function isUpToDate(remote) {
+  const server = remote.getServer();
+  return (server ? true : false) && (remote._stand_alone
```
sublimator: `Boolean(server)` ?

src/api/server/server.js
```diff
@@ -1,33 +1,45 @@
...
+    || (Date.now() - server._lastLedgerClose) <= CONNECTION_TIMEOUT);
+}
+
 function isConnected() {
-  return common.server.isConnected(this.remote);
+  return !isNaN(this.remote._ledger_current_index) && isUpToDate(this.remote);
```
sublimator: `0` is not a valid ledger index, and is `falsy` along with undefined ( the default value ). I'd probably just go with `this.remote._ledger_current_index && $expr`.  `Not is Not A Number` is mental mouthful.

sublimator:
LGTM

PR #393, opened 2015-07-02
Replace "totalPrice" with "price" in orders

clark800:


clark800:
This introduces rounding errors when inverting quality, but using totalPrice (calculated from TakerGets and TakerPays) will also have rounding errors when orders are partially executed because quality is fixed when the order is created. The only way I can think of to prevent rounding errors is to expose the value of "quality" directly, which is antithetical to the goal of having a higher-level interface.

test/fixtures/get-orderbook-response.json
```diff
@@ -324,9 +324,9 @@
           "value": "0.5",
           "counterparty": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
         },
-        "totalPrice": {
+        "price": {
```
sublimator: Hrmmmm, this doesn't quite feel right. I feel like `price` alone is somewhat ambiguous. Also, it just feels odd looking at json with just the input/output quantity and not the totalPrice.

Just throwing this out there: What do you think about leaving `totalPrice` and setting the `price` as just a string field `rate`?  Another option is to frame `quality` as `output_rate`, as offers regardless of buy/sell direction, have constant inputs and outputs.

src/api/ledger/parse/utils.js
```diff
@@ -1,10 +1,25 @@
 /* @flow */
 'use strict';
 const _ = require('lodash');
+const BigNumber = require('bignumber.js');
 const transactionParser = require('ripple-lib-transactionparser');
 const toTimestamp = require('../../../core/utils').toTimestamp;
 const utils = require('../utils');
 
+/*:: type XRPAmount = {currency: string, value: string} */
```
sublimator: Can these definitions be placed somewhere accessible so they don't need to be repeated locally?

src/api/ledger/parse/orderbook-order.js
```diff
@@ -4,22 +4,25 @@ const utils = require('./utils');
...
+}
+
 function parseOrderbookOrder(order: Object): Object {
   const direction = (order.Flags & flags.Sell) === 0 ? 'buy' : 'sell';
   const takerGetsAmount = parseAmount(order.TakerGets);
   const takerPaysAmount = parseAmount(order.TakerPays);
   const quantity = (direction === 'buy') ? takerPaysAmount : takerGetsAmount;
   const totalPrice = (direction === 'buy') ? takerGetsAmount : takerPaysAmount;
+  const price = replaceValue(totalPrice, (direction === 'buy') ?
+    utils.invertQuality(order.quality) : order.quality);
```
sublimator: Wait, so the `quality` when a native amount is in terms of drops. If pays is native, then you need to scale(-6) and so on.

sublimator: It might pay to have some fixtures that include XRP

sublimator: It might be interesting to port [xrpvalue.com](http://xrpvalue.com/) to this new API.

test/fixtures/get-orders-response.json
```diff
@@ -7,9 +7,9 @@
         "value": "17.70155237781915",
         "counterparty": "rMwjYedjc7qqtKYVLiAccJSmCwih4LnE2q"
       },
-      "totalPrice": {
+      "price": {
```
sublimator: I noticed the properties doesn't have `maker`.  So I guess this is basically a wrapper around `account_orders` ?

sublimator: Err `account_offers`

test/fixtures/get-orderbook-response.json
clark800: What is ambiguous? I don't see the advantage of including totalPrice if we already are specifying price. I think calling it outputRate is just as confusing as calling it quality.

test/fixtures/get-orders-response.json
clark800: Yeah. I'm not sure which is better, having consistency with orderbook orders by showing maker or reducing the redundancy by not showing the same maker in all the results. Do you feel strongly either way?

src/api/ledger/parse/orderbook-order.js
clark800: Ok, will fix

src/api/ledger/parse/utils.js
clark800: Maybe, I haven't had time to try it yet: https://github.com/facebook/flow/issues/267

test/fixtures/get-orderbook-response.json
sublimator: Yeah, it's just what I'm used to looking at I think. The advantage of doing
output_rate or whatever you'd call it is it wouldn't be inverted, and
depending on what the api client wanted to do, could lead to more accurate
math.

Consider that you've grouped the orders into 4 buckets. To determine how
much X you'd get for n Y (execution prediction) you'd need to do two
different kinds of math anyway. One using offers that are 'selling x' and
the other for 'buying y'. There's no nice and simple 'price' to apply.

I think output_rate is at least clearer than quality fwiw. I think most ppl
could deduce the meaning, once they worked out that an offer has
inputs/outputs.

clark800: I think it might be reasonable to call "quality" `exchangeRate`. What do you think?

clark800: One serious problem though is that we only know the quality for orderbook orders, not for offercreate transactions or account orders. We can compute it for offercreate transactions, but I think for account orders it might not be possible to show the correct exchange rate for orders that have been partially executed. Maybe we have to ask rippled to show it? Or do account_offers not get updated after a partial execution?

sublimator: Wait I was confusing myself anyway I think? output_rate is not really right
huh?

exchangeRate might work. It would always be what is named quality ?

Basically, I think ideally the quality, uninverted would be available for
whatever math. If you can find a good name for it ?

sublimator: account_offers doesn't include the quality?

clark800: Yeah, that's what I was thinking. `exchangeRate` is always equal to quality, and is computed in javascript for offercreate transactions and account_offers, which only specify takerGets and takerPays.

I'm also thinking maybe orders can be specified by "totalPrice" and "totalPrice" can also be returned in queries as before, but we can add "exchangeRate" to the "properties" section so that users have access to the more robust math.

clark800: > account_offers doesn't include the quality?

Not according to the docs

sublimator: If not, it can be added. It's (quality) packed into the BookDirectory field
so even though aren't walking offer directories, but rather the owner
directories shouldn't be a problem.

sublimator: Yeah, it's not included:
https://github.com/ripple/rippled/blob/develop/src/ripple/rpc/handlers/AccountOffers.cpp#L119-L124

sublimator: Nor is the *funded eh

clark800: Even `exchangeRate` could be deceptive because of the issue of XRP being expressed in drops... I guess we could just try to document that snag.

clark800: If we scale it we introduce rounding errors again.

clark800: I'm not sure if it would actually cause problems, but the mantissa wouldn't be the same because the exponent of floats represents the binary exponent, so you only get lossless multiplication by powers of 2.

sublimator: Wah ? I thought we had a bignumber lib that used decimals ?

sublimator: Basically:

(desired) input divideBy quality == output
(desired) output multiply quality == input

There's not a term for `quality` that neatly expresses the above in one
word?

(Though I agree exchangeRate would be easily recognised)

sublimator: > I'm not sure if it would actually cause problems

Anyway, if it's not a problem, it would probably be convenient to be in XRP
scale, given drops aren't really a thing in this api.

clark800: Oh yeah, I was thinking in terms of javascript math. I guess we could do it losslessly with the bignumber library. Users would have to use our math libraries to reproduce the calculations anyways, so I guess it's reasonable to assume they won't be using javascript math if they want accurate results.

test/fixtures/get-orders-response.json
sublimator: Consistency is generally a nice thing. @wltsmrz  ?

test/fixtures/get-orderbook-response.json
sublimator: Can just suffix the drops string with `'e6'` or `'e-6'` probably. Assuming that syntax works with bignumber ??

sublimator: Although rippled might already be doing that ...

sublimator:
Where are we with this? What's the plan to go forward

##  Summary

1.  Considering restoring `totalPrice`
2.  Considering exposing the `quality`, drops->xrp adjusted, as `exchangeRate` in the properties object
3.  For consistency sake need `quality` to be available in `account_offers` also.

## quality -> properties.exhangeRate -> ??? 

I was reading some stellar issues, and they too have had  confusion around rate/quality properties etc. I think we should be on the lookout for an alternative to `exchangeRate` that neatly expresses the following:

```
(desired) input divideBy quality == output
(desired) output multiply quality == input
```

`priceForOneOut` doesn't really roll off the tongue, but I think once a mind was oriented to the meaning of `out` it would be pretty easy to remember the meaning. 

Ideas?

sublimator:
I opened up a pull request adding `quality` to account_offers: https://github.com/ripple/rippled/pull/1168

clark800:
Awesome, that will be very helpful!

test/fixtures/get-orders-response.json
clark800: Yeah, I'm thinking consistency is more important.

clark800:
I think `makerExchangeRate` works to resolve the ambiguity of `exchangeRate`. I'm closing this pull request and replacing it with: https://github.com/ripple/ripple-lib/pull/398

Issue #28, opened 2013-11-18
Get a error when submitting a trade offer to the network.


enkows:
``` javascript
/* Loading ripple-lib Remote and Amount modules in Node.js */ 
var Remote = require('ripple-lib').Remote;
var Amount = require('ripple-lib').Amount;

/* Loading ripple-lib Remote and Amount modules in a webpage */
// var Remote = ripple.Remote;
// var Amount = ripple.Amount;

var MY_ADDRESS = 'rrrMyAddress';
var MY_SECRET  = 'secret';

var BUY_AMOUNT = Amount.from_human('100XRP');
var SELL_AMOUNT = Amount.from_human('1USD');

var remote = new Remote({ /* Remote options */ });

remote.connect(function() {
  remote.set_secret(MY_ADDRESS, MY_SECRET);

  var transaction = remote.transaction();

  transaction.offer_create({
    from: MY_ADDRESS, 
    buy: BUY_AMOUNT, 
    sell: SELL_AMOUNT
  });

  transaction.submit(function(err, res) {
    /* handle submission errors / success */
  });
});
```

I run the demo code with my ripple address and secret, but got this error.

```
TypeError: Error serializing "TakerGets": Cannot read property 'length' of null
```

PR #397, opened 2015-07-03
Added issuer to payment example in guide

lukeburns:
I added issuer information to the payment example. Without it, you get an unhelpful serializing error. 

Solution due to https://github.com/ripple/ripple-lib/issues/28

### DIFF
--- a/docs/GUIDES.md
+++ b/docs/GUIDES.md
@@ -187,7 +187,7 @@ var Amount = require('ripple-lib').Amount;
 var MY_ADDRESS = 'rrrMyAddress';
 var MY_SECRET  = 'secret';
 var RECIPIENT  = 'rrrRecipient';
-var AMOUNT     = Amount.from_human('1XRP');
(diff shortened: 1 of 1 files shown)


sublimator:
I think the from_human parser was changed at one point to require a space between the amount value and the currency. Maybe it just needs to be changed from `1XRP` to `1 XRP`.

lukeburns:
I was wrong. The example works fine (and `from_human` parses both the same). If you're using any other currency, you need to use `set_issuer`, otherwise you get the serializing error.

sublimator:
That makes sense, any serialized amounts, must have issuer specified, even if it's just some neutral issuer ( e.g. the issuer in Balance fields in a RippleState (trustline))

lukeburns:
I think the guide could use some work, so people can play with the examples without immediately getting caught up in errors. Not sure this is the best solution, but it would have saved me and the people in issue #28 a couple hours.

sublimator:
Noted :)

clark800:
LGTM

PR #398, opened 2015-07-07
Add makerExchangeRate (quality) to orders, and maker to account orders

clark800:


src/api/ledger/parse/utils.js
```diff
@@ -4,6 +4,15 @@ const _ = require('lodash');
 const transactionParser = require('ripple-lib-transactionparser');
 const toTimestamp = require('../../../core/utils').toTimestamp;
 const utils = require('../utils');
+const BigNumber = require('bignumber.js');
+
+function adjustQualityForXRP(quality: string, takerGetsCurrency: string,
+    takerPaysCurrency: string) {
+  const shift = (takerGetsCurrency === 'XRP' ? 1 : 0)
```
sublimator: This code seems kind of convoluted, but more importantly, would not the double negative here cause the set of realistic values to be [0, 1] ?  

Suggest:
```javascript
const shift =  takerGetsCurrency === 'XRP' ? 6 : 
               takerPaysCurrency === 'XRP' ? -6 :  0;

return shift ? (new BigNumber(quality)).shift(shift).toString() : 
               quality;
```

Maybe a test?

clark800: Oops, yeah should only be one negative.

clark800: Now fixed, including changes to fixtures that were using it. I didn't use the nested ternary because that seems just as convoluted and doesn't give the expected result for XRP/XRP, which doesn't occur, but can still be misleading to the reader.

src/api/ledger/parse/orderbook-order.js
```diff
@@ -11,19 +11,22 @@ function parseOrderbookOrder(order: Object): Object {
   const quantity = (direction === 'buy') ? takerPaysAmount : takerGetsAmount;
   const totalPrice = (direction === 'buy') ? takerGetsAmount : takerPaysAmount;
 
+  // note: immediateOrCancel and fillOrKill orders cannot enter the order book
+  // so we can omit those flags here
```
sublimator: Helpful

src/api/ledger/parse/utils.js
```diff
@@ -4,6 +4,15 @@ const _ = require('lodash');
 const transactionParser = require('ripple-lib-transactionparser');
 const toTimestamp = require('../../../core/utils').toTimestamp;
 const utils = require('../utils');
+const BigNumber = require('bignumber.js');
+
+function adjustQualityForXRP(quality: string, takerGetsCurrency: string,
+    takerPaysCurrency: string) {
+  const shift = (takerGetsCurrency === 'XRP' ? 6 : 0)
```
sublimator: It's correct at least now :) Arguably it was only incorrect before due to this convoluted construct. Do we have some hard and fast rule to avoid nested ternary expressions or something?

test/fixtures/get-orders-response.json
```diff
@@ -14,7 +14,9 @@
       }
     },
     "properties": {
-      "sequence": 719930
+      "maker": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
```
sublimator: Went with adding the maker eh

src/api/ledger/parse/utils.js
clark800: No, nested ternary is legal, but to make it *logically* correct, it would be:

`const shift =  (takerGetsCurrency === 'XRP') ? (takerPaysCurrency === 'XRP' ? 0 :  6) : 
               (takerPaysCurrency === 'XRP' ? -6 :  0);`

The subtraction method looks more readable to me.

test/fixtures/get-orders-response.json
clark800: Yeah

src/api/ledger/parse/utils.js
sublimator: > The subtraction method looks more readable to me.

Can't account for taste.

> logically correct

Actually, it's illegal for both to be XRP, so you can do basically `if (isXRP(gets)) { ... } else if (isXRP(pays)) { ... } else  {...}`

clark800: > it's illegal for both to be XRP

Yeah, as I said in the prior comment, even though it doesn't happen, it can still be misleading if the logic doesn't follow what makes sense.

> Can't account for taste.

It's also fewer characters :)

sublimator:
I have doubts about the way this is heading, but it seems correct, and I don't have experience with order/books etc outside of ripple, so I can only trust that this is aligning with more common convention.

src/api/ledger/parse/utils.js
sublimator: > It's also fewer characters :)

Only if you try and account for nonsense input

sublimator: lol, you could actually argue that it's more misleading to account for cases that are illegal. the fact that a bug slipped through, kinda spoke to the readability of the construct. 

I also found inspecting the orders kind of hard. When I noticed the bug, I looked at the test cases, assuming that we must have been missing the case the bug messed up. But it seems some of the test cases were also wrong. 

The way that the quantity is oriented towards the direction makes it a little harder to eyeball the `makerExchangeRate` for correctness. I'm not sure if it's just a habit, but I feel myself wanting to orient towards inputs/outputs and takerPays/takerGets.

A lot of this doesn't really feel right to me. To me the sell flag just defines when the order stops executing in the case when the maker gets a better exchange than what they proposed. If they wanted to get 500 XRP and give 1 USD, and they could make use of an offer giving 1000 XRP for 1 USD, do they give up their whole USD? or just 50 cents?

I'd like to see how the default grouping of bids/asks and inverseBids/inverseAsks is going to be used. To me, a more useful distinction for asks/bids is inputs(takerPays) and outputs(takerGets).  Admittedly, I'm coming from limited experience with (mostly just reading) order book display code.

clark800: You can do a google search for "order book"s and you'll see that 100% of them are expressed as quantity/price. I've never seen one use "taker gets" and "taker pays" besides in Ripple. Here's an example: https://www.bitstamp.net/market/order_book/

There is still a question about what to do with the redundancy of the flipped book. Traditional financial markets just have conventions to not use one of the two books.

sublimator: It's not so much the terms as much as their semantics. I'll try and
articulate it better, but on something else atm. I did look at bitstamp api
the other day, and noticed it returned asks/bids with a type buy/sell.

Anyway, as you said, you'll be seeking feedback from people after this
initial phase is complete, so I'm happy to take it on faith for the moment
and reserve judgement.

sublimator: Hrmm, it seems I was thinking of this: `https://www.bitstamp.net/api/transactions/` when I recalled a "type" 

It is interesting to look at that schema though:
```
date - unix timestamp date and time
tid - transaction id
price - BTC price
amount - BTC amount
type - buy or sell (0 - buy; 1 - sell)
```
That's certainly simpler:
* How much do you want to buy or sell? (amount)
* How much will be paid for each? (price)

We aren't trying to smooth/wash over something that's just fundamentally different?

> Traditional financial markets just have conventions to not use one of the two books

Do you have any examples I could look at?

clark800: > That's certainly simpler

It's also exactly what the new API is trying to do! The only differences are that:
1. we are forced to use `totalPrice` instead of `price` if we want to avoid rounding errors
2. we're using the name `direction` instead of `type`, and I think `direction` is better because it's more specific
3. we're using the name `quantity` instead of `amount`, and `quantity` is the more standard term

In summary, the data structures are:
RippleAPI: direction/quantity/totalPrice
"Textbook": direction/quantity/price
Bitstamp: type/amount/price
rippled: tfSell/takerGets/takerPays (different basis; takerGets/takerPays don't map directly to quantity/totalPrice)

> Do you have any examples I could look at?

It might be hard to find examples, but you can easily find explanations, like here: "the EUR/USD currency pair is always quoted indirectly. The reason for this is mostly convention." http://www.investopedia.com/ask/answers/06/eurusd.asp

sublimator: @mDuo13 made an interesting observation recently in #cpp-informal that the sell flag is meaningless once an order is placed on to the books, as at that point it's not possible to get a better rate. 

Again, I wonder if the {asks, bids, flipped: {asks, bids}} grouping is really ideal, and whether the self flag is a great basis for orienting the quantity to either of takerGets and takerPaid. 

With a more consistent orientation, that's not dependent on the sell flag, the makerExchangeRate would also have a consistent relationship with the quantity and totalPrice fields.

> also exactly what the new API is trying to do

Whether it's a buy or sell type, the amount is always, the amount of BTC.  I guess that's equivalent to the `quantity` being whatever is being sold or bought.

Still, if the sell flag is ultimately meaningless, for orders in a book, I wonder what *is* meaningful? To me it's that the maker is wanting to exchange A for B. 

I'm trying to keep an open mind, but this still feels like an unnatural superimposition (or as I put it before, "trying to smooth/wash over something that's just fundamentally different" )

clark800: Yeah, this is just another way of saying that the `flipped` book is totally redundant, which is true.

Thinking about this makes me realize that we do have the option of merging the flipped book into the non-flipped book. The only disadvantage would be that some orders would appear differently than the way they were placed (they would have their direction flipped). I think it's probably better to merge because I don't think anyone really cares how the order was placed. I'll make a pull request.

PR #400, opened 2015-07-08
Refactor to make quality adjustment for XRP more clear

clark800:


src/api/ledger/parse/utils.js
```diff
@@ -8,10 +8,11 @@ const BigNumber = require('bignumber.js');
 
 function adjustQualityForXRP(quality: string, takerGetsCurrency: string,
     takerPaysCurrency: string) {
-  const shift = (takerGetsCurrency === 'XRP' ? 6 : 0)
-              - (takerPaysCurrency === 'XRP' ? 6 : 0);
-  return shift === 0 ? quality :
-    (new BigNumber(quality)).shift(shift).toString();
+  const log10NumeratorAdjustment = (takerPaysCurrency === 'XRP' ? -6 : 0);
```
sublimator: I think the log10*Adjustment names are a bit heavy handed. People know what a shift is, or at least would, with a comment including "e-6 XRP". The brackets around the *adjustment ternary expressions are simply noise. Also, any non zero number is truthy, so you can invert the logic of the return expression:

```javascript
// quality = takerPays/takerGets, using drops (1e-6 XRP) for native operands
const numeratorShift = takerPaysCurrency === 'XRP' ? -6 : 0;
const denominatorShift = takerGetsCurrency === 'XRP' ? -6 : 0;
const shift = numeratorShift - denominatorShift;
return shift ? (new BigNumber(quality)).shift(shift).toString() : 
               quality;
```

As clever as that is, XRP/XRP is ultimately illegal, and asserting the invariants would make more sense than being so accommodating: 

```javascript
assert(!(takerPaysCurrency === 'XRP' &&  takerGetsCurrency === 'XRP'));
```

And having done that it seems like you could jump to the chase, and simply handle the 3 realistic cases:
```javascript
const shift = takerPaysCurrency === 'XRP' ? -6 :  
              takerGetsCurrency === 'XRP' ? 6 : 0;
return shift ? (new BigNumber(quality)).shift(shift).toString() : 
               quality;
```

clark800: I definitely will add the comment, I think that helps a lot.

I think explicitly using the "numerator"/"denominator" helps a lot here for clarity, so I don't want to get rid of that. I'm ok with `numeratorShift`/`denominatorShift`, though it is technically less clear. The `log10` naming makes it more obvious why we are subtracting.

I generally prefer to avoid using "truthiness" as it is more bug-prone than strict evaluation, though I do cheat on that rule quite a bit.

My take on the question of whether to handle the XRP/XRP case is that it's usually better to solve the general problem if it's just as simple as solving a specific problem. For someone who isn't familiar with qualities, I'm guessing it would be confusing why the XRP/XRP case doesn't seem to be handled. I would probably initially think it was a bug. And if the protocol ever changes to make XRP/XRP quality valid, then this function will work unchanged. Or someone might have a ratio of amounts and think "quality is a ratio of amounts", but then be surprised when the function fails on the XRP/XRP case. These aren't supposed to be realistic examples, but I'm just trying to describe the general principle that guides my opinion.

sublimator:
> I'm guessing it would be confusing why the XRP/XRP case doesn't seem to be handled

Hence an assert 

> if the protocol ever changes to make XRP/XRP quality valid,

I can't imagine it ever making sense for people to make offers giving 1000 XRP for 500 XRP. There's already support for giving money away ... with a direct Payment :)

PR #403, opened 2015-07-08
Get all results when limit is not specified

clark800:


src/api/ledger/orders.js
```diff
@@ -10,7 +10,7 @@ function requestAccountOffers(remote, address, ledgerVersion, options,
   remote.requestAccountOffers({
     account: address,
     marker: marker,
-    limit: limit,
+    limit: Math.min(Math.max(limit, 10), 400),
```
sublimator:
```javascript
> Math.max(undefined, 10)
NaN
```
An issue?

src/api/ledger/utils.js
```diff
@@ -29,7 +29,7 @@ function getRecursiveRecur(getter, marker, limit, callback) {
 }
 
 function getRecursive(getter, limit, callback) {
-  getRecursiveRecur(getter, undefined, limit, callback);
+  getRecursiveRecur(getter, undefined, limit || Infinity, callback);
```
sublimator:
```javascript
> NaN || Infinity
Infinity
```

sublimator: Wait, so I was confused I think, it won't be NaN because limit can't be passed in as undefined, only Infinity. And else where, the Math.min/Math.max stuff should always work fine with `limit=Infinity` to clamp so `10 <= limit <= 400`

src/api/ledger/transactions.js
```diff
@@ -43,7 +42,7 @@ function getAccountTx(remote, address, options, marker, limit, callback) {
     ledger_index_max: options.maxLedgerVersion || -1,
     forward: options.earliestFirst,
     binary: options.binary,
-    limit: Math.max(limit || DEFAULT_LIMIT, 10),
+    limit: Math.min(Math.max(limit, 10), 400),
```
sublimator: Maybe some clamp helper?

src/api/ledger/utils.js
clark800: Right

src/api/ledger/transactions.js
clark800: Added

PR #407, opened 2015-07-09
Add support for efficient range checking to RangeSet

clark800:


src/core/rangeset.js
```diff
@@ -1,67 +1,61 @@
...
-  ledger = Number(ledger);
+  addValue(value: number) {
+    this.addRange(value, value);
+  }
 
-  return this._ranges.some(function(r) {
-    return ledger >= r.start && ledger <= r.end;
-  });
-};
+  addParsedRanges(rangesString: string) {
```
wltsmrz: You're really adding unparsed ledger ranges, right? Maybe `addRanges` is good enough

test/rangeset-test.js
```diff
@@ -1,78 +1,49 @@
...
-      r.add(void(0));
-    });
-    assert.throws(function() {
-      r.add('a');
-    });
+  it('addValue()/addRange() -- malformed', function() {
+    const r = new RangeSet();
     assert.throws(function() {
-      r.add('2-1');
+      r.addRange(2, 1);
```
wltsmrz: How about a test for `addParsedRanges`?

src/core/rangeset.js
clark800: An unparsed range would be a string, a parsed range is a range. The other name I considered was `parseRangesAndAdd`, but I thought `addParsedRanges` sounded simpler.

clark800: I can see that it's ambiguous though. Is `parseRangesAndAdd` better?

wltsmrz: I'd vote for either addRanges or addSerializedRanges; addParsedRanges seems confusing

wltsmrz:
LGTM otherwise

src/core/rangeset.js
clark800: I went with `parseAndAddRanges` since it's the least ambiguous.

PR #412, opened 2015-07-10
Orderbook br8k

wltsmrz:


clark800:
Needs rebase

sublimator:
Right, it seems that "add switch to disable transaction submission" commit is unintentional?

sublimator:
What's `br8k` mean?

src/core/orderbook.js
```diff
@@ -933,41 +933,44 @@ OrderBook.prototype.insertOffer = function(node) {
...
-  }
 
-  if (this._offers.length === originalLength) {
-    this._offers.push(offer);
-  }
-
-  this.incrementOwnerOfferCount(offer.Account);
-
-  this.updateOwnerOffersFundedAmount(offer.Account);
+    if (this._offers.length === originalLength) {
```
clark800: Is there a reason why this part got moved around?

src/core/orderbook.js
```diff
@@ -933,41 +933,44 @@ OrderBook.prototype.insertOffer = function(node) {
...
+      this._offers.push(offer);
+    }
 
-  this.emit('offer_added', offer);
+    this.incrementOwnerOfferCount(offer.Account);
+    this.updateOwnerOffersFundedAmount(offer.Account);
+    this.emit('offer_added', offer);
+  } catch (e) {
```
clark800: Does all of this have to be inside the try? Isn't it just `subtractOwnerOfferTotal` that will cause the exception? https://gist.github.com/clark800/265e01d057bf751c38c9

src/core/orderbook.js
```diff
@@ -933,41 +933,44 @@ OrderBook.prototype.insertOffer = function(node) {
...
 
-  offer.LedgerEntryType = node.entryType;
-  offer.index = node.ledgerIndex;
+    offer.LedgerEntryType = node.entryType;
+    offer.index = node.ledgerIndex;
 
-  // We're safe to calculate quality for newly created offers
-  offer.quality = takerPays.divide(takerGets).to_text();
+    // We're safe to calculate quality for newly created offers
+    offer.quality = takerPays.divide(takerGets).to_text();
```
sublimator: Outside of scope of this pull request, but we could look at the BookDirectory for the quality too

src/core/orderbook.js
sublimator: Personally I'd rather something blow up (fail fast) than try and continue on after these sorts of errors.  Wouldn't it be better to fix the bug?

clark800: The issue is that this bug is really hard to figure out and it's really hard just to reproduce (Brandon couldn't reproduce after running for over a day).

sublimator: How often has it been seen in the wild? When/where was it first reported?

clark800: Apparently pretty rarely. https://github.com/ripple/ripple-lib/issues/352

sublimator: That guy was subscribing to a (probably large) product of currency/issuer pairs. I wonder what the reproduction frequency would be like doing something similar?

I don't think just swallowing exceptions and battling on in a broken state makes much sense.

>>> after 20 minutes more or less the error appears

That's common enough to warrant a fix.

clark800: Take a crack at it.

sublimator: So I just got an old ledger dump of 14.2e6 or near about, and ran a java program I had laying about, which finds all the order books, and then subscribed to ~50 issue pairs that had the most offers, and got a repro in about 30 seconds of running the script.  I saved the output, but am running again ...

clark800: Nice. If you can set it up so that you can reproduce it offline on demand, that would make things a lot easier.

sublimator: Yeaaahp

sublimator: So, looking at the logs from the repro just then, and also at your gist, I notice that both shortly before show a `server connect`, which may indicate dropped transactions, and thus lost state. It may not be handling disconnects/reconnects properly. Looking more ...

sublimator:
```javascript
  this._remote.on('disconnect', function() {
    self.resetCache();
```

```javascript
/**
 * Reset cached owner's funds, offer counts, and offer sums
 */

OrderBook.prototype.resetCache = function() {
  this._ownerFunds = {};
  this._ownerOffersTotal = {};
```

Seems later on that the getOwnerOfferTotal will return 0 by default, rather than undefined. So if upon disconnect, and we're already subscribed to `transactions`, and on reconnect we get a OfferDelete, and in `subtractOwnerOfferTotal` it tries to subtract from 0 and ends up negative?

I guess we need to ignore/buffer transactions until we are ready to process them (once subscribe() async series has finished)

Another thing I noticed is that we aren't requesting the book offers using ledger_index validated or the like, and will be getting some random state of an open ledger, due to ledger_index being undefined in the book_offers requests.

sublimator:
```javascript
OrderBook.prototype.notify = function(transaction) {
  const self = this;

  if (!this._subscribed) {
    return;
  }
```

There's some mechanics to ignore txn notifications when not in a `subscribed` state it seems, but even that doesn't seem to be set very meaningfully. I think it is set as soon as the subscribe command responds, and only ever unset when manually unsubscribed.

sublimator: I think we need to rejig the subscription mechanisms for all this if we want to
have any hope that it's solid.

We need to work from a set of offers (and account infos), using a validated
ledger, and then buffer the transactions, while we piece all that together,
until we are ready to apply them over consistent state.

I can't really think of any other way that this would work. Currently this looks
like it would be pretty random, hence the assert triggering now and then.

sublimator: I you run reconnect() on a timeout from inside the connect() method of Server, you can fairly easily reproduce the bug without elaborate setup etc. If you do this, another bug will be exposed, in the _sendMessage method, where it tries to send a message when the socket isn't open.

```javascript
Server.prototype._sendMessage = function(message) {
  if (this._ws) {
    if (this._remote.trace) {
```

You need to add something like  `&& this._socket_open` (set true of course when the, err, socket is open ) 

Here's a list of (relatively) active orderbooks to subscribe to:
https://gist.github.com/sublimator/995713d2fcff383e7221

You can `fix` the bug (read: make it much much less likely to happen) with:
```
OrderBook.prototype.resetCache = function() {
  this._subscribed = false;
```

A solid fix would need ensuring we have consistent states before applying transactions, and making sure we apply all transactions in a ledger, in the right order (say we are connected to multiple servers, even though we dedupe transactions by hash before notifying listeners, I don't recall that we ensured any order )

PR #414, opened 2015-07-10
Merge flipped orderbook in getOrderbook

clark800:
Orders in the order book are now "synthetic" in that the specification of the order does not necessarily match the specification used to initially specify the order (buy and sell can be flipped). This allows us to merge the X/Y order book with the Y/X order book, which is good because one of the two order books is redundant.

### DIFF
--- a/test/fixtures/get-orderbook-response.json
+++ b/test/fixtures/get-orderbook-response.json
@@ -5,123 +5,19 @@
         "direction": "buy",
         "quantity": {
           "currency": "USD",
-          "value": "10",
+          "value": "93.030522464522",
           "counterparty": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
         },
         "totalPrice": {
           "currency": "BTC",
-          "value": "3205.1",
-          "counterparty": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
-        }
-      },
-      "properties": {
-        "maker": "r49y2xKuKVG2dPkNHgWQAV61cjxk8gryjQ",
-        "sequence": 434,
-        "makerExchangeRate": "0.003120027456241615"
-      }
-    },
-    {
-      "specification": {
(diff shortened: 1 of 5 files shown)


test/fixtures/get-orderbook-response.json
```diff
@@ -169,6 +169,26 @@
         "direction": "buy",
         "quantity": {
           "currency": "USD",
+          "value": "0.003967400879423823",
```
sublimator: This looks a bit off somehow. I'm supposed to read quantity as the amount the person wants to buy? Here it's "want to buy/get 0.00397 USD and pay 1.242 BTC ? It seems like the issue is the wrong way around somehow ??

(Checking now)

sublimator: Ah, I see, https://github.com/ripple/ripple-lib/blob/develop/test/fixtures/book-offers-response.js#L228 (one little typo there, by the looks )

And: https://github.com/ripple/ripple-lib/blob/develop/test/fixtures/book-offers-response.js#L7-L13
The results were taken from real fixtures, but they are flipped, such that the real life ratios are inverted.

test/fixtures/get-orderbook-response.json
```diff
@@ -184,13 +204,13 @@
         "makerExchangeRate": "0.003222279177208227"
       },
       "state": {
-        "availableQuantity": {
-          "currency": "USD",
+        "fundedAmount": {
```
sublimator: Because `quantity` doesn't always relate to TakerGets huh

clark800: Yeah, `fundedAmount` is the "quantity available" for sell orders and "funds available" for buy orders.

test/fixtures/get-orderbook-response.json
clark800: Ok, I'll flip the currencies so to avoid confusion.

PR #415, opened 2015-07-11
Port integration tests from ripple-rest

darkdarkdragon:


test/fixtures/payments.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+const _ = require('lodash');
+
+
+module.exports.payment = function(_options) {
```
clark800: Use es6 default param: `options={}`

test/integration/integration-test.js
```diff
@@ -0,0 +1,193 @@
...
+    assert.strictEqual(data.address, wallet.getAddress());
+    assert.strictEqual(data.outcome.result, 'tesSUCCESS');
+    done();
+  });
+}
+
+function testTransaction(testcase, txData, type, currentLedgerVersion, done) {
+  const signedData = testcase.api.sign(txData, wallet.getSecret());
+  console.log('PREPARED...');
+  testcase.api.submit(signedData.signedTransaction, (_err, data) => {
```
clark800: Looks like we can rename `_err` to `error`

test/integration/integration-test.js
```diff
@@ -0,0 +1,193 @@
...
+function verifyResult(transactionType, transaction) {
+  assert(transaction);
+  assert.strictEqual(transaction.Account, wallet.getAddress());
+  assert.strictEqual(transaction.TransactionType, transactionType);
+}
+
+
+function setup(done) {
+  this.api = new RippleAPI({servers: ['wss://s1.ripple.com:443']});
+  // this.api = new RippleAPI({servers: ['wss://s-east.ripple.com:443']});
```
clark800: Might as well just delete this

test/integration/integration-test.js
```diff
@@ -0,0 +1,193 @@
...
+  this.api.connect(() => {
+    this.api.remote.getServer().once('ledger_closed', () => done());
+  });
+}
+
+function teardown(done) {
+  this.api.remote.disconnect(() => {
```
clark800: Could simplify to `this.api.remote.disconnect(done)`

test/integration/integration-test.js
```diff
@@ -0,0 +1,193 @@
...
+  const instructions = {maxLedgerVersionOffset: 100};
+  this.timeout(TIMEOUT);
+
+  beforeEach(setup);
+  afterEach(teardown);
+
+  it('settings', function(done) {
+    const currentLedgerVersion = this.api.getLedgerVersion();
+    this.api.prepareSettings(wallet.getAddress(), settingsSpecification, instructions, (err, txData) => {
+      if (err) {
```
clark800: This pattern repeats in every test case, we can reduce the redundancy with the `composeAsync` helper.

test/integration/integration-test.js
```diff
@@ -0,0 +1,193 @@
...
+        return;
+      }
+      verifyResult('TrustSet', txData);
+      testTransaction(this, txData, 'trustline', currentLedgerVersion, done);
+    });
+  });
+
+
+  it('payment', function(done) {
+    const paymentsSpecification = payments.payment({
```
clark800: Should be `paymentSpecification`; `payments` make it sound like pathfinding.

test/integration/integration-test.js
```diff
@@ -0,0 +1,193 @@
...
+
+
+const TIMEOUT = 20000;   // how long before each test case times out
+const INTERVAL = 2000;   // how long to wait between checks for validated ledger
+
+
+function verifyTransaction(testcase, hash, type, options, done) {
+  const currentOptions = {
+    minLedgerVersion: options.minLedgerVersion,
+    maxLedgerVersion: testcase.api.getLedgerVersion()
```
clark800: Why not options.maxLedgerVersion?

test/integration/integration-test.js
```diff
@@ -0,0 +1,193 @@
...
+
+describe.skip('integration tests', function() {
+  const instructions = {maxLedgerVersionOffset: 100};
+  this.timeout(TIMEOUT);
+
+  beforeEach(setup);
+  afterEach(teardown);
+
+  it('settings', function(done) {
+    const currentLedgerVersion = this.api.getLedgerVersion();
```
clark800: Unfortunately "current ledger" has been defined as the ledger that is currently going through consensus, whereas getLedgerVerison gives you the last closed ledger.

darkdarkdragon: do you mean that name 'currentLedgerVersion ' us misleading, right?

test/integration/integration-test.js
darkdarkdragon: Can't use composeAsync - it calls 'callback' parameter both in case of error  and in case if all goes well, but I need it to be called only in case of error.

test/integration/integration-test.js
darkdarkdragon: Do you mean why not just use options parameter as it is? 
Because it has maxLedgerVersion set to txData.LastLedgerSequence, which is somewhere in the future. And with that value I'm getting MissingLedgerHistoryError from getTransaction because it thinks that server does not have all needed history.
So we must fix getTransaction or just use current ledger version

test/integration/integration-test.js
clark800: Right

test/integration/integration-test.js
clark800: I think probably the best solution is to make MissingLedgerHistoryError only happen when ledgers are missing prior to the last closed ledger. So in this case I think it should just return NotFoundError.

test/integration/integration-test.js
clark800: Oh right, because the subsequent function is also async. I guess this is actually a situation for `async.waterfall` and `_.partial`. Do you think it would make it more readable to use `async.waterfall`?

darkdarkdragon: Yes, you are right. Personally I don't like waterfall with partial very much, but it is best of what was invented for javascript.

clark800: I agree, it's not very pretty. I'm ok with leaving it as it is if you prefer.

darkdarkdragon: I've changed it already.

test/integration/integration-test.js
darkdarkdragon: https://github.com/ripple/ripple-lib/pull/417

clark800:
LGTM

PR #416, opened 2015-07-13
Ternary operator syntax fix

darentuzi:
Ternary operator syntax fix

### DIFF
--- a/src/core/serializedtypes.js
+++ b/src/core/serializedtypes.js
@@ -257,7 +257,7 @@ function parse(so) {
   }
 
   const field_bits = tag_byte & 0x0f;
-  var field_name = (field_bits === 0)
+  const field_name = (field_bits === 0)
     ? binformat.fields[type_bits][so.read(1)[0]]
     : binformat.fields[type_bits][field_bits];
 
--- a/src/core/serializedtypes.js
+++ b/src/core/serializedtypes.js
@@ -258,8 +258,8 @@ function parse(so) {
 
   const field_bits = tag_byte & 0x0f;
   let field_name = (field_bits === 0)
-    ? field_name = binformat.fields[type_bits][so.read(1)[0]]
-    : field_name = binformat.fields[type_bits][field_bits];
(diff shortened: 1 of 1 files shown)


clark800:
CI tests are failing; eslint wants the `let` on line 250 to be `const`. There also seems to be a problem with npm that I don't understand yet.

sublimator:
Looks like the `problem with npm` was just connection issues. I just triggered a rebuild to see if that helps with anything.

src/core/serializedtypes.js
```diff
@@ -258,8 +258,8 @@ function parse(so) {
 
   const field_bits = tag_byte & 0x0f;
   let field_name = (field_bits === 0)
-    ? field_name = binformat.fields[type_bits][so.read(1)[0]]
```
sublimator: woah, what was going on here ??? :)

darentuzi: is the syntax correct? filed_name = ( test ? expression1 : expression2 )
if (field_bits===0) assign the value of line 261 to field_name otherwise line 262

sublimator: While technically the other syntax was OK, this fix looks much better, it's just that the original style was **so** strange, no?

darentuzi: yes, I think line 261&262 get executed before declaration at 260 so field_name could get implicitly declared as global? anyway simpler is better

sublimator: Oh, interesting :)

sublimator: Anyway, you just need to make the var a `const` and not `let` and it can be merged.

sublimator:
Can you squash it to one commit also once you're happy and force push?

sublimator:
If it's easier, could just create a new pull request with the substance of these changes. Thanks.

darentuzi:
sorry for this question but how do i squash it to one commit or do you mean close this request and open a new one and do the changes in one go? also i don't have write permissions to push

darentuzi:
okay gotit

PR #417, opened 2015-07-13
throw MissingLedgerHistoryError in getTransaction only when ledgers a…

darkdarkdragon:
…re missing prior to the last closed ledger.


clark800:
There's a lint error

clark800:
LGTM

PR #418, opened 2015-07-13
Refine ledger gap detection to reduce false positives

clark800:


src/api/ledger/transactions.js
```diff
@@ -62,24 +62,45 @@ function getAccountTx(remote, address, options, marker, limit, callback) {
...
+
+  if (!utils.hasCompleteLedgerRange(remote, minLedgerVersion,
+      maxLedgerVersion)) {
+    throw new utils.common.errors.MissingLedgerHistoryError();
+  }
+}
+
+function formatResponse(remote, options, transactions) {
   const compare = options.earliestFirst ? utils.compareTransactions :
     _.rearg(utils.compareTransactions, 1, 0);
```
sublimator: This just switches around the txn args, passed from sort, such that it will sort in reverse order?

src/api/ledger/transactions.js
```diff
@@ -62,24 +62,45 @@ function getAccountTx(remote, address, options, marker, limit, callback) {
...
 }
 
-function getTransactionsInternal(remote, address, options, callback) {
+function checkForLedgerGaps(remote, options, transactions) {
+  let {minLedgerVersion, maxLedgerVersion} = options;
+
+  // if we reached the limit on number of transactions, then we can shrink
+  // the required ledger range to only guarantee that there are no gaps in
+  // the range of ledgers spanned by those transactions
+  if (options.limit && transactions.length === options.limit) {
```
sublimator: How often were these false positives a problem? Can you describe the issue? What are the clients for this `getTransactionsInternal` etc? 

This is used for transaction submission in any way? limits should be respected I guess when searching a particular ledger range for txns that made it into a ledger. And I guess it makes sense that if you only want say 200 of the most recent transactions, and you just put an all possible ledgers ledger range, that if you were missing some, it shouldn't really matter.  This should still detect any gaps in any of the pages though yeah?

src/api/ledger/transactions.js
sublimator: I guess this is nice and simple, but I wonder if just at the point of response, just simply reversing them and swapping the `pages = prior_pages + new_page` to `pages = new_page + prior_pages` might not get the job done.  I thought they came already sorted ASC/DESC from rippled?

clark800: > This just switches around the txn args, passed from sort, such that it will sort in reverse order?

Yeah

> I thought they came already sorted ASC/DESC from rippled?

I think that's right, but it's probably best to be defensive and make sure.

src/api/ledger/transactions.js
clark800: Well one important use case is "get the next/previous transaction after this specified transaction". We just want to make sure that there are no gaps between the two transactions, so it's ok if there is a gap further back in the history.

> This should still detect any gaps in any of the pages though yeah?

Not sure what you mean by pages as there are no pages here, just a list of all transactions about to be returned. It is supposed to check that there are no gaps in the range of ledgers spanned by these transactions.

src/api/ledger/transactions.js
sublimator: By page I meant each response containing n transactions. I was under the impression that getRecursive automatically handles picking up from markers returned and sending more requests. The resume/marker system is kind of like pagination. Hence pages.

sublimator: Oops, this was answer to above "not sure what you meant by pages"

src/api/ledger/transactions.js
sublimator: So this checkForLedgerGaps just gets run once, on a list of transactions that could be a concat of multiple responses?

clark800: Yes

sublimator: Are there any use cases for this where this behaviour would be potentially dangerous that you know of? This seems OK to me, but I feel like I'm missing something.

clark800: Not that I can think of. If you hit the limit without encountering any ledger gaps, I don't think it matters if there are ledger gaps elsewhere.

geertweening:
@lumberj does this impact the amount of ledger_gap issues we've seen?

PR #420, opened 2015-07-14
Test all methods of RippleApi in integration-test

darkdarkdragon:


test/integration/integration-test.js
```diff
@@ -140,6 +154,20 @@ describe.skip('integration tests', function() {
       _.partial(verifyResult, 'OfferCreate'),
       _.partial(testTransaction, this, 'order', lastClosedLedgerVersion),
       (txData, callback) => {
+        this.api.getOrders(wallet.getAddress(), {}, composeAsync(orders => {
+          assert(orders && orders.length > 0);
+          const createdOrder = _.first(_.filter(orders, (order) => {
+            return order.properties.sequence === txData.Sequence;
+          }));
+          assert(createdOrder);
+          assert.strictEqual(createdOrder.specification.direction, orderSpecification.direction);
```
clark800: Would it work to do `assert.strictEqual(createdOrder.specification, orderSpecification);`?

test/integration/integration-test.js
```diff
@@ -147,10 +175,145 @@ describe.skip('integration tests', function() {
     ], () => done());
   });
 
+
+  it('isconnected', function(done) {
```
clark800: Can we camelcase the test names for consistency?

test/integration/integration-test.js
```diff
@@ -147,10 +175,145 @@ describe.skip('integration tests', function() {
...
+
+
+  it('serverinfo', function(done) {
+    this.api.getServerInfo(composeAsync(data => {
+      assert(data && data.info && data.info.pubkey_node);
+    }, done));
+  });
+
+
+  it('getfee', function(done) {
```
clark800: For synchronous test cases, you can omit the `done` parameter.

test/integration/integration-test.js
```diff
@@ -147,10 +175,145 @@ describe.skip('integration tests', function() {
...
+    const fee = this.api.getFee();
+    assert.strictEqual(typeof fee, 'string');
+    done();
+  });
+
+
+  it('getledgerversion', function(done) {
+    const ledgerVersion = this.api.getLedgerVersion();
+    assert.strictEqual(typeof ledgerVersion, 'number');
```
clark800: Could also check that it's greater than a fixed constant set to the current ledger version.

test/integration/integration-test.js
```diff
@@ -147,10 +175,145 @@ describe.skip('integration tests', function() {
...
+
+
+  it('gettransactions', function(done) {
+    const options = {
+      outgoing: true,
+      minLedgerVersion: this.startLedgerVersion
+    };
+    this.api.getTransactions(wallet.getAddress(), options, composeAsync(transactionsData => {
+      assert(transactionsData);
+      assert.strictEqual(transactionsData.length, this.transactions.length);
```
clark800: Nice

test/integration/integration-test.js
darkdarkdragon: No. orderSpecification has field immediateOrCancel, and createdOrder.specification - doesn't.

test/integration/integration-test.js
darkdarkdragon: No. orderSpecification has field immediateOrCancel, and createdOrder.specification - doesn't.

test/integration/integration-test.js
darkdarkdragon: ok

test/integration/integration-test.js
clark800: You can erase `immediateOrCancel: false` because it defaults to `false`.

clark800:
Can we remove the `eslint-disable: max-len`?

test/integration/integration-test.js
```diff
@@ -80,8 +86,16 @@ function teardown(done) {
 describe.skip('integration tests', function() {
   const instructions = {maxLedgerVersionOffset: 100};
   this.timeout(TIMEOUT);
+  const transactions = [];
 
+  // beforeEach(setup);
   beforeEach(setup);
+  beforeEach(function() {
+    if (!this.startLedgerVersion) {
```
clark800: How does this work? Is `this.startLedgerVersion` ever set at this point? Can we move this into the `setup` method so that `beforeEach` is only called once?

test/integration/integration-test.js
```diff
@@ -80,8 +86,16 @@ function teardown(done) {
 describe.skip('integration tests', function() {
   const instructions = {maxLedgerVersionOffset: 100};
   this.timeout(TIMEOUT);
+  const transactions = [];
 
+  // beforeEach(setup);
```
clark800: Looks like a leftover comment

test/integration/integration-test.js
darkdarkdragon: Good point.

test/integration/integration-test.js
```diff
@@ -147,10 +175,145 @@ describe.skip('integration tests', function() {
...
+  it('serverinfo', function(done) {
+    this.api.getServerInfo(composeAsync(data => {
+      assert(data && data.info && data.info.pubkey_node);
+    }, done));
+  });
+
+
+  it('getfee', function(done) {
+    const fee = this.api.getFee();
+    assert.strictEqual(typeof fee, 'string');
```
clark800: Could check to make sure that it parses to a valid number.

test/integration/integration-test.js
darkdarkdragon: BTW, found here some inconvenience - getTransactions (and getTransaction) doesn't return transaction hash, and so I can't match transactions from getTransactions with transactions that I recorded earlier. That pretty much make getTransactions method useless.

sublimator: say wah?

clark800: Yeah, they should return the hash. The hash is not really an "outcome", so maybe we should add a "properties" section to the response. Do you want to add that?

sublimator: Where actually building stuff with an api helps

test/integration/integration-test.js
darkdarkdragon: This code intends to initialize `this.startLedgerVersion` only one, very first time.
And I just realized that `this` object that gets passed to each `it` test is the same one, so you right, I can move this to `setup`.

clark800: `setup` is for initialization code for _each_ test case. If it should only be run once, then you could just put it at the beginning of the suite

sublimator: mocha has a suite setup iirc too, can't ever remember what it's name is though

darkdarkdragon: Yeah, you're right, it has `before` method, but for `startLedgerVersion` I need value from first `this.api.connect` run, so no other way.

clark800: Ok, I guess it's ok to put in `setup` then. It's probably worth adding a comment to explain why it has to be there.

test/integration/integration-test.js
```diff
@@ -124,45 +144,201 @@ describe.skip('integration tests', function() {
...
+    assert(!isNaN(+fee));
+  });
+
+
+  it('getLedgerVersion', function() {
+    const ledgerVersion = this.api.getLedgerVersion();
+    assert.strictEqual(typeof ledgerVersion, 'number');
+    assert(ledgerVersion > this.startLedgerVersion);
```
clark800: Should be `>=` right?

darkdarkdragon: This could happen only if it will be only one test enabled ) So I felt that > would be ok,
But I can change to >= to be safe )

clark800: Yeah, we should always assume that all the test cases are independent, so it should be possible to run it by itself, which is often done for debugging using `mocha -g`.

darkdarkdragon: ok, fixed

clark800:
LGTM

PR #421, opened 2015-07-15
Fix parsing of settings transactions

clark800:


src/api/ledger/parse/settings.js
```diff
@@ -2,27 +2,69 @@
...
+  if (tx.TransactionType !== 'AccountSet') {
+    return settings;
+  }
+
+  const modifiedNodes = tx.meta.AffectedNodes.filter(node =>
+    node.ModifiedNode.LedgerEntryType === 'AccountRoot');
+  assert(modifiedNodes.length === 1);
+  const node = modifiedNodes[0].ModifiedNode;
+
+  const oldFlags = _.get(node.PreviousFields, 'Flags');
```
sublimator: Nice, so if the first arg to `get` is undefined, it won't cause any error ?

clark800: Yeah

sublimator: Sweet, I kind of came to like the `?` operator in coffee-script. I saw they added it to C# recently. 

`oldFlags = node.PreviousFields?.Flags`

src/api/ledger/parse/settings.js
```diff
@@ -2,27 +2,69 @@
...
+        settings[flagName] = true;
+      } else if (clearFlags & flagValue) {
+        settings[flagName] = false;
+      }
+    });
+  }
+
+  // enableTransactionIDTracking requires a special case because it
+  // does not affect the Flags field; instead it adds/removes a field called
+  // "AccountTxnID" to/from the account root. Checking SetFlag and ClearFlag
```
sublimator: I think you could just check for the AccountTxnID in the metadata for a more meaningful check of what the transaction actually changed. This would be in line with checking the Flags in the metadata as above rather than assuming the transaction did what was intended. Basically following the general rule of always looking at the metadata.

clark800: That makes sense, will do

reedrosenbluth:
LGTM

PR #427, opened 2015-07-15
Add getAccountInfo method and move sequence field in response

clark800:


src/api/ledger/accountinfo.js
```diff
@@ -0,0 +1,30 @@
...
+const composeAsync = utils.common.composeAsync;
+
+function formatAccountInfo(response) {
+  const data = response.account_data;
+  return {
+    sequence: data.Sequence,
+    xrpBalance: utils.common.dropsToXrp(data.Balance),
+    ownerCount: data.OwnerCount,
+    previousTransactionID: data.PreviousTxnID,
+    previousTransactionLedgerVersion: data.PreviousTxnLgrSeq
```
sublimator: AccountTxnID ?

clark800: Ah, yes: "Note: The deprecated PreviousTxnID transaction parameter was removed entirely in rippled 0.28.0. Use AccountTxnID instead." - https://ripple.com/build/transactions/#common-fields

sublimator: Well PreviousTxnID is still the account root, it's used for something else than transaction chaining. It shows the last transaction that affected the account. 

PreviousTxnID as a txn field was removed though.

clark800: Oh ok, so no change needed then?

sublimator: I thought I saw you supporting AccountTxnID for Flags elsewhere? It's a `thing` in RippleAPI ?

sublimator: i.e.  Do you need to add it here?

clark800: Oh, it wasn't listed in the docs for the account root (https://wiki.ripple.com/Ledger_Format#AccountRoot) so I forgot about it. How is it different than PreviousTxnID? It sounds like it's the same. Does that mean if the asfAccountTxnID wasn't set, then the PreviousTxnID will be undefined?

sublimator: PreviousTxnID is always set in the account root, it's part of [threading](https://wiki.ripple.com/Transaction_Metadata#Threading) (link refers to metadata but it's a ledger entry thing).  The rules are basically (actually more nuanced), that PreviousTxnID  gets updated for ANY txn that affects the account, and that AccountTxnID (if enabled via AccountSet asfAccountTxnID) is updated with the id of every Account initiated txn.

clark800: Yeah, should be added then. I just updated.

src/api/ledger/accountinfo.js
```diff
@@ -0,0 +1,32 @@
...
+const utils = require('./utils');
+const removeUndefined = require('./parse/utils').removeUndefined;
+const validate = utils.common.validate;
+const composeAsync = utils.common.composeAsync;
+
+function formatAccountInfo(response) {
+  const data = response.account_data;
+  return removeUndefined({
+    sequence: data.Sequence,
+    xrpBalance: utils.common.dropsToXrp(data.Balance),
```
sublimator: There's also things like TransferRate (very likely want that), and the EmailHash (or whatever it was called, might be deprecated)

sublimator: There's a list of all the fields in an AccountRoot: https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/LedgerFormats.cpp#L27-L41

I'm not sure what it means that a lot of them aren't documented here: https://ripple.com/build/rippled-apis/#account-info

You can't assume that omission means deprecation because TransferRate is missing and that's definitely not.  Shrug.

src/api/ledger/accountinfo.js
```diff
@@ -0,0 +1,32 @@
...
+    ownerCount: data.OwnerCount,
+    previousInitiatedTransactionID: data.AccountTxnID,
```
sublimator: And regularKey is missing, something that I've definitely seen around the RippleAPI traps.

src/api/ledger/accountinfo.js
clark800: transferRate and regularKey are in `getSettings`. "settings" refers to all the properties on the account root that can be set by the user; "account info" refers to the properties on the account root that are not settings.

clark800: Looks like we have everything in that first link covered now.

sublimator: Hrmmm. Split into two requests now huh.

sublimator: Have u considered just one that contained settings&properties? getAccount
or the like?

clark800: Well I think it's nice to maintain the symmetry of having a "get" method that corresponds to every "prepare" method. If we broke that symmetry, people would look for the symmetry and get confused when there are a bunch of fields in one but not the other. We could merge the settings into the "getAccountInfo" response, but then people will argue that "getSettings" is totally redundant and want to get rid of it. So I don't know, it seems like this is the simplest solution.

clark800: Here you can see how `sequence` got interpreted as a "setting" here, and ended up in the `specification` for a settings transaction on accident: https://github.com/ripple/ripple-lib/pull/427/files#diff-62ddd0ef536cb84665c26471f1a09fa4L9

sublimator: Getting corfee :) Will look in a bit.

sublimator: I see what you're saying but I would probably just have went with 1 getAccount*, with {settings: {...}, properties: {...}} somewhat inline with what you had now.

clark800: I'll bring this question up in the feedback seminar, which I'll probably schedule for next week. Also, send me an email if you can think of any other points that you think I should bring up about the API design.

PR #428, opened 2015-07-16
fix attachTransactionDate in getTransaction

darkdarkdragon:


clark800:
Mutating function parameters is bad practice. Generally I think mutation is best avoided whenever possible, but mutation of function-local variables is tolerable sometimes for performance.

> There's no test for this?

Apparently not. There's still a lot of code without test coverage. I only wrote minimal tests for the common use cases because testing every path would have slowed down development a lot and the goal is to get something usable out the door quickly. Ivan is working on increasing test coverage now.

clark800:
LGTM

sublimator:
> Mutating function parameters is bad practice. 

I knew you'd say something like that :)

clark800:
There was a good article about mutation on Hacker News today:
https://underthehood.myob.com/functional-programming-in-a-dysfunctional-world/

sublimator:
Oh, I'm sold on immutability and defensive copies etc, but used judiciously.  But I guess you only get the "functional" code reasonability effects if it's a "practice".

Have you played with any of the shared data structures I've seen floating about the traps? Facebook had some lib iirc. I wonder what the impetus was behind these libs?

clark800:
Are you talking about ImmutableJS? https://facebook.github.io/immutable-js/

I was considering whether it's worth using it to enforce immutability more. It would be interesting if eslint had a rule that required all data structures to be created with ImmutableJS :)

sublimator:
Yeah, that's the one

sublimator:
> all data structures to be created with ImmutableJS

Have you seen any codebases out there on your travels that use it pervasively?

clark800:
No, I haven't really been looking though.

PR #431, opened 2015-07-17
test parseTrustline in getTransaction

darkdarkdragon:


test/fixtures/api/responses/get-transaction-trust-set.json
```diff
@@ -0,0 +1,32 @@
+{
+  "type": "trustline",
+  "address": "r9UHu5CWni1qRY7Q4CfFZLGvXo2pGQy96b",
+  "sequence": 449,
+  "id": "635A0769BD94710A1F6A76CDE65A3BC661B20B798807D1BBBDADCEA26420538D",
+  "specification": {
+    "limit": "10000",
+    "currency": "USD",
+    "counterparty": "rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM",
+    "qualityIn": 500000000,
```
sublimator: This has nothing to with this pull request, but as I'm eye-balling the RippleAPI <-> Rippled forms:

We recently had this curious "seeming" bug where someone set the QualityIn to 1 and with a partial payment the DeliveredAmount showed some tiny amount that wasn't in line with the balance changes in `face value` terms.

Turns out Amount (and in accordance DeliveredAmount) is interpreted not at face value, but at how the recipient (Destination) values the money. So say if they value a certain gateway's dollar at 98c ( which would correspond to 980,000,000 / 1e9 ths ) then, if you sent $1 and allowed partial payment, the DeliveredAmount would be 98c, while the trust line balance changes (face value) would show $1. Vertigo inducing ... (The wiki recommends keeping QualityIn at default value - so as to accept things at face value )

I'm actually not entirely sure but I believe a QualityOut out of say 1.02e9 would mean that you value it higher (than face value), and thus to ripple through would require $1.02 for every $1 pushed `out`. 

Note the terms I'm using above: `"1.02" + "e9"`.  I think "1.02" conveys things better than `1002000000`.

When you're looking at the tx_json "raw" fields and formats, you need to keep in mind that you're looking at a low level hashing form. While we're going to all this trouble to smooth over these bumps, I think it would be worth perhaps using a decimal string to describe this ratio, and maybe different terms.

With regards to QualityIn, our docs actually state [this](https://ripple.com/build/transactions/#trustset) (which I believe is wrong):
```
(Optional) % fee for incoming value on this line, represented as an integer over 1,000,000,000. A value of 0 is shorthand for no fee
```

test/api-test.js
```diff
@@ -120,6 +120,13 @@ describe('RippleAPI', function() {
...
 
+  it('getTransaction - trust set', function(done) {
+    const hash =
+      '635A0769BD94710A1F6A76CDE65A3BC661B20B798807D1BBBDADCEA26420538D';
+    this.api.getTransaction(hash, {},
+      _.partial(checkResult, responses.getTransaction.trust, done));
+  });
+
   it('getTransactions', function(done) {
     const options = {types: ['payment', 'order'], outgoing: true, limit: 2};
```
sublimator: getAccountInfo#previousInitiated* ~ outgoing

test/fixtures/api/responses/get-transaction-trust-set.json
```diff
@@ -0,0 +1,32 @@
...
+    "qualityOut": 500000000,
+    "disableRippling": true,
+    "frozen": false,
+    "authorized": false
```
sublimator: I'm not exactly sure what the overarching rules are yet, but I see code like this elsewhere:

```
src/api/ledger/parse/payment.js:44:    allowPartialPayment: isPartialPayment(tx) || undefined,
src/api/ledger/parse/payment.js:45:    noDirectRipple: isNoDirectRipple(tx) || undefined
```

test/fixtures/api/responses/get-transaction-trust-set.json
clark800: Good point about expressing quality in a higher-level format, will do that.

I filed a JIRA ticket for the docs, add a comment if this isn't right: https://ripplelabs.atlassian.net/browse/DEV-138

test/api-test.js
clark800: I can't figure out what this comment means, please elaborate.

test/fixtures/api/responses/get-transaction-trust-set.json
clark800: Yeah, we should consistently stick to the rule of not showing things that are default, will fix that.

test/api-test.js
```diff
@@ -120,6 +120,13 @@ describe('RippleAPI', function() {
       _.partial(checkResult, responses.getTransaction.orderCancellation, done));
   });
 
+  it('getTransaction - trust set', function(done) {
+    const hash =
+      '635A0769BD94710A1F6A76CDE65A3BC661B20B798807D1BBBDADCEA26420538D';
+    this.api.getTransaction(hash, {},
+      _.partial(checkResult, responses.getTransaction.trust, done));
```
clark800: `trust` -> `trustline`

darkdarkdragon: fixed

test/api-test.js
```diff
@@ -120,6 +120,13 @@ describe('RippleAPI', function() {
       _.partial(checkResult, responses.getTransaction.orderCancellation, done));
   });
 
+  it('getTransaction - trustline set', function(done) {
+    const hash =
+      '635A0769BD94710A1F6A76CDE65A3BC661B20B798807D1BBBDADCEA26420538D';
+    this.api.getTransaction(hash, {},
+      _.partial(checkResult, responses.getTransaction.trust, done));
```
clark800: Doesn't seem to be fixed yet

darkdarkdragon: Now fixed.

clark800:
LGTM

test/api-test.js
sublimator: It's a tiny thing. Just pointing out that in the account info command, the AccountTxnID uses `initiated` to mean the account's own transactions while getTransactions uses `outgoing`.

test/fixtures/api/responses/get-transaction-trust-set.json
sublimator: I guess the same for TransferRate eh? I just looked at the schema, and it looks like it's an integer

test/api-test.js
clark800: Oh, yeah `initiated` is more appropriate, here's a new pull request: https://github.com/ripple/ripple-lib/pull/437

PR #432, opened 2015-07-17
testing - add memo field into payment transaction so parsePaymentMemo…

darkdarkdragon:
…s from parsePayment is tested

### DIFF
--- a/test/fixtures/api/responses/get-transactions.json
+++ b/test/fixtures/api/responses/get-transactions.json
@@ -3,8 +3,14 @@
     "type": "payment",
     "address": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
     "sequence": 4,
-    "id": "4C37C92576DEB000D13B07F4D3F99F968BD86B6B83A840BEFFB2BFC8A042A81B",
+    "id": "99404A34E8170319521223A6C604AF48B9F1E3000C377E6141F9A1BF60B0B865",
     "specification": {
+      "memos": [
+        {
+          "type": "client",
+          "format": "rt1.5.2"
+        }
+      ],
       "source": {
         "address": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
         "amount": {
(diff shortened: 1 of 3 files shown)


test/fixtures/api/responses/get-transactions.json
```diff
@@ -3,8 +3,16 @@
     "type": "payment",
     "address": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
     "sequence": 4,
-    "id": "4C37C92576DEB000D13B07F4D3F99F968BD86B6B83A840BEFFB2BFC8A042A81B",
+    "id": "99404A34E8170319521223A6C604AF48B9F1E3000C377E6141F9A1BF60B0B865",
     "specification": {
+      "memos": [
+        {
```
sublimator: In the context of RippleAPI carte blanche, is there much reason to expose these low level fields (`MemoType` / `MemoFormat`) vs something like `{"type"  : "...",  "format" : "...", "data" : "..."}` ?

sublimator: There seems to be a schema:  https://github.com/ripple/ripple-lib/blob/develop/src/api/common/schemas/memo.json

sublimator:
Again, not specifically related to this pull request, but on the topic of memos, I notice that RippleAPI only seems to support memos for Payment transactions, when they actually work for any transaction: https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/TxFormats.cpp#L113

test/fixtures/api/responses/get-transactions.json
clark800: Yeah, looks like memo parsing is not working yet. We should fix that before merging this pull request. @darkdarkdragon Can you fix parsePaymentMemos to return memos in the above schema format?

clark800:
Supporting memos on non-payment transaction types is low-enough priority that we can leave that for later. Will make a note of it.

test/fixtures/api/responses/get-transactions.json
darkdarkdragon: ok
BTW, as I understand, that one
https://github.com/ripple/ripple-client/blob/develop/src/js/tabs/send.controller.js#L1121
should be 
`tx.addMemo('client', 'plain/text', 'rt' + $scope.version);`

clark800: Looks like it

darkdarkdragon: done

clark800:
LGTM

PR #436, opened 2015-07-20
Fix orderbook notification while disconnected

wltsmrz:
OrderBook's `notify()` function is the entry point for orderbook notifications (created, modified, deleted offers). In this function,  `_subscribed` property is checked to be truthy, however, this property only means that the end user subscribed or unsubscribed. The orderbook can also disconnect automatically, if a connection to rippled is lost. When this happens, all of the orderbook's account funds caches are cleared, but `notify()` would still execute as normal, because the end user hasn't unsubscribed. So, with cleared caches, the default return value for cached account funds is `0`. But if `notify()` detects a deleted offer... you get it. The error is `Offer total cannot be negative` https://github.com/ripple/ripple-lib/issues/352

This pull request simply adds a check for `_synced` property, which is set to `false` on disconnect

### DIFF
--- a/src/core/orderbook.js
+++ b/src/core/orderbook.js
@@ -62,9 +62,9 @@ function OrderBook(remote,
   this._ownerFunds = {};
   this._ownerOffersTotal = {};
 
-  // We consider ourselves synchronized if we have a current
+  // We consider ourselves synced if we have a current
   // copy of the offers, we are online and subscribed to updates
-  this._synchronized = false;
+  this._synced = false;
 
   // Transfer rate of the taker gets currency issuer
   this._issuerTransferRate = null;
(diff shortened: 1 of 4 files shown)


src/core/orderbook.js
```diff
@@ -842,7 +842,7 @@ OrderBook.prototype.updateOwnerOffersFundedAmount = function(account) {
 OrderBook.prototype.notify = function(transaction) {
   const self = this;
 
-  if (!this._subscribed) {
+  if (!(this._subscribed && this._synced)) {
```
clark800: I was just looking at this issue too. It was possible for a `notify` to create a `model` event before it gets synchronized.

clark800:
Can you summarize what the problem was and what the fix was? It's not obvious from the changes.

wltsmrz:
I think Nicholas Dudfield nailed it earlier, but I'll summarize in the description of this pull request

clark800:
LGTM

PR #441, opened 2015-07-21
Add flow to api

lumberj:
1. Add flow to ledger/parse/account-order, trustline, orderbook-order
Use eslint plugin flowtype to allow flow type annotations. Babel-eslint emits errors when using the type keyword. This plugin works around the issue by stripping flow annotations before linting. Source: https://www.npmjs.com/package/eslint-plugin-flowtype

2. Add flow typecheck to api/ledger/pathfind

3. Add flow check to src/api: index, accountinfo, balances, orderbook-orders, settings, transaction, transactions

4. Add flow to api/server/server.js

5. Add flow-bin to dev dependencies and ignore it in flowconfig

### DIFF
--- a/src/api/ledger/pathfind.js
+++ b/src/api/ledger/pathfind.js
@@ -1,3 +1,4 @@
+/* @flow */
 'use strict';
 const _ = require('lodash');
 const async = require('async');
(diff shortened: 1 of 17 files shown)


src/api/ledger/pathfind.js
```diff
@@ -16,8 +22,13 @@ function addParams(params, result) {
   });
 }
 
-function requestPathFind(remote, pathfind, callback) {
-  const params = {
+type PathFind = {
+  source: {address: string, currencies: Array<number>},
```
lumberj: Array<number> doesn't look right, should be Array<string>

clark800:
LGTM after squash

PR #442, opened 2015-07-21
check if proxy option is undefined, not if it exists

reedrosenbluth:


clark800:
LGTM

clark800:
Can you reword the commit message so that it makes sense without looking at the code?

reedrosenbluth:
done

PR #443, opened 2015-07-22
cover src/api/transaction/payment.js with unit tests

darkdarkdragon:


test/api-test.js
```diff
@@ -50,6 +50,33 @@ describe('RippleAPI', function() {
       _.partial(checkResult, responses.preparePayment, done));
   });
 
+  it('preparePayment - 2', function(done) {
```
clark800: Can you use more descriptive names?

darkdarkdragon: Yeah, I can try..
But here it will be something like 'preparePayment with lot of flags and options'...

clark800: That's much better than just a meaningless number

darkdarkdragon: Ok

test/api-test.js
```diff
@@ -50,6 +50,33 @@ describe('RippleAPI', function() {
...
+    this.api.preparePayment(address, requests.preparePayment2, instructions,
+      _.partial(checkResult, responses.preparePayment2, done));
+  });
+
+  it('preparePayment - 3', function(done) {
+    const paymentRequest = _.cloneDeep(requests.preparePayment2);
+    paymentRequest.source.amount.currency = 'USD';
+    paymentRequest.destination.amount.currency = 'LTC';
+    const expected = _.cloneDeep(responses.preparePayment2);
```
clark800: I'm thinking it might be better to just make new fixtures for this case even if there is some redundancy in the fixtures since it will get rid of all these lines of code.

clark800: It's nice to be able to look at the fixture file and directly be able to see exactly what's being tested too.

clark800:
Your slashes are backwards :)

darkdarkdragon:
Ha! That's my native windows slashes!

clark800:
Can't let that into our commit log, people will think we still use windows!

darkdarkdragon:
Windows is cool!

clark800:
LGTM

PR #445, opened 2015-07-22
fix settings.json schema to allow empty stings to

darkdarkdragon:
'emailHash' and 'walletLocator' fields so they
can be cleared

cover api/transaction/settings.js with tests

### DIFF
--- a/src/api/common/schemas/settings.json
+++ b/src/api/common/schemas/settings.json
@@ -12,12 +12,22 @@
     "noFreeze": {"type": "boolean"},
     "globalFreeze": {"type": "boolean"},
     "defaultRipple": {"type": "boolean"},
-    "emailHash": {"$ref": "hash128"},
-    "walletLocator": {"$ref": "hash256"},
-    "walletSize": {"type": "integer"},
+    "emailHash": {
+      "oneOf": [
+        {"type": "null"},
+        {"$ref": "hash128"}
+      ]
+    },
+    "walletLocator": {
+      "oneOf": [
+        {"type": "null"},
+        {"$ref": "hash256"}
+      ]
+    },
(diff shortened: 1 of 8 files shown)


src/api/common/schemas/settings.json
```diff
@@ -12,8 +12,18 @@
     "noFreeze": {"type": "boolean"},
     "globalFreeze": {"type": "boolean"},
     "defaultRipple": {"type": "boolean"},
-    "emailHash": {"$ref": "hash128"},
-    "walletLocator": {"$ref": "hash256"},
+    "emailHash": {
+      "oneOf": [
+        {"type": "string", "maxLength": 0},
```
clark800: Do you think it would be better if we made `null` the value to clear the field?

darkdarkdragon: Yes, that would be good - right now you can't clear integer fields - but I don't know how schema validator will react.

src/api/common/schemas/settings.json
```diff
@@ -12,12 +12,22 @@
...
+    },
+    "walletLocator": {
+      "oneOf": [
+        {"type": "null"},
+        {"$ref": "hash256"}
+      ]
+    },
+    "walletSize": {"type": ["integer", "null"]},
     "messageKey": {"type": "string"},
     "domain": {"type": "string"},
```
clark800: Do you know if these other non-boolean fields are clearable?

darkdarkdragon: I've just took information from [here](https://github.com/ripple/ripple-lib/blob/develop/src/api/common/constants.js#L29) - one with 'defaults' field are clearable. Other from that - don't know.

clark800:
LGTM

PR #446, opened 2015-07-22
Increase tests coverage

darkdarkdragon:


src/api/common/schemas/pathfind.json
```diff
@@ -24,7 +24,32 @@
...
+            },
+            "currency": {
+              "description": "The three-character code or hex string used to denote currencies",
+              "$ref": "currency"
+            },
+            "counterparty": {
+              "description": "The Ripple account address of the currency's issuer or gateway",
+              "$ref": "address"
+            },
+            "required": ["currency", "value"]
```
clark800: So this change is to make counterparty optional? Is there documentation to confirm that the destination counterparty is optional for non-XRP?

test/api-test.js
```diff
@@ -224,8 +235,10 @@ describe('RippleAPI', function() {
 
   it('getOrderbook - direction is correct for bids and asks', function(done) {
     this.api.getOrderbook(address, orderbook, {}, (error, data) => {
-      assert(_.every(data.bids, bid => bid.specification.direction === 'buy'));
-      assert(_.every(data.asks, ask => ask.specification.direction === 'sell'));
+      assert.ok(
```
clark800: What is the `.ok` for?

test/api-test.js
```diff
@@ -262,7 +275,134 @@ describe('RippleAPI', function() {
...
-      _.partial(checkResult, responses.getPaths, done));
+      _.partial(checkResult, responses.getPaths.XrpToUsd, done));
+  });
+
+  // @TODO
+  // need decide what to do with currencies/XRP:
+  // if add 'XRP' in currencies, then there will be exception in
+  // xrpToDrops function (called from toRippledAmount)
+  it('getPaths USD 2 USD', function(done) {
+    const pathfind = {
```
clark800: Maybe we should put these request objects in the fixtures?

darkdarkdragon:
@wltsmrz Turns  out I've got old version of assert-diff installed

src/api/common/schemas/pathfind.json
darkdarkdragon: Yes, that make it optional.
Doc is [here](https://github.com/ripple/ripple-lib/blob/develop/src/api/ledger/pathfind.js#L37)

clark800: Ok thanks

test/api-test.js
darkdarkdragon: Same as just assert().
Turned out that I has old assert-diff installed where just assert() doesn't work.
I will change assert.ok back to normal in next commit.

test/api-test.js
darkdarkdragon: Yes, good point.

darkdarkdragon:
@clark800 done fixes

test/api-test.js
```diff
@@ -11,6 +11,7 @@ const hashes = require('./fixtures/hashes');
 const MockPRNG = require('./mock-prng');
 const sjcl = require('../src').sjcl;
 const address = addresses.ACCOUNT;
+const NotFoundError = require('../src/api/common/errors').NotFoundError;
```
clark800: Should use `this.api.errors.NotFoundError` to use the real interface

test/api-test.js
```diff
@@ -225,7 +236,8 @@ describe('RippleAPI', function() {
   it('getOrderbook - direction is correct for bids and asks', function(done) {
     this.api.getOrderbook(address, orderbook, {}, (error, data) => {
       assert(_.every(data.bids, bid => bid.specification.direction === 'buy'));
-      assert(_.every(data.asks, ask => ask.specification.direction === 'sell'));
+      assert(
```
clark800: Unnecessary change?

darkdarkdragon: too long line

test/api-test.js
darkdarkdragon: @clark800  Fixed

clark800:
LGTM

clark800:
Can you squash to get rid of the commit about assert.ok?

darkdarkdragon:
yep

PR #449, opened 2015-07-23
Express trustline quality and account transferRate as floats

clark800:


sublimator:
For the next person curious why circle crossed red:
```
Bad response: 500 {"message":"Coveralls is in Read Only mode while we work on updating the system. Sorry for the inconvenience.Check Status: http://status.coveralls.io/","error":true,"url":""}
```

src/api/common/schemas/quality.json
```diff
@@ -1,8 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   "title": "quality",
-  "description": "Ratio for incoming/outgoing transit fees represented in billionths. (For example, a value of 500 million represents a 0.5:1 ratio.) As a special case, 0 is treated as a 1:1 ratio.",
-  "type": "integer",
-  "minimum": 0,
-  "maximum": 1000000000
+  "description": "Ratio for incoming/outgoing transit fees. As a special case, 0 is treated as a 1:1 ratio.",
+  "type": "number",
+  "minimum": 0
```
sublimator: The transferRate schema has a maximum defined by the UInt32 upper bound. Why not quality also?  As well as lower/upper bounds I guess it should also check if there's more than 9 decimal places? What was the reason to use a float instead of a string, consistent with amount@value, and trustline@limit, etc?

clark800: I asked in the development channel if there was a similar limit for quality, but I haven't gotten an answer yet. I may just assume that it has the same limit. I don't think our schema validator properly supports checking for the correct number of decimal places, though I believe json schema supports it with the `multipleOf` property. Maybe I'll submit a bug report.

The only reason we are using strings for currency values is because javascript floats don't have enough precision for them. So for numbers where they do have enough precision, there's no need to use strings. One could argue for consistency or one could argue for minimizing the "damage" and keeping it limited to just currency values. Currently it is choosing the latter.

sublimator: I don't know what application level bounds there are, but Quality(In|Out) are serialised as UInt32 fields, so there's at least that: 4.294967295 

> One could argue for consistency

Probably what I would have gone for.

clark800: Yeah, Rome just confirmed that "trustline quality is [1, 4294967295] and account transfer rate is [1000000000, 4294967295]" so I updated it.

src/api/common/schemas/quality.json
```diff
@@ -1,8 +1,8 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
   "title": "quality",
-  "description": "Ratio for incoming/outgoing transit fees represented in billionths. (For example, a value of 500 million represents a 0.5:1 ratio.) As a special case, 0 is treated as a 1:1 ratio.",
-  "type": "integer",
-  "minimum": 0,
-  "maximum": 1000000000
+  "description": "Ratio for incoming/outgoing transit fees. As a special case, 0 is treated as a 1:1 ratio.",
+  "type": "number",
+  "minimum": 0.000000001,
```
sublimator: > As a special case, 0 is treated as a 1:1 ratio.

This minimum would disallow the  special case of 0 yeah?  "Just use 1" ?

clark800: Yeah, I want to get rid of all the special cases.

sublimator: Does it still deserve a mention in the `description` then?

sublimator:
LGTM

src/api/common/schemas/quality.json
clark800: Oh, missed that. Updated.

PR #452, opened 2015-07-24
Increase tests coverage even more

darkdarkdragon:


src/api/server/server.js
```diff
@@ -28,7 +28,9 @@ function isConnected(): boolean {
 function getServerInfo(callback: (err: any, data: any) => void): void {
   this.remote.requestServerInfo((error, response) => {
     if (error) {
-      callback(new common.errors.RippledNetworkError(error.message));
+      const message = error && error.remote && error.remote.error_message ?
```
clark800: Could use `_.get(error, 'remote', 'error_message')`

test/fixtures/api/rippled/account-tx.js
```diff
@@ -196,15 +198,37 @@ module.exports = function(request, options={}) {
...
     id: request.id,
     status: 'success',
     type: 'response',
     result: {
-      marker: request.marker === undefined ? 'ABC' : undefined,
+      marker: marker === undefined ? undefined : String(marker),
       transactions: [
         {
-          ledger_index: 348860,
+          ledger_index: 348860 - +marker,
```
clark800: We're going to disallow this soon, so better to change it now: https://github.com/eslint/eslint/issues/1621

test/api-test.js
```diff
@@ -341,4 +490,55 @@ describe('RippleAPI', function() {
   it('getLedgerVersion', function() {
     assert.strictEqual(this.api.getLedgerVersion(), 8819951);
   });
+
+  it('ledget utils - compareTransactions', function() {
```
clark800: `ledget` -> `ledger` and same below

test/api-test.js
```diff
@@ -341,4 +490,55 @@ describe('RippleAPI', function() {
...
+        callback(new Error(), null);
+      }
+    }
+    utils.getRecursive(getter, 10, (error) => {
+      assert(error instanceof Error);
+      done();
+    });
+  });
+
+  it('validator', function() {
```
clark800: Can you add a case to test a valid address & secret pair?

sublimator: On that topic: I noticed previously we've been validating that an address (used for tx_json.Account) matches the secret used for signing, but what about regular keys?

clark800: I'm not totally clear on how regular keys work. Is the issue that validation will fail if someone signs the transaction using the regular key? If that's the case, then maybe we should just get rid of that validation check?

darkdarkdragon: If I understand right, signing with regular keys doesn't work, so that validation check is ok.

clark800: This reference seems to suggest you can: https://ripple.com/knowledge_center/generating-and-setting-a-regular-key/

darkdarkdragon: It doesn't work in ripple-lib
https://github.com/ripple/ripple-lib/issues/243
https://github.com/ripple/ripple-lib/issues/226

darkdarkdragon: Actually, with little changes to [sign.js](https://github.com/ripple/ripple-lib/blob/develop/src/api/transaction/sign.js) I was able to sign transaction using regular key.

sublimator: If it doesn't work, it's just a bug that could/should be fixed.

clark800:
LGTM

PR #457, opened 2015-07-28
allow to sign transaction in api.sign using regular key

darkdarkdragon:
but you still cannot sign transaction with regular key by usign
core.Transaction.sign method

### DIFF
--- a/src/api/common/validate.js
+++ b/src/api/common/validate.js
@@ -24,6 +24,20 @@ function validateAddressAndSecret(obj) {
   }
 }
 
+function validateSecret(secret) {
+  if (!secret) {
+    throw error('Parameter missing: secret');
+  }
+  if (typeof secret !== 'string' || secret[0] !== 's') {
+    throw error('Invalid parameter');
+  }
+
+  const seed = new core.Seed().parse_base58(secret);
+  if (!seed.is_valid()) {
+    throw error('invalid seed');
+  }
+}
+
 function validateLedgerRange(options) {
   if (!_.isUndefined(options.minLedgerVersion)
       && !_.isUndefined(options.maxLedgerVersion)) {
@@ -41,6 +55,7 @@ function validateOptions(schema, options) {
 module.exports = {
   address: _.partial(schemaValidate, 'address'),
   addressAndSecret: validateAddressAndSecret,
+  secret: validateSecret,
   currency: _.partial(schemaValidate, 'currency'),
   identifier: _.partial(schemaValidate, 'hash256'),
   sequence: _.partial(schemaValidate, 'sequence'),
--- a/src/api/transaction/sign.js
+++ b/src/api/transaction/sign.js
@@ -18,8 +18,8 @@ const HASH_TX_ID = 0x54584E00; // 'TXN'
 const HASH_TX_SIGN = 0x53545800; // 'STX'
 const HASH_TX_SIGN_TESTNET = 0x73747800; // 'stx'
 
-function getKeyPair(address, secret) {
-  return core.Seed.from_json(secret).get_key(address);
+function getKeyPair(secret) {
+  return core.Seed.from_json(secret).get_key();
 }
 
 function getPublicKeyHex(keypair) {
@@ -51,9 +51,9 @@ function sign(txJSON: {Account: string; SigningPubKey: string,
     TxnSignature: string}, secret: string):
     {signedTransaction: string; id: string} {
   validate.txJSON(txJSON);
-  validate.addressAndSecret({address: txJSON.Account, secret: secret});
+  validate.secret(secret);
 
-  const keypair = getKeyPair(txJSON.Account, secret);
+  const keypair = getKeyPair(secret);
   if (txJSON.SigningPubKey === undefined) {
     txJSON.SigningPubKey = getPublicKeyHex(keypair);
   }
--- a/src/core/seed.js
+++ b/src/core/seed.js
@@ -28,17 +28,31 @@ Seed.prototype.parse_json = function(j) {
   if (typeof j === 'string') {
     if (!j.length) {
       this._value = NaN;
-    // XXX Should actually always try and continue if it failed.
-    } else if (j[0] === 's') {
-      this._value = Base.decode_check(Base.VER_FAMILY_SEED, j);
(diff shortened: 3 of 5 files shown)


src/api/transaction/sign.js
```diff
@@ -48,12 +48,14 @@ function computeSignature(txJSON, keypair) {
 }
 
 function sign(txJSON: {Account: string; SigningPubKey: string,
-    TxnSignature: string}, secret: string):
+    TxnSignature: string}, secret: string, isRegular = false):
```
clark800: I think we should avoid the extra complexity of this parameter. We can just delete the `validate.addressAndSecret` and maybe we can just remove the address parameter below?

sublimator: address param to Seed.get_key is optional yes if u meant that

clark800: What is the address parameter for, just optimization?

sublimator: Some ridiculous search among 4B possible keys/accounts from a seed, which
is limited to one by default anyway. It's for the account families
deprecated feature.

sublimator: Say your acvount id was account 30 from a seed, it loops until the keypair
matches the passed in address (hash of pubkey == address). Ill conceived.
Don't need to pass it.

darkdarkdragon: We can. I just don't know what difference it make for transactions to be all signed with `get_key()` instead of `get_key(address)`, so I left it intact.
But it works, I've checked.

clark800: Ok, let's do that then to keep things simple. Do you know how to validate a secret by itself?

sublimator: Seed.from_json will parse passphrases too. In case you strictly want to
accept bs58 only ....

sublimator: Seed has is_valid() iirc

src/api/common/validate.js
```diff
@@ -24,6 +24,19 @@ function validateAddressAndSecret(obj) {
   }
 }
 
+function validateSecret(secret) {
+  if (!secret) {
+    throw error('Parameter missing: secret');
+  }
+  try {
+    if (!core.Seed.from_json(secret).is_valid()) {
```
sublimator: 'passphrase' would be a valid secret here, which I think is discouraged outside of unit tests

sublimator: The nested catches are in case Seed.from_json throws somehow? Possible I guess, but I don't think it will happen intentionally (sadly), hence the `is_valid()`.

darkdarkdragon: Yeah, it really doesn't throw anything.

And I've made some changes to Seed.parse_json so it try to parse as different types

test/api-test.js
```diff
@@ -562,6 +562,12 @@ describe('RippleAPI', function() {
       secret: 'shzjfakiK79YQdMjy4h8cGGfQSV6u'
     };
     assert.doesNotThrow(_.partial(validate.addressAndSecret, goodWallet));
+    assert.throws(_.partial(validate.secret, ''),
+      this.api.errors.ValidationError);
+    //  'passphrase' that starts from 's', like secret
+    assert.doesNotThrow(_.partial(validate.secret, 's!!!'));
```
clark800: Shouldn't this one be invalid based on length? Also, can you add a test case for `passphrase` as sublimator mentioned?

darkdarkdragon: That's one is treated like passphrase, and that what I'm testing here - that passphrase starting from 's' is allowed.

clark800: `passphrase` doesn't start with `s`

darkdarkdragon: I don't see any restriction on passphrase.

clark800: There's no test case that is invalid, non-empty, and does not start with an `s`. `passphrase` covers that case.

darkdarkdragon: Sorry, don't understand. Any string that is not secret and not hex will be counted as `passphrase` and that would be valid. 
What I miss?

clark800: I mean adding `assert.throws(_.partial(validate.secret, 'passphrase'), this.api.errors.ValidationError);`

Code is better than english huh :smiley: 

Also, separate issue is that I think we want:

 `assert.throws(_.partial(validate.secret, 's!!!'), this.api.errors.ValidationError);`

darkdarkdragon: Code is universal :smiley: 

But for this we must remove parsing of `passphrase` from `Seed.parse_json`

clark800: Not necessarily, we can just not use the `parse_json` function. We don't have to allow hex or passphrases here, just the `s...` form.

darkdarkdragon: OK, got it.

darkdarkdragon: done.
only one thing to note - I was forced to add new method to `Seed`
hope this not a prolem :smiley:

src/core/seed.js
```diff
@@ -44,6 +48,19 @@ Seed.prototype.parse_json = function(j) {
...
 };
 
+Seed.prototype.parse_base = function(j) {
+  if (typeof j !== 'string') {
+    throw new Error('Value must be a string');
+  }
+  if (!j.length || j[0] !== 's') {
+    this._value = NaN;
+  } else {
+    this._value = Base.decode_check(Base.VER_FAMILY_SEED, j);
```
clark800: Instead of making a new function, could we just directly use `Base.decode_check(Base.VER_FAMILY_SEED, j);`

darkdarkdragon: We could, but it feels so wrong to duplicate functionality :worried:

clark800: What needs to be duplicated?

darkdarkdragon: That string
`Base.decode_check(Base.VER_FAMILY_SEED, j)`
Now it is just one string, but it belongs to Seed.
And what if in future there will be added another encoding? Then validate would needed to be changed too.

clark800: Ok, that sounds fine. `parse_base58` might make more sense though.

darkdarkdragon: renamed

clark800:
LGTM

PR #458, opened 2015-07-28
Fix counterparty logic in pathfind parsing

clark800:
I'm not entirely sure this is the correct way to do it, but I'm pretty sure the previous behavior was wrong.

### DIFF
--- a/src/api/common/schemas/adjustment.json
+++ b/src/api/common/schemas/adjustment.json
@@ -4,7 +4,16 @@
   "type": "object",
   "properties": {
     "address": {"$ref": "address"},
-    "amount": {"$ref": "amount"},
+    "amount": {
+      "type": "object",
+      "properties": {
+        "currency": {"$ref": "currency"},
+        "counterparty": {"$ref": "address"},
+        "value": {"$ref": "$value"}
+      },
+      "required": ["currency", "value"],
+      "additionalProperties": false
+    },
     "tag": {
       "description": "A string representing an unsigned 32-bit integer most commonly used to refer to a sender's hosted account at a Ripple gateway",
       "$ref": "uint32"
--- a/src/api/common/schemas/pathfind.json
+++ b/src/api/common/schemas/pathfind.json
@@ -24,32 +24,7 @@
       "additionalProperties": false,
       "required": ["address"]
     },
-    "destination": {
-      "type": "object",
-      "properties": {
-        "address": {"$ref": "address"},
-        "amount": {
-          "type": "object",
-          "properties": {
-            "value": {
-              "description": "The quantity of the currency, denoted as a string to retain floating point precision",
-              "$ref": "value"
-            },
-            "currency": {
-              "description": "The three-character code or hex string used to denote currencies",
-              "$ref": "currency"
-            },
-            "counterparty": {
-              "description": "The Ripple account address of the currency's issuer or gateway",
-              "$ref": "address"
-            },
-            "required": ["currency", "value"]
-          },
-          "additionalProperties": false
-        }
-      },
-      "required": ["address", "amount"]
-    }
+    "destination": {"$ref": "adjustment"}
   },
   "required": ["source", "destination"],
   "additionalProperties": false
--- a/src/api/ledger/parse/payment.js
+++ b/src/api/ledger/parse/payment.js
@@ -1,5 +1,6 @@
 /* @flow */
 'use strict';
+const _ = require('lodash');
 const assert = require('assert');
 const utils = require('./utils');
 const parseAmount = require('./amount');
@@ -30,18 +31,24 @@ function parsePaymentMemos(tx) {
   });
 }
 
+function removeGenericCounterparty(amount, address) {
+  return amount.counterparty === address ?
+    _.omit(amount, 'counterparty') : amount;
+}
+
 function parsePayment(tx: Object): Object {
   assert(tx.TransactionType === 'Payment');
 
   const source = {
     address: tx.Account,
-    amount: parseAmount(tx.SendMax || tx.Amount),
+    amount: removeGenericCounterparty(
+      parseAmount(tx.SendMax || tx.Amount), tx.Account),
     tag: tx.SourceTag
   };
 
   const destination = {
     address: tx.Destination,
-    amount: parseAmount(tx.Amount),
+    amount: removeGenericCounterparty(parseAmount(tx.Amount), tx.Destination),
     tag: tx.DestinationTag
   };
 
--- a/src/api/ledger/pathfind.js
+++ b/src/api/ledger/pathfind.js
@@ -78,17 +78,10 @@ function conditionallyAddDirectXRPPath(remote, address, paths, callback) {
   }
 }
 
-function addDestinationCounterparty(amount, address) {
-  return amount.currency === 'XRP' || amount.counterparty ? amount :
-    _.assign({counterparty: address}, amount);
-}
-
 function formatResponse(pathfind, paths) {
   if (paths.alternatives && paths.alternatives.length > 0) {
     const address = pathfind.source.address;
-    const destinationAmount = addDestinationCounterparty(
-      pathfind.destination.amount, pathfind.destination.address);
-    return parsePathfind(address, destinationAmount, paths);
+    return parsePathfind(address, pathfind.destination.amount, paths);
   }
   if (!_.includes(paths.destination_currencies,
       pathfind.destination.amount.currency)) {
--- a/test/fixtures/api/responses/get-paths-send-usd.json
+++ b/test/fixtures/api/responses/get-paths-send-usd.json
@@ -12,7 +12,6 @@
       "address": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
       "amount": {
         "value": "0.000001",
-        "counterparty": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
         "currency": "USD"
       }
     },
--- a/test/fixtures/api/responses/get-transaction-payment.json
+++ b/test/fixtures/api/responses/get-transaction-payment.json
@@ -15,8 +15,7 @@
       "address": "rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM",
       "amount": {
         "currency": "USD",
-        "value": "0.001",
-        "counterparty": "rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM"
+        "value": "0.001"
       }
     },
     "paths": "[[{\"currency\":\"USD\",\"issuer\":\"rpZc4mVfWUif9CRoHRKKcmhu1nx2xktxBo\",\"type\":48,\"type_hex\":\"0000000000000030\"},{\"account\":\"rpZc4mVfWUif9CRoHRKKcmhu1nx2xktxBo\",\"currency\":\"USD\",\"issuer\":\"rpZc4mVfWUif9CRoHRKKcmhu1nx2xktxBo\",\"type\":49,\"type_hex\":\"0000000000000031\"}]]"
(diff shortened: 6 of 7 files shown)


src/api/ledger/parse/pathfind.js
```diff
@@ -8,17 +8,25 @@ function parsePaths(paths) {
...
     return {
       source: {
         address: sourceAddress,
-        amount: parseAmount(alternative.source_amount)
+        amount: sourceAmount
       },
       destination: {
         address: pathfindResult.destination_account,
-        amount: destinationAmount
+        amount: ensureIOUCounterparty(destinationAmount, sourceAmount)
```
sublimator: The `destination.amount` is what's used for the `tx_json.Amount` right?

This doesn't look quite right as elsewhere:
```js
if (typeof params.dst_amount === 'object' && !params.dst_amount.issuer) {
    // Convert blank issuer to sender's address
    // (Ripple convention for 'any issuer')
    // https://ripple.com/build/transactions/
    //     #special-issuer-values-for-sendmax-and-amount
    // https://ripple.com/build/ripple-rest/#counterparties-in-payments
    params.dst_amount.issuer = params.dst_account;
  }
```

It uses the destination account as default counterparty, rather than the source amounts'

sublimator: Consider:  https://github.com/ripple/ripple-lib-extensions/blob/master/transactionparser/test/fixtures/payment-iou-multipath.json#L6-L9

clark800: Check out the link in the comment: https://ripple.com/build/transactions/#special-issuer-values-for-sendmax-and-amount

Setting it to the destination address is a rippled convention that means "any issuer", but in the new API, that convention should never be exposed.

sublimator: Define `never be exposed`?  I tried pondering what you meant, but can only come up with `blank counterparty in source/destination amount implies the source/destination addresses`. That however, is not what I'm seeing here.

Consider that the source amount could be BTC/$gateway and the destination amount in the path find USD/$recipient where the recipient doesn't even hold a trustline for USD/$gateway. And what if the source amount is XRP? It looks like counter party wouldn't even be set?

sublimator: > counterparty wouldn't even be set?

Which wouldn't serialise. 

One sense of `never be exposed` means removing functionality. If you're ok with that, then you should just demand a counterparty for IOU destination amounts. However, it's the only way you can send $100 across multiple trust lines, no?  A Payment's paths is specified not only by Paths but by Amount and SendMax I believe. (Also, what about old transactions loaded into the new format? See: https://github.com/ripple/ripple-lib/blob/develop/src/api/ledger/parse/payment.js#L38-L44)

sublimator: > Which wouldn't serialise.

Hrmm, not true: https://github.com/ripple/ripple-lib/blob/develop/src/api/transaction/payment.js#L26

clark800: "Never be exposed" refers to this part "When the issuer field of the destination Amount field matches the Destination account address, it is treated as a special case meaning “any issuer that the destination accepts.""

If you interpret it literally, it's actually _wrong_ because it is not an IOU issued by the destination, that's just a special code. A user of the new API should never see that in the requests of responses.

sublimator: Why are you setting the default counterparty to the source amount counter party?

clark800: If Alice is sending "Gateway/USD" and Bob is receiving "?/USD", it seems like the simplest assumption.

sublimator: But it could be XRP -> BTC, or BTC -> USD etc no ?

sublimator: Would leaving the counterparty simply undefined in those cases be more palatable?

sublimator: And if people do explicitly set source.amount.counterparty === address , should/would it be illegal?

clark800: I really don't know enough to answer those questions, I made this pull request hoping that somebody could tell me.

sublimator: * A Payment's paths is built from Amount/SendMax/Paths

* A Payment can be cross currency/issuer, executing offers

* issuer of self(destination|source) is code for any trustline
  * this is meaningful, and allows sending from multiple source trustlines to 
    multiple destination trustlines (see: [payment-iou-multipath.json](https://github.com/ripple/ripple-lib-extensions/blob/master/transactionparser/test/fixtures/payment-iou-multipath.json))
 

Assuming the above points were true (ask around/investigate I guess), would you find simply omitting the counterparty a better representation for the semantics ?

sublimator: lol, I guess strictly speaking, paths are built with Account/Destination too, but I think you got what I was trying to convey!

clark800: Oh, so maybe there isn't a unique counterparty because it could be delivered through trustlines with different counterparties? If that's the case, then yeah, I think omitting the counterparty makes sense.

clark800:
@sublimator updated

src/api/common/schemas/get-paths.json
```diff
@@ -6,7 +6,23 @@
     "type": "object",
     "properties": {
       "source": {"$ref": "adjustment"},
-      "destination": {"$ref": "adjustment"},
+      "destination": {
+        "type": "object",
```
sublimator: Will you bother to strip counterparty where appropriate here? https://github.com/ripple/ripple-lib/blob/develop/src/api/ledger/parse/payment.js#L38

clark800: Can you give an example where it would be appropriate to strip?

sublimator: Basically LGTM, just curious about whether you plan to use/enforce a canonical representation everywhere?

> an example where it would be appropriate to strip?

Well, when in rippled format, it will have the undefined source.amount.counterparty, filled out with Amount.issuer as Account right? And if you want to show transactions sourced from rippled data it in the RippleAPI format that it was declared in, it might make sense to set amount.counterparty to undefined when Amount.issuer == Account?

sublimator: re: `convention should never be exposed`

clark800: So the line you pointed at was parsing a payment transaction, which means it has already been executed and I think the counterparty should be set properly. Is that right?

sublimator: Yes it would be set "correctly", but now you have two ways of saying the same thing in RippleAPI, via omission or setting to the wildcard counterparty. Just pointing it out.

sublimator: And if people look at the wildcard form for already submitted transactions, they have to understand the semantics (which you said was wrong), at which point ?

clark800: By "correctly" I mean it should be the actual counterparty, not any special case. The only option I see is that we could throw a validation error if the destination counterparty is the destination address.

clark800: Oh, I thought the wildcard form only applied to pathfinding, but I just noticed the docs say it works for payments. Maybe the right answer is to parse the metadata to figure out the actual counterparty?

sublimator:
```js
function applyAnyCounterpartyEncoding(payment) {
  // Convert blank counterparty to sender or receiver's address
  //   (Ripple convention for 'any counterparty')
  // https://ripple.com/build/transactions/
  //    #special-issuer-values-for-sendmax-and-amount
  // https://ripple.com/build/ripple-rest/#counterparties-in-payments
  if (isIOUWithoutCounterparty(payment.source.amount)) {
    payment.source.amount.counterparty = payment.source.address;
  }
  if (isIOUWithoutCounterparty(payment.destination.amount)) {
    payment.destination.amount.counterparty = payment.destination.address;
  }
}
```

It does that when converting a RippleAPI payment to a rippled Payment. 
You could ponder disallowing doing it explicitly and do an inverse of that when parsing.

sublimator: > parse the metadata 

Could be more than one, and also, those fields are part of the (RippleAPI)  `specification` yeah?

clark800: Yeah specification is not always "original specification", it's more like "specification that would yield the same result". How do you know there could be more than one counterparty in a single payment, is that mentioned in the docs?

sublimator: Look at that multi path fixture from ripple-lib-extensions

sublimator: https://github.com/ripple/ripple-lib-extensions/blob/master/transactionparser/test/fixtures/payment-iou-multipath.json

It sends $300 from 3 source trustlines to 3 destination trustlines

sublimator: > it's more like "specification that would yield the same result". 

Ah, I see.  In any case, wildcard doesn't really mean whatever it happened to deliver no? There's the balanceChanges etc in the outcome for that.

I think you could pretty easily have a 1:1 mapping here from the RippleAPI <-> Rippled Payment source/destination amounts. (It would imply disallowing explicit wildcarding)

clark800: > You could ponder disallowing doing it explicitly and do an inverse of that when parsing.

I don't know if it's possible to do the inverse because it is actually a valid possibility in the case of a "redeem" payment where you actually are paying to the issuer.

sublimator: Hrrmm? Getting breakfast. Not sure what you mean off hand. biab

sublimator: Forcing a redemption is done by setting SendMax.issuer to Destination yes?

sublimator: I still don't get what you're saying sorry. It seems fairly clear that If
undefined means X, on the other side X can be translated to undefined,
cause it will translate back to X.

Are you talking a simple redemption case with no Paths? It ought to still
work?

clark800: Taking a step back, David said in the chat that:

> It's not a special case. The destination is the issuer. Which end of a trust line you consider the "issuer" is arbitrary. You are an issuer for all your trust lines.

With this interpretation we could say that it's not actually wrong or even really a special case, so we could leave it as is.

Basically we are adding a syntactic sugar that allows the user to just not specify the counterparty when they don't care about the counterparty, which I think is desirable for user-generated requests.

The question here is whether we should apply the syntactic sugar even for non-user-generated responses. I think it's confusing either way; but I guess it's better to do it as you suggest.

sublimator: "If you interpret it literally, it's actually wrong because it is not an
IOU issued by the destination"

Does 'counterparty' hold up as well as Dave is saying issuer does?

I have no strong preference on this subject. Needs further thought.

Maybe the only real advantage to omitting the counterparty would be less
clutter? I guess you might be able to argue it's more accessible and self
contained semantics.

An amount with counterparty B could mean different things when its used as
a source or destination amount where source account is B.

An amount with no counterparty? It's just an amount of some currency.

Hrrmmmm.

clark800: I think `counterparty` is just a more accurate name for `issuer`, emphasizing the symmetry between the two sides.

sublimator: What do other people think?

sublimator: Anyway, the only real objection I had with this PR was using sourceamount
counterparty for the dest amount when unspecified iirc. Which is fixed. I
don't want to hold you up.

clark800:
@sublimator updated

sublimator:
LGTM

PR #459, opened 2015-07-28
Fix transaction construction for SetRegularKey transactions

wltsmrz:
Change `SetRegularkey` to `SetRegularKey` and add test...


sublimator:
LGTM, squash?

PR #460, opened 2015-07-28
Fix transaction sequencing

wltsmrz:
Bug was introduced during linting that would cause the transaction Sequence to be ahead, causing terPRE_SEQ and AccountSet no-op filler transactions for every transaction submitted

### DIFF
--- a/src/core/transactionmanager.js
+++ b/src/core/transactionmanager.js
@@ -728,8 +728,7 @@ TransactionManager.prototype.submit = function(tx) {
 
   if (typeof tx.tx_json.Sequence !== 'number') {
     // Honor manually-set sequences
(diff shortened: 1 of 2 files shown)


src/core/transactionmanager.js
```diff
@@ -728,8 +728,7 @@ TransactionManager.prototype.submit = function(tx) {
 
   if (typeof tx.tx_json.Sequence !== 'number') {
     // Honor manually-set sequences
-    this._nextSequence += 1;
-    tx.tx_json.Sequence = this._nextSequence;
+    tx.tx_json.Sequence = this._nextSequence++;
```
sublimator: Good catch

sublimator:
LGTM

PR #464, opened 2015-07-29
Add DeliverMin setter

wltsmrz:
https://github.com/ripple/rippled/commit/4dc573f195c943e4d43e4eff4e09bde18cf6178f
https://github.com/ripple/rippled/commit/7c2478480d90c3d843ba4aa24b329ad430913c2c

### DIFF
--- a/src/core/transaction.js
+++ b/src/core/transaction.js
@@ -892,10 +892,6 @@ Transaction.prototype.accountSet = function(options_) {
 };
 
 Transaction.prototype.setAccountSetFlag = function(name, value) {
-  // if (this.getType() !== 'AccountSet') {
-    // throw new Error('TransactionType must be AccountSet to use ' + name);
-  // }
-
   const accountSetFlags = Transaction.set_clear_flags.AccountSet;
   let flagValue = value;
 
@@ -930,12 +926,6 @@ Transaction.prototype.setClearFlag = function(flag) {
 
 Transaction.prototype.setTransferRate =
 Transaction.prototype.transferRate = function(rate) {
-  /* eslint-disable max-len */
-  // if (this.getType() !== 'AccountSet') {
-    // throw new Error('TransactionType must be AccountSet to use TransferRate');
-  // }
-  /* eslint-enable max-len */
-
   const transferRate = rate;
 
   if (transferRate === 0) {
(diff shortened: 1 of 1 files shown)


src/core/transaction.js
```diff
@@ -1186,9 +1162,6 @@ Transaction.prototype.pathAdd = function(path) {
   if (!Array.isArray(path)) {
     throw new Error('Path must be an array');
   }
-  // if (this.getType() !== 'Payment') {
```
sublimator: Why were these removed?

src/core/transaction.js
```diff
@@ -1037,26 +1027,14 @@ Transaction.prototype.rippleLineSet = function(options_) {
...
 Transaction.prototype.setQualityIn = function(quality) {
-  // if (this.getType() !== 'TrustSet') {
-    // throw new Error('TransactionType must be TrustSet to use QualityIn');
-  // }
-
   return this._setUInt32('QualityIn', quality);
 };
 
 Transaction.prototype.setQualityOut = function(quality) {
-  // if (this.getType() !== 'TrustSet') {
```
sublimator: This is now handled by the "field unknown for type" like messages in SerializedObject.from_json ?

src/core/transaction.js
wltsmrz: Eventually we'll subclass Transaction so they'd be unnecessary... I doubt that we'll ever switch these on prior to that, so just removing some clutter

sublimator: Fair enough

src/core/transaction.js
```diff
@@ -1129,14 +1099,20 @@ Transaction.prototype.setDestination = function(destination) {
...
-    // throw new Error('TransactionType must be Payment to use SendMax');
-  // }
-
   return this._setAmount('SendMax', send_max);
 };
 
 /**
+ * Set DeliverMin for Payment
+ *
+ * @param {String|Object} deliver min  amount
```
sublimator: jsdoc name doesn't match the function parameter

src/core/transaction.js
wltsmrz: Yeah, the serialization stuff should catch it, but I added this (although it's not enabled) along with other transaction validation things in the event that someone doesn't use local_signing

src/core/transaction.js
wltsmrz: Me fix

sublimator: Me Jane

sublimator:
L+G+T+M

PR #469, opened 2015-07-30
Update Flow to 0.14

lumberj:
Fixes issue with fs.readFileSync return type
See: https://github.com/facebook/flow/issues/416
https://github.com/facebook/flow/compare/v0.13.1...v0.14.0


clark800:
LGTM

PR #477, opened 2015-08-03
Add suspended payment

wltsmrz:
https://github.com/ripple/rippled/commit/3f0eacf5e70c43715a1e726d7a228d0ee307ee4e


src/core/binformat.js
```diff
@@ -116,7 +120,8 @@ const FIELDS_MAP = exports.fields = {
     17: 'InvoiceID',
     18: 'Nickname',
     19: 'Amendment',
-    20: 'TicketID'
+    20: 'TicketID',
```
sublimator: @scottschurr  What's the status of TicketID?

src/core/binformat.js
```diff
@@ -396,7 +422,19 @@ exports.ledger = {
     ['LedgerIndex',         OPTIONAL],
     ['Balance',             REQUIRED],
     ['LowLimit',            REQUIRED],
-    ['HighLimit',           REQUIRED]])
+    ['HighLimit',           REQUIRED]]),
+  SusPay: [117].concat(sleBase,[
```
sublimator: I think this probably wants to be `SuspendedPayment`, as `SusPay` is just an abbreviation for internals

@vinniefalco ?

src/core/transaction.js
```diff
@@ -1112,6 +1112,77 @@ Transaction.prototype.setDeliverMin = function(deliver_min) {
   return this._setAmount('DeliverMin', deliver_min);
 };
 
+Transaction.prototype.suspendedPayment = function(options_) {
```
sublimator: `function(options={})` ?

src/core/transaction.js
```diff
@@ -1112,6 +1112,77 @@ Transaction.prototype.setDeliverMin = function(deliver_min) {
...
+  const options = _.merge({}, options_);
```
sublimator: Oh, I see you're allowing `cancelAfter` and `cancel_after`, and don't want to mutate the passed in options_.

src/core/transaction.js
```diff
@@ -1112,6 +1112,77 @@ Transaction.prototype.setDeliverMin = function(deliver_min) {
...
+};
+
+Transaction.prototype.setCancelAfter = function(cancel_after) {
+  return this._setUInt32('CancelAfter', cancel_after);
+};
+Transaction.prototype.setFinishAfter = function(finish_after) {
+  return this._setUInt32('FinishAfter', finish_after);
+};
+
+Transaction.prototype.suspendedPaymentCancel = function(options_) {
```
sublimator: I guess this func is a better candiate for default args feature. 

We don't seem to be consistently allowing camelCase or snake_case though? See `offer_sequence`

test/fixtures/suspay.json
```diff
@@ -0,0 +1,59 @@
+{
+  "SUSPAY_CREATE_META_JSON": {
+    "TransactionIndex": 2,
+    "AffectedNodes": [
+      {
+      "CreatedNode": {
+        "LedgerEntryType": "SusPay",
```
sublimator: This is from rippled? It uses `SusPay` ? I thought I recalled @justmoon request the full monty in the json.

test/remote-test.js
```diff
@@ -2073,4 +2073,97 @@ describe('Remote', function() {
...
+      method: 255,
+      proof: 'DEADBEEF'
+    });
+    assert.deepEqual(transaction.tx_json, {
+      Flags: 0,
+      TransactionType: 'SuspendedPaymentFinish',
+      Account: 'rsLEU1TPdCJPPysqhWYw9jD97xtG5WqSJm',
+      Owner: 'r36xtKNKR43SeXnGn7kN4r4JdQzcrkqpWe',
+      OfferSequence: 1,
+      Digest: '5'.repeat(64),
```
sublimator: repeat ? I guess is a babel-runtime nicety? noted in my tricks book :)

test/fixtures/suspay.json
sublimator: It is from rippled: https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/LedgerFormats.cpp#L84

sublimator: https://github.com/ripple/rippled/pull/1224

src/core/transaction.js
```diff
@@ -1112,6 +1112,77 @@ Transaction.prototype.setDeliverMin = function(deliver_min) {
   return this._setAmount('DeliverMin', deliver_min);
 };
 
+Transaction.prototype.suspendedPayment = function(options_) {
+  const options = _.merge({}, options_);
+
+  if (_.isUndefined(options.cancel_after)) {
```
clark800: Why are we adding new code that uses snake case?

wltsmrz: It's an option rather than a method name, but I could be convinced to use whateverCaseThisIs for everything

clark800: I was told that there was a consensus (from before I joined) that javascript code would be written in camelCase. I think we should stick to it so we can get rid of all these ugly special case checks.

src/core/binformat.js
scottschurr: @sublimator As far as I know Tickets are in suspended animation.  When we revisit out-of-order transactions we'll see whether Tickets are the right approach or if there's a better way.

src/core/transaction.js
wltsmrz: I can't recall the details, but we did agree at some point that method names should be whateverCaseThisIs. rippled itself expects under_scored request properties so this is unavoidable to some degree

clark800: Right, but there's never a need for `if` statements if it's always going to convert from camel to snake at one point.

src/core/binformat.js
wltsmrz: Any interest in changing @vinniefalco?

PR #478, opened 2015-08-04
Convert API to promises

clark800:


package.json
```diff
@@ -18,6 +18,7 @@
     "async": "~0.9.0",
     "babel-runtime": "^5.5.4",
     "bignumber.js": "^2.0.3",
+    "es6-promisify": "^2.0.0",
```
sublimator: I guess this will play nicely with yield and friends :) ?

wltsmrz:
LGTM

package.json
clark800: I'm not sure, is this what you mean? http://stackoverflow.com/questions/21846390/getting-a-promises-value-via-yield-co

src/api/common/schemas/settings-transaction.json
```diff
@@ -10,6 +10,6 @@
     "address": {"$ref": "address"},
     "sequence": {"$ref": "sequence"}
   },
-  "required": ["type", "id", "address", "sequence", "specification", "outcome"],
+  "required": ["type", "id", "address", "sequence", "specification"],
```
lumberj: Why was outcome removed?

clark800: For non-validated transacitons.

lumberj: I see, so you won't show an outcome on an unvalidated ledger?

clark800: Yeah

src/api/ledger/accountinfo.js
```diff
@@ -31,4 +31,8 @@ function getAccountInfo(account, options, callback) {
     composeAsync(formatAccountInfo, callback));
 }
 
-module.exports = utils.wrapCatch(getAccountInfo);
+function getAccountInfo(account: string, options={}) {
+  return utils.promisify(getAccountInfoAsync.bind(this))(account, options);
```
lumberj: What's the value of `this` here?

clark800: The instance of `RippleAPI` that this function gets attached to.

PR #479, opened 2015-08-04
Fix RangeSet for validated_ledger as single ledger

wltsmrz:


src/core/rangeset.js
```diff
@@ -45,7 +45,7 @@ class RangeSet {
     const rangeStrings = rangesString.split(',');
     _.forEach(rangeStrings, rangeString => {
       const range = rangeString.split('-').map(Number);
-      this.addRange(range[0], range[1]);
+      this.addRange(range[0], range[1] || range[0]);
```
clark800: This would fail if we ever get bounds checking. It might be more clear to branch on `range.length`.

wltsmrz: Crossed my mind. Ok...

clark800:
LGTM

PR #481, opened 2015-08-04
Add test case for createPathFind and convert snake case function calls to camel case

clark800:


clark800:
This is intended to resolve: https://ripplelabs.atlassian.net/browse/RLJS-273

src/core/pathfind.js
```diff
@@ -54,16 +58,16 @@ PathFind.prototype.create = function () {
...
   this.remote.request_path_find_close().broadcast().request();
   this.emit('end');
   this.emit('close');
 };
 
-PathFind.prototype.notify_update = function (message) {
-  var src_account = message.source_account;
-  var dst_account = message.destination_account;
-  var dst_amount  = Amount.from_json(message.destination_amount);
+PathFind.prototype.notify_update = function(message) {
```
geertweening: should this be camelCased?

src/core/pathfind.js
```diff
@@ -74,7 +78,7 @@ PathFind.prototype.notify_update = function (message) {
   }
 };
 
-PathFind.prototype.notify_superceded = function () {
+PathFind.prototype.notify_superceded = function() {
```
geertweening: should this be camelCased as well?

clark800: I only changed calls that were broken by removing the attachUnderscored converter. Changing all underscore functions would probably be a bigger project.

geertweening:
LGTM

PR #482, opened 2015-08-04
Deprecate core

clark800:


sublimator:
This seems a bit premature somehow

src/index.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
-const _ = require('lodash');
 const core = require('./core');
 const RippleAPI = require('./api');
 
-module.exports = _.assign({}, core, {RippleAPI: RippleAPI});
+module.exports = {
+  RippleAPI,
+  _DEPRECATED: core   // WARNING: this will be removed soon
+};
```
sublimator: I'm actually a little confused as to how the tests all pass with this change. What's going on here?

sublimator: Oh right, it's because we have a `test/node_modules/ripple-lib` symbolic link to core: https://github.com/clark800/ripple-lib/blob/deprecate/test/node_modules/ripple-lib

src/core/index.js
```diff
@@ -37,31 +37,3 @@ exports._test = {
 exports.sjcl = require('./utils').sjcl;
 exports.Wallet = require('ripple-wallet-generator')({sjcl: exports.sjcl});
 exports.types = require('./serializedtypes');
-
-// camelCase to under_scored API conversion
-function attachUnderscored(name) {
```
sublimator: Why remove this? This will cause breakage beyond needing to fix the require to dereference _DEPRECATED such that `deprecating core` doesn't fully describe the change ?  The `core` will now be internals that you will be free to change the implementation of, at will ?

> // WARNING: this will be removed soon`

How soon? What's the plan here? (Sorry if this was mentioned on architects. I wasn't really paying attention)

clark800: Yeah, the idea is to unexpose core so that we can freely refactor/delete code without any concern about breaking things. There is no set timeframe for removal, but we're trying to keep things moving fast.

geertweening:
LGTM

PR #491, opened 2015-08-07
Don't set empty paths

lumberj:
"The Paths field must not be an empty array, nor an array whose members are all
empty arrays."
https://ripple.com/build/transactions/#paths

### DIFF
--- a/src/core/transaction.js
+++ b/src/core/transaction.js
@@ -344,8 +344,8 @@ Transaction.prototype._computeFee = function() {
   }
 
   switch (fees.length) {
-    case 0: return undefined;
-    case 1: return String(fees[0]);
+  case 0: return undefined;
+  case 1: return String(fees[0]);
   }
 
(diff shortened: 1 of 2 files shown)


src/core/transaction.js
```diff
@@ -344,8 +344,8 @@ Transaction.prototype._computeFee = function() {
   }
 
   switch (fees.length) {
-    case 0: return undefined;
-    case 1: return String(fees[0]);
+  case 0: return undefined;
```
wltsmrz: Oh no you di-int!

wltsmrz: j/k, this looks fine

lumberj: Haha it was a linter error

wltsmrz:
LGTM

wltsmrz:
Probably want to squarsh these commits huh?

lumberj:
Squashed

PR #492, opened 2015-08-10
[WIP] Only output demurrage form from to_human()

sublimator:
@shekenahglory 

Messing around in relation to your serialisation problem.  The human readable form, as per `XAU (-0.5%pa)` shows a rounded, approximate percentage only, and does not include the start date that's included in the full 20 bytes of a currency.  This basically sets {force_hex : true} as a default option in `to_json()` and then makes `to_human()` output the human form. 

Could definitely be tweaked/polished more.

### DIFF
--- a/src/core/amount.js
+++ b/src/core/amount.js
@@ -941,7 +941,7 @@ Amount.prototype.to_text_full = function(opts) {
   }
   return this._is_native
       ? this.to_human() + '/XRP'
-      : this.to_text() + '/' + this._currency.to_json()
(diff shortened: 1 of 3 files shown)


sublimator:
We should probably just rip out the demurrage form and in the general sense, any lossy form ?

clark800:
> We should probably just rip out the demurrage form and in the general sense, any lossy form?

That sounds reasonable

clark800:
Closing per @sublimator's request

PR #494, opened 2015-08-10
Use streaming pathfinding: rippled path_find instead of ripple_path_find

clark800:


sublimator:
Few points worth highlighting:

1. You must `close` or `destroy` or whatever it's called the request when you're done with it, otherwise it just keeps ticking along. 

2. Note that you can only ever have one path_find request active per websocket connection. If you request to open a new path finding stream, when one is already open, the previous will be closed automatically. 

Point 2 is the reason why ripple-rest always used the http command rather than websockets.

src/api/ledger/pathfind.js
```diff
@@ -46,8 +46,11 @@ function requestPathFind(remote, pathfind: PathFind, callback) {
       _.omit(utils.common.toRippledAmount(amount), 'value'));
   }
 
-  remote.requestRipplePathFind(params,
-    composeAsync(_.partial(addParams, params), callback));
+  remote.createPathFind(params, data => {
+    if (data.full_reply) {
+      callback(null, addParams(params, data));
```
sublimator: This `callback ` might be called more than once? cause the stream hasn't been told to be closed ?

src/core/remote.js
```diff
@@ -2277,7 +2281,7 @@ Remote.prototype.requestRipplePathFind = function(options_, callback_) {
       destination_account: options_.dst_account,
       destination_amount: options_.dst_amount,
       source_currencies: options_.src_currencies
-      }, options_);
+    }, options_);
   } else {
     _.merge(options, makeOptions(
       'ripple_path_find',
```
sublimator: A lingering mention of `ripple_path_find`

clark800: Is that a problem? It's unclear whether it's supposed to refer to the rippled name, which is `ripple_path_find`.

sublimator: nup, just pointing it, `up 2 u` as they say in cambodge

sublimator: Other points (only 1 request at a time, needing closing etc) seem important though

clark800: Yes, working on that still

clark800:
@sublimator updated

src/api/ledger/pathfind.js
```diff
@@ -46,8 +46,12 @@ function requestPathFind(remote, pathfind: PathFind, callback) {
       _.omit(utils.common.toRippledAmount(amount), 'value'));
   }
 
-  remote.requestRipplePathFind(params,
-    composeAsync(_.partial(addParams, params), callback));
+  remote.createPathFind(params, function(data) {
+    if (data.full_reply) {
```
wltsmrz: Wouldn't it be preferred for this check to be in createPathFind()?

src/core/pathfind.js
```diff
@@ -59,7 +59,8 @@ PathFind.prototype.create = function() {
 };
 
 PathFind.prototype.close = function() {
-  this.remote.request_path_find_close().broadcast().request();
+  this.removeAllListeners('update');
```
wltsmrz: This addresses my other comment

src/api/ledger/pathfind.js
clark800: Well, I think that would prevent the ability to use it for actual streaming. Here we just want to get the first full reply, but other use cases might want to continue streaming.

wltsmrz: Passing a callback as last argument which is meant to be called multiple times is... not a common pattern, I think. I'd assume if the last argument to createPathFind is a function, that the user doesn't want to streams... also, first argument to the callback should be an error or null

sublimator: @wltsmrz  raises a good point I think. What's the error handling situation here?

src/core/remote.js
```diff
@@ -2369,6 +2377,11 @@ Remote.prototype.requestPathFindClose = function(callback) {
 
   request.message.subcommand = 'close';
   request.callback(callback);
+  this._cur_path_find = null;   // TODO: wait for confirmation (RIPD-1013)
+  if (this._queued_path_finds.length > 0) {
```
sublimator: Is `ripple_path_find` deprecated or something ?

clark800: No, why do you ask? We're using `path_find` because rippled's implementation of `ripple_path_find` has an issue that makes it fail if a few simultaneous requests are made.

src/api/ledger/pathfind.js
clark800: All use cases should be passing the `onUpdate` function because without it I don't think you can guarantee that you will get the first `update` event. I just made it optional to avoid breaking things. I am tempted to just ignore that issue though and use a node-style callback.

clark800:
@wltsmrz @sublimator updated

clark800:
Just noticed a bug

clark800:
Fixed

wltsmrz:
LGTM

PR #495, opened 2015-08-11
Use ripple-keypairs and ripple-address-codec

sublimator:
Use elliptic for signing


sublimator:
```
src/core/base.js
  43:5  error  Value of 'e' may be overwritten in IE 8 and earlier  no-catch-shadow
  60:5  error  Value of 'e' may be overwritten in IE 8 and earlier  no-catch-shadow
  95:5  error  Value of 'e' may be overwritten in IE 8 and earlier  no-catch-shadow

src/core/seed.js
  64:7  error  Value of 'e' may be overwritten in IE 8 and earlier  no-catch-shadow

src/core/transaction.js
  404:7   error  Value of 'e' may be overwritten in IE 8 and earlier  no-catch-shadow
  819:11  error  Value of 'e' may be overwritten in IE 8 and earlier  no-catch-shadow
```

I'm not sure what to do about the above, other than disabling no-catch-shadow. I tried just using some obviously unique name, that couldn't possibly be in an upper scope (can't see `e` anywhere either ... ) but it wasn't of much use.

sublimator:
I was having a bit of grief with circle, with failing tests due to old versions of dependencies in the cache. I hit `rebuild with cleared cache` on the circle ui, and now have tests passing, but with eslint now failing (as I have locally, with the v1.1.0 of eslint and the no-catch-shadow)

sublimator:
https://github.com/eslint/eslint/issues/3322

sublimator:
https://github.com/eslint/eslint/pull/3360

wltsmrz:
LGTM

clark800:
LGTM

PR #496, opened 2015-08-11
annotate api functions with types

darkdarkdragon:


src/api/ledger/accountinfo.js
```diff
@@ -6,7 +6,43 @@ const removeUndefined = require('./parse/utils').removeUndefined;
...
+  OwnerCount: number,
+  PreviousTxnID: string,
+  AccountTxnID: string,
+  PreviousTxnLgrSeq: number,
+  index: string
+}
+
+type AccountDataResponse = {
+  account_data: AccountData,
+  ledger_current_index: number,
```
sublimator: Is `ledger_current_index` only for open ledgers?

darkdarkdragon: don't know. but I've checked here 
https://ripple.com/build/websocket-tool/#account_info
`ledger_current_index` is returned in response

sublimator: Try this:

```json
{
  "id": 1,
  "command": "account_info",
  "account": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
  "ledger_index" : "validated"
}
```

src/api/ledger/accountinfo.js
```diff
@@ -6,7 +6,43 @@ const removeUndefined = require('./parse/utils').removeUndefined;
...
-function formatAccountInfo(response) {
+type AccountData = {
+  Sequence: number,
+  Account: string,
+  Balance: string,
+  Flags: number,
+  LedgerEntryType: string,
+  OwnerCount: number,
+  PreviousTxnID: string,
+  AccountTxnID: string,
```
sublimator: I have no idea how the flow typechecking system works, as I've no experience with, but I can say AccountTxnID may not always be present. 

I saw some `flow` types, with names suffixed with `?`, which I  assumed meant optional?

src/api/ledger/accountinfo.js
darkdarkdragon: okay, then I will make `ledger_current_index` optional, and add optional `ledger_hash` field

and, `getAccountInfoAsync` doesn't allow to specify `validated', or ledger hash, as ledger index in request - maybe we should change `ledgerVersion` schema?

src/api/ledger/accountinfo.js
darkdarkdragon: yes, `?` means optional

src/api/ledger/accountinfo.js
sublimator: @darkdarkdragon  I'm not really super up to speed with RippleAPI 

I guess that's a question for @clark800

clark800: We want `ledgerVersion` to always be a number. Currently there is no support for lookups by ledger hash, but if we do add support for that, we should add a new field in the request called `ledgerHash`. I don't think there is any need to specify `validated` because that should be the default when no `ledgerVersion` is specified.

sublimator: > I don't think there is any need to specify validated because that should be the default when no ledgerVersion is specified.

Oh, nice! So the new default will be to get the last validated ledger, rather than an open ledger

clark800: Yeah, that's the intended design, though I haven't confirmed that it's working that way. There's not even a way to get non-validated ledgers with this API.

src/api/ledger/accountinfo.js
```diff
@@ -31,7 +68,7 @@ function getAccountInfoAsync(account, options, callback) {
     composeAsync(formatAccountInfo, callback));
 }
 
-function getAccountInfo(account: string, options={}) {
+function getAccountInfo(account: string, options:AccountInfoOptions={}) {
```
clark800: Missing space after colon

src/api/ledger/transaction.js
```diff
@@ -7,10 +7,156 @@ const parseTransaction = require('./parse/transaction');
...
+type SettingsTransaction = {
+  type: string,
+  specification: GetSettings,
+  outcome: Outcome,
+  id: string,
+  address: string,
+  sequence: number
+}
+
+type GetTransactionResponse = PaymentTransaction | OrderTransaction |OrderCancellationTransaction | TrustlineTransaction | SettingsTransaction
```
clark800: Missing space after `|`

src/api/ledger/transaction.js
```diff
@@ -7,10 +7,156 @@ const parseTransaction = require('./parse/transaction');
...
+
+type GetTransactionResponseCallback = (err?: Error, data?: GetTransactionResponse) => void
+
+type CallbackType = (err?: Error, data?: Object) => void
+
+function attachTransactionDate(remote: Remote, tx: Object, callback: CallbackType) {
   if (tx.date) {
-    callback(null, tx);
+    callback(undefined, tx);
```
clark800: I think we should stick with `null`: http://stackoverflow.com/questions/22657880/idiomatic-successful-callback-in-node-js

src/api/ledger/transaction.js
```diff
@@ -22,29 +168,31 @@ function attachTransactionDate(remote, tx, callback) {
...
           || tx.ledger_index <= options.maxLedgerVersion);
 }
 
-function getTransactionAsync(identifier, options, callback) {
+function getTransactionAsync(identifier: string, options: TransactionOptions, callback: GetTransactionResponseCallback) {
   validate.identifier(identifier);
   validate.getTransactionOptions(options);
 
   const remote = this.remote;
-  const maxLedgerVersion = Math.min(options.maxLedgerVersion,
```
clark800: Nice, caught a bug. To avoid using `let`, how about `options.maxLedgerVersion || Infinity`?

src/api/server/server.js
```diff
@@ -4,11 +4,43 @@
...
 const CONNECTION_TIMEOUT = 1000 * 30;
 
-function isUpToDate(remote): boolean {
+type GetServerInfoResponse = {
+  buildVersion: string,
+  completeLedgers: string,
+  hostid: string,
+  ioLatencyMs: number,
+  load: {
+    job_types: Array<Object>,
```
clark800: I guess we missed this before, we should probably do the camel case conversion recursively

clark800:
Have you tried moving the type aliases to separate modules? http://flowtype.org/blog/2015/02/18/Import-Types.html

clark800:
This issue was from another pull request, but it would be good to fix it here so we don't forget about it. These should be numbers instead of strings: https://github.com/ripple/ripple-lib/blob/develop/src/api/ledger/utils.js#L84

src/api/server/server.js
```diff
@@ -37,19 +69,19 @@ function getLedgerVersion(): number {
...
 
-function disconnect() {
+function disconnect(): Promise {
   return common.promisify(callback => {
     this.remote.disconnect(() => callback(null));
   })();
 }
 
-function getServerInfo() {
+function getServerInfo(): Promise {
```
clark800: Is there any way to annotate what the promise returns? I didn't see much in the docs about it.

src/api/ledger/transaction.js
darkdarkdragon: yeah, that's one thing I wanted to ask you about.
I haven't found the way how to specify in flow that value has type and can be null.
In this case - for `CallbackType` - I've specified first parameter as `err?: Error`, but for 
`callback(null, tx)` flow gives error
Maybe you know right way how to type parameter in this case?

clark800: The question mark notation allows `null`, maybe the flow error you're seeing is due to something else. http://flowtype.org/docs/nullable-types.html

src/api/ledger/transaction.js
darkdarkdragon: Thanks! Forgot about `Infinity`

src/api/ledger/accountinfo.js
darkdarkdragon: No, it is just returns open ledger - I've checked :smile:

clark800: Thanks for checking! We have to fix that then. I guess wherever we call into core we have to specify `validated` then.

src/api/ledger/transaction.js
darkdarkdragon: Argh! Thanks! Missed that one in documentation.

src/api/server/server.js
darkdarkdragon: I've didn't find it too.
But now, finally, by looking into flow's code I've found how to annotate it - it is just `Promise<Type>`
I'll make changes to PR

clark800: Nice

darkdarkdragon:
Should I just change types here https://github.com/ripple/ripple-lib/blob/develop/src/api/ledger/utils.js#L84
Or we can remove calls to `Number` constructor in `compareTransactions` and rely on server that `ledgerVersion` and `indexInLedger` will always be present and be number?
Or make that check in `parseOutcome`?

clark800:
Yeah, I think it should be checked in `parseOutcome`

darkdarkdragon:
ok

src/api/server/server.js
darkdarkdragon: My bad. It is recursive, I just took that field from documentation and didn't changed.

src/api/ledger/accountinfo.js
```diff
@@ -18,21 +55,23 @@ function formatAccountInfo(response) {
...
+function getAccountInfoAsync(account: string, options: AccountInfoOptions,
+  callback: AccountInfoCallback) {
+    validate.address(account);
+    validate.getAccountInfoOptions(options);
 
-  const request = {
-    account: account,
-    ledger: options.ledgerVersion || 'validated'
-  };
+    const request = {
```
clark800: I don't think we should have to indent the whole function an extra level.

src/api/ledger/transaction-types.js
```diff
@@ -0,0 +1,147 @@
...
+  address: string,
+  sequence: number
+}
+
+export type TransactionOptions = {
+  minLedgerVersion?: number,
+  maxLedgerVersion?: number
+}
+
+export type GetTransactionResponse = PaymentTransaction | OrderTransaction |
```
clark800: This is really cool having the types in a separate file!

darkdarkdragon: I like it too :smile:

src/api/ledger/accountinfo.js
darkdarkdragon: How to do it right?
First line of function 
`function getAccountInfoAsync(account: string, options: AccountInfoOptions,
  callback: AccountInfoCallback) {`
is too long, so I moved part of it to new line and indented it, but that requires to indent more all other function lines.

clark800: I don't know the best solution, but what I did was put `) {` on a line by itself.

src/api/ledger/parse/utils.js
```diff
@@ -62,8 +62,9 @@ function parseOutcome(tx: Object): ?Object {
     fee: utils.common.dropsToXrp(tx.Fee),
     balanceChanges: balanceChanges,
     orderbookChanges: orderbookChanges,
-    ledgerVersion: tx.ledger_index,
-    indexInLedger: tx.meta.TransactionIndex
+    ledgerVersion: typeof tx.ledger_index === 'number' ? tx.ledger_index : 0,
```
clark800: Could you do `tx.ledger_index || 0`? And the same below?

src/api/ledger/accountinfo.js
clark800: Here is the eslint issue. Looks like someone else had the same idea: https://github.com/eslint/eslint/issues/3173

src/api/ledger/parse/utils.js
darkdarkdragon: We are assuming that server will always send number?
Ok, I'll make change

clark800: I think it should be a number. Have you ever seen it as something else?

sublimator: I've only seen 'ledger_index' as a string in the ledger header (i.e.
'ledger' cmd responses) but there might be others

@rec is looking at making it consistently a number

PR #497, opened 2015-08-12
[WIP] Replace sjcl.bn with bn.js

sublimator:


sublimator:
This is going to be an eslint moan fest!

src/core/serializedtypes.js
```diff
@@ -308,25 +311,25 @@ const STInt64 = exports.Int64 = new SerializedType({
...
     } else if (isString(value)) {
       if (!isHexInt64String(value)) {
         throw new Error('Not a valid hex Int64.');
       }
-      bigNumObject = new SJCL_BN(value, 16);
-    } else if (value instanceof SJCL_BN) {
-      if (!value.greaterEquals(0)) {
+      bigNumObject = new BN(value, 16);
+    } else if (value instanceof BN) {
+      if (value.ucmp(new BN(0)) == - 1) {
```
sublimator: We need to use `ucmp` which compares the values as if they are both unsigned. We should probably make a BN(0) const somewhere.

scripts/verify_ledger_json.js
```diff
@@ -2,8 +2,9 @@
 'use strict';
 
 var fs = require('fs');
-var Amount = require('../dist/npm').Amount;
-var Ledger = require('../dist/npm').Ledger;
+var ripple = require('../dist/npm')._DEPRECATED;
```
sublimator: ouch

src/core/uint.js
```diff
@@ -109,15 +111,19 @@ UInt.prototype.copyTo = function(d) {
 };
 
 UInt.prototype.equals = function(d) {
-  return this.is_valid() && d.is_valid() && this._value.equals(d._value);
+  return this.is_valid() && d.is_valid() && this.cmp(d) === 0;
+};
+
+UInt.prototype.cmp = function(o) {
```
sublimator: What should this return if any of the operands is not valid ?

src/core/serializedobject.js
```diff
@@ -257,11 +249,8 @@ SerializedObject.jsonify_structure = function(structure, field_name) {
 
       if (typeof structure.to_json === 'function') {
         output = structure.to_json();
-      } else if (structure instanceof sjcl.bn) {
-        output = ('0000000000000000' +
-                   normalize_sjcl_bn_hex(structure.toString())
-                  .toUpperCase()
-                  ).slice(-16);
+      } else if (structure instanceof BN) {
+        output = utils.arrayToHex(structure.toArray('bn', 8));
```
sublimator: This code kinda of assumes any BN is a UInt64

src/core/ledger.js
```diff
@@ -47,11 +46,11 @@ Ledger.prototype.calcAccountRootEntryHash = function(account) {
  * @return {UInt256}
  */
 Ledger.calcOfferEntryHash =
-Ledger.prototype.calcOfferEntryHash = function(account, sequence) {
-  account = UInt160.from_json(account);
-  sequence = parseInt(sequence, 10);
+Ledger.prototype.calcOfferEntryHash = function(_account, _sequence) {
```
sublimator: Do we prefer `param_` or `_param` ?

I almost tried sneaking in a directive to disable that check :)

src/core/serializedtypes.js
```diff
@@ -308,25 +296,25 @@ const STInt64 = exports.Int64 = new SerializedType({
...
     } else if (isString(value)) {
       if (!isHexInt64String(value)) {
         throw new Error('Not a valid hex Int64.');
       }
-      bigNumObject = new SJCL_BN(value, 16);
-    } else if (value instanceof SJCL_BN) {
-      if (!value.greaterEquals(0)) {
+      bigNumObject = new BN(value, 16);
+    } else if (value instanceof BN) {
+      if (value.ucmp(new BN(0)) === -1) {
```
sublimator: Should probably habitually use `$comparison < 0` eh?

src/core/serializedtypes.js
```diff
@@ -308,25 +296,25 @@ const STInt64 = exports.Int64 = new SerializedType({
...
         throw new Error('Negative value for unsigned Int64 is invalid.');
       }
       bigNumObject = value;
     } else {
-      throw new Error('Invalid type for Int64');
+      throw new Error('Invalid type for Int64: ' + (typeof value) + ' value');
     }
-    serializeBits(so, bigNumObject.toBits(64), true); // noLength = true
+    serializeBytes(so, bigNumObject.toArray('be', 8), true); // noLength = true
```
sublimator: 'be' means big endian, and the following arg is the byte length, which it will pad with 0s if not enough bytes, or throw if over

src/core/utils.js
```diff
@@ -77,7 +77,7 @@ function hexToArray(h) {
 
 function arrayToHex(a) {
   return a.map(function(byteValue) {
-    const hex = byteValue.toString(16);
+    const hex = byteValue.toString(16).toUpperCase();
```
sublimator: We use upper case everywhere right, so may as well just make this upper case off the bat. It's better to do this per byte, or once joined?

scripts/verify_ledger_json.js
sublimator: Maybe we should add some script usage to the CI script?

src/core/uint.js
```diff
@@ -225,26 +228,13 @@ UInt.prototype.parse_bytes = function(j) {
   return this;
 };
 
-
 UInt.prototype.parse_json = UInt.prototype.parse_hex;
 
-UInt.prototype.parse_bn = function(j) {
```
sublimator: Do we need to_bn/ parse_bn ?

src/core/ledger.js
clark800: I think the ideal solution is to use strict types and more descriptive names like `const sequence = parseInt(sequenceString, 10)`

src/core/uint.js
clark800: Throw an exception?

src/core/serializedtypes.js
clark800: How about adding `lessThan` and `greaterThan` helper functions?

src/core/serializedtypes.js
clark800: Might be good to put this comment in the code.

src/core/utils.js
clark800: I'm ok either way

clark800:
LGTM

src/core/uint.js
sublimator: sounds good to me

src/core/ledger.js
sublimator: Agree with the descriptive names, in the general sense but I think `sequence` would most likely be passed in as a Number here.

src/core/serializedtypes.js
sublimator: will do

PR #499, opened 2015-08-12
Get validated ledger in getSettings and getAccountInfo

clark800:
I think the other "get" methods already are getting the last validated ledger.

### DIFF
--- a/src/api/ledger/accountinfo.js
+++ b/src/api/ledger/accountinfo.js
@@ -24,7 +24,7 @@ function getAccountInfoAsync(account, options, callback) {
 
   const request = {
     account: account,
-    ledger: options.ledgerVersion
+    ledger: options.ledgerVersion || 'validated'
   };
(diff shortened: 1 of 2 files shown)


wltsmrz:
Pretty much, might still be better to pass in 'validated' though

https://github.com/ripple/ripple-lib/blob/002102ce626fecf33e123002fe215f22cfe483aa/src/api/ledger/balances.js#L37-L38
https://github.com/ripple/ripple-lib/blob/002102ce626fecf33e123002fe215f22cfe483aa/src/core/remote.js#L529

wltsmrz:
LGTM

clark800:
I believe the methods that use remote.getLedgerSequence can't use the "validated" string because they have to lock in a specific ledger version because they are making multiple requests.

wltsmrz:
Oh yeah

PR #500, opened 2015-08-13
Add getLedgerHeader method

clark800:
Added parent_close_time and close_flags to the rippled ledger response because they are needed for computing the ledger hash; they are being added to rippled with this pull request: https://github.com/ripple/rippled/pull/1237


src/api/ledger/ledger-header.js
```diff
@@ -0,0 +1,40 @@
...
+    ledgerVersion: parseInt(header.ledger_index || header.seqNum, 10),
+    parentLedgerHash: header.parent_hash,
+    parentCloseTime: header.parent_close_time,
+    totalDrops: header.total_coins || header.totalCoins,
+    transactionHash: header.transaction_hash
+  };
+}
+
+function getLedgerHeaderAsync(ledgerVersion, callback) {
+  validate.ledgerVersion(ledgerVersion);
```
wltsmrz: This doesn't break if ledgerVersion is undefined, right?

clark800: Hmmm, seems like it should break, but it's not. Seems like a bug in is-my-json-valid. I'll add an `if` around this so it won't break when they fix it.

clark800:
Updated

wltsmrz:
LGTM

PR #501, opened 2015-08-13
Replace sjcl.bn with bn.js

sublimator:
Something went awry with the last pull wreck:  #497


src/core/ledger.js
```diff
@@ -65,22 +61,23 @@ Ledger.prototype.calcOfferEntryHash = function(account, sequence) {
...
-  account1 = UInt160.from_json(account1);
-  account2 = UInt160.from_json(account2);
-
-  if (!account1.is_valid()) {
+Ledger.prototype.calcRippleStateEntryHash = function(account1,
+                                                     account2,
+                                                     currencyArg) {
+  const currency = Currency.from_json(currencyArg);
+  let lowAccount = UInt160.from_json(account1);
+  let hiAccount = UInt160.from_json(account2);
```
clark800: We don't know that it is the high account yet, so we shouldn't give it this name. I also fixed this in another pull request, here is what I did: https://github.com/ripple/ripple-lib/pull/498/files#diff-d2006f6549215f8372ec508459e5ce44R93

src/core/serializedtypes.js
```diff
@@ -308,25 +296,27 @@ const STInt64 = exports.Int64 = new SerializedType({
...
     } else if (isString(value)) {
       if (!isHexInt64String(value)) {
         throw new Error('Not a valid hex Int64.');
       }
-      bigNumObject = new SJCL_BN(value, 16);
-    } else if (value instanceof SJCL_BN) {
-      if (!value.greaterEquals(0)) {
+      bigNumObject = new BN(value, 16);
+    } else if (value instanceof BN) {
+      if (value.sign) {
```
clark800: It doesn't look like `sign` is in the docs, maybe it's better to use `cmp`

src/core/uint.js
```diff
@@ -108,16 +110,32 @@ UInt.prototype.copyTo = function(d) {
...
+UInt.prototype.equals = function(o) {
+  return this.is_valid() &&
+         o.is_valid() &&
+         // This throws but the expression will short circuit
+         this.cmp(o) === 0;
+};
+
+UInt.prototype.cmp = function(o) {
+  assert(this.is_valid() && o.is_valid());
+  return this._value.ucmp(o._value);
```
clark800: Why ucmp here?

src/core/uint.js
```diff
@@ -108,16 +110,32 @@ UInt.prototype.copyTo = function(d) {
...
 };
 
 UInt.prototype.is_valid = function() {
-  return this._value instanceof sjcl.bn;
+  return this._value instanceof BN;
 };
 
 UInt.prototype.is_zero = function() {
-  return this.is_valid() && this._value.equals(new sjcl.bn(0));
+  return this.is_valid() && this._value.cmpn(0) === 0;
```
clark800: Why cmpn here? bn.js docs are pretty sparse, so we might need comments to clarify what these are

sublimator: cmpn means cmp with n)umber

src/core/uint.js
sublimator: They are UInt, so it's fine. I think I maybe have been confused that it was required when I used it. `u` means u)nsigned comparison. cmp actually calls it, after checking `sign`

clark800: Probably better to use `cmp` throughout to avoid confusion

sublimator: pushed some address-ness-ness

clark800:
LGTM

src/core/serializedtypes.js
sublimator: Hrmmm, it's not using `_private` naming, so it's probably OK, but I can use cmp, as it's easier to read at the least

clark800: Not everyone uses that convention though. It's generally best to only use explicitly documented features.

PR #502, opened 2015-08-13
Remove sjcl-extended/ripple-wallet-generator. Use hash.js & sjcl-codec.

sublimator:
This is base on bn.js branch / pull request #501. See the last commit for the diff rather than the `changes` tab

Note that we now use a tiny build of sjcl, that includes only bitArray and sjcl.codec for hex, bytes and utf8.  We could extract this into a module on npm and access the functionality from that module rather than from "sjcl" which, when it's cut down build, is misleading.

hash.js from the elliptic family is used for all hashing.


src/api/common/utils.js
```diff
@@ -27,6 +28,11 @@ function toRippledAmount(amount: Amount): string|Amount {
   };
 }
 
+function generateWallet(options) {
```
sublimator: flow errors around here

sublimator: and yeah, still happening

clark800: Didn't it say "missing annotation"? Have you tried adding an annotation?

sublimator: Yeah, I did at some point, but then it complained about module missing for `ripple-keypairs`. I'll look into it more.

test/api-test.js
```diff
@@ -413,9 +413,11 @@ describe('RippleAPI', function() {
   });
 
   it('generateWallet', function() {
-    withDeterministicPRNG(() => {
-      assert.deepEqual(this.api.generateWallet(), responses.generateWallet);
-    });
+    function rand() {
+      return _.fill(Array(16), 0);
+    }
+    assert.deepEqual(this.api.generateWallet({randGen: rand}),
```
clark800: Do you think it would look better if we called it `random`?

sublimator: Sure

.flowconfig
```diff
@@ -1,7 +1,7 @@
 [ignore]
 .*/src/api/.*
 .*/src/core/.*
-.*/dist/.*
+.*/(?<!ripple-keypairs)/dist/.*
```
clark800: Does something like this work? `.*/(?<!node_modules/.*)/dist/.*`

sublimator: I tried briefly with an online regex debug, before moving on. If you can confirm it works I'll change it.

sublimator: Hopefully facebook will fix the issue

test/api-test.js
sublimator: Wait till I add the brorand module lol :)

src/api/common/utils.js
```diff
@@ -27,6 +28,11 @@ function toRippledAmount(amount: Amount): string|Amount {
   };
 }
 
+function generateWallet(options: Object): Object {
```
sublimator: How do I denote that options is err ... optional?  `options?: Object`  ?

src/core/account.js
```diff
@@ -41,7 +42,7 @@ function Account(remote, account) {
   this._entry = { };
 
   function listenerAdded(type) {
-    if (~Account.subscribeEvents.indexOf(type)) {
+    if (Account.subscribeEvents.indexOf(type) !== -1) {
```
sublimator: I guess there's an app for that ? lodash.contains or something?

src/api/common/utils.js
clark800: yeah

src/core/account.js
clark800: `_.includes`

.flowconfig
```diff
@@ -1,7 +1,7 @@
 [ignore]
 .*/src/api/.*
 .*/src/core/.*
-.*/dist/.*
+.*/(?<!ripple-keypairs/)dist/.*
```
clark800: I don't think this is working as intended. I tried changing 'ripple-keypairs' to some other string and it still worked. Maybe we should just delete the "dist" line; it still works if you do that. It would be slightly slower because it's checking more files, but at least we wouldn't have this problem again.

src/core/utils.js
```diff
@@ -166,6 +166,6 @@ exports.fromTimestamp = fromTimestamp;
 exports.getMantissaDecimalString = getMantissaDecimalString;
 exports.getMantissa16FromString = getMantissa16FromString;
 
-exports.sjcl = require('sjcl-extended');
+exports.sjcl = require('sjcl-codecs');
```
clark800: Is this temporary?

test/api-test.js
```diff
@@ -413,9 +413,11 @@ describe('RippleAPI', function() {
   });
 
   it('generateWallet', function() {
-    withDeterministicPRNG(() => {
-      assert.deepEqual(this.api.generateWallet(), responses.generateWallet);
-    });
+    function random() {
+      return _.fill(Array(16), 0);
+    }
+    assert.deepEqual(this.api.generateWallet({randGen: random}),
```
clark800: Oh, I meant `randGen` -> `random`

sublimator: Ah, right, that would be nice eh: `{random}`

src/core/utils.js
sublimator: Yes, definitely temporary, just a sketch. What do you think about an npm
package with a lite sjcl build with a unique name though? I didn't want to
just depend on sjcl and a postinstall script because it can be misleading
to require "sjcl" then.

Having said that, could just add a './sjclcodecs.js' module that adds some
indirection.

I wish sjcl were more like lodash and had highly modular npm packages.

Hrmmm?

.flowconfig
sublimator: Yeah, I wasn't sure if it was really working. I'll just update ripple-keypairs to take `random` in the generateWallet, and use some different folder instead of dist, and get on with life.

src/core/uint.js
```diff
@@ -217,16 +216,7 @@ UInt.prototype.parse_hex = function(j) {
...
-    const bytes = sjcl.codec.bytes.fromBits(j);
-    this.parse_bytes(bytes);
-  } else {
-    this._value = NaN;
-  }
-
-  this._update();
-
-  return this;
+  return this.parse_bytes(sjclcodec.bytes.fromBits(j));
```
sublimator: Other than these bitty *bits routines, all we depend upon in sjcl is the utf8 codec really. We could just cull these, and find some other utf8 codec. I'm leaning ever so lightly towards just keeping these though.

clark800: I think we should fully eliminate sjcl. I don't know what the UTF-8 codec does exactly, but it sounds like it's probably only a few lines of code.

.flowconfig
clark800: That means we might run into this confusing issue again though, it's hard to know if/when the flow team will fix it: https://github.com/facebook/flow/issues/510

Isn't deleting the "dist" line a better option for now?

clark800:
LGTM

PR #503, opened 2015-08-14
annotate api functions with types

darkdarkdragon:


src/api/ledger/parse/utils.js
```diff
@@ -62,8 +62,8 @@ function parseOutcome(tx: Object): ?Object {
     fee: utils.common.dropsToXrp(tx.Fee),
     balanceChanges: balanceChanges,
     orderbookChanges: orderbookChanges,
-    ledgerVersion: tx.ledger_index,
-    indexInLedger: tx.meta.TransactionIndex
+    ledgerVersion: tx.ledger_index || 0,
```
clark800: Thinking about this again makes me realize we probably shouldn't put 0 there. When are these fields not present? I believe it's when you get a transaction that is not yet validated right? In that case I think we want these fields to be not present.

clark800: Actually that's not right, the whole `outcome` section should be not present if it's not validated. When is outcome present but not these fields?

src/api/ledger/transaction-types.js
```diff
@@ -0,0 +1,147 @@
...
+  counterparty: string,
+  limit: string,
+  qualityIn?: number,
+  qualityOut?: number,
+  ripplingDisabled?: boolean,
+  authorized?: boolean,
+  frozen?: boolean
+}
+
+type GetSettings = {
```
clark800: Could we call this `Settings` since the others don't have `Get` in the name?

darkdarkdragon: Not a problem. It was just copy from schema :smile:

src/api/ledger/parse/utils.js
darkdarkdragon: Originally I've added this because I've removed checks from `compareTransactions` function.
If we sure what this field always will be present we can remove this.

BTW, can unvalidated transactions be part of `getTransactions` response? In that case `compareTransactions` will throw exception.

clark800: Yes, I think if `outcome` is present, then `ledgerVersion` and `indexInLedger` should always be present.

> can unvalidated transactions be part of getTransactions response?

Yes

darkdarkdragon:
@clark800 I did the changes, you can review it now.

clark800:
LGTM

PR #504, opened 2015-08-14
Add getLedger method, remove getLedgerHeader method

clark800:


src/api/ledger/ledger.js
```diff
@@ -0,0 +1,26 @@
...
+    transactions: options.includeTransactions,
+    accounts: options.includeAccounts
+  };
+
+  this.remote.requestLedger(request,
+    composeAsync(response => parseLedger(response.ledger), callback));
+}
+
+function getLedger(options={}) {
+  return utils.promisify(getLedgerAsync.bind(this))(options);
```
wltsmrz: Any reason you chose

`utils.promisify(getLedgerAsync.bind(this))(options);`

rather than

`utils.promisify(getLedgerAsync).call(this, options);`

src/api/ledger/parse/ledger.js
```diff
@@ -0,0 +1,50 @@
...
+
+function parseTransactions(transactions) {
+  if (!transactions || transactions.length === 0) {
+    return {};
+  }
+  if (typeof transactions[0] === 'string') {
+    return {transactionHashes: transactions};
+  }
+  return {
+    transactions: _.map(transactions, parseTransaction),
```
wltsmrz: Why do we bother using lodash here?

sublimator: Could just use `transactions.map` eh?

src/api/ledger/ledger.js
sublimator: Latter form does indeed read a little nicer

src/api/common/schemas/get-ledger.json
```diff
@@ -14,7 +14,12 @@
     "parentLedgerHash": {"$ref": "hash256"},
     "parentCloseTime": {"type": "integer", "minimum": 0},
     "totalDrops": {"$ref": "value"},
-    "transactionHash": {"$ref": "hash256"}
+    "transactionHash": {"$ref": "hash256"},
+    "transactions": {"type": "array", "items": {"type": "object"}},
+    "rawTransactions": {"type": "string"},
```
sublimator: What's the idea behind stringifying these?

src/api/common/schemas/get-ledger.json
```diff
@@ -14,7 +14,12 @@
...
+    "transactionHashes": {"type": "array", "items": {"$ref": "hash256"}},
+    "rawAccounts": {"type": "string"},
```
sublimator: With regard to stringifying, it's not like the Paths in the transactions that just get passed around, and are supposed to be treated as an opaque string. 

If one of the design goals is that Most RippleAPI users can be oblivious to rippled json formats, they simply don't need to see the raw versions so perhaps their inclusion should be opt in.  And if they have opted in, it doesn't make much sense to stringify it IMO. Anyone who wants to do something with the rawAccounts will have to parse that (probably big) string.

src/api/common/schemas/ledger-options.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "ledger-options",
+  "description": "Options for getLedger",
+  "type": "object",
+  "properties": {
+    "ledgerVersion": {"$ref": "ledgerVersion"},
+    "includeAllData": {"type": "boolean"},
+    "includeTransactions": {"type": "boolean"},
+    "includeAccounts": {"type": "boolean"}
```
sublimator: Is there a better term than `accounts` in `includeAccounts`, `rawAccounts` etc ?

test/mock-rippled.js
```diff
@@ -24,6 +25,27 @@ function createResponse(request, response, overrides={}) {
...
+    if (!request.accounts) {
+      delete newResponse.result.ledger.accountState;
+    }
+    // the following fields were not in the ledger response in the past
+    if (newResponse.result.ledger.close_flags === undefined) {
+      newResponse.result.ledger.close_flags = 0;
+    }
+    if (newResponse.result.ledger.parent_close_time === undefined) {
+      newResponse.result.ledger.parent_close_time =
+        newResponse.result.ledger.close_time - 10;
```
sublimator: Why guess ?

sublimator: Oh, this is just mock-rippled code huh

src/api/ledger/ledger.js
wltsmrz: More importantly, `bind` is much slower

src/api/ledger/parse/ledger.js
wltsmrz: Seems weird that the new API code tends to use lodash for array iteration methods strictly... in this parseTransactions() function we could use _.isEmpty() for the first conditional and _.isString() for the second...

src/api/ledger/ledger.js
sublimator: Probably only really matters in code where the function gets called thousands of times a second?

wltsmrz: IMO performance shouldn't ever be ignored, at least obvious stuff like this. The performance difference here is fairly significant, and it adds up when you've got a whole stack of functions that call Function.bind (as the case seems to be here), and are called many times throughout the lifetime of a program.  I also find the `call` version slightly more readable as you also mentioned, so it seems like a no brainer

sublimator: You're right, `call` is a clear winner if it's more readable and has perf advantages

src/api/common/schemas/get-ledger.json
sublimator: So I guess what I'm asking is, what's the purpose of these raw* fields? Who is going to use them?

src/api/ledger/parse/ledger.js
clark800: The advantage of Lodash is that it can optimize for each browser/interpreter. Using `isEmpty` and `isString` is a good idea.

src/api/ledger/ledger.js
clark800: Not sure if I agree it's more readable, I actually kind of like having the context binding separated from the function arguments. Regarding optimization, I tend to agree with the quote: "We should forget about small efficiencies, say about 97% of the time: premature optimization is the root of all evil”. http://www.catb.org/esr/writings/taoup/html/ch01s06.html#ftn.id2878872

In this case, I don't think it's a problem, so I'll change it since you both prefer it.

src/api/common/schemas/get-ledger.json
clark800: The `raw` forms will be used in `computeLedgerHash`. The convention is to keep rippled format objects as strings to emphasize that the user should never look at them. The `rawTransactions` string is mostly redundant with the `transactions` array. There is not yet a parsed version of `rawAccounts`, but that may be added later if demand exists.

src/api/common/schemas/ledger-options.json
clark800: Possibly, suggestions?

test/mock-rippled.js
clark800: Yeah

src/api/common/schemas/get-ledger.json
clark800: I updated `computeLedgerHash` to use the `raw` fields.

clark800:
@sublimator @wltsmrz updated

src/api/common/schemas/ledger-options.json
sublimator: I don't have any suggestions, but I remember `accountState` never really going down well on the first bite. Do you have a new term for ledger entries ?

clark800: Ideas: `cumulativeState`, `ledgerState`, `state`

sublimator: My first vote is for ledgerState out of those, then state, and finally cumulativeState, but only if you put a gun to my head.

src/api/common/schemas/ledger-options.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "ledger-options",
+  "description": "Options for getLedger",
+  "type": "object",
+  "properties": {
+    "ledgerVersion": {"$ref": "ledgerVersion"},
+    "includeAllData": {"type": "boolean"},
```
sublimator: While you're tinkering around, considering stuff in the area: includeAllData might be interpreted as a shorthand for includeLedgerState and includeTransactions

src/api/ledger/orderbook.js
```diff
@@ -77,7 +78,7 @@ function getOrderbookAsync(account, orderbook, options, callback) {
 }
 
 function getOrderbook(account: string, orderbook: Object, options={}) {
-  return utils.promisify(getOrderbookAsync.bind(this))(
+  return utils.promisify(getOrderbookAsync).call(this,
```
sublimator: Forgive me if I asked this before, I'm getting senile, but would there be any advantage to stashing away a promisified func? Or they have state or something?

src/api/common/schemas/ledger-options.json
sublimator: Although, `ledger` is implied by the context, so maybe I like that more cause it's short. Shrug.

src/api/common/schemas/get-ledger.json
sublimator: rot13 the strings! (j/k)

sublimator: Out of scope of this P.R. but have you considered making it so that it's possible to convert losslessly between rippled <-> RippleAPI json formats?

src/api/ledger/orderbook.js
clark800: What do you mean by stashing away?

src/api/common/schemas/ledger-options.json
clark800: Ideas?

sublimator:
Basically LGTM

src/api/common/schemas/get-ledger.json
clark800: Yeah, but there are some things in the rippled format that we don't want to expose because they are meaningless to users like `BookNode` and `OwnerNode`, but I'm not sure if they factor into the hash. So it may be possible to eliminate the `raw` fields; it would just require some research to see.

src/api/ledger/orderbook.js
sublimator: basically `const ref = utils.promisify(getOrderbookAsync);` or whatever form makes sense

not a big deal, just throwing it out there

src/api/common/schemas/ledger-options.json
sublimator: Another one to consider: ledgerEntries

src/api/common/schemas/get-ledger.json
sublimator: > I'm not sure if they factor into the hash

Every single byte factors into the hash

sublimator: > there are some things in the rippled format that we don't want to expose

I see. Hrmmm

src/api/common/schemas/ledger-options.json
sublimator: 4am here, just woke up from weird dreams, drawing blanks, nothing I like popping out

You could flip it out around to saying something about including only ids/hashes,  e.g. itemIdentifiersOnly which defaults to true. 

I got nothing haha

src/api/common/schemas/ledger-options.json
sublimator: Or simply entries

clark800:
@sublimator updated

sublimator:
L:G:T:M

src/api/offline/ledgerhash.js
```diff
@@ -28,25 +28,47 @@ function hashLedgerHeader(ledgerHeader) {
...
+  const ledgerObject = common.core.Ledger.from_json({accountState: state});
+  const stateHash = ledgerObject.calc_account_hash().to_hex();
+  if (ledger.stateHash !== undefined && ledger.stateHash !== stateHash) {
+    throw new common.errors.ValidationError('stateHash in header'
+      + ' does not match computed hash of state');
+  }
+  return stateHash;
+}
+
+function computeLedgerHash(ledger: Object): string {
```
sublimator: Is this a method on the RippleAPI object? If rippled (or a satellite app, if rippled were reduced to p2p txn processor) were suddenly serving this api, could you see computeLedgerHash as an exposed method?

clark800: > Is this a method on the RippleAPI object?

Yes

> could you see computeLedgerHash as an exposed method?

It's possible, but clearly it is in a different category in the sense that it is not strictly necessary for interacting with the ripple network. It would be nice if it were a module in ripple-lib-extensions, but that would require extracting all the serialization code too. Maybe we could make a serialization module in ripple-lib-extensions.

sublimator: Maybe it would be nice to clearly delineate (at least with comments/docs)
what is part of the new json api and what are client functions for using
it.

clark800: Those two categories aren't necessarily disjoint.

sublimator: You're being pedantic, you know what I meant :)

clark800: Online vs offline? It's still not decided for sure, but I think the offline methods will be moved to separate modules.

sublimator: What I'm trying to get at is that there's this new ripple api which which
has a request/response interface speaking json primitives for parameters
and return values that would allow it to be easily exported from an RPC
server.

There's exports like generateWallet, where the analogous rippled command
"wallet_create" will only serve to local connections. computeLedgerHash
seems somewhat in the same class. Both could feasibly be seen to be
exported from a rippled (or a satellite app if rippled became p2p txn
processing only) served json rpc API.

But there's little guys like isConnected, connect, disconnect which are
clearly `client` routines.

Take for example the getTransactions api, where it's conceivable people may
want an event handler to be executed for each page of transactions as they
come in to allow them to update a UI. They might build a wrapper around the
getTransactions service. That would be a `client` routine.

And `errors`?

I think if say in the src/api/index.js what are server and what are client
routines were clearly marked out, it would help communicate the ultimate
intent of some of these apis one day being accessed via RPC which unless I
got confused somewhere is the goal?

It's getting a little weird to see things like
RippleAPI.prototype.isValidAddress, (which would likely not be exported
from a json rpc) sort of tacked on as a method, seemingly only because
RippleAPI is currently the only namespace exported.

sublimator: RippleAPI, with its connect/disconnect etc seems like a class that might
have in another life been called a Client. isValidAddress() doesn't seem
like a method for a client, but rather a function from some utility or
validation module. And if `isValidAddress` why not `isValidAmount` and so
on, and where will they be exported from ?

clark800: Do you have any specific ideas? We already have plans to create a Wallet module, which will likely be exported parallel to RippleAPI and would have generateAddress and isValidAddress. We also have plans to make more documentation, but that is lower priority until it's more stable. I don't have an idea for where to put computeLedgerHash yet.

sublimator: I'm just suggesting something simple for the moment, like keeping the `RPC
handlers` grouped with a comment. You don't have to go so far as creating a
separate map, to extend RippleAPI.prototype from with the fancy _.assign
magic etc.

i.e. Just draw some lines around the shapes that are emerging so people
know where you're going. Broadly you have 3 classes of routines here,
future rpc handlers, rpc client methods, and utils for working with the new
api formats. You've said there's Wallet utils, and at some point XRPValue &
IOUAmount will be exposed, but maybe from another repo.

I guess it might not hurt to pull out your crystal ball a bit and try and
work out what sort of features you plan to expose?

PR #505, opened 2015-08-15
[WIP] Fix reserve calculation

sublimator:


src/core/server.js
```diff
@@ -905,6 +907,9 @@ Server.prototype._feeTxUnit = function() {
  */
 
 Server.prototype._reserve = function(ownerCount) {
+  // We should be in a valid state before calling this method
+  assert(this._reserve_base && this._reserve_inc);
```
sublimator: I'm in favour of blowing up here rather than returning some invalid Amount. It seemed like the tests before showed it was both throwing errors when not connected and returning NaN invalid amounts.

test/remote-test.js
```diff
@@ -935,11 +935,10 @@ describe('Remote', function() {
...
-  it('Get reserve', function() {
+  it('reserve() before reserve rate known', function() {
     remote._connected = true;
     remote._servers[0]._connected = true;
-    assert.strictEqual(remote.reserve(1).to_json(), 'NaN');
-    remote._servers = [];
+    // Throws because the server has not had reserve_inc, reserve_base set
     assert.throws(function() {
       remote.reserve(10).to_json();
     });
```
sublimator: guess could change the assert to an identifiable error

src/core/server.js
```diff
@@ -913,7 +918,7 @@ Server.prototype._reserve = function(ownerCount) {
     throw new Error('Owner count must not be negative.');
   }
 
-  return reserve_base.add(reserve_inc.product_human(owner_count));
+  return reserve_base.add(reserve_inc.multiply(owner_count));
```
sublimator: product_human, once upon a time had special casing for integer Numbers, coercing them into strings, then tacking a '.0' on the end, so the amount would be parsed as XRP rather than drops.

clark800:
LGTM

sublimator:
haha, you merged [WIP] [WIP] :) 

Now we even from that time back when :+1:

clark800:
Oh haha, I even checked the commit log, but I guess I instinctively ignore anything inside square brackets.

PR #506, opened 2015-08-16
Remove UInt#from_bn

sublimator:
parse_bn was removed in the move to elliptic so removing from_bn to match

If no one noticed before, and rues the day it was excised, we could always add it back in.


clark800:
LGTM

PR #509, opened 2015-08-17
Rename generateWallet to generateAddress

clark800:
In the future, we may make a "wallet" class that contains multiple addresses.


src/api/common/index.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
   dropsToXrp: utils.dropsToXrp,
   xrpToDrops: utils.xrpToDrops,
   toRippledAmount: utils.toRippledAmount,
-  generateWallet: utils.generateWallet,
+  generateAddress: utils.generateAddress,
```
sublimator: `generateAddress` is growing on me

@darkdarkdragon  What you think about this name?

sublimator: Oh, this might want to support ed25519 at some point huh?

darkdarkdragon: `generateAddress` is definitely better than `generateWallet`.
Maybe `generateRippleAddress` will be even better?

darkdarkdragon: Isn't `ed25519` is supported now?
Looking into code I see you could call 
`generateAddress(type='ed25519')`

sublimator: haha, lol I added that days ago, I'm getting senile

sublimator: I think 'ripple' is kinda implied?

darkdarkdragon: It is because you get used to it.

Words `RippleName` and `RippleAddress` used all over documentation, and casual developer will expect to find same in code.



On Mon, Aug 17, 2015 at 8:00 PM -0700, "sublimator" <notifications@github.com> wrote:
> @@ -9,7 +9,7 @@ module.exports = {
>    dropsToXrp: utils.dropsToXrp,
>    xrpToDrops: utils.xrpToDrops,
>    toRippledAmount: utils.toRippledAmount,
> -  generateWallet: utils.generateWallet,
> +  generateAddress: utils.generateAddress,

I think 'ripple' is kinda implied?


---
Reply to this email directly or view it on GitHub:
https://github.com/ripple/ripple-lib/pull/509/files#r37260251

sublimator: That's a good point. @clark800 ?

clark800: I'd prefer `generateAddress` over `generateRippleAddress`. I don't think there is any ambiguity because there is only one type of address in the API.

vinniefalco: I'm pushing for these names in rippled:
`AccountID` - 160 bit account identifier
`AccountPublicKey` - Public key for an account
`PublicKey` - A general public key

clark800: @vinniefalco I assume you're suggesting `generateAccountID`? I consider "address" a synonym for "accountID", so I think technically either works, but "address" is a bit easier to say and familiar from bitcoin.

vinniefalco: I'm not really suggesting anything specific just providing some info. "Address" is a heavily abused term in the rippled code base. Specifically, a `RippleAddress` is a "kitchen sink" type object which can hold almost anything. It can hold a "node id", AccountID, public key, private key, validator public key, validator private key, and more. Which breaks the type-safety of the C++ language and leads to all kinds of unnecessary ugliness.

clark800: Ah, I see. Thanks for the info. I agree that we should have precise names. In this API "address" should always mean "accountID".

sublimator: I prefer accountID to address too

sublimator:
LGTM

sublimator:
> This branch has conflicts that must be resolved

src/api/common/index.js
clark800: Maybe we should get some more opinions on the naming then

sublimator: I mean in the general sense. Here where it's generating an accountID and a
secret it feels somehow better to use generateAddress. Only alternative
name that pops to mind is generateAccountKeys which I guess would make
sense when RegularKey joins the party. But then, "just use the accountID in
the RegularKey field" ...

vinniefalco: `generateAccountPair`

sublimator: for when you wanna make two accounts?

vinniefalco: There is not a 1 to 1 mapping of rippled functions to these JS routines but to stimulate ideas here are the names and return types we went with:
https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/SecretKey.h#L100

PR #511, opened 2015-08-18
fix transactionFilter

darkdarkdragon:


src/api/ledger/transactions.js
```diff
@@ -29,6 +29,7 @@ function transactionFilter(address, filters, tx) {
     return false;
   }
   if (filters.counterparty && tx.address !== filters.counterparty
+      && tx.specification.destination
```
clark800: To be defensive, it might be good to also check that `tx.specification` exists in case someone adds a transaction type that doesn't have a specification in the future.

Also, we should probably add support for trustline transactions here.

And for transaction types that do not have a counterparty, this should return false because if there is no counterparty, then no transactions can match the specified counterparty.

darkdarkdragon: done

src/api/ledger/transactions.js
```diff
@@ -28,8 +28,15 @@ function transactionFilter(address, filters, tx) {
...
-  if (filters.counterparty && tx.address !== filters.counterparty
-      && tx.specification.destination.address !== filters.counterparty) {
+  if (filters.counterparty) {
+    if (tx.address === filters.counterparty || (
+      tx.specification && (
+        (tx.specification.destination &&
+          tx.specification.destination.address === filters.counterparty) ||
+        (tx.specification.counterparty === filters.counterparty)
+      ))) {
+      return true;
```
clark800: It's a bit risky to put a `return true` here because someone might add another condition after this one. Could you create a new function to check for counterparty match so that it can keep the same pattern of only returning false?

darkdarkdragon: done

src/api/ledger/transactions.js
```diff
@@ -28,13 +28,24 @@ function transactionFilter(address, filters, tx) {
...
     return false;
   }
-  if (filters.counterparty && tx.address !== filters.counterparty
-      && tx.specification.destination.address !== filters.counterparty) {
-    return false;
-  }
   return true;
 }
 
+function counterpartyFilter(address, filters, tx) {
```
clark800: Should still be called from inside the transactionFilter

darkdarkdragon: done

src/api/ledger/transactions.js
```diff
@@ -15,6 +15,21 @@ function parseAccountTxTransaction(tx) {
   return parseTransaction(tx.tx);
 }
 
+function counterpartyFilter(address, filters, tx) {
```
clark800: `address` is an unused parameter

darkdarkdragon: fixed

clark800:
LGTM

PR #512, opened 2015-08-18
Add isValidAddress

clark800:


clark800:
This method will probably be moved to a separate wallet library soon

lumberj:
LGTM

src/api/common/schema-validator.js
```diff
@@ -11,7 +11,7 @@ const ValidationError = require('./errors').ValidationError;
 
 let SCHEMAS = {};
 
-function isValidAddress(address) {
+function isValidAddress(address: string): boolean {
   return core.UInt160.is_valid(address);
```
darkdarkdragon: Do we want values like `skdkfjekcsdkdfkgxcmvnlmmoesdfgvvcdgfkjfv` or `6e3efa86a5eb0a3c5dc9beb3a204783bb00e1913` to be considered as valid addresses here?
Because UInt160 considers it valid.

clark800: Nice catch, I'd consider that a bug. Would you mind fixing it?

darkdarkdragon: Of course.

PR #513, opened 2015-08-19
fix parseTrustline for case when no QualityIn or QualityOut are set i…

darkdarkdragon:
…n transaction

### DIFF
--- a/src/api/ledger/parse/trustline.js
+++ b/src/api/ledger/parse/trustline.js
@@ -15,8 +15,11 @@ function parseFlag(flagsValue, trueValue, falseValue) {
   return undefined;
 }
 
-function parseQuality(quality) {
-  return (new BigNumber(quality)).shift(-9).toNumber();
+function parseQuality(quality?: number) {
+  if (typeof quality === 'number') {
+    return (new BigNumber(quality)).shift(-9).toNumber();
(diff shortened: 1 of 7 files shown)


src/api/ledger/parse/trustline.js
```diff
@@ -15,8 +15,10 @@ function parseFlag(flagsValue, trueValue, falseValue) {
   return undefined;
 }
 
-function parseQuality(quality) {
-  return (new BigNumber(quality)).shift(-9).toNumber();
+function parseQuality(quality?: number) {
+  if (typeof quality === 'number') {
```
clark800: We don't need typechecking here because flow should already be checking that. Can just check `quality !== undefined`. Also, the function should explicitly return `undefined` otherwise.

darkdarkdragon: This function is called with data that comes from server, so flow really can't check anything here.

clark800: Good point. Rippled docs say it should be a number, but I suppose it's safer to check. Can you add the `return undefined` though?

darkdarkdragon: done

clark800:
LGTM

PR #514, opened 2015-08-19
Update eslint to 1.2.0

sublimator:


package.json
```diff
@@ -36,10 +36,10 @@
     "assert-diff": "^1.0.1",
     "babel": "^5.8.21",
     "babel-core": "^5.8.22",
-    "babel-eslint": "^4.0.5",
+    "babel-eslint": "^4.0.10",
```
sublimator: npm will just install 4.0.10 anyway via `^` ?

sublimator: Will drop this commit, I think

wltsmrz:
LGTM

PR #515, opened 2015-08-19
in isValidAddress check that address starting with 'r'

darkdarkdragon:
because UInt160 considers valid hex values

### DIFF
--- a/src/api/common/schema-validator.js
+++ b/src/api/common/schema-validator.js
@@ -12,7 +12,9 @@ const ValidationError = require('./errors').ValidationError;
 let SCHEMAS = {};
 
 function isValidAddress(address: string): boolean {
-  return core.UInt160.is_valid(address);
+  return typeof address === 'string' && address.length > 0 &&
+    address[0] === 'r' &&
+    core.UInt160.is_valid(address);
 }
 
 function isValidLedgerHash(ledgerHash) {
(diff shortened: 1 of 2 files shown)


src/api/common/schema-validator.js
```diff
@@ -12,7 +12,8 @@ const ValidationError = require('./errors').ValidationError;
 let SCHEMAS = {};
 
 function isValidAddress(address: string): boolean {
-  return core.UInt160.is_valid(address);
+  return typeof address === 'string' && address.length && address[0] === 'r' &&
```
sublimator: Flow seems to be complaining about:
```
>>> true && 0 && false
0
```

You might need to wrap the return in Boolean to shut it up?

darkdarkdragon: Just `length > 0` is shorter :smile:

wltsmrz:
LGTM

src/api/common/schema-validator.js
```diff
@@ -12,7 +12,9 @@ const ValidationError = require('./errors').ValidationError;
 let SCHEMAS = {};
 
 function isValidAddress(address: string): boolean {
-  return core.UInt160.is_valid(address);
+  return typeof address === 'string' && address.length > 0 &&
```
clark800: `typeof address === 'string'` is redundant

darkdarkdragon: This function is exposed as public api, so it can be called with any value.

clark800: Ah good point, flow doesn't help if client code isn't using it!

clark800:
LGTM

src/api/common/schema-validator.js
darkdarkdragon: And in most cases it will not :smile:

PR #518, opened 2015-08-21
Allow to specify server url without explicit port number

darkdarkdragon:


sublimator:
LGTM

src/core/server.js
```diff
@@ -126,7 +130,9 @@ function Server(remote, opts_) {
     self._updateScore('ledgerclose', ledger);
   });
 
+/* eslint-disable no-unused-vars */
   this.on('response_ping', function onPingResponse(message, request) {
+/* eslint-enable no-unused-vars */
```
clark800: Here is a somewhat cleaner way of dealing with these situations: https://github.com/ripple/javascript-style-guide/pull/20

PR #524, opened 2015-08-24
Deprecate positional request constructor API

wltsmrz:
* abbreviated arguments (src_  / dst_) to requestPathFindCreate() etc. dropped
* requestTxHistory() dropped, requestTransactionHistory() lives on


src/core/orderbook.js
```diff
@@ -1142,20 +1142,20 @@ OrderBook.prototype.getOffersSync = function() {
 OrderBook.prototype.toJSON =
 OrderBook.prototype.to_json = function() {
   const json = {
-    taker_gets: {
+    gets: {
```
clark800: Isn't it ambiguous if you don't specify that it is the "taker" as opposed to the "maker"?

src/core/remote.js
```diff
@@ -1121,6 +1121,7 @@ Remote.prototype.requestTransactionEntry = function(options, callback) {
  * @return {Request} request
  */
 
+Remote.prototype.requestTx =
```
sublimator: Should update the jsdoc comments too, so they don't show positional arguments. I forgot to update them last time, and opened a subsequent WIP p.r. fixing them up.

src/core/orderbook.js
sublimator: While the ambiguity is unlikely to be a problem the potential is there, and it's also true the RPC arguments as recognized by rippled follow a taker_* pattern, so when you add it all up it seems like we should go with taker_* ?

sublimator:
LGTM

PR #528, opened 2015-08-26
remove redundant abs from MIN_IOU_VALUE

mDuo13:
First, let it be known I AM NOT SURE THIS WORKS. I am raising this PR in part to ask a question.

I do not see why we the code uses `-1000000000000000e-96` as consts.min_value, only to go and .abs() it a few lines later. Is it not valid to instantiate `IOUValue('1000000000000000e-96')`?

### DIFF
--- a/src/core/amount.js
+++ b/src/core/amount.js
@@ -49,13 +49,13 @@ const consts = {
   // Maximum possible amount for non-XRP currencies using the maximum mantissa
   // with maximum exponent. Corresponds to hex 0xEC6386F26FC0FFFF.
   max_value: '9999999999999999e80',
-  // Minimum possible amount for non-XRP currencies.
-  min_value: '-1000000000000000e-96'
+  // Minimum nonzero absolute value for non-XRP currencies.
+  min_value: '1000000000000000e-96'
 };
 
 const MAX_XRP_VALUE = new XRPValue(1e11);
 const MAX_IOU_VALUE = new IOUValue(consts.max_value);
-const MIN_IOU_VALUE = new IOUValue(consts.min_value).abs();
+const MIN_IOU_VALUE = new IOUValue(consts.min_value);
 
 const bi_xns_unit = new IOUValue(1e6);
 
--- a/test/amount-test.js
+++ b/test/amount-test.js
@@ -1225,7 +1225,7 @@ describe('Amount', function() {
 
     it('from_json minimum IOU', function() {
       const amt = Amount.from_json('-1e-81/USD/rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh');
-      assert.strictEqual(amt.to_text(), '-1000000000000000e-96');
+      assert.strictEqual(amt.to_text(), '1000000000000000e-96');
       assert.strictEqual(amt.to_text(), Amount.min_value);
     });
 
--- a/test/amount-test.js
+++ b/test/amount-test.js
@@ -1224,15 +1224,21 @@ describe('Amount', function() {
     });
 
     it('from_json minimum IOU', function() {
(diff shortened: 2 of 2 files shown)


test/amount-test.js
```diff
@@ -1225,7 +1225,7 @@ describe('Amount', function() {
 
     it('from_json minimum IOU', function() {
       const amt = Amount.from_json('-1e-81/USD/rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh');
-      assert.strictEqual(amt.to_text(), '-1000000000000000e-96');
```
sublimator: - is right there yeah?

test/amount-test.js
```diff
@@ -1225,7 +1225,7 @@ describe('Amount', function() {
...
+      assert.strictEqual(amt.to_text(), '1000000000000000e-96');
       assert.strictEqual(amt.to_text(), Amount.min_value);
```
sublimator: This wants a '-' tacked onto the front.  Amount.min_abs_value might be clearer than Amount.min_value?

src/core/amount.js
```diff
@@ -49,13 +49,13 @@ const consts = {
   // Maximum possible amount for non-XRP currencies using the maximum mantissa
   // with maximum exponent. Corresponds to hex 0xEC6386F26FC0FFFF.
   max_value: '9999999999999999e80',
-  // Minimum possible amount for non-XRP currencies.
-  min_value: '-1000000000000000e-96'
+  // Minimum nonzero absolute value for non-XRP currencies.
```
sublimator: nonzero!

test/amount-test.js
```diff
@@ -1225,7 +1225,7 @@ describe('Amount', function() {
 
     it('from_json minimum IOU', function() {
       const amt = Amount.from_json('-1e-81/USD/rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh');
```
sublimator: Or you could just remove the `-` here in the `-1e81 .... `

clark800: Better to keep the `-` case, or keep both

sublimator: makes sense

test/amount-test.js
```diff
@@ -1224,15 +1224,21 @@ describe('Amount', function() {
...
       assert.strictEqual(amt.to_text(), '-1000000000000000e-96');
-      assert.strictEqual(amt.to_text(), Amount.min_value);
+      assert.strictEqual(amt.negate().to_text(), Amount.min_value);
     });
 
     it('from_json exceed minimum IOU', function() {
       assert.throws(function() {
-        Amount.from_json('-1e-82/USD/rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh');
-      }, 'Exceeding min value of ' + Amount.min_value);
+        Amount.from_json('1e-82/USD/rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh');
```
clark800: Why is the negative sign removed?

mDuo13: Because the first test is actually measuring minimum nonzero absolute value, not minimum value. Actual minimum value (as in, "most negative") would be something like -1000000000000000e80 or -999999999999999e80.

clark800: I agree with what you said, but I don't understand how it relates to the removal of the negative sign.

mDuo13: My latest update should make it more clear, by renaming the tests and adding all the possibilities:

- meeting, exceeding maximum absolute value
- meeting, exceeding negative maximum absolute value
- meeting, exceeding minimum absolute value
- meeting, exceeding negative minimum absolute value

Let me know if it still doesn't make sense.

clark800:
LGTM

PR #529, opened 2015-08-26
[WIP] Only emit `model` event after ledger closed

sublimator:


src/core/orderbook.js
```diff
@@ -123,6 +125,13 @@ function OrderBook(remote,
     self.updateFundedAmounts(transaction);
   }
 
+  function ledgerClosed(message) {
+    if (message.ledger_index === self._ledgerLastNotified) {
+      // We had transaction notifications, so notify and recompute
```
sublimator: We might do something similar with all the `offer_added` events, renaming them to `_offer_added` and buffering these events until ledger close.

@shekenahglory  

How are the offer_added etc events used in ripple-charts (if at all)?

sublimator:
This would only really help if you had lots of transactions for the same book in one ledger.

src/core/orderbook.js
```diff
@@ -38,9 +38,10 @@ function assertValidNumber(number, message) {
  * @param {String} orderbook key
  */
 
-function OrderBook(remote,
-                   currencyGets, issuerGets, currencyPays, issuerPays,
-                   key) {
+/* eslint-disable max-len*/
+// Otherwise it gets confused by the indentation!
```
clark800: One option that I've been using is to put the `) {` on a line of its own.

clark800:
@sublimator can you add a comment explaining why this was closed?

sublimator:
It was a quick kludge, and I believe Ivan and Mathew are working on fixes
for the perf issue, in the context of actual apps they are working on.

The idea was basically to buffer txn notifications, until the ledger
closed.  Note all the `"type" : "transaction"` messages are for validated
messages, and arrive in quick succession, followed the ledgerClose.

I think Matt ended up going with a more elegant simple throttle. I haven't
really been following along.

clark800:
I think the buffering solution seems more elegant than the throttling solution. I think Ivan may be working on changing to a buffering solution now. I linked him to this pull request.

sublimator:
Whatever works

sublimator:
If we are spending a lot of time on this, we should also update the book requests to request a validated ledger.

PR #534, opened 2015-08-28
Return instructions in prepare responses

clark800:


src/api/transaction/utils.js
```diff
@@ -23,8 +24,22 @@ function getFeeDrops(remote) {
...
+    maxLedgerVersion: txJSON.LastLedgerSequence
+  };
+  return {
+    txJSON: JSON.stringify(txJSON),
+    instructions: _.omit(instructions, _.isUndefined)
+  };
+}
+
+function prepareTransaction(transaction: any, remote: any, instructions: any,
+    callback: (err: ?(typeof Error),
```
lumberj: Might be helpful to use a `type` alias for callbacks to save some lines

clark800: done

lumberj:
LGTM :+1:

PR #535, opened 2015-08-28
Return promise error if submit result is an immediate failure

clark800:


src/api/transaction/submit.js
```diff
@@ -5,6 +5,16 @@ const validate = utils.common.validate;
 const Request = utils.common.core.Request;
 const convertErrors = utils.common.convertErrors;
 
+function convertSubmitErrors(callback) {
+  return function(error, data) {
+    if (data.engineResult !== 'tesSUCCESS') {
```
lumberj: This isn't right. See here: https://ripple.com/build/transactions/#result-categories

The immediate response only indicates an immediate failure if the result code is `tej` or `tem`. Anything else means the transaction _may_ appear in a ledger.

clark800: What's an example of a code that is not tesSUCCESS that could be a successfully submitted transaction?

lumberj: Anything other than `tej` or `tem`

clark800:
@lumberj updated

lumberj:
LGTM

PR #536, opened 2015-08-31
suspended payments

sentientwaffle:
New API methods:
* `prepareSuspendedPaymentCreate`
* `prepareSuspendedPaymentFinish`
* `prepareSuspendedPaymentCancellation`


wltsmrz:
Might want to add tests for ripple-lib "core." Otherwise LGTM, except for the commit messages of course

src/api/common/schemas/suspended-payment-finish.json
```diff
@@ -0,0 +1,20 @@
...
+  "properties": {
+    "memos": {
+      "type": "array",
+      "items": {
+        "$ref": "memo"
+      }
+    },
+    "owner": {"$ref": "address"},
+    "offerSequence": {"$ref": "uint32"},
+    "method": {"type": "integer", "minimum": 0, "maximum": 256},
```
sublimator: 0 -- 255 are valid values for a uint8, so maximum should be 255 I think

sentientwaffle: Thanks, good catch!

src/api/ledger/parse/suspended-payment-cancel.js
```diff
@@ -0,0 +1,29 @@
...
+    return utils.removeUndefined({
+      type: m.Memo.parsed_memo_type,
+      format: m.Memo.parsed_memo_format,
+      data: m.Memo.parsed_memo_data
+    });
+  });
+}
+
+function parsePayment(tx: Object): Object {
+  assert(tx.TransactionType === 'SuspendedPaymentCancel');
```
sublimator:
```
SuspendedPaymentCancel 
SusPayCancel
```

I recall during discussions of this that @justmoon requested the non abbreviated form was present in the txn json (as seen here with the TransactionType). I think in the general sense, he wanted any public apis to use SuspendedPayment rather than SusPay.  

It probably doesn't really matter, but I went out of my way to [patch](https://github.com/ripple/rippled/pull/1224) rippled in accordance with Stefan's wishes, so I figured I'd bring it up again.

src/api/index.js
```diff
@@ -62,6 +65,9 @@ RippleAPI.prototype = {
   prepareTrustline,
   prepareOrder,
   prepareOrderCancellation,
```
sublimator: prepareOrder**Cancellation** vs prepareSusPay**Cancel**

src/api/ledger/parse/suspended-payment-cancel.js
```diff
@@ -0,0 +1,29 @@
+/* @flow */
+'use strict';
+const assert = require('assert');
+const utils = require('./utils');
+
+function parsePaymentMemos(tx) {
```
sublimator: We could probably do with this guy in some utils module somewhere.

src/api/transaction/suspay-finish.js
```diff
@@ -0,0 +1,48 @@
...
+
+function createSusPayFinishTransaction(account, payment) {
+  validate.address(account);
+  validate.suspendedPaymentFinish(payment);
+
+  const transaction = new Transaction();
+  transaction.susPayFinish({
+    from: account,
+    owner: payment.owner,
+    offerSequence: payment.offerSequence
```
sublimator: offerSequence?

src/api/transaction/suspay-finish.js
```diff
@@ -0,0 +1,48 @@
...
+const validate = utils.common.validate;
+const Transaction = utils.common.core.Transaction;
+
+function createSusPayFinishTransaction(account, payment) {
+  validate.address(account);
+  validate.suspendedPaymentFinish(payment);
+
+  const transaction = new Transaction();
+  transaction.susPayFinish({
+    from: account,
```
wltsmrz: I think adding `from` might've been a mistake. We'll probably never get rid of it, but I'd use `account` here instead

src/api/transaction/suspay-finish.js
```diff
@@ -0,0 +1,48 @@
...
+}
+
+function prepareSusPayFinishAsync(account, payment, instructions, callback) {
+  const transaction = createSusPayFinishTransaction(account, payment);
+  utils.createTxJSON(transaction, this.remote, instructions, callback);
+}
+
+function prepareSusPayFinish(account: string, payment: Object,
+instructions = {}) {
+  return utils.promisify(prepareSusPayFinishAsync.bind(this))(
```
wltsmrz: `utils.promisify(fn).call(this, ...)` will be arguably more readable and certainly better-performing

src/api/transaction/suspay-finish.js
sublimator: @vinniefalco  

I know you gonna tell me to RTFM but where's the `offer` part come in?

src/api/transaction/suspay-cancel.js
```diff
@@ -0,0 +1,38 @@
+/* @flow */
+'use strict';
+const _ = require('lodash');
+const utils = require('./utils');
+const validate = utils.common.validate;
+const Transaction = utils.common.core.Transaction;
+
+function createSusPayCancelTransaction(account, payment) {
+  validate.address(account);
+  validate.suspendedPaymentCancel(payment);
```
sublimator: And here's the full monty !

src/api/transaction/suspay-finish.js
wltsmrz: Oh wait, can we just do `utils.promisify(fn).apply(this, arguments)` in most places?

src/api/transaction/suspay-finish.js
vinniefalco: I reused the existing `SField` instead of creating a new one. If there are sound objections, I'm open to changing it.

sublimator: I don't think we should be letting the implementation details bubble up such that we have a transaction field with a misleading name. I think we should just add the new generic field if these SusPay critters have nothing to do with offers. 

Ashame we have bollix fields like this:
```
    WalletSize(12, Type.UInt32),
```

Probably TargetSeq should be TargetSequence for consistency with OfferSequence, LastLedgerSequence etc?

src/api/ledger/parse/suspended-payment-finish.js
```diff
@@ -0,0 +1,25 @@
...
+  return sjclcodec.utf8String.fromBits(bits);
+}
+
+function parsePayment(tx: Object): Object {
+  assert(tx.TransactionType === 'SuspendedPaymentFinish');
+
+  return utils.removeUndefined({
+    memos: utils.parsePaymentMemos(tx),
+    owner: tx.Owner,
+    offerSequence: tx.OfferSequence,
```
sublimator: Given RippleAPI washes over the oddities in rippled already, maybe we could use better name for `offerSequence` ?

src/api/index.js
```diff
@@ -62,6 +65,9 @@ RippleAPI.prototype = {
...
+  prepareSuspendedPaymentCreate: prepareSusPayCreate,
+  prepareSuspendedPaymentFinish: prepareSusPayFinish,
+  prepareSuspendedPaymentCancellation: prepareSusPayCancel,
```
sublimator: @clark800 will almost certainly want you to `const prepareSuspendedPaymentCreate = require('./transaction/suspay-create');` to keep his pretty export list.

src/api/ledger/parse/suspended-payment-finish.js
sentientwaffle: I think that using different names for the same field would be more confusing, not less -- though I agree that the name `OfferSequence` is misleading.

sublimator: Maybe, but note that one of the goals of this new API is that users aren't
supposed to be required to understand the rippled json. A lot of things
have been renamed already. Consider at some point documentation will be
written for this feature and @mduo13 or someone will have to spin
`offerSequence`. Ideally we'd change the name in rippled eh?

If you could name it anything, what would you name it?

sentientwaffle: Well, it refers to the sequence number of the corresponding `SuspendedPaymentCreate` transaction, so `paymentSequence`.

lumberj: >  note that one of the goals of this new API is that users aren't
supposed to be required to understand the rippled json

:+1: on renaming

src/api/index.js
lumberj: Maybe the filenames can reflect the naming too?

src/api/transaction/suspay-finish.js
vinniefalco: I disagree, fields are precious and we've been consuming them with disregard for their scarcity. @JoelKatz what do you think?

I don't have a preference for TargetSequence versus TargetSeq.

sublimator: Seems like swinging from frivolous to overly stingy. At 38/256 I think we're some ways from being forced to use **Offer**Sequence for Payments. We haven't painted ourselves that far into the corner yet.

Also, have these actually been used historically: 
```
SF_U32 const sfStampEscrow         = make::one<SF_U32::type>(&sfStampEscrow,         STI_UINT32, 22, "StampEscrow");
SF_U32 const sfBondAmount          = make::one<SF_U32::type>(&sfBondAmount,          STI_UINT32, 23, "BondAmount");
SF_U32 const sfOperationLimit      = make::one<SF_U32::type>(&sfOperationLimit,      STI_UINT32, 29, "OperationLimit");
```

?

At one point @scottschurr and I asked the data team if they could search for TicketID uses. Maybe we could recycle an ordinal.

src/api/index.js
```diff
@@ -62,6 +66,9 @@ RippleAPI.prototype = {
   prepareTrustline,
   prepareOrder,
   prepareOrderCancellation,
+  prepareSuspendedPaymentCreate,
```
wltsmrz: Sadly, this might need to look like `prepareSuspendedPaymentCreation` for consistency

src/api/ledger/parse/suspended-payment-finish.js
JoelKatz: Given that this is already an uncommon field, adding another uncommon field is not a big deal. The problem is that "OfferSequence" should have been named something else from the start -- maybe "OtherSequence" or something. So if we add a new field, it will wind up being "PaymentSequence" and we'll multiply the bad choice. But I think that's about the best we can do.

src/api/index.js
sublimator: Yeah :+1: to what @lumberj said

src/api/ledger/parse/suspended-payment-finish.js
sublimator: @JoelKatz  @vinniefalco 

What's the deployment ETA for this feature? Do we have time to add the new field?

yana-novikova: We should have rippled 0.29.1 release this month but suspended payments feature will not be enabled. We will enable it in rippled release 0.3 which will probably happen next month.

clark800: > I think that using different names for the same field would be more confusing, not less -- though I agree that the name OfferSequence is misleading.

This API is supposed to provide a level of abstraction over rippled, so there's no reason to preserve rippled naming; all names have been chosen to be the most clear name we could think of.

clark800:
I see that you renamed "SusPay" to "SuspendedPayment", but can you remove all occurrences of "SusPay" everywhere?

src/api/ledger/parse/suspended-payment-cancel.js
```diff
@@ -0,0 +1,16 @@
+/* @flow */
+'use strict';
+const assert = require('assert');
+const utils = require('./utils');
+
+function parsePayment(tx: Object): Object {
```
clark800: `parseSuspendedPaymentCancellation`

src/api/ledger/parse/suspended-payment-create.js
```diff
@@ -0,0 +1,39 @@
...
+const assert = require('assert');
+const utils = require('./utils');
+const parseAmount = require('./amount');
+
+function removeGenericCounterparty(amount, address) {
+  return amount.counterparty === address ?
+    _.omit(amount, 'counterparty') : amount;
+}
+
+function parsePayment(tx: Object): Object {
```
clark800: `parseSuspendedPayment`

src/api/ledger/parse/suspended-payment-finish.js
```diff
@@ -0,0 +1,25 @@
...
+const assert = require('assert');
+const sjclcodec = require('sjcl-codec');
+const utils = require('./utils');
+
+function convertHexToString(hexString) {
+  const bits = sjclcodec.hex.toBits(hexString);
+  return sjclcodec.utf8String.fromBits(bits);
+}
+
+function parsePayment(tx: Object): Object {
```
clark800: `parseSuspendedPaymentFulfillment`?

src/api/common/schemas/suspended-payment.json
```diff
@@ -0,0 +1,28 @@
...
+        "$ref": "memo"
+      }
+    },
+    "digest": {"$ref": "hash256"},
+    "cancelAfter": {
+      "type": "integer",
+      "minimum": 0,
+      "description": "milliseconds since unix epoch"
+    },
+    "finishAfter": {
```
clark800: I would interpret this name to mean "if this time is reached, automatically finish the transaction", which doesn't seem to be the correct interpretation, so I think there might be a better name. Maybe `validAfter`?

clark800: Another idea is: finishAfter/cancelAfter -> activationDate/expirationDate

src/api/ledger/parse/suspended-payment-finish.js
```diff
@@ -0,0 +1,25 @@
...
+  assert(tx.TransactionType === 'SuspendedPaymentFinish');
+
+  return utils.removeUndefined({
+    memos: utils.parsePaymentMemos(tx),
+    owner: tx.Owner,
+    offerSequence: tx.OfferSequence,
+    method: tx.Method,
+    digest: tx.Digest,
+    proof: tx.Proof && convertHexToString(tx.Proof)
```
clark800: I think the ternary operator would be better.

src/api/ledger/parse/utils.js
```diff
@@ -67,8 +67,22 @@ function parseOutcome(tx: Object): ?Object {
   };
 }
 
+function parsePaymentMemos(tx: Object): ?Array<Object> {
```
clark800: We could rename to `parseMemos` since all transaction types support memos.

src/api/common/schemas/suspended-payment.json
wltsmrz: That's exactly how I originally interpreted the rippled parameters. It wasn't until I started writing tests that I realized that it means transaction can only be in one of two modes (Cancel/Finish) at a time. `allowCancelAfter`/`allowFinishAfter` might be even more clear

src/api/transaction/suspended-payment-creation.js
```diff
@@ -0,0 +1,76 @@
...
+const validate = utils.common.validate;
+const toRippledAmount = utils.common.toRippledAmount;
+const Transaction = utils.common.core.Transaction;
+
+function isIOUWithoutCounterparty(amount) {
+  return amount && amount.currency !== 'XRP'
+    && amount.counterparty === undefined;
+}
+
+function applyAnyCounterpartyEncoding(payment) {
```
clark800: Seems like too much copy/paste going on

src/core/transaction.js
```diff
@@ -1501,6 +1501,157 @@ Transaction.prototype.summary = function() {
...
+ *
+ *  @param {String} source account
+ *  @param {String} destination account
+ *  @param {Amount} payment amount
+ */
+
+Transaction.prototype.suspendedPaymentCreate = function(options_) {
+  let options;
+
+  if (typeof options_ === 'object') {
```
clark800: I don't think we need to support multiple calling syntaxes in new code. I think Brandon has been moving to object-only convention, so we should require that.

src/core/transaction.js
```diff
@@ -1501,6 +1501,157 @@ Transaction.prototype.summary = function() {
...
+    options = _.merge({}, options_);
+
+    if (_.isUndefined(options.account)) {
+      options.account = options.src || options.from;
```
clark800: Why two different names?

sublimator: Seems like Xopy/paste from legacy Payment

clark800: Would be good to avoid mutation of `options`.

clark800: If it is copy/pasted, then I think we should get rid of it because we don't need legacy support in new code.

src/api/common/schemas/suspended-payment-cancel.json
```diff
@@ -0,0 +1,17 @@
...
+  "type": "object",
+  "properties": {
+    "memos": {
+      "type": "array",
+      "items": {
+        "$ref": "memo"
+      }
+    },
+    "owner": {"$ref": "address"},
+    "offerSequence": {"$ref": "uint32"}
```
clark800: Weren't we going to rename this?

src/api/index.js
```diff
@@ -61,6 +67,9 @@ RippleAPI.prototype = {
   prepareTrustline,
   prepareOrder,
   prepareOrderCancellation,
+  prepareSuspendedPaymentCreation,
+  prepareSuspendedPaymentFinish,
```
clark800: "Execution" seems more symmetrical with "Cancellation"

src/api/ledger/parse/suspended-payment-finish.js
```diff
@@ -0,0 +1,25 @@
...
+const assert = require('assert');
+const sjclcodec = require('sjcl-codec');
+const utils = require('./utils');
+
+function convertHexToString(hexString) {
+  const bits = sjclcodec.hex.toBits(hexString);
+  return sjclcodec.utf8String.fromBits(bits);
+}
+
+function parseSuspendedPaymentFulfillment(tx: Object): Object {
```
clark800: Need to be consistent: "SuspendedPaymentFinish" vs. "SuspendedPaymentFulfillment". I think "Fulfillment" or "Execution" are both better than "Finish".

src/api/ledger/parse/transaction.js
```diff
@@ -27,7 +33,10 @@ function parseTransaction(tx: Object): Object {
     'trustline': parseTrustline,
     'order': parseOrder,
     'orderCancellation': parseOrderCancellation,
-    'settings': parseSettings
+    'settings': parseSettings,
+    'suspendedPaymentCreate': parseSuspendedPaymentCreate,
```
clark800: Consistency

src/api/transaction/suspended-payment-cancellation.js
```diff
@@ -0,0 +1,39 @@
+/* @flow */
+'use strict';
+const _ = require('lodash');
+const utils = require('./utils');
+const validate = utils.common.validate;
+const Transaction = utils.common.core.Transaction;
+
+function createSuspendedPaymentCancelTransaction(account, payment) {
```
clark800: Consistency

clark800:
I didn't comment on all the naming inconsistencies, but we need to fix all of them. The names in "core" should mimic the names in rippled, but the names in API should be the "long form" names.

sentientwaffle:
@clark800 ok, fixed.

src/api/common/schemas/suspended-payment-creation.json
```diff
@@ -0,0 +1,28 @@
...
+        "$ref": "memo"
+      }
+    },
+    "digest": {"$ref": "hash256"},
+    "allowCancelAfter": {
+      "type": "integer",
+      "minimum": 0,
+      "description": "milliseconds since unix epoch"
+    },
+    "allowFinishAfter": {
```
clark800: allowExecuteAfter

sentientwaffle:
@clark800 ok, `allowFinishAfter` has been replaced with `allowExecuteAfter`

clark800:
LGTM, can you squash (one for lint, one for the rest)?

sentientwaffle:
@clark800 ok, squashed.

PR #537, opened 2015-08-31
Optimize performance (for 0.12 master branch)

darkdarkdragon:
Optimize performance of OrderBook

### DIFF
--- a/src/amount.js
+++ b/src/amount.js
@@ -388,7 +388,7 @@ Amount.prototype._copy = function(value) {
 Amount.prototype.compareTo = function(to) {
   var toAmount = Amount.from_json(to);
   if (!this.is_comparable(toAmount)) {
-    return new Amount(NaN);
+    return 0;
   }
   return this._value.comparedTo(toAmount._value);
 };
--- a/src/autobridgecalculator.js
+++ b/src/autobridgecalculator.js
@@ -20,7 +20,8 @@ function assertValidLegOneOffer(legOneOffer, message) {
 function AutobridgeCalculator(currencyGets, currencyPays,
   legOneOffers, legTwoOffers, issuerGets, issuerPays) {
   this._currencyGets = currencyGets;
-  this._currencyPays = currencyPays;
+  this._currencyGetsHex = currencyGets.to_hex();
+  this._currencyPaysHex = currencyPays.to_hex();
   this._issuerGets = issuerGets;
   this._issuerPays = issuerPays;
   this.legOneOffers = _.cloneDeep(legOneOffers);
@@ -28,6 +29,8 @@ function AutobridgeCalculator(currencyGets, currencyPays,
 
   this._ownerFundsLeftover = {};
 }
+
+AutobridgeCalculator.NULL_AMOUNT = Utils.normalizeAmount('0');
 
 /**
  * Calculates an ordered array of autobridged offers by quality
@@ -39,7 +42,7 @@ AutobridgeCalculator.prototype.calculate = function() {
   let legOnePointer = 0;
   let legTwoPointer = 0;
 
-  let offersAutobridged = [];
+  const offersAutobridged = [];
 
   this.clearOwnerFundsLeftover();
 
@@ -70,14 +73,17 @@ AutobridgeCalculator.prototype.calculate = function() {
       continue;
     }
 
-    if (legOneTakerGetsFunded.compareTo(legTwoTakerPaysFunded) > 0) {
+    //  using private fields for speed
+    if (legOneTakerGetsFunded._value.comparedTo(
+      legTwoTakerPaysFunded._value) > 0) {
       autobridgedOffer = this.getAutobridgedOfferWithClampedLegOne(
         legOneOffer,
         legTwoOffer
       );
 
       legTwoPointer++;
-    } else if (legTwoTakerPaysFunded.compareTo(legOneTakerGetsFunded) > 0) {
+    } else if (legTwoTakerPaysFunded._value.comparedTo(
+      legOneTakerGetsFunded._value) > 0) {
       autobridgedOffer = this.getAutobridgedOfferWithClampedLegTwo(
         legOneOffer,
         legTwoOffer
@@ -210,8 +216,6 @@ AutobridgeCalculator.prototype.clearOwnerFundsLeftover = function() {
  */
 
 AutobridgeCalculator.prototype.resetOwnerFundsLeftover = function(account) {
-  assert(UInt160.is_valid(account), 'Account is invalid');
-
   this._ownerFundsLeftover[account] = Utils.normalizeAmount('0');
 
   return this._ownerFundsLeftover[account];
@@ -226,12 +230,10 @@ AutobridgeCalculator.prototype.resetOwnerFundsLeftover = function(account) {
  */
 
 AutobridgeCalculator.prototype.getLeftoverOwnerFunds = function(account) {
-  assert(UInt160.is_valid(account), 'Account is invalid');
-
   let amount = this._ownerFundsLeftover[account];
 
   if (!amount) {
-    amount = Utils.normalizeAmount('0');
+    amount = AutobridgeCalculator.NULL_AMOUNT.clone();
   }
 
   return amount;
@@ -248,7 +250,6 @@ AutobridgeCalculator.prototype.getLeftoverOwnerFunds = function(account) {
 
 AutobridgeCalculator.prototype.addLeftoverOwnerFunds =
 function(account, amount) {
-  assert(UInt160.is_valid(account), 'Account is invalid');
   assert(amount instanceof Amount, 'Amount is invalid');
 
   this._ownerFundsLeftover[account] = this.getLeftoverOwnerFunds(account)
@@ -266,7 +267,6 @@ function(account, amount) {
 
 AutobridgeCalculator.prototype.setLeftoverOwnerFunds =
 function(account, amount) {
-  assert(UInt160.is_valid(account), 'Account is invalid');
   assert(amount instanceof Amount, 'Amount is invalid');
 
   this._ownerFundsLeftover[account] = amount;
@@ -291,13 +291,13 @@ function(takerGets, takerPays) {
 
   autobridgedOffer.TakerGets = {
     value: takerGets.to_text(),
-    currency: this._currencyGets.to_hex(),
+    currency: this._currencyGetsHex,
     issuer: this._issuerGets
   };
 
   autobridgedOffer.TakerPays = {
     value: takerPays.to_text(),
-    currency: this._currencyPays.to_hex(),
+    currency: this._currencyPaysHex,
     issuer: this._issuerPays
   };
 
@@ -308,7 +308,8 @@ function(takerGets, takerPays) {
 
   autobridgedOffer.autobridged = true;
 
-  autobridgedOffer.BookDirectory = Utils.convertOfferQualityToHex(quality);
+  autobridgedOffer.BookDirectory = Utils.convertOfferQualityToHexFromText(autobridgedOffer.quality);
+  autobridgedOffer.qualityHex = autobridgedOffer.BookDirectory;
 
   return autobridgedOffer;
 };
--- a/src/baseconverter.js
+++ b/src/baseconverter.js
@@ -1,28 +1,39 @@
 'use strict';
 
+
 function normalize(digitArray) {
-  while (digitArray[0] === 0) {
-    digitArray.shift();
+  let i = 0;
+  while (digitArray[i] === 0) {
+    ++i;
+  }
+  if (i > 0) {
+    digitArray.splice(0, i);
   }
   return digitArray;
 }
 
 function divmod(digitArray, base, divisor) {
-  var remainder = 0;
-  var quotient = [];
-  for (var j = 0; j < digitArray.length; j++) {
-    var temp = remainder * base + parseInt(digitArray[j], 10);
-    quotient.push(Math.floor(temp / divisor));
+  let remainder = 0;
+  let temp, divided;
+  let j = -1;
+
+  const length = digitArray.length;
+  const quotient = new Array(length);
+
+  while (++j < length) {
+    temp = remainder * base + digitArray[j];
+    divided = temp / divisor;
+    quotient[j] = divided << 0;
     remainder = temp % divisor;
   }
   return {quotient: normalize(quotient), remainder: remainder};
 }
 
 function convertBase(digitArray, fromBase, toBase) {
-  var result = [];
-  var dividend = digitArray;
+  const result = [];
+  let dividend = digitArray, qr;
   while (dividend.length > 0) {
-    var qr = divmod(dividend, fromBase, toBase);
+    qr = divmod(dividend, fromBase, toBase);
     result.unshift(qr.remainder);
     dividend = qr.quotient;
   }
--- a/src/orderbook.js
+++ b/src/orderbook.js
@@ -23,10 +23,6 @@ const AutobridgeCalculator = require('./autobridgecalculator');
 const OrderBookUtils = require('./orderbookutils');
 const log = require('./log').internal.sub('orderbook');
 
-function assertValidNumber(number, message) {
-  assert(!_.isNull(number) && !isNaN(number), message);
-}
-
 /**
  * @constructor OrderBook
  * @param {Remote} remote
@@ -35,11 +31,12 @@ function assertValidNumber(number, message) {
  * @param {String} bid currency
  * @param {String} bid issuer
  * @param {String} orderbook key
+ * @param {Boolean} fire 'model' event after receiving transaction
+                    only once in 10 seconds
  */
 
 function OrderBook(remote,
-                   currencyGets, issuerGets, currencyPays, issuerPays,
-                   key) {
+                   currencyGets, issuerGets, currencyPays, issuerPays, key) {
   EventEmitter.call(this);
 
   const self = this;
@@ -60,6 +57,8 @@ function OrderBook(remote,
   this._ownerFundsUnadjusted = {};
   this._ownerFunds = {};
   this._ownerOffersTotal = {};
+  this._validAccounts = {};
+  this._validAccountsCount = 0;
 
   // We consider ourselves synchronized if we have a current
   // copy of the offers, we are online and subscribed to updates
@@ -72,13 +71,44 @@ function OrderBook(remote,
   // books that we must keep track of to compute autobridged offers
   this._legOneBook = null;
   this._legTwoBook = null;
+  this._gotOffersFromLegOne = false;
+  this._gotOffersFromLegTwo = false;
+
+  this.sortOffers = this._currencyGets.has_interest() ?
+    _sortOffers.bind(this) : _sortOffersQuick;
+
+  this.notifyDirectOffersChanged =
+    _.debounce(
+      this.notifyDirectOffersChangedInternal,
+      OrderBook.NOTIFY_TIMEOUT,
+      {maxWait: OrderBook.NOTIFY_MAXWAIT});
 
   this._isAutobridgeable = !this._currencyGets.is_native()
     && !this._currencyPays.is_native();
 
-  function computeAutobridgedOffersWrapper() {
-    self.computeAutobridgedOffers();
-    self.mergeDirectAndAutobridgedBooks();
+  this._autobridgeThrottleTimeMultiplier = 1;
+  this.createDebouncedOffersWrapper();
+
+  function computeAutobridgedOffersWrapperOne() {
+    self._gotOffersFromLegOne = true;
+    self.computeAutobridgedOffersThrottled();
+  }
+
+  function computeAutobridgedOffersWrapperTwo() {
+    self._gotOffersFromLegTwo = true;
+    self.computeAutobridgedOffersThrottled();
+  }
+
+  function onDisconnect() {
+    self.resetCache();
+    self._gotOffersFromLegOne = false;
+    self._gotOffersFromLegTwo = false;
+    if (!self._destroyed) {
+      self._remote.once('disconnect', onDisconnect);
+      self._remote.once('connect', function() {
+        self.subscribe();
+      });
+    }
   }
 
   if (this._isAutobridgeable) {
@@ -88,15 +118,15 @@ function OrderBook(remote,
       issuer_pays: issuerPays
     });
 
-    this._legOneBook.on('model', computeAutobridgedOffersWrapper);
-
     this._legTwoBook = remote.createOrderBook({
       currency_gets: currencyGets,
       issuer_gets: issuerGets,
       currency_pays: 'XRP'
     });
+  }
 
-    this._legTwoBook.on('model', computeAutobridgedOffersWrapper);
+  function updateFundedAmountsWrapper(transaction) {
+    self.updateFundedAmounts(transaction);
   }
 
   function listenersModified(action, event) {
@@ -106,7 +136,16 @@ function OrderBook(remote,
       switch (action) {
         case 'add':
           if (++self._listeners === 1) {
+            self._shouldSubscribe = true;
             self.subscribe();
+
+            self._remote.on('transaction', updateFundedAmountsWrapper);
+            self._remote.once('disconnect', onDisconnect);
+
+            if (self._isAutobridgeable) {
+              self._legOneBook.on('model', computeAutobridgedOffersWrapperOne);
+              self._legTwoBook.on('model', computeAutobridgedOffersWrapperTwo);
+            }
           }
           break;
         case 'remove':
@@ -118,10 +157,6 @@ function OrderBook(remote,
     }
   }
 
-  function updateFundedAmountsWrapper(transaction) {
-    self.updateFundedAmounts(transaction);
-  }
-
   this.on('newListener', function(event) {
     listenersModified('add', event);
   });
@@ -130,23 +165,21 @@ function OrderBook(remote,
     listenersModified('remove', event);
   });
 
-  this._remote.on('transaction', updateFundedAmountsWrapper);
-
   this.on('unsubscribe', function() {
     self.resetCache();
 
     self._remote.removeListener('transaction', updateFundedAmountsWrapper);
-  });
+    self._remote.removeListener('disconnect', onDisconnect);
 
-  this._remote.once('prepare_subscribe', function() {
-    self.subscribe();
-  });
+    self._gotOffersFromLegOne = false;
+    self._gotOffersFromLegTwo = false;
 
-  this._remote.on('disconnect', function() {
-    self.resetCache();
-    self._remote.once('prepare_subscribe', function() {
-      self.subscribe();
-    });
+    if (self._isAutobridgeable) {
+      self._legOneBook.removeListener('model',
+        computeAutobridgedOffersWrapperOne);
+      self._legTwoBook.removeListener('model',
+        computeAutobridgedOffersWrapperTwo);
+    }
   });
 
   return this;
@@ -164,7 +197,21 @@ OrderBook.EVENTS = [
   'offer_changed', 'offer_funds_changed'
 ];
 
-OrderBook.DEFAULT_TRANSFER_RATE = 1000000000;
+OrderBook.DEFAULT_TRANSFER_RATE = Amount.from_json(1000000000);
+
+OrderBook.NOTIFY_TIMEOUT = 100;
+
+OrderBook.NOTIFY_MAXWAIT = 250;
+
+OrderBook.AUTOBRIDGE_CALCULATE_THROTTLE_TIME = 1000;
+
+OrderBook.AUTOBRIDGE_CALCULATE_DEBOUNCE_TIME = 250;
+
+OrderBook.AUTOBRIDGE_CALCULATE_DEBOUNCE_MAXWAIT = 500;
+
+OrderBook.ZERO_NATIVE_AMOUNT = Amount.from_json('0');
+
+OrderBook.ZERO_NORMALIZED_AMOUNT = OrderBookUtils.normalizeAmount('0');
 
 /**
  * Normalize offers from book_offers and transaction stream
@@ -191,18 +238,20 @@ OrderBook.offerRewrite = function(offer) {
   result.Flags = result.Flags || 0;
   result.OwnerNode = result.OwnerNode || new Array(16 + 1).join('0');
   result.BookNode = result.BookNode || new Array(16 + 1).join('0');
+  result.qualityHex = result.BookDirectory.slice(-16);
 
   return result;
 };
 
 /**
  * Initialize orderbook. Get orderbook offers and subscribe to transactions
+ * @api private
  */
 
 OrderBook.prototype.subscribe = function() {
   const self = this;
 
-  if (!this._shouldSubscribe) {
+  if (!this._shouldSubscribe || this._destroyed) {
     return;
   }
 
@@ -215,7 +264,7 @@ OrderBook.prototype.subscribe = function() {
       self.requestTransferRate(callback);
     },
     function(callback) {
-      self.requestOffers(callback);
+      self.requestOffers(callback, true);
     },
     function(callback) {
       self.subscribeTransactions(callback);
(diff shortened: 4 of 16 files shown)


src/baseconverter.js
```diff
@@ -1,18 +1,30 @@
 'use strict';
 
+
 function normalize(digitArray) {
-  while (digitArray[0] === 0) {
-    digitArray.shift();
+  const length = digitArray.length;
```
sublimator: I don't think this is actually used any more by anything, as we now (indirectly) use this: https://github.com/cryptocoinjs/base-x

darkdarkdragon: in 0.13 or in 0.12?
because it is used in 0.12, and RippleTrade is using 0.12, and divmod was slowest part of the code

sublimator: Ha! Interesting. I noticed the base58 in the profile too. Yes, in 0.12 I
think this is used yes. Are we going to patch the old version?

darkdarkdragon: 0.13 has completely changed api, so I can't imagine that RippleTrade will use it.
and most of my changes here went to OrderBook and AutobridgeCalculator, which, I believe, doesn't changed much in 0.13, so it will be easy port my changes to 0.13

darkdarkdragon:
@sublimator can you review and merge it? I want it to make into new RippleTrade release

sublimator:
@shekenahglory said he was going to take a look over it soon. He's been hand's on with this code quite recently and understands the issues better than me.

darkdarkdragon:
@sublimator ok, thanks

wltsmrz:
Thanks for looking into this. Are there any client-side performance results to share?

src/orderbook.js
```diff
@@ -72,15 +71,59 @@ function OrderBook(remote,
...
+  function computeAutobridgedOffersWrapper() {
+    clearTimeout(self._autobridgeComputeTimeout);
+    self._autobridgeComputeTimeout = null;
+    const passed = Date.now() - self._lastAutobridgeComputeTime;
+    if (passed >= OrderBook.UPDATES_PERIOD) {
+      computeAutobridgedOffersWrapperInternal();
+    } else {
+      self._autobridgeComputeTimeout =
+        setTimeout(_.partial(computeAutobridgedOffersWrapperInternal, true),
+          OrderBook.UPDATES_PERIOD - passed);
```
shekenahglory: Am I understanding this correctly, the autobridged books will be processed at an interval of not less than 15 seconds? (as currently configured)

src/orderbook.js
```diff
@@ -1185,6 +1309,14 @@ OrderBook.prototype.computeAutobridgedOffers = function() {
   assert(!this._currencyGets.is_native() && !this._currencyPays.is_native(),
     'Autobridging is only for IOU:IOU orderbooks');
 
+  if (this._destroyed) {
+    return;
+  }
+
+  if (!this._gotOffersFromLegOne || !this._gotOffersFromLegTwo) {
```
shekenahglory: I looks like this function is only ever called when one of the leg books is updated, so always one or the other of these would be true

src/orderbook.js
```diff
@@ -1204,20 +1343,24 @@ OrderBook.prototype.computeAutobridgedOffers = function() {
...
 
   this._mergedOffers = this._offers
     .concat(this._offersAutobridged)
-    .sort(function(a, b) {
-      const aQuality = OrderBookUtils.getOfferQuality(a, self._currencyGets);
-      const bQuality = OrderBookUtils.getOfferQuality(b, self._currencyGets);
-
-      return aQuality.compareTo(bQuality);
-    });
+    .sort(_sortOffers.bind(this));
```
shekenahglory: Does this offer a performance benefit?  Its still recreating 2 amount objects with each function call, which looked to be the main hinderance to performance

shekenahglory:
the `_autobridgeComputeTimeout` seems unnecessarily complicated - debouncing `notifyDirectOffersChanged` is much less invasive, and gives performance benefits for both autobridged and non-autobridged books, and only negligibly slows the frequency of updates:

```
OrderBook.prototype.notifyDirectOffersChanged = function() {
  const self = this;

  // debounce
  clearTimeout(this._notifyTimeout);
  this._notifyTimeout = setTimeout(function() {
    console.log('notify', self._key);

    if (self._isAutobridgeable) {
      self.mergeDirectAndAutobridgedBooks();
    } else {
      self.emit('model', self._offers);
    }
  }, 100);
};

```

src/orderbook.js
shekenahglory: from earlier processing, each order has a `quality` field as a numerical string - can we or should we just use that for the sort rather than creating Amount objects? 

```
function (a, b) {
  return Number(a.quality) - Number(b.quality);
}
```

shekenahglory: Another option would be to create the amounts once separate from the `mergedOffers` array, sort and then remap the offers array

```
  const offers = self._offers.concat(self._offersAutobridged);
  const qualities = offers.map(function(o, i) {
    return {
      index: i,
      value: OrderBookUtils.getOfferQuality(o, self._currencyGets)
    }
  });

  qualities.sort(function(a, b) {
    return a.value.compareTo(b.value);
  });

  self._mergedOffers = qualities.map(function(q) {
    return offers[q.index];
  });
```

src/orderbook.js
darkdarkdragon: Yes. But I've changed that to 5 seconds.

src/orderbook.js
darkdarkdragon: `_gotOffersFromLegOne` and `_gotOffersFromLegTwo` will both be true when data from both legs will arrive

src/orderbook.js
darkdarkdragon: >Does this offer a performance benefit? Its still recreating 2 amount objects with each function call, which looked to be the main hinderance to performance

Here I've just tried to get rid of closure creation.
And thanks to @sublimator , I've reduced sort function to just comparison of strings, so it's very fast now.

darkdarkdragon:
@shekenahglory  Good point about throttling `notifyDirectOffersChanged`, but it will not be enough.
`AutobridgeCalculator.calculate` can take up to a second, and is called on update from every leg, so it will be called twice per ledger. Plus RippleTrade has two autobridged books on screen, so four times per ledger. And if one ledger takes four seconds to close, just `AutobridgeCalculator.calculate` will take longer.
In new commit I've throttled `OrderBook.computeAutobridgedOffersWrapper` and made a system that dynamically increases throttle pause if calculation takes too long.

darkdarkdragon:
> Thanks for looking into this. Are there any client-side performance results to share?

@wltsmrz I've made benchmarks using node.js, and now it is more than 3 times faster.

src/orderbook.js
```diff
@@ -164,7 +201,15 @@ OrderBook.EVENTS = [
   'offer_changed', 'offer_funds_changed'
 ];
 
-OrderBook.DEFAULT_TRANSFER_RATE = 1000000000;
+OrderBook.DEFAULT_TRANSFER_RATE = Amount.from_json(1000000000);
+
+OrderBook.NOTIFY_THROTTLE_TIME = 4000;
```
shekenahglory: This only needs to be 250ms or maybe less to catch all the transactions - they all get emitted in a cluster when the ledger closes then nothing happens for several seconds.  My guess is you might get very slightly more calls to compute autobridged offers at 250 vs 4000, but significantly less than not throttling (when the volume is heavy).  As a trader want to be as close to real time as possible on orderbook updates.

shekenahglory: After thinking about it, we don't want to use _.throttle for notifyDirectOffersChanged, we need to use _.debounce (with a max wait time) - because throttle will fire on the *first* transaction per ledger that updates the orderbook, then not fire again until after the throttle time - we want to wait until the *last* transaction from an incoming ledger has been processed before further calculation.

The same might be true for computing autobridged offers - what happens when compute is triggered by leg book 1, then in the same ledger, leg book 2 is modified? In this implementation, wouldn't leg book 2's  modifications not be reflected in the orderbook until the next time computing is triggered, on a later ledger?

src/orderbook.js
```diff
@@ -1197,6 +1334,37 @@ OrderBook.prototype.computeAutobridgedOffers = function() {
...
 };
 
+OrderBook.prototype.computeAutobridgedOffersWrapper = function() {
+  var startTime = Date.now();
+  this.computeAutobridgedOffers();
+  this.mergeDirectAndAutobridgedBooks();
+  var lasted = (Date.now() - startTime) * 5;
+  let m = this._autobridgeThrottleTimeMultiplier;
+  while (lasted > m * OrderBook.AUTOBRIDGE_CALCULATE_THROTTLE_TIME) {
+    ++m;
```
shekenahglory: Don't we also need a mechanism for allowing this to decrease back to the original?

src/orderbook.js
```diff
@@ -106,7 +140,16 @@ function OrderBook(remote,
       switch (action) {
         case 'add':
           if (++self._listeners === 1) {
+            self._shouldSubscribe = true;
             self.subscribe();
+
+            self._remote.on('transaction', updateFundedAmountsWrapper);
+            self._remote.once('disconnect', onDisconnect);
+
+            if (self._isAutobridgeable) {
```
shekenahglory: Adding the model's here would mean that calling the autobridged books's subscribe method would not result in subscription to the leg books - only the 'auto subscribe' method of adding a listener or its initial creation would result in a proper subscription (not a problem for trade since it subscribes by adding a listener)

src/orderbook.js
darkdarkdragon: I don't think we needed that. I think autobridged book size will be not change much during user session, so time needed to process it will be constant too. And if user device is slow, it will not get faster.
This code was written with mobile devices in mind.

src/orderbook.js
darkdarkdragon: I specifically marked subscribe/unsubscribe methods as private in comments - anyway there is no point to subscribe without adding listeners, and anyway they not work right if called from outside.

src/orderbook.js
darkdarkdragon: In this implementation leg book 2's modifications will be reflected after `AUTOBRIDGE_CALCULATE_THROTTLE_TIME` time will pass - specifically for that I've used
`{leading: true, trailing: true}` as `_.throttle` options - so `computeAutobridgedOffersWrapper` will be called at the end of period in this case.
And we cannot use debounce here, because we risk to never see changes if there will be high volume of transactions.

shekenahglory: lodash's .debounce has a maxWait param that will ensure that the function will be called.

darkdarkdragon: And it is good idea to change `NOTIFY_THROTTLE_TIME` to something like 100 ms, but autobridge calculation still must be throttled with something big, because it is too slow, and we risk to get same situation like now - when autobridge calculation gets called more often than it can complete

shekenahglory: running your code (with debounce instead of throttle) with a node.js process, its using way less resources than before, I am simultaneously subscribed to 6 very active autobridged books (+ 2 XRP books) - its not mobile but I am not seeing anything close to 1 second to do the autobridge calc, (i did see .7 sec once, and fairly regularly .2-.3 sec.

Using debounce, autobridged calculation is never being called more than once per book per ledger, though it could under heavy volume.

darkdarkdragon: Still don't see the use for debounce here. If I will set wait and maxWait to same value, it effectively be `_throttle`, and if wait will be smaller than maxWait, then it will not serve its main purpose - to make sure not to overload users cpu by not allowing to run autobridge compute to often.

shekenahglory: It has to do with why autobridge compute is being called.  Its being called multiple times needlessly , and debouncing it fixes that.  Before any optimizations, if in 1 ledger 5 transactions affected 1 leg, and 3 transactions affected another, and 2 transactions affected the IOU/IOU book, compute and merge would be called 10 times. 

Throttling it to every 5 seconds with trailing set to true in this case will result in compute being called 2 times for that ledger - once at the very first transaction, and then once again 5 seconds later with all the changes from the ledger.  

Debouncing it will (generally) result in notify being called once per leg and IOU/IOU book per ledger, and (generally) compute being called only once per autobridged book per ledger.  And it gets called immediately after *all* the transactions have been applied, rather than 5 seconds later.

Throttling reduces the calls to compute, but debouncing fixes the reasons why compute is being called so often in the first place.

src/orderbook.js
shekenahglory: should we rename them to _subscribe and _unsubscribe then?

darkdarkdragon: I didn't found the pattern of naming private functions with underscore here, so I left it as is. 
But yes, it would be better to rename all private function to start with underscore.

src/orderbook.js
darkdarkdragon: So you suggest to debounce only `notifyDirectOffersChanged` and not `computeAutobridgedOffersWrapper`?
With what concrete parameters to debounce?

shekenahglory: No definitely debounce computeAutobridgedOffersWrapper - having both legs modified in one ledger happens quite regularly.  right now I have `notify` at timeout = 100, maxWait at 500, and `compute` at 250/500 and it seems to work very well with your other optimizations.  I don't have `compute` adjusting dynamically

I think we should go with something like that, and get it into ripple trade and charts, and then consider further optimization.  This is already phenomenal improvement.

darkdarkdragon: In my firefox it takes 300-600 ms for `autobridgeCalculator.calculate` to complete, and 
`computeAutobridgedOffersWrapper` can take 400-1500 ms
With you suggested change, compute can be called every 250 ms, so it will definitely will kill users cpu in case of high load.
What we can do is debounce throttled function - that way we will get benefit of both methods.

shekenahglory: under heavy load, the debounce wouldn't get a sufficient pause, so the the # of calls to  `compute` would be limited to the `maxWait` interval which would be at 500 ms - we could have a system that dynamically increases the maxWait if the compute is taking too long.

shekenahglory: but I don't think compute time is constant, it seems to vary alot on my machine, so it should have the ability to normalize

darkdarkdragon: Can new transactions arrive every 300ms? That will saturate cpu easily.

What we really need - is to make sure that there is some pause between `computeAutobridgedOffersWrapper` calls, so cpu can process something else. Maybe we should do that check?

shekenahglory: My thinking is, no a transaction cannot arrive every 300ms, they all arrive at the same time the ledger closes, because we are looking only at validated ones which require the ledger to be validated.  That seems to be what is happening when I a following an orderbook, if anyone knows differently please say so.

darkdarkdragon: Ok, but we still need to be make sure not to start compute too often, and increasing `maxWait` will not do the trick - we must start function if pause wasn't long enough between runs.

darkdarkdragon: @shekenahglory I've made changes, please take a look. For notify I now use debounce, and both debounce and throttle for autobridge calculation. And it now decreases back throttled pause if calculation are fast again.

shekenahglory: LGTM, except I think maybe AUTOBRIDGE_CALCULATE_THROTTLE_TIME should be set to 1000 - it keeps the default faster and makes the escalation more granular at the low end, if I did my math correctly:

AUTOBRIDGE_CALCULATE_THROTTLE_TIME @ 4000:

| Compute Time | Multiplier | Throttle Timeout |
| ------------- | ----------- | ---------- |
| 1000 | 1 | 4000 |
| 2000 | 2 | 8000 | 
| 4000 | 3 | 12000 |

AUTOBRIDGE_CALCULATE_THROTTLE_TIME @ 1000:

| Compute Time | Multiplier | Throttle Timeout |
| ------------- | ----------- | ---------- |
| 250 | 1 | 1000 |
| 500 | 2 | 2000 | 
| 1000 | 3 | 3000 |
| 2000 | 4 | 5000 |
| 4000 | 9 | 9000 |

darkdarkdragon: I've changed formula in last commit, so now it looks like this
AUTOBRIDGE_CALCULATE_THROTTLE_TIME @ 4000:

Compute Time | Multiplier | Throttle Timeout
----------------------|---------------|------------------------
1999|1|4000
2000|2|8000
4000|3|12000

if we change AUTOBRIDGE_CALCULATE_THROTTLE_TIME @ 1000:
it will be

Compute Time | Multiplier | Throttle Timeout
----------------------|---------------|------------------------
499|1|1000
500|2|2000
1000|3|3000
1500|4|4000

I think that will be ok

shekenahglory: So you are good with 1000? AUTOBRIDGE_CALCULATE_THROTTLE_TIME is still at 4000 in the latest commit

darkdarkdragon: Yes, I'm good with 1000. I've made new commit right now.

shekenahglory:
LGTM

shekenahglory:
Can we get another review of this or get it merged?

wltsmrz:
Some of the commits ought to be fixed-up

darkdarkdragon:
@wltsmrz Do you want me to squash commits?

wltsmrz:
That'd be great

darkdarkdragon:
@wltsmrz done

darkdarkdragon:
@wltsmrz Can you give it new version and put into npm registry, so it can be used in RippleTrade?

lumberj:
@darkdarkdragon I'll do this first thing in the morning

darkdarkdragon:
@lumberj Thanks!

lumberj:
Can this be merged into develop?

lumberj:
>  Can this be merged into develop?

If not, I'd like to have a 0.12 branch to track continuing development on 0.12. I would branch off from here to create 0.12 branch, tag it, release, and revert this commit on master.

geertweening:
@darkdarkdragon @shekenahglory these changes need to be merged back into develop as well. As it stands right now I can't do so without a ton of merge conflicts. Can you guys work on porting these changes back to develop and open a PR when those changes are ready?

darkdarkdragon:
@geertweening @lumberj I will port it into develop, it will just take some time - there was some changes to OrderBookUtils that needs to be optimzed again

darkdarkdragon:
@lumberj What about giving new version to current master so it can be used in RT?

lumberj:
@darkdarkdragon if I try merging this commit into develop, it creates a lot of conflicts I don't know how to resolve. I want to avoid having to resolve these conflicts much later, when we are ready to release 0.13 and merge develop into master.

darkdarkdragon:
@lumberj I'll merge this over weekend, I just wanted to release new version of RT earlier

src/amount.js
```diff
@@ -388,7 +388,7 @@ Amount.prototype._copy = function(value) {
 Amount.prototype.compareTo = function(to) {
   var toAmount = Amount.from_json(to);
   if (!this.is_comparable(toAmount)) {
-    return new Amount(NaN);
+    return 0;
```
clark800: If they are not comparable you say that they are equal? That doesn't sound right.

darkdarkdragon: Any return here doesn't sound right. Only right solution is to throw exception, but I don't sure it is acceptable here.

clark800: The amount class uses a mixture of NaN's and exceptions; there's a task to make it consistent: https://ripplelabs.atlassian.net/browse/RLJS-217

I think an exception would be much better than returning an incorrect result.

darkdarkdragon: OK, I'll make new pr for it.

src/autobridgecalculator.js
```diff
@@ -70,14 +73,17 @@ AutobridgeCalculator.prototype.calculate = function() {
       continue;
     }
 
-    if (legOneTakerGetsFunded.compareTo(legTwoTakerPaysFunded) > 0) {
+    //  using private fields for speed
```
clark800: I think generally using private fields is not an acceptable solution for optimization. What was causing the slowdown, the from_json in comparedTo?

src/baseconverter.js
```diff
@@ -1,28 +1,39 @@
...
+  let temp, divided;
+  let j = -1;
+
+  const length = digitArray.length;
+  const quotient = new Array(length);
+
+  while (++j < length) {
+    temp = remainder * base + digitArray[j];
+    divided = temp / divisor;
+    quotient[j] = divided << 0;
```
clark800: Just curious, what was the speedup of these divmod changes. I remember trying something like this and didn't get too much speedup.

src/baseconverter.js
```diff
@@ -1,28 +1,39 @@
...
     remainder = temp % divisor;
   }
   return {quotient: normalize(quotient), remainder: remainder};
 }
 
 function convertBase(digitArray, fromBase, toBase) {
-  var result = [];
-  var dividend = digitArray;
+  const result = [];
+  let dividend = digitArray, qr;
```
clark800: The `qr` declarationis kind of hidden over there. I'm proposing an eslint rule to forbid this.

src/orderbook.js
```diff
@@ -1062,28 +1188,33 @@ OrderBook.prototype.deleteOffer = function(node, isOfferCancel) {
...
   assert(Array.isArray(offers), 'Offers is not an array');
 
-  const self = this;
-
   this.resetCache();
 
-  const newOffers = _.map(offers, function(rawOffer) {
-    const offer = OrderBook.offerRewrite(rawOffer);
+  let i = -1, offer;
+  const l = offers.length;
```
clark800: I think the common rule is that you can only use one-letter variable names for loop indices or mathematical symbols.

src/orderbook.js
```diff
@@ -1197,6 +1336,44 @@ OrderBook.prototype.computeAutobridgedOffers = function() {
...
 };
 
+OrderBook.prototype.computeAutobridgedOffersWrapper = function() {
+  var startTime = Date.now();
+  this.computeAutobridgedOffers();
+  this.mergeDirectAndAutobridgedBooks();
+  var lasted = (Date.now() - startTime);
+  
+  const newMult =
+    ((lasted * 2 / OrderBook.AUTOBRIDGE_CALCULATE_THROTTLE_TIME) << 0) + 1;
```
clark800: This trick for doing floor is not readable. I seem to remember it's also not generally more performant on modern engines. If we really have to use it we should probably wrap it in a function or at least add a comment.

src/autobridgecalculator.js
darkdarkdragon: Yes, from_json is generally very slow and must not be used in inner loops. AutobridgeCalculator is slowest part of the code, so I'm trying to squeeze every last bit of performance out of it.
And, if you look into `OrderBookUtils` - I'm using private field of `Amount` there, because I really need a fast way to construct `Amount` object.

clark800: Maybe you could add a guard statement to parse_json to return fast if the argument is already an Amount? And maybe there is a way to speed up Amount creation in the Amount class?

src/orderbook.js
```diff
@@ -1197,6 +1336,44 @@ OrderBook.prototype.computeAutobridgedOffers = function() {
...
+  if (newMult !== this._autobridgeThrottleTimeMultiplier) {
+    this._autobridgeThrottleTimeMultiplier = newMult;
+    this.createDebouncedOffersWrapper();
+  }
+}
+
+OrderBook.prototype.createDebouncedOffersWrapper = function() {
+  const m = this._autobridgeThrottleTimeMultiplier;
+  this.computeAutobridgedOffersThrottled = 
+    _.debounce(
```
clark800: I don't understand what the debounce/throttle are doing. Maybe a comment to explain the main idea?

src/autobridgecalculator.js
darkdarkdragon: This is here a really special cases, that shouldn't be used otherwise.
In compare, I'm sure that other object is `Amount` and comparable, so I don't want to create new object and run comparable checks on it.
As for creation, I want to bypass some checks too, because I'm reasonably sure that my data is valid, and also don't want to create more objects that necessary.

src/orderbook.js
darkdarkdragon: You can read a long our conversation with Matthew here.
debounce is used to group updates, and throttle to make sure that `AutobridgeCalculator.calculate` not run too often.

src/autobridgecalculator.js
clark800: So in the compareTo function, you could skip the from_json call if it is already an Amount.

The general point is that there is sometimes a tradeoff between code quality and performance, and we have to draw a line somewhere so that we don't let perf optimizations create too much complexity. Ideally javascript would have enforced private methods (like many other languages) so we wouldn't even have to discuss it, but I think we should just consider such tricks as off-limits and find other workarounds, which always exist.

darkdarkdragon: Other languages have notion of "friends" classes, specially for this case.
`AutobridgeCalculator` and `OrderBook` are internal to ripplelib code, so I think they are allowed to use some private parts of other classes.
I think we can create something like  `Amount._createFast` that encapsulate fast way of creation `Amount` object.

src/orderbook.js
darkdarkdragon: You're right, here it is no needed.
But my benchmark on latest firefox shows that shifting is 90 times faster.

darkdarkdragon: And, shifting is different operation from `Math.floor` - shifting make value to be integer, so it clamps it to 32 bits.

clark800: > But my benchmark on latest firefox shows that shifting is 90 times faster.

I get that it is only 3% faster in Firefox and slower in Chrome: http://jsperf.com/math-floor-alternatives

> And, shifting is different operation from Math.floor - shifting make value to be integer, so it clamps it to 32 bits.

All the more reason to not use shifting when you really mean floor?

src/autobridgecalculator.js
clark800: The convention for underscore is private though, not friends. If we need a `createFast` method, we can just make it a public method.

src/orderbook.js
clark800: I don't know if I'm understanding correctly, but the only real event is the ledger close event right? Why can't you just do all the processing once for each ledger close event and get rid of all the debouncing and throttling complexity?

shekenahglory: That might be the way to go - the main thing is this code was needlessly running multiple times in one ledger, and can take over 1 sec even on a desktop browser.  You would only want to throttle then in the case it takes over 4 seconds (generally ledger closes are about 4.5 seconds these days)

The reason it was done this way is because the trigger is `transaction` events - that bit would need to be reworked to get all the transactions once the ledger closes.  I guess it would introduce a little latency but then so does the debouncing and throttling.

src/orderbook.js
darkdarkdragon: I don't think we can rely on this this results - it shows me 25% slower on one run, and 6% faster on second.

>All the more reason to not use shifting when you really mean floor?

Thant's the bad thing in js - I want integer arithmetic here, I don't want floats.

lumberj: Then force conversion to integer using `| 0` and use `floor`?

clark800: > I don't think we can rely on this this results - it shows me 25% slower on one run, and 6% faster on second.

Maybe you have other processes running on your system? It is consistent for me.

> Thant's the bad thing in js - I want integer arithmetic here, I don't want floats.

I don't follow this. The only difference is whether it gets truncated to 32 bits right?

src/orderbook.js
darkdarkdragon: > I don't know if I'm understanding correctly, but the only real event is the ledger close event right? Why can't you just do all the processing once for each ledger close event and get rid of all the debouncing and throttling complexity?

Then our data will be always 4 seconds late, and we wanted to show to user data that as much recent as we can.

clark800: Why would it be 4 seconds late? When a ledger closes, you get a batch of new transactions and you process them as fast as you can.

shekenahglory: right, it wouldn't be late, assuming the `ledger_closed` event triggered at the same time the batch of transactions came in.  The only latency would be a call needs to be made back to rippled to get the transactions instead of them streaming in

darkdarkdragon: > right, it wouldn't be late, assuming the ledger_closed event triggered at the same time the batch of transactions came in. The only latency would be a call needs to be made back to rippled to get the transactions instead of them streaming in

You're saying that we just need  to re-request orderbook from server after each `ledger_closed` event?

clark800: > You're saying that we just need to re-request orderbook from server after each ledger_closed event?

Not necessarily, you can just request the transactions in that ledger and do the processing in the same way. But even that's not necessary either, you can just buffer the stream until you get the ledger closed event. Actually it looks like @sublimator was working on this already:

```
+    // We'll wait until the ledger close message which always comes AFTER all
+    // the {"type: "transaction", ... } messages in a well specified (if not
+    // documented) order.
```
https://github.com/ripple/ripple-lib/pull/529/files

I'm not sure why this pull request was closed.

darkdarkdragon: > Why would it be 4 seconds late? When a ledger closes, you get a batch of new transactions and you process them as fast as you can.

What batch? We are waiting for transaction to be sent to us by server, one by one, and we don't know when last will arrive, that's why we use debounce.

clark800: Validated transactions can only be created by a ledger close. So whenever a ledger close occurs, you will get a batch in the stream of all the transactions that were validated in that ledger version.

src/orderbook.js
darkdarkdragon: > Maybe you have other processes running on your system? It is consistent for me.

I don't know what's wrong with jsperf, but simple tests that just do `Math.floor` in loop shows that's it significantly slower than shifting or `| 0` in firefox, but slightly faster in node.js

> I don't follow this. The only difference is whether it gets truncated to 32 bits right?

I just mentioned this because I think it can be better optimized by engine

src/orderbook.js
darkdarkdragon: > Validated transactions can only be created by a ledger close. So whenever a ledger close occurs, you will get a batch in the stream of all the transactions that were validated in that ledger version.

Yeah, we're getting them after close event, and we getting them one by one, without knowing what transaction is last so we can start processing them.

clark800: There is a field `txn_count` in the `ledgerClosed` stream message. Could you buffer until you get that many transactions?

darkdarkdragon: If there is way to know what transaction is last in batch then yes, we don't need that debounce thing.
But we still need throttle, to save user's device from burning in case it's slow.

darkdarkdragon: I'll make new pr that will use that `txn_count` field.

shekenahglory: Some reworking would still need to be done, to ensure that the autobridged book is only merged and emitted once, even if both legs and the IOU/IOU book are modified.  Its pretty common for both legs in heavily used books to be updated on the same ledger, particularly when volume is up.

darkdarkdragon: Good point

clark800: > But we still need throttle

You mean in case it takes more than 4 secs to process the ledger close update? Even then I don't think throttling will help because you have to process all of the transactions to get the correct result.

darkdarkdragon: In case it takes more than 2 seconds - RippleTrade has two books on trade page.
Result will be correct because I'm throttling just `Autobridge.calculate` - it will put most recent data.

clark800: Instead of throttling, could you use a lock that skips calculate if calculate is currently running?

darkdarkdragon: Javascript is single threaded :smile:
I was thinking about putting it into web workers, but I'm worried about compatibility, and they have awkward syntax for creation.
And way I was thinking about to do it - is to create some global (for ripplelib) scheduler, that makes sure that there is some pause between runs, so computer can do things other from calculating orderbooks :smile:

clark800: > Javascript is single threaded

You can add setTimeout(fn, 0) inside long calculations to allow UI operations to execute. Then you need the lock to make sure it doesn't re-enter.

darkdarkdragon: Yeah, that can be done. I'll look into it.

PR #538, opened 2015-09-03
test that shows how

darkdarkdragon:
"Offer total cannot be negative" error appears

### DIFF
--- a/test/amount-test-error.js
+++ b/test/amount-test-error.js
@@ -1,0 +1,46 @@
+'use strict';
+const assert = require('assert');
+const Amount = require('ripple-lib').Amount;
+const Remote = require('ripple-lib').Remote;
+
+const data = require('./fixtures/negative-error');
+
+
+describe.skip('Amount ', function() {
+  it('Show "Offer total cannot be negative" error', function() {
+    const a1 = {
+      currency: 'JPY',
+      issuer: 'r94s8px6kSw1uZ1MV98dhSRTvc6VMPoPcN',
+      value: '66436.33517689175'
+    };
+    const a2 = {
+      currency: 'JPY',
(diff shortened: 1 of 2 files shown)


test/amount-test-error.js
```diff
@@ -0,0 +1,26 @@
...
+    var a2 = {
+        "currency": "JPY",
+        "issuer": "r94s8px6kSw1uZ1MV98dhSRTvc6VMPoPcN",
+        "value": "66435.49665972557"
+    };
+    var a1a = Amount.from_json(a1);
+    var res = a1a.add(a2).subtract(a2).subtract(a1);
+
+    console.log(res.to_human());
+    assert(!res.is_negative(), 'Offer total cannot be negative');
```
sublimator: You're saying it's tiny fractional amounts ?

darkdarkdragon: Yep. I've pushed new commit that shows events that lead to the error.

sublimator: Good catch

darkdarkdragon: Thanks. It was a long hunt :smile:

clark800: Awesome job finding a test case to reproduce it!

test/amount-test-error.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable max-len */
```
clark800: Seems like this is unnecessary

clark800:
@darkdarkdragon Can we fix up the lint errors and add a skip for the unit test so we can get this merged?

clark800:
And squash please

darkdarkdragon:
ok

darkdarkdragon:
@clark800 done

clark800:
LGTM

PR #541, opened 2015-09-08
Add tecOVERSIZE transaction result

wltsmrz:


src/core/binformat.js
```diff
@@ -423,7 +423,7 @@ exports.ter = {
   tecNO_LINE_REDUNDANT     : 127,
   tecPATH_DRY              : 128,
   tecUNFUNDED              : 129,  // Deprecated, old ambiguous unfunded.
-  tecMASTER_DISABLED       : 130,
+  tecNO_ALTERNATIVE_KEY    : 130,
```
sublimator: Hrmm, this was changed huh ?

sublimator: A little bit `hrmmmmmmmm` cause now old json won't serialise properly. When did rippled make this change?

shekenahglory: it wont fail right? just the output will be tecNO_ALTERNATIVE_KEY when it used to say tecMASTER_DISABLED?

wltsmrz:
Let's wait until https://github.com/ripple/ripple-lib/pull/545

src/core/binformat.js
sublimator: Talking about old json -> binary.

sublimator:
LGTM

PR #543, opened 2015-09-08
resend request in case server was disconnected after request was sent

darkdarkdragon:


wltsmrz:
Shouldn't this already occur? https://github.com/ripple/ripple-lib/blob/50acc4c7087bcef6ac9c8a25a0829f926b6b5d58/src/core/remote.js#L769-L771

wltsmrz:
Oh nevermind, I see what you're doing now

wltsmrz:
Might it be preferred to add a `response` event or so in [this neighborhood](https://github.com/ripple/ripple-lib/blob/50acc4c7087bcef6ac9c8a25a0829f926b6b5d58/src/core/server.js#L673) and moving this reconnect logic to request.request() or remote.request()?

darkdarkdragon:
@wltsmrz I've moved logic into request.request

wltsmrz:
It would be great while you're at it to allow `request.request(callback);` again. This was broken a while back. Aside from that, squash commits & LGTM

src/core/server.js
```diff
@@ -686,6 +687,7 @@ Server.prototype._handleResponse = function(message) {
       error_message: 'Remote reported an error.',
       remote: message
     });
+    request.emit('response');
```
wltsmrz: I was imagining `response` as a catch-all... you could put it outside this success/failure check and rename it "message" if that sounds better

darkdarkdragon:
@wltsmrz I've made changes, please take a look - if this is what you wanted, I'll sqash commits

wltsmrz:
LGTM

clark800:
It looks like it's possible that a transaction could be submitted twice. If the first one succeeds, then the second one will fail because it has the same sequence number, but the user will only see the failure result and may mistakenly conclude that they have to resubmit it.

One option could be to check for `tefALREADY` errors in the reconnect case and convert them to success responses.

darkdarkdragon:
ok, I'll make the change

wltsmrz:
There is already reconnect logic in transactionmanager

wltsmrz:
Probably wouldn't be a tefALREADY either, isn't that for same ledger requests?

darkdarkdragon:
Maybe I'm wrong, but I don't transactionmanager being used in new `RippleApi`

wltsmrz:
Maybe not

clark800:
The docs say `tefALREADY` means "The same exact transaction has already been applied."

PR #545, opened 2015-09-10
Add multisign helpers

wltsmrz:
Re: https://github.com/ripple/rippled/commit/9e69bd5c56ae8b4ceb44b353fc75d8d9f6b21834

+ Update binformat
+ Add `SignerListSet` transaction constructors
+ Add transaction methods

**New transaction methods**

`multiSigningData(signing account) -> serializedobject` Internal method, called in multiSign() from reconstituted transaction to prepare signing data

 `getMultiSigningJson() -> tx_json` Returns a normalized tx_json with empty SigningPubKey and canonical signature flag auto-filled

`multiSign(account, secret) -> signer` Multi-signs a transaction reconstituted from getMultiSigningJson()

`addMultiSigner(signer)` Add and sort new multi-signer in the format returned by multiSign(): {Account, SigningPubKey, TxnSignature}

 `hasMultiSigners() -> boolean` Self-explanatory

`getMultiSigners() -> array` Self-explanatory

**Proposed "flow"**

```js
var tx = remote.createTransaction('AccountSet', {
  account: accountID1
});

tx.setSequence(1);
tx.setFee(1000);

// Auto-fill SigningPubKey and possibly set canonical signing flag
// Return normalized tx_json & don't tamper with tx.tx_json
var multiSigningJson = tx.getMultiSigningJson();

// Reconstitute transaction
var multiSigningTx = Transaction.from_json(multiSigningJson);

// Multisign
var signer = multiSigningTx.multiSign(accountID2, seed2);

// Back at the origin tx, attach signer
tx.addSigner(signer);

// Transaction should be submittable as normal
tx.submit(function(err, res) {});
```

**Note**

+ Tested successfully with `local_signing` enabled/disabled
+ Multi-signed transactions are marked as non-resubmittable. Resubmission can be allowed in the future
+ `LastLedgerSequence` isn't set automatically
+ The signers array is indeed uppercase `Signers` and outside `submit_multisigned` request tx_json (unconventional but probably changed soon)
+ `Transaction.from_json()` will now clone the argument

@sublimator @scottschurr

### DIFF
--- a/src/core/transaction.js
+++ b/src/core/transaction.js
@@ -1,7 +1,8 @@
 'use strict';
 
+const assert = require('assert');
 const util = require('util');
-const lodash = require('lodash');
+const _ = require('lodash');
 const EventEmitter = require('events').EventEmitter;
 const utils = require('./utils');
 const sjclcodec = require('sjcl-codec');
@@ -9,6 +10,7 @@ const Amount = require('./amount').Amount;
 const Currency = require('./amount').Currency;
 const UInt160 = require('./amount').UInt160;
 const Seed = require('./seed').Seed;
+const KeyPair = require('ripple-keypairs').KeyPair;
 const SerializedObject = require('./serializedobject').SerializedObject;
 const RippleError = require('./rippleerror').RippleError;
 const hashprefixes = require('./hashprefixes');
@@ -199,7 +201,7 @@ Transaction.from_json = function(j) {
 
 Transaction.prototype.setJson =
 Transaction.prototype.parseJson = function(v) {
-  this.tx_json = v;
+  this.tx_json = _.merge({}, v);
   return this;
 };
 
@@ -358,7 +360,6 @@ Transaction.prototype._computeFee = function() {
   });
 
   const midInd = Math.floor(fees.length / 2);
-
   const median = fees.length % 2 === 0
   ? Math.floor(0.5 + (fees[midInd] + fees[midInd - 1]) / 2)
   : fees[midInd];
@@ -377,56 +378,86 @@ Transaction.prototype._computeFee = function() {
  * return `false`
  */
 
+Transaction.prototype.err = function(error, errorMessage) {
+  this.emit('error', new RippleError(error, errorMessage));
+  return false;
+};
+
 Transaction.prototype.complete = function() {
-  if (this.remote) {
-    if (!this.remote.trusted && !this.remote.local_signing) {
-      this.emit('error', new RippleError(
-        'tejServerUntrusted', 'Attempt to give secret to untrusted server'));
-      return false;
-    }
+  // Auto-fill the secret
+  this._secret = this._secret || this.getSecret();
+
+  if (_.isUndefined(this._secret)) {
+    return this.err('tejSecretUnknown', 'Missing secret');
   }
 
-  if (!this._secret) {
-    this._secret = this.getSecret();
+  if (this.remote && !(this.remote.local_signing || this.remote.trusted)) {
+    return this.err(
+      'tejServerUntrusted',
+      'Attempt to give secret to untrusted server');
   }
 
-  // Try to auto-fill the secret
-  if (!this._secret) {
-    this.emit('error', new RippleError('tejSecretUnknown', 'Missing secret'));
-    return false;
-  }
-
-  if (typeof this.tx_json.SigningPubKey === 'undefined') {
+  if (_.isUndefined(this.tx_json.SigningPubKey)) {
     try {
-      const seed = Seed.from_json(this._secret);
-      const key = seed.get_key();
-      this.tx_json.SigningPubKey = key.pubKeyHex();
-    } catch(e) {
-      this.emit('error', new RippleError(
-        'tejSecretInvalid', 'Invalid secret'));
-      return false;
+      this.setSigningPubKey(this.getSigningPubKey());
+    } catch (e) {
+      return this.err('tejSecretInvalid', 'Invalid secret');
     }
   }
 
-  // If the Fee hasn't been set, one needs to be computed by
-  // an assigned server
-  if (this.remote && typeof this.tx_json.Fee === 'undefined') {
-    if (this.remote.local_fee || !this.remote.trusted) {
-      this.tx_json.Fee = this._computeFee();
-      if (!this.tx_json.Fee) {
-        this.emit('error', new RippleError('tejUnconnected'));
-        return false;
+  // Auto-fill transaction Fee
+  if (_.isUndefined(this.tx_json.Fee)) {
+    if (this.remote && (this.remote.local_fee || !this.remote.trusted)) {
+      const computedFee = this._computeFee();
+
+      if (!computedFee) {
+        // Unable to compute fee due to no connected servers
+        return this.err('tejUnconnected');
       }
+
+      this.tx_json.Fee = computedFee;
     }
   }
 
   if (Number(this.tx_json.Fee) > this._maxFee) {
-    this.emit('error', new RippleError(
-      'tejMaxFeeExceeded', 'Max fee exceeded'));
-    return false;
+    return this.err('tejMaxFeeExceeded', 'Max fee exceeded');
   }
 
   // Set canonical flag - this enables canonicalized signature checking
+  this.setCanonicalFlag();
+
+  return this.tx_json;
+};
+
+Transaction.prototype.getKeyPair = function(secret_) {
+  if (this._keyPair) {
+    return this._keyPair;
+  }
+
+  const secret = secret_ || this._secret;
+  assert(secret, 'Secret missing');
+
+  const keyPair = Seed.from_json(secret).get_key();
+  this._keyPair = keyPair;
+
+  return keyPair;
+};
+
+Transaction.prototype.getSigningPubKey = function(secret) {
+  return this.getKeyPair(secret).pubKeyHex();
+};
+
+Transaction.prototype.setSigningPubKey = function(key) {
+  if (_.isString(key)) {
+    this.tx_json.SigningPubKey = key;
+  } else if (key instanceof KeyPair) {
+    this.tx_json.SigningPubKey = key.pubKeyHex();
+  }
+
+  return this;
+};
+
+Transaction.prototype.setCanonicalFlag = function() {
   if (this.remote && this.remote.local_signing && this.canonical) {
     this.tx_json.Flags |= Transaction.flags.Universal.FullyCanonicalSig;
 
@@ -435,7 +466,7 @@ Transaction.prototype.complete = function() {
     this.tx_json.Flags = this.tx_json.Flags >>> 0;
   }
 
-  return this.tx_json;
+  return this;
 };
 
 Transaction.prototype.serialize = function() {
@@ -450,6 +481,14 @@ Transaction.prototype.signingData = function() {
   const so = new SerializedObject();
   so.append(hashprefixes.HASH_TX_SIGN_BYTES);
   so.parse_json(this.tx_json);
+  return so;
+};
+
+Transaction.prototype.multiSigningData = function(account) {
+  const so = new SerializedObject();
+  so.append(hashprefixes.HASH_TX_MULTISIGN_BYTES);
+  so.parse_json(this.tx_json);
+  so.append(UInt160.from_json(account).to_bytes());
   return so;
 };
 
@@ -470,7 +509,11 @@ Transaction.prototype.hash = function(prefix_, asUINT256, serialized) {
 };
 
 Transaction.prototype.sign = function() {
-  const seed = Seed.from_json(this._secret);
+  if (this.hasMultiSigners()) {
+    return this;
+  }
+
+  const keyPair = this.getKeyPair();
   const prev_sig = this.tx_json.TxnSignature;
 
   delete this.tx_json.TxnSignature;
@@ -483,9 +526,7 @@ Transaction.prototype.sign = function() {
     return this;
   }
 
-  const key = seed.get_key();
-  const hex = key.signHex(this.signingData().buffer);
-  this.tx_json.TxnSignature = hex;
+  this.tx_json.TxnSignature = keyPair.signHex(this.signingData().buffer);
   this.previousSigningHash = hash;
 
   return this;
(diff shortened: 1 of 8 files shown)


src/core/transaction.js
```diff
@@ -1413,6 +1461,24 @@ Transaction.prototype.summary = function() {
   return txSummary;
 };
 
-exports.Transaction = Transaction;
+Transaction.prototype.addSigner = function(signer) {
+  if (_.isUndefined(this.signers)) {
+    this.signers = [];
+  }
+  this.signers.push({Signer: signer});
```
sublimator: I believe at some point these need to be sorted

src/core/transactionmanager.js
```diff
@@ -539,6 +540,10 @@ TransactionManager.prototype._prepareRequest = function(tx) {
     submitRequest.txJson(tx.tx_json);
   }
 
+  if (!_.isEmpty(tx.signers)) {
+    submitRequest.message.Signers = tx.signers;
```
sublimator: Does this actually work for locally signed transactions, i.e. a txn submitted via `tx_blob` property? I'm guessing not

src/core/transaction.js
wltsmrz: Haven't confirmed that it matters. rippled responds with a tesSUCCESS & sorted signers

src/core/transactionmanager.js
wltsmrz: On further testing I'm unable to get a regular `submit` command to work

src/core/transaction.js
sublimator: Using submit-multisigned rpc command ?

wltsmrz: Yeah

src/core/transactionmanager.js
wltsmrz: Scratch that

scottschurr: Does that mean you eventually successfully got a regular `submit` to work with a multisigned blob?  If that doesn't work work I'll need to debug it.  But @sublimator is right, if you submit a blob the `Signers` array must be sorted.

src/core/transaction.js
scottschurr: As a convenience, `submit_multisigned` sorts the Signers array before submitting it.  If the transaction is submitted as a blob the Signers must arrive pre-sorted.  This gives the transaction a canonical serialized representation.

scottschurr:
I'm not a javascript guy, but the parts I could make sense out of looked right to me.

It sounds like the `submit_multisigned` implementation in rippled has a bug.  Either 1) the `Signers` array outside the `tx_json` should be lower case, or 2) the `Signers` array should be inside the `tx_json`.  Is that correct?  If so, which is the preferred fix?

src/core/transactionmanager.js
wltsmrz: Yeah, I eventually got `submit` to work by working backward from a correct `tx_blob`... there was an issue with Transaction.complete() / canonical signing flag. Purely ripple-lib API issue

wltsmrz:
I think an uppercase `Signers` inside the tx_json might be better as the response from rippled will always be in this format

src/core/transactionmanager.js
scottschurr: Nice.  Thanks for the confirmation.

scottschurr:
I'm good with moving the upper case `Signers` inside the tx_json for `submit_multisigned`.  @sublimator, would you care to weigh in on the proposed change?

Also we'll need to consider how to coordinate the change between the two of us.  I can put together a branch with the change and you can use that for development while I shepherd the pull request through.  Do we need to do something more complicated than that?

wltsmrz:
That sounds fine to me @scottschurr

src/core/transaction.js
```diff
@@ -1413,6 +1489,51 @@ Transaction.prototype.summary = function() {
...
+
+Transaction.prototype.getMultiSigningJson = function() {
+  assert(this.tx_json.Sequence, 'Sequence must be set before multi-signing');
+  assert(this.tx_json.Fee, 'Fee must be set before multi-signing');
 
-// vim:sw=2:sts=2:ts=8:et
+  const signingTx = Transaction.from_json(this.tx_json);
+  signingTx.remote = this.remote;
+  signingTx.setSigningPubKey('');
+  signingTx.setCanonicalFlag();
```
wltsmrz: Should we just default this?

sublimator: What you mean?

src/core/transaction.js
```diff
@@ -377,56 +377,73 @@ Transaction.prototype._computeFee = function() {
...
+  this.setCanonicalFlag();
+
+  return this.tx_json;
+};
+
+Transaction.prototype.getSigningPubKey = function(secret_) {
+  const secret = secret_ || this._secret;
+  assert(secret, 'Secret missing');
+  const seed = Seed.from_json(secret);
+  return seed.get_key().pubKeyHex();
```
sublimator: Maybe we can cache the keyPair somewhere at least so it's shared between sign() and complete()?

src/core/transaction.js
wltsmrz: Eh, nevermind

sublimator: Just always set the canonical sig flag ?

src/core/transaction.js
```diff
@@ -1333,16 +1373,52 @@ Transaction.prototype.offerCancel = function(options_) {
...
 
+Transaction._prepareSigner = function(signer) {
+  const {account, weight} = signer;
+
+  if (!UInt160.is_valid(account)) {
+    throw new Error(`Signer account invalid: ${signer.account}`);
+  }
+
+  if (!_.isNumber(weight) || weight < 1 || weight >= Math.pow(256, 2)) {
+    throw new Error(`Signer weight invalid: ${signer.weight}`);
```
sublimator: Something about:  `1 <= ${signer.weight} <= 65535` ?

src/core/transaction.js
wltsmrz: Sure

src/core/transaction.js
wltsmrz: Ok

test/transaction-test.js
```diff
@@ -2066,4 +2106,80 @@ describe('Transaction', function() {
...
+
+    transaction.setSigningPubKey('');
+
+    const a1 = 'rPMh7Pi9ct699iZUTWaytJUoHcJ7cgyziK';
+    const d1 = transaction.multiSigningData(a1);
+
+    const tbytes = ripple.SerializedObject.from_json(
+      lodash.merge(transaction.tx_json, {SigningPubKey: ''})).buffer;
+    const abytes = ripple.UInt160.from_json(a1).to_bytes();
+    const prefix = [83, 77, 84, 0];
```
sublimator: Maybe reference the hashprefixes.js ?

test/transaction-test.js
```diff
@@ -2066,4 +2106,80 @@ describe('Transaction', function() {
...
+
+    const multiSigningJson = transaction.getMultiSigningJson();
+    const t1 = Transaction.from_json(multiSigningJson);
+
+    const a1 = 'rPMh7Pi9ct699iZUTWaytJUoHcJ7cgyziK';
+    const a2 = 'rH4KEcG9dEwGwpn6AyoWK9cZPLL4RLSmWW';
+
+    const s1 = t1.multiSign(a1, 'alice');
+    assert.strictEqual(s1.Account, a1);
+    assert.strictEqual(s1.SigningPubKey, '0388935426E0D08083314842EDFBB2D517BD47699F9A4527318A8E10468C97C052');
```
sublimator: Nice with the rfc6979 deterministic sigs eh :) No need to mess around with patching the rng

test/transaction-test.js
wltsmrz: Ok

src/core/transactionmanager.js
```diff
@@ -520,6 +521,17 @@ TransactionManager.prototype._resubmit = function(ledgers_, pending_) {
...
+    if (this._remote.local_signing) {
+      tx.tx_json.Signers = tx.getMultiSigners();
+    } else {
+      submitRequest.message.command = 'submit_multisigned';
+      submitRequest.message.Signers = tx.getMultiSigners();
+    }
+  }
+
   if (this._remote.local_signing) {
     tx.sign();
```
sublimator: Hrmm, does this try and single sign too? or is there something inside the sign() method that will stop it from creating the TxnSignature when multi signing?

wltsmrz: Presence of TxnSignature apparently doesn't have an effect, but I've added a check in sign()

sublimator: hrmmm, it should likely be a tem ? 

@scottschurr

sublimator:
LGTM

src/core/transaction.js
```diff
@@ -1413,6 +1501,59 @@ Transaction.prototype.summary = function() {
...
+Transaction.prototype.getMultiSigners = function() {
+  return this.multi_signers;
+};
+
+Transaction.prototype.getMultiSigningJson = function() {
+  assert(this.tx_json.Sequence, 'Sequence must be set before multi-signing');
+  assert(this.tx_json.Fee, 'Fee must be set before multi-signing');
+
+  const signingTx = Transaction.from_json(this.tx_json);
+  signingTx.remote = this.remote;
```
sublimator: Why do we make a signingTx Transaction object? and then set the remote? But return tx_json?

src/core/transaction.js
```diff
@@ -1413,6 +1501,59 @@ Transaction.prototype.summary = function() {
...
+  signingTx.setSigningPubKey('');
+  signingTx.setCanonicalFlag();
+
+  return signingTx.tx_json;
+};
+
+Transaction.prototype.multiSign = function(account, secret) {
+  const signingData = this.multiSigningData(account);
```
sublimator: We could make secret the first arg, and make account optional, using the seed -> keyPair.accountID()

src/core/transaction.js
wltsmrz: It clones the original transaction and goes through the motions, sets canonical flag as the original transaction ultimately will before submission... it's a little weird, but I'm ok with it

src/core/transaction.js
wltsmrz: True

clark800:
Looks like quite a few steps in the flow. Would something like this be possible?
```
tx.multiSign(address, secret);
tx.submit();
```

wltsmrz:
You can currently do `tx.addSigner(tx.multiSign(account, secret))` but I don't know why you would want to

clark800:
I'm not that familiar with the rippled multisign design, but if I wanted to make a transaction that required 2 of 3 signatures and provide the first signature myself, would that be easy to do with this interface?

I would expect something like:
```
const tx = createTransaction(...);
tx.setupMultiSign([pubkey1, pubkey2, pubkey3], 2);
tx.sign(privkey1);
// then later:
tx.sign(privkey3);
tx.submit();
```

I'm just thinking that the API doesn't seem to line up directly with this simple API, so I'm wondering if it might be possible to simplify it.

wltsmrz:
I don't know how useful multisign is if the originator already has the secrets for all multisigners. I'm sure there are use cases but I've assumed that the multisigners may want to be separate parties. I don't think it's radically complex to call `getMultiSigningJson()` on the side of the initiator and `Transaction.from_json()` on the side of the multisigner... but again, this should be optional currently.

In your example `setupMultiSign()` requires another transaction to rippled: `SignerListSet`

src/core/transactionmanager.js
scottschurr: Sorry, I haven't got all of the context.  Is the question here whether a `sign` RPC command should fail if there's a `Signers` array inside the `tx_json`?  I don't believe I currently check for that, but I certainly could.

If the transaction actually gets submitted then the checking prevents a transaction from being both single- and multi-signed.

sublimator: Scott, I think Brandon was able to 'submit' a tx_blob with both
TxnSignature AND the Signers array.

wltsmrz: Reconfirmed

clark800:
Ok, I think that makes sense, thanks.

src/core/transactionmanager.js
scottschurr: Yuck.  Good find gents.  I need to fix that.  I'll see if I can figure out what's wrong with my checks.

sublimator: I think in the original multi sign impl it was checked.

scottschurr: I think I may understand what's going on.  The decision between regular- and multi-signing is based on the state of the `SigningPubKey` field.  If the field has a non-empty value then we're regular signing.  If the field is empty then we're multi-signing.  I have a unit test that looks for a `temINVALID` if the `SigningPubKey` is non-empty but `Signers` is present.

I don't (yet) have a test with an empty `SigningPubKey` and both `Signers` and `TxnSignature` filled in.  I bet that is the case you're seeing.

The good news is that the `SigningPubKey` value determines which kind of signing we do.  If it's non-empty it's regular signing.  If it's empty we're multisigning.  So the bug could have been worse.  Still, I'll try to figure out a way to wrangle the unit test framework into performing this test.  Assuming the test fails then I'll fix the bug.  If the test doesn't fail then I'll still be puzzled.

wltsmrz:
I realized we should add a check to make sure `LastLedgerSequence` is set before submission and we can automatically set `LastLedgerSequence` the same way canonical signing flag is currently set. Currently, it must be set manually, but there are no checks. Will have another pull request for these changes

sublimator:
LastLedgerSequence must be signed yah?

wltsmrz:
Yeah, should return it in `getMultiSigningJson()`

sublimator:
It could take quite some ledgers to get all the sigs

wltsmrz:
Seems like it should still always be set though, right? People should be aware of the low default LastLedgerSequence offset (3), which is probably not realistic for multisigning

sublimator:
Maybe use a different default for multi signing?

wltsmrz:
Seems like it could be done. Should we then expose a `multisign_last_ledger_offset` Remote config? I might be satisfied with a `tx.setLastLedgerSequenceOffset()`

wltsmrz:
https://github.com/ripple/ripple-lib/pull/552

PR #546, opened 2015-09-14
make Remote.getLedgerSequence asynchronous

darkdarkdragon:
copy message from Error instance into RippleError

### DIFF
--- a/src/api/ledger/balances.js
+++ b/src/api/ledger/balances.js
@@ -7,6 +7,9 @@ const getTrustlines = require('./trustlines');
 const validate = utils.common.validate;
 const composeAsync = utils.common.composeAsync;
 const convertErrors = utils.common.convertErrors;
+import type {Remote} from '../../core/remote';
+import type {GetLedgerSequenceCallback} from '../../core/remote';
+
 
 function getTrustlineBalanceAmount(trustline) {
   return {
@@ -31,14 +34,26 @@ function getTrustlinesAsync(account, options, callback) {
     .catch(callback);
 }
 
+function getLedgerVersion(remote: Remote, optionValue?: number,
+  callback: GetLedgerSequenceCallback
+) {
+  if (typeof optionValue === 'number') {
+    callback(null, optionValue);
+  } else {
+    remote.getLedgerSequence(callback);
+  }
+}
+
+
 function getBalancesAsync(account, options, callback) {
   validate.address(account);
   validate.getBalancesOptions(options);
 
-  const ledgerVersion = options.ledgerVersion
-                      || this.remote.getLedgerSequence();
   async.parallel({
-    xrp: _.partial(utils.getXRPBalance, this.remote, account, ledgerVersion),
+    xrp: async.compose(
+      _.partial(utils.getXRPBalance, this.remote, account),
+      _.partial(getLedgerVersion, this.remote, options.ledgerVersion)
+    ),
     trustlines: _.partial(getTrustlinesAsync.bind(this), account, options)
   }, composeAsync(formatBalances, convertErrors(callback)));
 }
--- a/src/api/ledger/balances.js
+++ b/src/api/ledger/balances.js
@@ -34,25 +34,24 @@ function getTrustlinesAsync(account, options, callback) {
     .catch(callback);
 }
 
-function getLedgerVersion(remote: Remote, optionValue?: number,
+function getLedgerVersionHelper(remote: Remote, optionValue?: number,
   callback: GetLedgerSequenceCallback
 ) {
-  if (typeof optionValue === 'number') {
+  if (optionValue !== undefined && optionValue !== null) {
     callback(null, optionValue);
   } else {
     remote.getLedgerSequence(callback);
   }
 }
 
-
 function getBalancesAsync(account, options, callback) {
   validate.address(account);
   validate.getBalancesOptions(options);
 
   async.parallel({
-    xrp: async.compose(
-      _.partial(utils.getXRPBalance, this.remote, account),
-      _.partial(getLedgerVersion, this.remote, options.ledgerVersion)
+    xrp: async.seq(
+      _.partial(getLedgerVersionHelper, this.remote, options.ledgerVersion),
+      _.partial(utils.getXRPBalance, this.remote, account)
     ),
     trustlines: _.partial(getTrustlinesAsync.bind(this), account, options)
   }, composeAsync(formatBalances, convertErrors(callback)));
--- a/src/api/ledger/orders.js
+++ b/src/api/ledger/orders.js
@@ -26,10 +26,21 @@ function getOrdersAsync(account, options, callback) {
   validate.address(account);
   validate.getOrdersOptions(options);
 
-  const ledgerVersion = options.ledgerVersion
-                      || this.remote.getLedgerSequence();
+  if (!options.ledgerVersion) {
+    const self = this;
+    this.remote.getLedgerSequence((err, seq) => {
+      if (err) {
+        convertErrors(callback)(err);
+        return;
+      }
+      const newOptions = _.extend(options, {ledgerVersion: seq});
+      getOrdersAsync.call(self, account, newOptions, callback);
+    });
+    return;
+  }
+
   const getter = _.partial(requestAccountOffers, this.remote, account,
-                           ledgerVersion);
+                           options.ledgerVersion);
   utils.getRecursive(getter, options.limit,
     composeAsync((orders) => _.sortBy(orders,
       (order) => order.properties.sequence), callback));
--- a/src/api/ledger/orders.js
+++ b/src/api/ledger/orders.js
@@ -1,6 +1,7 @@
 /* @flow */
 'use strict';
 const _ = require('lodash');
+const async = require('async');
 const utils = require('./utils');
 const validate = utils.common.validate;
 const composeAsync = utils.common.composeAsync;
@@ -26,19 +27,6 @@ function getOrdersAsync(account, options, callback) {
   validate.address(account);
   validate.getOrdersOptions(options);
 
-  if (!options.ledgerVersion) {
-    const self = this;
-    this.remote.getLedgerSequence((err, seq) => {
-      if (err) {
-        convertErrors(callback)(err);
-        return;
-      }
-      const newOptions = _.extend(options, {ledgerVersion: seq});
-      getOrdersAsync.call(self, account, newOptions, callback);
-    });
-    return;
-  }
-
   const getter = _.partial(requestAccountOffers, this.remote, account,
                            options.ledgerVersion);
   utils.getRecursive(getter, options.limit,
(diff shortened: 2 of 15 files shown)


src/api/ledger/balances.js
```diff
@@ -31,14 +34,26 @@ function getTrustlinesAsync(account, options, callback) {
     .catch(callback);
 }
 
+function getLedgerVersion(remote: Remote, optionValue?: number,
+  callback: GetLedgerSequenceCallback
+) {
+  if (typeof optionValue === 'number') {
```
clark800: I'd prefer `optionValue !== undefined && optionValue !== null` to avoid `typeof`.

src/api/ledger/balances.js
```diff
@@ -31,14 +34,26 @@ function getTrustlinesAsync(account, options, callback) {
     .catch(callback);
 }
 
+function getLedgerVersion(remote: Remote, optionValue?: number,
```
clark800: Confusing name because `getLedgerVersion` already exists in the API, maybe something like `getLedgerVersionHelper`?

src/api/ledger/balances.js
```diff
@@ -31,14 +34,26 @@ function getTrustlinesAsync(account, options, callback) {
...
+
 function getBalancesAsync(account, options, callback) {
   validate.address(account);
   validate.getBalancesOptions(options);
 
-  const ledgerVersion = options.ledgerVersion
-                      || this.remote.getLedgerSequence();
   async.parallel({
-    xrp: _.partial(utils.getXRPBalance, this.remote, account, ledgerVersion),
+    xrp: async.compose(
```
clark800: I think `async.seq` is a bit more intuitive since it orders the functions in execution order.

src/api/ledger/orders.js
```diff
@@ -26,10 +26,21 @@ function getOrdersAsync(account, options, callback) {
   validate.address(account);
   validate.getOrdersOptions(options);
 
-  const ledgerVersion = options.ledgerVersion
-                      || this.remote.getLedgerSequence();
+  if (!options.ledgerVersion) {
+    const self = this;
```
clark800: You don't need `self` when using arrow functions.

src/api/ledger/orders.js
```diff
@@ -26,10 +26,21 @@ function getOrdersAsync(account, options, callback) {
   validate.address(account);
   validate.getOrdersOptions(options);
 
-  const ledgerVersion = options.ledgerVersion
-                      || this.remote.getLedgerSequence();
+  if (!options.ledgerVersion) {
```
clark800: I think it would simplify things a lot if we used the API's getLedgerVersion method and use promise.then. Also, could use getLedgerVersionHelper so that it only calls getLedgerVersion when ledgerVersion is not specified in the options. Then you can get rid of the branching here.

clark800:
I don't think we have to make remote.getLedgerSequence async. I thought we just wanted to make RippleAPI.getLedgerVersion return a promise. If you do just that, most of this pull request is unnecessary I think. Is there some advantage to making remote.getLedgerSequence async too?

darkdarkdragon:
I will still need to replace all the calls to `remote.getLedgerSequence` with the async version, be it `remote.getLedgerSequence` or `api.getLedgerVersion`, so there is no difference.

clark800:
> I will still need to replace all the calls to remote.getLedgerSequence with the async version

Why?

darkdarkdragon:
Because original goal of this change was to make api behavior consistent - right now, if you call `getServerInfo` while server is not connected - it will wait for connection and then return result, but if call `getTransaction` it will fail immediately if not connected.

clark800:
Couldn't that also be accomplished by adding a wrapper called `ensureConnected` around `getServerInfoAsync`?

darkdarkdragon:
Maybe. But this feels like hacks. All the core api implemented in such way that it will wait for connection, and RippleApi relays on that, and only use of `getLedgerSequence` ruins all that cool system. And using wrappers over that feels like making hack and not fixing problem in first place.

clark800:
That's true. I think if we go with this solution, we would have to factor out the `if (!options.ledgerVersion) {` blocks. It's too complex for the task. Can you try to refactor that logic into a helper function? I think we should only have to add one or two lines to each of the methods.

darkdarkdragon:
OK, I'll refactor it

darkdarkdragon:
@clark800 done

src/api/ledger/transaction.js
```diff
@@ -78,11 +78,22 @@ function getTransactionAsync(identifier: string, options: TransactionOptions,
     }
   }
 
+
+  /* eslint-disable no-unused-vars, handle-callback-err */
```
clark800: We can avoid the disabling comment if we use the technique described here to selectively mark individual parameters as unused: https://github.com/ripple/javascript-style-guide/pull/20

Also, why is the error case not being handled?

src/api/ledger/transactions.js
```diff
@@ -121,19 +121,26 @@ function getTransactionsAsync(account, options, callback) {
...
-      const bound = options.earliestFirst ?
-        {minLedgerVersion: ledgerVersion} : {maxLedgerVersion: ledgerVersion};
-      const newOptions = _.assign(defaults, options, {startTx: tx}, bound);
-      getTransactionsInternal(this.remote, account, newOptions, callback);
-    }).catch(callback);
-  } else {
-    const newOptions = _.assign(defaults, options);
-    getTransactionsInternal(this.remote, account, newOptions, callback);
-  }
+  const self = this;
```
clark800: Can we use an arrow function to eliminate this?

src/api/server/server.js
```diff
@@ -63,28 +62,43 @@ function getServerInfoAsync(
   });
 }
 
-function getFee(): number {
-  return common.dropsToXrp(this.remote.createTransaction()._computeFee());
+function getFee(): ?number {
+  if (!this.remote._servers.length) {
```
clark800: Underscores are private

src/api/server/server.js
```diff
@@ -63,28 +62,43 @@ function getServerInfoAsync(
...
+    throw new common.errors.RippledNetworkError('No servers available.');
+  }
+  const fee = this.remote.createTransaction()._computeFee();
+  if (typeof fee !== 'string') {
```
clark800: Isn't the only non-string return value `undefined`? If so, it's better to test for undefined than to test the type.

src/api/transaction/utils.js
```diff
@@ -39,42 +43,64 @@ function formatPrepareResponse(txJSON) {
...
   const txJSON = transaction.tx_json;
 
-  if (instructions.maxLedgerVersion !== undefined) {
-    txJSON.LastLedgerSequence = parseInt(instructions.maxLedgerVersion, 10);
-  } else {
-    const offset = instructions.maxLedgerVersionOffset !== undefined ?
-      parseInt(instructions.maxLedgerVersionOffset, 10) : 3;
-    txJSON.LastLedgerSequence = remote.getLedgerSequence() + offset;
+
+  function prepare1(callback_) {
```
clark800: Needs more descriptive names than prepare1, prepare2, prepare3: prepareMaxLedgerVersion, prepareFee, prepareSequence

src/api/ledger/transaction.js
```diff
@@ -78,11 +78,22 @@ function getTransactionAsync(identifier: string, options: TransactionOptions,
...
+    });
+  }
+  /* eslint-enable no-unused-vars, handle-callback-err */
+
   async.waterfall([
     _.partial(remote.requestTx.bind(remote),
       {hash: identifier, binary: false}),
     _.partial(attachTransactionDate, remote)
-  ], callbackWrapper);
+  ], callbackWrapper2);
```
clark800: Can these be named more descriptively?

src/core/remote.js
```diff
@@ -521,7 +523,35 @@ Remote.prototype._handleMessage = function(message, server) {
...
+  if (!this._servers.length) {
+    callback(new Error('No servers available.'));
+    return;
+  }
+
+  if (_.isFinite(this._ledger_current_index)) {
+    // the "current" ledger is the one after the most recently closed ledger
+    callback(null, this._ledger_current_index - 1);
+  } else {
+    const self = this;
```
clark800: Arrow function to eliminate `self`

test/api-test.js
```diff
@@ -562,8 +564,13 @@ describe('RippleAPI', function() {
     });
   });
 
-  it('getLedgerVersion', function() {
-    assert.strictEqual(this.api.getLedgerVersion(), 8819951);
+  it('getLedgerVersion', function(done) {
+    this.api.getLedgerVersion().then((ver) => {
+      assert.strictEqual(ver, 8819951);
+      done();
+    }, (err) => {
```
clark800: I think you can just write `done` here

clark800:
Looks like this needs a rebase

clark800:
I'm not sure if it's necessary, but do you think it would be better if we started using promises more instead of `async`?

src/api/ledger/transaction.js
darkdarkdragon: > Also, why is the error case not being handled?

There is no point to check it. `getLedgerSequence` returns error only in one case - servers was not specified in `Remote` constructor, and in this case error will be thrown much earlier.

darkdarkdragon:
> I'm not sure if it's necessary, but do you think it would be better if we started using promises more instead of async?

But this would require changes to core api? 
I would prefer using promises, but I don't see how.

src/api/ledger/transaction.js
clark800: It's bad practice because getLedgerSequence could change in the future in unexpected ways. We should always follow best practices and not take shortcuts. Generally I think there are very few cases where disabling eslint rules is justified.

clark800:
So for example, instead of using `remote.getLedgerSequence` everywhere, you could use `getLedgerVersion`.

darkdarkdragon:
Yeah, you right.

src/api/ledger/transaction.js
darkdarkdragon: Yeah, you're right about change, but here I really don't care if `getLedgerSequence` returns an error.

src/api/ledger/transaction.js
```diff
@@ -89,11 +89,20 @@ function getTransactionAsync(identifier: string, options: TransactionOptions,
     }
   }
 
+
+  function maxLedgerGetter(error_?: Error, tx?: Object) {
+    remote.getLedgerSequence(function(err?, seq: number) {
```
clark800: Since this is in "src/api", we should consistently use "ledgerVersion" instead of "seq"/"sequence"/"ledgerSequence"/"sequence" (here and below)

darkdarkdragon: Do you mean use `api.getLedgerVersion` instead of `remote.getLedgerSequence`?

clark800: I mean renaming the parameter "seq"

darkdarkdragon:
@clark800 I've made changes, please take a look

clark800:
LGTM

PR #548, opened 2015-09-15
Add PendingLedgerVersionError

lumberj:
NonContiguousLedgerHistoryError:
When requesting a tx, if maxLedgerVersion and minLedgerVersion provided, check
if min and max are within a non-contiguous range in the ledger's history

PendingLedgerVersionError:
If maxLedgerVersion provided, check if ledger is ahead of the server's last
validated ledger

### DIFF
--- a/src/api/common/errors.js
+++ b/src/api/common/errors.js
@@ -58,6 +58,13 @@ function MissingLedgerHistoryError(message) {
 MissingLedgerHistoryError.prototype = new RippleError();
 MissingLedgerHistoryError.prototype.name = 'MissingLedgerHistoryError';
 
+function PendingLedgerVersionError(message) {
+  this.message = message ||
+    'maxLedgerVersion is greater than server\'s most recent validated ledger';
+}
+PendingLedgerVersionError.prototype = new RippleError();
+PendingLedgerVersionError.prototype.name = 'PendingLedgerVersionError';
+
 /**
  * Request timed out
  */
@@ -82,6 +89,7 @@ module.exports = {
   TransactionError,
   RippledNetworkError,
   NotFoundError,
+  PendingLedgerVersionError,
   MissingLedgerHistoryError,
   TimeOutError,
   ApiError,
--- a/src/api/ledger/transaction.js
+++ b/src/api/ledger/transaction.js
@@ -52,8 +52,8 @@ function getTransactionAsync(identifier: string, options: TransactionOptions,
   validate.getTransactionOptions(options);
 
   const remote = this.remote;
-  const maxLedgerVersion = Math.min(options.maxLedgerVersion || Infinity,
-    remote.getLedgerSequence());
+  const maxLedgerVersion =
+    options.maxLedgerVersion || remote.getLedgerSequence();
 
   function callbackWrapper(error_?: Error, tx?: Object) {
     let error = error_;
@@ -67,13 +67,19 @@ function getTransactionAsync(identifier: string, options: TransactionOptions,
       error = new errors.NotFoundError('Transaction not found');
     }
 
+    // Missing complete ledger range
     if (error instanceof errors.NotFoundError
-        && !utils.hasCompleteLedgerRange(remote,
-            options.minLedgerVersion, maxLedgerVersion)) {
-      callback(new errors.MissingLedgerHistoryError('Transaction not found,'
-        + ' but the server\'s ledger history is incomplete'));
+        && !utils.hasCompleteLedgerRange(remote, options.minLedgerVersion,
+                                         maxLedgerVersion)) {
+      if (utils.isPendingLedgerVersion(remote, maxLedgerVersion)) {
+        callback(new errors.PendingLedgerVersionError());
+      } else {
+        callback(new errors.MissingLedgerHistoryError());
+      }
+      // Transaction is found, but not in specified range
     } else if (!error && tx && !isTransactionInRange(tx, options)) {
       callback(new errors.NotFoundError('Transaction not found'));
+      // Transaction is not found
     } else if (error) {
       convertErrors(callback)(error);
     } else if (!tx) {
--- a/src/api/ledger/utils.js
+++ b/src/api/ledger/utils.js
@@ -99,10 +99,17 @@ function compareTransactions(first: Outcome, second: Outcome): number {
 function hasCompleteLedgerRange(remote: Remote, minLedgerVersion?: number,
                                 maxLedgerVersion?: number
 ): boolean {
+
   const firstLedgerVersion = 32570; // earlier versions have been lost
   return remote.getServer().hasLedgerRange(
     minLedgerVersion || firstLedgerVersion,
     maxLedgerVersion || remote.getLedgerSequence());
+}
+
+function isPendingLedgerVersion(remote: Remote, maxLedgerVersion: ?number
+): boolean {
+  const currentLedger = remote.getLedgerSequence();
+  return currentLedger < (maxLedgerVersion || 0);
 }
 
 module.exports = {
@@ -112,6 +119,7 @@ module.exports = {
   renameCounterpartyToIssuerInOrder,
   getRecursive,
   hasCompleteLedgerRange,
+  isPendingLedgerVersion,
   promisify: common.promisify,
   clamp: clamp,
   common: common
--- a/test/api-test.js
+++ b/test/api-test.js
@@ -160,157 +160,184 @@ describe('RippleAPI', function() {
       _.partial(checkResult, responses.getBalances, 'getBalances'));
   });
 
-  it('getTransaction - payment', function() {
-    return this.api.getTransaction(hashes.VALID_TRANSACTION_HASH).then(
-      _.partial(checkResult, responses.getTransaction.payment,
-        'getTransaction'));
-  });
-
-  it('getTransaction - settings', function() {
-    const hash =
-      '4FB3ADF22F3C605E23FAEFAA185F3BD763C4692CAC490D9819D117CD33BFAA1B';
-    return this.api.getTransaction(hash).then(
-      _.partial(checkResult, responses.getTransaction.settings,
-        'getTransaction'));
-  });
-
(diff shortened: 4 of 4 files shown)


lumberj:
Requires fixing the tests

src/api/ledger/transaction.js
```diff
@@ -62,11 +62,19 @@ function getTransactionAsync(identifier: string, options: TransactionOptions,
...
+        && !utils.hasCompleteLedgerRange(remote, minLedgerVersion,
+                                         maxLedgerVersion)) {
+      if (minLedgerVersion && maxLedgerVersion
+          && utils.isNonContiguousLedgerRange(remote, minLedgerVersion, 
+                                         maxLedgerVersion)) {
+        callback(new errors.NonContiguousLedgerHistoryError());
+      } else if (maxLedgerVersion 
+                 && utils.isPendingLedgerVersion(remote, maxLedgerVersion)) {
+        callback(new errors.PendingLedgerError());
+      }
```
clark800: The if/else doesn't cover all the cases. I think it's probably easiest to check for PendingLedgerVersionError (looks like typo above), then have an else (not else if) for NonContiguousLedgerHistoryError.

lumberj:
The typechecking errors are not obvious to me what they are..

src/api/ledger/transaction.js
lumberj: Will update

clark800:
For the typechecking errors, I think you just have to add the `?` to the parameter annotations.

lumberj:
The parameter annotations are correct, though. They should not be optional (indicated by the `?`)

src/api/ledger/transaction.js
lumberj: > The if/else doesn't cover all the cases.

The cases it covers:

* tx is missing
  * server has incomplete ledger history in the specified range, 
    * ledger history is noncontiguous in the specified range
    * maxLedgerVersion is ahead of the server's last known ledger
  * server has complete ledger history in the specified range
* tx is not missing
  * ...

What's missing?

clark800: For example, there is no case that covers `maxLedgerVersion = undefined`

src/api/ledger/utils.js
```diff
@@ -105,6 +105,25 @@ function hasCompleteLedgerRange(remote: Remote, minLedgerVersion?: number,
...
+}
+
+function isNonContiguousLedgerRange(remote: Remote, minLedgerVersion?: number,
+                                    maxLedgerVersion?: number
+): boolean {
+  if (minLedgerVersion === undefined || maxLedgerVersion === undefined) {
+    return false;
+  }
+  return !hasCompleteLedgerRange(remote, minLedgerVersion, maxLedgerVersion)
+    && hasLedger(minLedgerVersion) && hasLedger(maxLedgerVersion);
```
lumberj: This is the typechecking issue. `minLedgerVersion` and `max` may be undefined.

src/api/ledger/transaction.js
lumberj: Isn't that this case? https://github.com/ripple/ripple-lib/pull/548/files#diff-c9b59b930eec9b6808a63ab2fbfab346R86

It's just a NotFoundError in that case. If you don't specify a range, or a max

src/api/ledger/utils.js
clark800: There are two `?` syntaxes, one for optional parameters, and one for maybe types. Here you need the `?` before the type name for a maybe type.

src/api/ledger/transaction.js
clark800: No, I mean once you are inside the first `if` branch.

src/api/ledger/utils.js
lumberj: That doesn't work. I get an error like this instead now:

```
ripple-lib/build/flow/api/ledger/utils.js:107:11,76: function call
Error:
ripple-lib/build/flow/api/ledger/utils.js:103:72,77: null
This type is incompatible with
ripple-lib/build/flow/api/ledger/utils.js:94:68,73: number
```

lumberj: Ok now working

src/api/common/errors.js
```diff
@@ -58,6 +58,21 @@ function MissingLedgerHistoryError(message) {
...
+  this.message = message ||
+    'Server has a gap in the ledger history in the specified range';
+}
+NonContiguousLedgerHistoryError.prototype = new RippleError();
+NonContiguousLedgerHistoryError.prototype.name =
+  'NonContiguousLedgerHistoryError';
+
+function PendingLedgerVersionError(message) {
+  this.message = message ||
+    'maxLedgerVersion is ahead of the server\'s ledger history';
```
clark800: Might be more clear to say something like "maxLedgerVersion is greater than server's most recent validated ledger"

lumberj: I like that, will change

src/api/ledger/utils.js
```diff
@@ -105,6 +105,25 @@ function hasCompleteLedgerRange(remote: Remote, minLedgerVersion?: number,
...
+}
+
+function isPendingLedgerVersion(remote: Remote, maxLedgerVersion?: number
+): boolean {
+  return remote.getServer().getLastLedger < maxLedgerVersion || 0;
```
clark800: Should probably use remote.getLedgerSequence(). Also, parens could be good: `(maxLedgerVersion || 0)`.

src/api/ledger/utils.js
```diff
@@ -105,6 +105,28 @@ function hasCompleteLedgerRange(remote: Remote, minLedgerVersion?: number,
...
 
+function hasLedger(remote: Remote, ledger: ?number) {
+  return remote.getServer().hasLedger(ledger);
+}
+
+function isNonContiguousLedgerRange(remote: Remote, minLedgerVersion: ?number,
+                                    maxLedgerVersion: ?number
+): boolean {
+  if (!minLedgerVersion || !maxLedgerVersion) {
+    return false;
```
clark800: This could cause issues because the hasCompleteLedgerRange function fills in default values for the min and max. You might want to do the same here.

clark800: But if you use my suggestion above you could delete this whole function anyways.

lumberj: What issues would it cause? If min and max are not provided, then this function returns false.

src/api/ledger/transaction.js
lumberj: It is covered. If `maxLedgerVersion` is undefined, both conditionals in the first `if` branch return false. Then you get to line 86 and it returns with the original NotFoundError

clark800: `utils.hasCompleteLedgerRange` can be either true or false even if `maxLedgerVersion` is undefined, so it may never reach line 86. Also more generally, you should try to be defensive and avoid this kind of coupling between different branches of an if/else statement. A good rule of thumb is to usually not end with an "else if" because it quite often indicates that there are missing cases.

src/api/ledger/utils.js
clark800: It's inconsistent that `hasCompleteLedgerRange` provides defaults, but `isNonContiguousLedgerRange` doesn't even though `isNonContiguousLedgerRange` calls `hasCompleteLedgerRange`. Let's just delete this function though.

lumberj: Delete which function? `hasCompleteLedgerRange` or `isNonContigousLedgerRange`

clark800: I was thinking delete `isNonContigousLedgerRange` because it's just the `else` case above whenever its not a pending error (doesn't need to be an else if).

lumberj: Makes sense, will do

test/api-test.js
```diff
@@ -160,157 +160,184 @@ describe('RippleAPI', function() {
...
     });
-  });
 
-  it('getTransaction - not found by hash', function() {
-    const hash = hashes.NOTFOUND_TRANSACTION_HASH;
-    return this.api.getTransaction(hash).then(() => {
-      assert(false, 'Should throw NotFoundError');
-    }).catch(error => {
-      assert(error instanceof this.api.errors.NotFoundError);
+    it('getTransaction - missing ledger history with ledger range', function() {
```
lumberj: New test

test/api-test.js
```diff
@@ -160,157 +160,184 @@ describe('RippleAPI', function() {
...
 
-  it('getTransaction - missing ledger history', function() {
-    const hash = hashes.NOTFOUND_TRANSACTION_HASH;
-    // make gaps in history
-    this.api.remote.getServer().emit('message', ledgerClosed);
-    return this.api.getTransaction(hash).then(() => {
-      assert(false, 'Should throw MissingLedgerHistoryError');
-    }).catch(error => {
-      assert(error instanceof this.api.errors.MissingLedgerHistoryError);
+    it('getTransaction - not found - future maxLedgerVersion', function() {
```
lumberj: New test

lumberj:
@clark800 Updated and squashed

src/api/ledger/utils.js
```diff
@@ -99,12 +99,21 @@ function compareTransactions(first: Outcome, second: Outcome): number {
...
   const firstLedgerVersion = 32570; // earlier versions have been lost
   return remote.getServer().hasLedgerRange(
     minLedgerVersion || firstLedgerVersion,
     maxLedgerVersion || remote.getLedgerSequence());
 }
 
+function isPendingLedgerVersion(remote: Remote, maxLedgerVersion: ?number
+): boolean {
+  const currentLedger = remote.getLedgerSequence();
+  console.log('CURENT_LEDGER', currentLedger);
```
clark800: Left some debugging statements here

lumberj: Good catch. Embarrassing typo

lumberj:
Removed debug statements

clark800:
LGTM

PR #549, opened 2015-09-16
Do not return non-validated transaction from RippleAPI#getTransaction

lumberj:


src/api/ledger/transaction.js
```diff
@@ -57,6 +57,11 @@ function getTransactionAsync(identifier: string, options: TransactionOptions,
 
   function callbackWrapper(error_?: Error, tx?: Object) {
     let error = error_;
+
+    if (!error && tx && tx.validated === false) {
```
clark800: Would be more defensive to do `tx.validated !== true`

lumberj: Agreed, will change

test/fixtures/api/rippled/tx/not-validated.json
```diff
@@ -40,6 +40,7 @@
       ],
       "TransactionIndex": 5,
       "TransactionResult": "tesSUCCESS"
-    }
+    },
+    "validated": false
```
clark800: Do you know why these were missing? Does rippled ever omit "validated"?

lumberj: It shouldn't

lumberj: Checking

clark800:
LGTM

PR #552, opened 2015-09-17
Autofill LastLedgerSequence for multisigned transactions

wltsmrz:
+ `Signers` has been moved to tx_json for submit/submit_multisigned
+ Automatically set LastLedgerSequence, honor existing values
+ By default, LastLedgerSequence offset is unrealistically low. Users should call either `tx.setLastLedgerSequence()` or `tx.setLastLedgerSequenceOffset()` before `tx.getMultiSigningJson()` / multisigning

### DIFF
--- a/src/core/transaction.js
+++ b/src/core/transaction.js
@@ -45,6 +45,7 @@ function Transaction(remote) {
   ? this.remote.automatic_resubmission
   : true;
   this._maxFee = remoteExists ? this.remote.max_fee : undefined;
+  this._lastLedgerOffset = remoteExists ? this.remote.last_ledger_offset : 3;
   this.state = 'unsubmitted';
   this.finalized = false;
   this.previousSigningHash = undefined;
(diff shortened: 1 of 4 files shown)


src/core/transaction.js
```diff
@@ -1512,30 +1524,44 @@ Transaction.prototype.setSigners = function(signers) {
...
   assert(this.tx_json.Fee, 'Fee must be set before multi-signing');
 
-  const signingTx = Transaction.from_json(this.tx_json);
+  if (_.isUndefined(this.tx_json.LastLedgerSequence)) {
+    // Auto-fill LastLedgerSequence
+    this.setLastLedgerSequence();
+  }
+
+  const cleanedJson = _.omit(this.tx_json, [
+    'SigningPubKey',
```
sublimator: The empty SigningPubKey needs to be signed

src/core/transaction.js
```diff
@@ -1512,30 +1524,44 @@ Transaction.prototype.setSigners = function(signers) {
...
+    'Signers',
+    'TxnSignature'
+  ]);
+
+  const signingTx = Transaction.from_json(cleanedJson);
   signingTx.remote = this.remote;
   signingTx.setSigningPubKey('');
```
sublimator: Oh, it's set here huh

src/core/transaction.js
```diff
@@ -582,19 +583,30 @@ Transaction.prototype.clientID = function(id) {
...
-  this._setUInt32('LastLedgerSequence', sequence);
+Transaction.prototype.setLastLedgerSequence = function(sequence) {
+  if (!_.isUndefined(sequence)) {
+    this._setUInt32('LastLedgerSequence', sequence);
+  } else {
+    // Autofill LastLedgerSequence
+    assert(this.remote, 'Unable to set LastLedgerSequence, missing Remote');
+
+    this._setUInt32('LastLedgerSequence',
+                    this.remote.getLedgerSequence() + 1
```
sublimator: getLedgerSequence is .... ? the last validated ledger seen ?

scottschurr:
@wltsmrz It looks like `submit_multisign` is working for you with the `Signers` moved inside the `tx_json`.  Is that a go for me to submit a pull request for that change in RippleD?

wltsmrz:
Yep, looks good

src/core/transaction.js
wltsmrz: Yeah

src/core/transaction.js
wltsmrz: Indeed

wltsmrz:
Adding a note that this will probably break if the following pull request is mergedhttps://github.com/ripple/ripple-lib/pull/546

PR #553, opened 2015-09-17
Merge OrderBook optimizations from master to develop

darkdarkdragon:


src/core/orderbook.js
```diff
@@ -1203,6 +1355,33 @@ OrderBook.prototype.computeAutobridgedOffers = function() {
...
+  if (newMult !== this._autobridgeThrottleTimeMultiplier) {
+    this._autobridgeThrottleTimeMultiplier = newMult;
+    this.createDebouncedOffersWrapper();
+  }
+};
+
+OrderBook.prototype.createDebouncedOffersWrapper = function() {
+  const m = this._autobridgeThrottleTimeMultiplier;
+  this.computeAutobridgedOffersThrottled =
+    _.debounce(
```
clark800: What's the plan regarding buffering as opposed to debounce/throttle? I don't know if there's a good reason to merge this if we're planning on removing it immediately.

clark800: Oh, I guess this is so that we can merge develop into master without rewriting history because this already got into master.

darkdarkdragon: Yes. And I'm already started to work on merge, so there wasn't any sense to stop, anyway develop branch got bunch of own optimizations - Amount and OrderBookUtils was significantly changed, so I've did optimization on them again.

src/core/amount.js
```diff
@@ -105,6 +105,19 @@ Amount.NaN = function() {
   return result;                      // but let's be careful
 };
 
+Amount.createFast = function(value: Value, currency: Currency, issuer: UInt160,
+  isNative: boolean
+) {
+  const res = new Amount(value);
```
clark800: i think we should try to avoid abbreviations in most cases

src/core/amount.js
```diff
@@ -105,6 +105,19 @@ Amount.NaN = function() {
   return result;                      // but let's be careful
 };
 
+Amount.createFast = function(value: Value, currency: Currency, issuer: UInt160,
```
clark800: to make the name more consistent with other names, maybe we could call it `from_components`?

src/core/amount.js
```diff
@@ -743,8 +762,9 @@ Amount.prototype.to_text = function() {
   // not native
   const offset = this._value.getExponent() - 15;
   const sign = this._value.isNegative() ? '-' : '';
-  const mantissa = utils.getMantissa16FromString(
-    this._value.abs().toString());
+  const mantissa = this._value.isNegative() ?
```
clark800: Is this faster? It seems like it wouldn't be worth it.

clark800:
LGTM as long as we address all the feedback right after

src/core/amount.js
darkdarkdragon: Faster. `abs` trigger creation of `BigNumber` object, and `BigNumber` constructor is slowest single part of ther code, so most of my optimizations went into removing unnecessary calls to it constructor.

clark800: So consider that if I come back to this in a few months, I might refactor it back to the way it was before. I think it's worth making a helper function called `abs` to do this and giving it a comment explaining that the BigNumber constructor is very slow.

src/core/amount.js
darkdarkdragon: I wanted to emphasis somehow that it is not regular function that can be used anywhere because I'm skipping call to `_check_limits` here.
Maybe `from_components_unsafe`?

clark800: sure

darkdarkdragon:
done

PR #554, opened 2015-09-17
Update es6-promisify to 3.0.0

lumberj:
https://github.com/digitaldesignlabs/es6-promisify/releases/tag/v3.0.0


clark800:
Is there a specific reason why we need this update?

lumberj:
It updates es6-promisify to use native promises instead of a shim

clark800:
It looks like both versions 2 and 3 use native when available and polyfill otherwise.

PR #557, opened 2015-09-18
Add ledgerClosed event

clark800:


src/api/index.js
```diff
@@ -34,11 +36,17 @@ const getLedger = require('./ledger/ledger');
 
 function RippleAPI(options: {}) {
   common.validate.remoteOptions(options);
+  EventEmitter.call(this);
   const _options = _.assign({}, options, {automatic_resubmission: false});
   this.remote = new common.core.Remote(_options);
+  this.remote.on('ledger_closed', message => {
+    this.emit('ledgerClosed', message);
```
lumberj: Maybe RippleAPI should format the message to be consistent with camelcasing convention https://github.com/ripple/ripple-lib/blob/develop/test/fixtures/api/rippled/ledger-close.json

lumberj: Also, rename variables, e.g., ledger_index becomes ledgerVersion

clark800: Yeah, good idea

clark800:
@lumberj updated

src/api/common/schemas/ledger-closed.json
```diff
@@ -0,0 +1,21 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "ledgerClosed",
+  "description": "A ledgerClosed event message",
+  "type": "object",
+  "properties": {
+    "feeBase": {"type": "integer", "minimum": 0},
+    "feeReference": {"type": "integer", "minimum": 0},
+    "ledgerHash": {"$ref": "ledgerHash"},
+    "ledgerVersion": {"$ref": "ledgerVersion"},
+    "ledgerTimestamp": {"type": "integer", "minimum": 0},
```
lumberj: Is this a Unix timestamp or Ripple time? https://wiki.ripple.com/JSON_format#time

lumberj: Would it make sense to convert this to unix time or ISO8601 utc time?

clark800: This is straight from rippled, so I'm guessing it's a ripple time.

Hmmm, I'm not sure about converting. By default I guess I would leave it.

lumberj: If we use iso8601, then we can also use, http://json-schema.org/latest/json-schema-validation.html#anchor108

clark800: It would be more user-friendly, think we should convert it?

lumberj: I think so, otherwise we should think about exposing a convenience method (maybe via ripple-lib-extensions) for working with [Ripple Epoch time](https://ripple.com/build/rippled-apis/#specifying-time)

wltsmrz: SGTM

clark800: Will do

clark800:
@lumberj updated

lumberj:
LGTM

PR #558, opened 2015-09-22
Add pattern for servers option

clark800:


wltsmrz:
Could also check port, at least. Something else that people might forget

wltsmrz:
Or does "format": "uri" take care of that?

clark800:
The port is optional now

wltsmrz:
LGTM

PR #560, opened 2015-09-22
Always call Remote.connect() callback

wltsmrz:


src/core/remote.js
```diff
@@ -421,22 +421,23 @@ Remote.prototype.reconnect = function() {
 /**
  * Connect to the Ripple network
  *
- * @param {Function} [callback]
+ * @param [Function] [callback]
```
lumberj: Changed by mistake?

lumberj:
LGTM

clark800:
Might be more clear if the commit message was "Fix: RippleAPI.connect() does not resolve if already connected"

src/core/remote.js
wltsmrz: Callback is optional

wltsmrz:
I'll add a note in the commit @clark800

PR #561, opened 2015-09-23
more OrderBook performance optimizations

darkdarkdragon:
OrderBook: emit 'model' event only after last transaction in closed ledger
run AutobridgeCalculator only once in a ledger

### DIFF
--- a/src/autobridgecalculator.js
+++ b/src/autobridgecalculator.js
@@ -2,7 +2,6 @@
 
 const _ = require('lodash');
 const assert = require('assert');
-const UInt160 = require('./uint160').UInt160;
 const Amount = require('./amount').Amount;
 const Utils = require('./orderbookutils');
 
@@ -38,15 +37,39 @@ AutobridgeCalculator.NULL_AMOUNT = Utils.normalizeAmount('0');
  * @return {Array}
  */
 
-AutobridgeCalculator.prototype.calculate = function() {
-  let legOnePointer = 0;
-  let legTwoPointer = 0;
+AutobridgeCalculator.prototype.calculate = function(callback) {
+  const legOnePointer = 0;
+  const legTwoPointer = 0;
 
   const offersAutobridged = [];
 
   this.clearOwnerFundsLeftover();
 
+  this._calculateInternal(legOnePointer, legTwoPointer, offersAutobridged,
+    callback);
+};
+
+AutobridgeCalculator.prototype._calculateInternal = function(
+  legOnePointer_, legTwoPointer_, offersAutobridged, callback
+) {
+
+  let legOnePointer = legOnePointer_;
+  let legTwoPointer = legTwoPointer_;
+
+  const startTime = Date.now();
+
   while (this.legOneOffers[legOnePointer] && this.legTwoOffers[legTwoPointer]) {
+    // manually implement cooperative multitasking that yields after 30ms
+    // of execution so user's browser stays responsive
+    const lasted = (Date.now() - startTime);
+    if (lasted > 30) {
+      setTimeout(() => {
+        this._calculateInternal(legOnePointer, legTwoPointer, offersAutobridged,
+          callback);
+      }, 0);
+      return;
+    }
+
     const legOneOffer = this.legOneOffers[legOnePointer];
     const legTwoOffer = this.legTwoOffers[legTwoPointer];
     const leftoverFunds = this.getLeftoverOwnerFunds(legOneOffer.Account);
@@ -103,7 +126,7 @@ AutobridgeCalculator.prototype.calculate = function() {
     offersAutobridged.push(autobridgedOffer);
   }
 
-  return offersAutobridged;
+  callback(offersAutobridged);
 };
 
 /**
@@ -308,7 +331,8 @@ function(takerGets, takerPays) {
 
   autobridgedOffer.autobridged = true;
 
-  autobridgedOffer.BookDirectory = Utils.convertOfferQualityToHexFromText(autobridgedOffer.quality);
+  autobridgedOffer.BookDirectory =
+    Utils.convertOfferQualityToHexFromText(autobridgedOffer.quality);
   autobridgedOffer.qualityHex = autobridgedOffer.BookDirectory;
 
   return autobridgedOffer;
@@ -432,7 +456,7 @@ function(legOneOffer, takerGets) {
   const legOneQuality = Utils.getOfferQuality(legOneOffer, this._currencyGets);
 
   legOneOffer.TakerGets = takerGets.to_text();
-  legOneOffer.TakerPays = takerGets.multiply(legOneQuality);
+  legOneOffer.TakerPays = takerGets.multiply(legOneQuality).to_json();
 };
 
 module.exports = AutobridgeCalculator;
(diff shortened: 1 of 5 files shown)


src/autobridgecalculator.js
```diff
@@ -38,7 +38,7 @@ AutobridgeCalculator.NULL_AMOUNT = Utils.normalizeAmount('0');
  * @return {Array}
  */
 
-AutobridgeCalculator.prototype.calculate = function() {
+AutobridgeCalculator.prototype.calculate = function(callback = function() {}) {
```
clark800: Why is the callback optional? Is there a way to get the results without using the callback?

darkdarkdragon: Good point :smile:

src/orderbook.js
```diff
@@ -828,6 +828,32 @@ OrderBook.prototype.isBalanceChangeNode = function(node) {
...
+OrderBook.prototype._canRunAutobridgeCalc = function(): boolean {
+  return !this._calculatorRunning && (Date.now() - this._lastCalcDuration >= 100);
+}
+
+OrderBook.prototype.onTransaction = function(transaction) {
+  this.updateFundedAmounts(transaction);
+
+
+  if (--this._transactionsLeft === 0 && !this._waitingForOffers) {
+    const lastClosedLedger = this._remote.getLedgerSequence() - 1;
```
clark800: Isn't the lastClosedLedger returned by getLedgerSequence now?

src/autobridgecalculator.js
```diff
@@ -46,7 +46,27 @@ AutobridgeCalculator.prototype.calculate = function() {
...
+};
+
+AutobridgeCalculator.prototype._calculateInternal = function(
+  legOnePointer, legTwoPointer, offersAutobridged, callback = function() {}
+) {
+
+  const startTime = Date.now();
+
   while (this.legOneOffers[legOnePointer] && this.legTwoOffers[legTwoPointer]) {
+    const lasted = (Date.now() - startTime);
```
clark800: Why do we need to care about the time? Also, I don't understand the purpose of this block of code, can you explain more?

src/orderbook.js
```diff
@@ -125,8 +128,12 @@ function OrderBook(remote,
     });
   }
 
-  function updateFundedAmountsWrapper(transaction) {
-    self.updateFundedAmounts(transaction);
+  function onTransactionWrapper(transaction) {
```
clark800: Could use arrow functions or just `bind` for this

src/orderbook.js
```diff
@@ -828,6 +828,32 @@ OrderBook.prototype.isBalanceChangeNode = function(node) {
   return true;
 };
 
+OrderBook.prototype._canRunAutobridgeCalc = function(): boolean {
+  return !this._calculatorRunning && (Date.now() - this._lastCalcDuration >= 100);
```
clark800: Seems very strange to have arbitrary numbers like this. Shouldn't the lock be sufficient?

src/orderbook.js
darkdarkdragon: In master branch it just returns `_ledger_current_index`

clark800: I didn't notice this is going to master. Why isn't this for develop?

src/autobridgecalculator.js
darkdarkdragon: Purpose is to split processing into chunks, so users browser can do something other than calculate offers. I'm using time instead of just splitting into predefined number of steps because user's device can have different speed.
Also, I've set timeout to 33ms so it slightly bigger than processing time - 30 - because RT client has two book open simultaneously, so two AutobridgeCalculator can be run at the same time, so I'm trying to give them change to work together  and not to burn users cpu.

src/orderbook.js
darkdarkdragon: because RT using master and I need this to be used in RT

src/autobridgecalculator.js
clark800: Ok, I think you should add a comment like "manually implement cooperative multitasking that yields after 30ms of execution and restarts after 33ms". Also, wouldn't it still work if it was `setTimeout(func, 0)`? With the 33ms delay it will be wasting a lot of time if you're only running with one orderbook.

darkdarkdragon: You're right, `0` will do.

src/orderbook.js
darkdarkdragon: If I use `bind` I will need to introduce new variable to keep `bind` result because I need to unsubscribe that function from event. So I've choose separate function as more descriptive.

src/orderbook.js
darkdarkdragon: >  Shouldn't the lock be sufficient?

I don't know, I haven't tested these extreme conditions. I wrote this in hope that it will give user's browser some chances in case that device is slow and load is high. It won't matter in normal conditions.

darkdarkdragon:
@clark800 I've made some refactoring

src/orderbook.js
clark800: I mean instead of making the wrapper function you can do `self._remote.on('transaction', self.onTransaction.bind(self));`

darkdarkdragon: in this case I must save result of `self.onTransaction.bind(self)` into var, because I need to do `self._remote.removeListener('transaction', onTransactionWrapper);`

src/orderbook.js
clark800: The browser should have a change to repaint and handle user interaction every time you call the `setTimeout(fn, 0)`, so I don't think we need to worry about it. Also, I think the logic is incorrect because it is comparing an absolute time with a duration.

src/orderbook.js
clark800: Oh ok, I see now

src/orderbook.js
darkdarkdragon: You're right, logic is incorrect. 
OK, I'll throw this out, this was just a precaution.

src/autobridgecalculator.js
```diff
@@ -38,15 +37,39 @@ AutobridgeCalculator.NULL_AMOUNT = Utils.normalizeAmount('0');
...
+  let legTwoPointer = legTwoPointer_;
+
+  const startTime = Date.now();
+
   while (this.legOneOffers[legOnePointer] && this.legTwoOffers[legTwoPointer]) {
+    // manually implement cooperative multitasking that yields after 30ms
+    // of execution so user's browser stays responsive
+    const lasted = (Date.now() - startTime);
+    if (lasted > 30) {
+      setTimeout(() => {
```
wltsmrz: Why not use `async.setImmediate()` ?

wltsmrz: https://github.com/caolan/async/blob/309eb0857a2083ef96f0f44157109b16726fd02f/lib/async.js#L195-L210

wltsmrz: Possibly gooder? https://github.com/YuzuJS/setImmediate

darkdarkdragon: I don't think it worth it.

darkdarkdragon:
@clark800 I'm removed that duration calculation/check

clark800:
LGTM

PR #563, opened 2015-09-23
Move to new ripple-keypairs API

clark800:
Breaking changes: remove Seed class, all secrets must be specified as base58 strings (passphrases and hex no longer supported)


wltsmrz:
LGTM

PR #566, opened 2015-09-25
Decouple UInt160 from account.js and amount.js

clark800:


src/core/amount.js
```diff
@@ -21,7 +22,7 @@ function Amount(value = new XRPValue(NaN)) {
   this._value = value;
   this._is_native = true; // Default to XRP. Only valid if value is not NaN.
   this._currency = new Currency();
-  this._issuer = new UInt160();
+  this._issuer = 'NaN';
```
lumberj: Why `'NaN'` string, and not `NaN`

clark800: Because it's a string field. I just used the string 'NaN' because that's what was required to not break the unit tests.

src/core/account.js
```diff
@@ -43,7 +42,7 @@ function Account(remote, account) {
     if (_.includes(Account.subscribeEvents, type)) {
       if (!self._subs && self._remote._connected) {
         self._remote.requestSubscribe()
-        .addAccount(self._account_id)
+        .addAccount(self._address)
         .broadcast().request();
```
wltsmrz: Not necessarily part of this pull request but we ought to have error handlers for these requests

clark800: Ok, can you add a JIRA task so we don't forget?

src/core/account.js
```diff
@@ -66,8 +65,8 @@ function Account(remote, account) {
   this.on('removeListener', listenerRemoved);
 
   function attachAccount(request) {
-    if (self._account.is_valid() && self._subs) {
-      request.addAccount(self._account_id);
+    if (isValidAddress(self._address) && self._subs) {
```
wltsmrz: Maybe we should just assert that it's valid in the constructor?

clark800: Might be a good idea, but not in this pull request because we should avoid/minimize behavioral changes in refactoring pull requests.

wltsmrz:
LGTM

src/core/account.js
wltsmrz: Sure

PR #569, opened 2015-09-27
Handle case when request was re-submitted due to reconnection

darkdarkdragon:


darkdarkdragon:
This one was discussed [here](https://github.com/ripple/ripple-lib/pull/543#issuecomment-140215334)

src/core/server.js
```diff
@@ -676,6 +676,17 @@ Server.prototype._handleResponse = function(message) {
     const result = message.result;
     const responseEvent = 'response_' + command;
 
+    if (command === 'submit' && result.engine_result === 'tefPAST_SEQ' &&
+        request.reconnectionsCount > 0
+    ) {
+      // handle case when request reached the server, but response was lost
+      // and request was automatically repeated
+      result.engine_result = 'tesSUCCESS';
```
wltsmrz: Although it kind of makes sense, this smells funny. I'm thinking we should undo the automatic re-request thing, make sure `Request.timeout()` works, add check for server has ledger range [here](https://github.com/ripple/ripple-lib/blob/87fdbc932f1369b8ce73478ddc9697d89d26799b/src/core/remote.js#L1312-L1318), let transactionmanager's reconnect logic do its thing, and defer other auto-re-requesting responsibility to clients

clark800: What do you think of having a default timeout and making request.js return a `tejTimeout` error?

src/core/server.js
```diff
@@ -676,6 +676,17 @@ Server.prototype._handleResponse = function(message) {
     const result = message.result;
     const responseEvent = 'response_' + command;
 
+    if (command === 'submit' && result.engine_result === 'tefPAST_SEQ' &&
```
clark800: Why not `tefALREADY`? Isn't it unsafe to convert to success for a `tefPAST_SEQ`?

src/core/server.js
wltsmrz: Sounds reasonable

src/core/server.js
darkdarkdragon: I didn't got `tefALREADY`. When I'm resubmitting request I'm getting `tefPAST_SEQ` from server.

clark800: So we don't know if the this transaction already got validated because it could be any new transaction at this point right?

darkdarkdragon: Yes, that's the problem here. So I think that @wltsmrz is right, we need to drop re-request logic and make timeout for requests.

darkdarkdragon:
I've made new pr https://github.com/ripple/ripple-lib/pull/577

FrRichard:
is it related to https://github.com/ripple/ripple-lib/issues/573 ?

PR #570, opened 2015-09-28
make sure that 'before' event emitted from Request only once

darkdarkdragon:


wltsmrz:
LGTM

src/core/request.js
```diff
@@ -44,13 +44,15 @@ Request.prototype.request = function(servers, callback_) {
   const callback = typeof servers === 'function' ? servers : callback_;
 
   this.emit('before');
+
+  const wasRequested = this.requested;
```
clark800: Is there a reason we have to call `callback` even if this was already requested? Could we just throw an exception if request gets called twice?

darkdarkdragon: Maybe user wants another callback added? From reading previous code I've suggested that logic.

clark800: Seems like unnecessary complexity unless there is a good use case, in which case we should add a comment to explain.

darkdarkdragon: I've made changes that throws error in case of request called twice
https://github.com/ripple/ripple-lib/pull/572

PR #574, opened 2015-10-01
Fix bug in Amount.ratio_human

clark800:


MatthewPhinney:
LGTM

wltsmrz:
What is the explanation

clark800:
It was passing an IOUValue to Amount.multiply, which expects either an Amount or JSON.

wltsmrz:
LGTM

PR #576, opened 2015-10-02
allow offers request on orderbook subscribe when disconnected

shekenahglory:
When the remote is offline and an orderbook is created with XRP as the `gets` currency, the orderbook never gets properly subscribed because requestOffers doesn't allow a request to go to the queue when remote is disconnected.  Its not a problem for non-native currencies because requestTransferRate waits for the remote to connect.

### DIFF
--- a/src/core/orderbook.js
+++ b/src/core/orderbook.js
@@ -23,6 +23,7 @@ const AutobridgeCalculator = require('./autobridgecalculator');
 const OrderBookUtils = require('./orderbookutils');
 const log = require('./log').internal.sub('orderbook');
 const {IOUValue} = require('ripple-lib-value');
+const RippleError = require('./rippleerror').RippleError;
 
 function _sortOffers(a, b) {
   const aQuality = OrderBookUtils.getOfferQuality(a, this._currencyGets);
@@ -256,6 +257,9 @@ OrderBook.offerRewrite = function(offer) {
 /**
  * Initialize orderbook. Get orderbook offers and subscribe to transactions
  * @api private
+ * NOTE: this method is not meant to be publicly used
+ * and it does not work for autobridged books since
+ * it does not add listeners for them
  */
 
 OrderBook.prototype.subscribe = function() {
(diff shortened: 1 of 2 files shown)


FrRichard:
is it related to https://github.com/ripple/ripple-lib/issues/573 ?

shekenahglory:
Seems likely - if you are subscribing before connecting

FrRichard:
hmm this is weird because I always remote.connect() before .. I'll try with those changes and see

FrRichard:
this bug didn't appear with 0.12.6

shekenahglory:
remote.connect() is async, so it may be that you initiated the connection but hasnt been completed yet - unless you are specifically waiting until you get a connection response

FrRichard:
got it thanks

clark800:
Is it possible to add a unit test to prevent regressions? Could you add a comment explaining what `internal` is and how it is used?

shekenahglory:
Yeah I can add that, I'm not entirely sure that the bit about ignoring requests while its not connected is ideal though...it doesn't seem like expected behavior does it?

shekenahglory:
It seems like we would want to ignore subsequent requests if one is already queued up, not ignore all

clark800:
I think the Request class will call this method again when it reconnects, until it gets a success or error response or times out.

shekenahglory:
As it is now, if requestOffers is called when its disconnected, a request object never gets created - it just returns undefined without even executing the callback

shekenahglory:
Actually isnt this always an internal call and therefor a private method? there is another method `getOffers` that is public

shekenahglory:
@darkdarkdragon I also just discovered that autobridged books don't load because `syncronized` is never set to `true`: https://github.com/ripple/ripple-lib/blob/develop/src/core/orderbook.js#L358 I thinking this supposed to be `synced` here and under `computeAutobridgedOffersWrapper` also?

darkdarkdragon:
@shekenahglory You're right, it must be `_synced`. Missed that one on merge :confused:

src/core/orderbook.js
```diff
@@ -333,16 +336,18 @@ OrderBook.prototype.destroy = function() {
...
 OrderBook.prototype.requestOffers = function(callback = function() {},
   internal = false) {
   const self = this;
 
-  if (!this._remote.isConnected()) {
+  if (!this._remote.isConnected() && !internal) {
     // do not make request if not online.
     // that requests will be queued and
     // eventually all of them will fire back
+    callback(new Error('remote is offline'));
```
clark800: All errors passed to callbacks should be instances of RippleError (which means they can inherit from RippleError).

clark800:
LGTM after squash

shekenahglory:
Rebased and squashed, but now I get seemingly unrelated test failures?

`ReferenceError: UInt160 is not defined
      at Remote.requestGatewayBalances (/home/ubuntu/ripple-lib/src/core/remote.js:1:14)`

clark800:
Yes, that was a merge failure that happened yesterday. I'll fix it soon, thanks

clark800:
I fixed the bug and tried rebuilding on CircleCI, but it failed again. Can you try rebasing?

PR #577, opened 2015-10-02
remove Request resubmit logic

darkdarkdragon:
add default timeout to Request

### DIFF
--- a/src/core/request.js
+++ b/src/core/request.js
@@ -34,14 +34,15 @@ function Request(remote, command) {
     command: command,
     id: undefined
   };
+  this._timeout = this.remote.submission_timeout;
 }
 
 util.inherits(Request, EventEmitter);
 
 // Send the request to a remote.
 Request.prototype.request = function(servers, callback_) {
-  const self = this;
   const callback = typeof servers === 'function' ? servers : callback_;
+  const self = this;
 
   if (this.requested) {
     throw new Error('Already requested');
@@ -70,17 +71,25 @@ Request.prototype.request = function(servers, callback_) {
     }
   }
 
-  function onReconnect() {
-    doRequest();
-  }
+  const timeout = setTimeout(() => {
+    if (typeof callback === 'function') {
+      callback(new RippleError('tejTimeout'));
+    }
+
+    this.emit('timeout');
+    // just in case
+    this.emit = _.noop;
+    this.cancel();
+  }, this._timeout);
 
   function onResponse() {
-    self.remote.removeListener('connected', onReconnect);
+    clearTimeout(timeout);
   }
 
   if (this.remote.isConnected()) {
-    this.remote.on('connected', onReconnect);
+    this.remote.on('connected', doRequest);
   }
+
   this.once('response', onResponse);
 
   doRequest();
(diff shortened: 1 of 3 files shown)


src/core/request.js
```diff
@@ -59,31 +59,31 @@ Request.prototype.request = function(servers, callback_) {
...
-    if (Array.isArray(servers)) {
-      servers.forEach(function(server) {
-        self.setServer(server);
-        self.remote.request(self);
-      }, self);
-    } else {
-      self.remote.request(self);
+  const timeout = setTimeout(() => {
+    if (typeof callback === 'function') {
+      callback(new RippleError('tejTimeout'));
```
clark800: Do you need a `return` here? Otherwise it looks like it will emit `timeout` and if no listener is registered it will raise an uncaught exception right?

src/core/request.js
```diff
@@ -59,31 +59,31 @@ Request.prototype.request = function(servers, callback_) {
...
     }
-  }
 
-  function onReconnect() {
```
clark800: So I think there's no need to remove the reconnect logic for non-transaction requests; it would be convenient to keep it for read-only requests.

clark800: Actually, I'm not really convinced that we have to remove it even for transaction requests. We already interpret "tef" errors as "maybe succeeded", so it might be easiest to just leave it as it was. But I still think it would be good to keep the default timeout.

darkdarkdragon: So I'm putting back reconnect logic, right?

clark800: Yeah, I think so

src/core/request.js
darkdarkdragon: Why it would raise exception?

darkdarkdragon:
@clark800 I've made changes

src/core/request.js
clark800: I was thinking of `error` events: "Error events are treated as a special case in Node.js. If there is no listener for it, then the default action is to print a stack trace and exit the program." So I guess it isn't a problem because it is a `timeout` event, not an `error` event.

src/core/request.js
```diff
@@ -264,40 +273,11 @@ Request.prototype.callback = function(callback, successEvent, errorEvent) {
...
-  const self = this;
-
-  function requested() {
-    self.timeout(duration, callback);
-  }
-
-  if (!this.requested) {
-    // Defer until requested
-    return this.once('request', requested);
+Request.prototype.setSubmissionTimeout = function(delay) {
```
clark800: How about just `setTimeout`?

src/core/request.js
```diff
@@ -34,14 +34,15 @@ function Request(remote, command) {
     command: command,
     id: undefined
   };
+  this._submissionTimeout = this.remote.submission_timeout;
```
clark800: "submission" makes it sound like it might only apply to "submit" for transactions, whereas this also applies for read-only requests. Maybe we should remove "submission"?

src/core/request.js
```diff
@@ -264,40 +273,11 @@ Request.prototype.callback = function(callback, successEvent, errorEvent) {
   return this;
 };
 
-Request.prototype.timeout = function(duration, callback) {
```
clark800: Since this is a breaking change, can you add a not in the commit message that says it contains a breaking change?

src/core/request.js
darkdarkdragon: You're right, I'll remove it

src/core/request.js
darkdarkdragon: OK

src/core/request.js
darkdarkdragon: OK

src/core/request.js
darkdarkdragon: Thanks for the info! Didn't know that.

darkdarkdragon:
@clark800 done

clark800:
LGTM

src/core/request.js
```diff
@@ -70,17 +71,25 @@ Request.prototype.request = function(servers, callback_) {
     }
   }
 
-  function onReconnect() {
-    doRequest();
-  }
+  const timeout = setTimeout(() => {
+    if (typeof callback === 'function') {
+      callback(new RippleError('tejTimeout'));
```
wltsmrz: `tej` is supposed to refer to a local transaction result, not a generic result for all types of requests, but...

PR #579, opened 2015-10-03
Balance sheet

wltsmrz:


src/api/common/schemas/get-balance-sheet.json
```diff
@@ -0,0 +1,67 @@
...
+        "properties": {
+          "required": ["counterparty", "balances"],
+          "additionalProperties": false,
+          "counterparty": {"$ref": "address"},
+          "balances": {
+            "type": "array",
+            "items": {
+              "type": "object",
+              "properties": {
+                "required": ["currency", "value"],
```
clark800: `required` and `additionalProperties` should be outside the `properties` object

test/api-test.js
```diff
@@ -202,6 +201,11 @@ describe('RippleAPI', function() {
       _.partial(checkResult, responses.getBalances, 'getBalances'));
   });
 
+  it('getBalanceSheet', function() {
+    return this.api.getBalanceSheet(address).then(
+      _.partial(checkResult, responses.getBalanceSheet, undefined));
```
clark800: Can hook up the response schema here by replacing `undefined` with the schema name. Then it will check that the response always matches the expected schema, testing the code against the schema and the schema against the code.

src/api/ledger/balance-sheet.js
```diff
@@ -0,0 +1,69 @@
...
+  if (!_.isUndefined(assets)) {
+    result.assets = Object.keys(assets).map((k) => {
+      return {
+        counterparty: k,
+        assets: assets[k]
+      };
+    });
+  }
+  if (!_.isUndefined(obligations)) {
+    result.obligations = Object.keys(obligations).map((k) => {
```
clark800: lodash map gives you both value and key, so you could do `result.obligations = _.map(obligations, (value, currency) => {currency, value})`, and similar for balances and assets

test/fixtures/api/responses/get-balance-sheet.json
```diff
@@ -0,0 +1,68 @@
+{
+  "balances": [
+    {
+      "counterparty": "rKm4uWpg9tfwbVSeATv4KxDe6mpE9yPkgJ",
+      "balances": [
```
clark800: How about making `balances` and object keyed by the counterparty address to avoid the redundant `balances` (and similar for assets)?

clark800: Another option would be to merge the `counterparty` into the amounts so that `balances` is just a list of amounts.

wltsmrz: I don't know how I'd write a schema for your first solution. I could make it a large one-dimensional array of amounts (second suggestion) but the current format would seem preferable

test/fixtures/api/rippled/index.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
   },
   account_offers: require('./account-offers'),
   account_tx: require('./account-tx'),
+  balance_sheet: require('./balance-sheet'),
```
clark800: the naming here is supposed to match rippled's names

test/fixtures/api/responses/get-balance-sheet.json
clark800: You can use `patternProperties` instead of `properties`. I'm kind of on the fence about which option is better; it's nice to have the explicit notation that the address is the counterparty in the second option, but it's also nice to have the lower redundancy in the first option. If you like the first option better, then we can go with that.

src/api/common/schemas/get-balance-sheet.json
wltsmrz: Done

test/api-test.js
wltsmrz: Done

src/api/ledger/balance-sheet.js
wltsmrz: Done

test/fixtures/api/rippled/index.js
wltsmrz: Done

test/fixtures/api/responses/get-balance-sheet.json
wltsmrz: patternProperties will work with `required`? Seems simpler to write schema for non-variable property names

wltsmrz: If it's between ripple address properties and one-dimensional array of amounts I choose the latter, but I see nothing wrong with the current format

clark800: We should avoid naming that suggests that something is a non-whole component of itself, which is a logical contradiction (balances = balances + counterparty). A one-dimensional array of amounts sounds good to me.

wltsmrz: Done

src/api/common/schemas/get-balance-sheet.json
```diff
@@ -0,0 +1,47 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "getBalanceSheet",
+  "description": "getBalanceSheet response",
+  "type": "object",
+  "properties": {
+    "balances": {
+      "type": "array",
+      "items": {
+        "type": "object",
```
clark800: Probably better to just do `{"$ref": "amount"}` here.

src/api/common/schemas/get-balance-sheet.json
```diff
@@ -0,0 +1,47 @@
...
+          "counterparty": {"$ref": "address"},
+          "currency": {"$ref": "currency"},
+          "value": {"$ref": "value"}
+        }
+      }
+    },
+    "assets": {
+      "type": "array",
+      "items": {
+        "type": "object",
```
clark800: And here

wltsmrz: Done

clark800:
LGTM

clark800:
Oops, forgot squash, oh well

wltsmrz:
lol

PR #582, opened 2015-10-06
add more flow annotations to RippleApi

darkdarkdragon:


src/api/common/utils.js
```diff
@@ -17,7 +17,7 @@ function xrpToDrops(xrp: string): string {
   return (new BigNumber(xrp)).times(1000000.0).floor().toString();
 }
 
-function toRippledAmount(amount: Amount): string|Amount {
+function toRippledAmount(amount: AmountRippled): string|AmountRippled {
```
sublimator: This doesn't seem right. It's taking an `AmountRippled` as input ? I believe this function takes a RippleAPI amount and converts it to the json form as spoken by rippled.

darkdarkdragon: You're right. My mistake.

darkdarkdragon:
this is errors in `server-test.js` (not related to this pr)

src/api/ledger/pathfind.js
```diff
@@ -77,13 +71,15 @@ function addDirectXrpPath(paths, xrpBalance) {
...
-function isRippledIOUAmount(amount) {
+function isRippledIOUAmount(amount: Amount) {
   // rippled XRP amounts are specified as decimal strings
   return (typeof amount === 'object') &&
     amount.currency && (amount.currency !== 'XRP');
 }
 
-function conditionallyAddDirectXRPPath(remote, address, paths, callback) {
+function conditionallyAddDirectXRPPath(remote: Remote, address: string,
+  paths, callback
```
clark800: Is there supposed to be an annotation on `paths`?

src/api/common/types.js
```diff
@@ -0,0 +1,52 @@
+/* @flow */
+'use strict';
+
+export type AmountRippled = {
```
clark800: Could we call it `RippledAmount`?

src/api/common/types.js
```diff
@@ -0,0 +1,52 @@
...
+
+// Amount where counterparty and value are optional
+export type LaxLaxAmount = {
+  currency: string,
+  value?: string,
+  counterparty?: string
+}
+
+// A currency-counterparty pair, or just currency if it's XRP
+export type Issue = LaxLaxAmount;
```
clark800: An `Issue` shouldn't allow `value`

src/api/ledger/utils.js
```diff
@@ -48,21 +58,20 @@ function getRecursive(getter: Getter, limit?: number, callback: Callback) {
...
   const issuer = amount.counterparty === undefined ?
-    amount.issuer : amount.counterparty;
+    (amount.issuer !== undefined ? amount.issuer : undefined) :
+    amount.counterparty;
   const withIssuer = _.assign({}, amount, {issuer: issuer});
   return _.omit(withIssuer, 'counterparty');
 }
 
-type Order = {taker_gets: Amount, taker_pays: Amount}
+type BOrder = {taker_gets: Issue, taker_pays: Issue}
```
clark800: I think we should avoid abbreviations. I don't know what "B" means.

src/api/transaction/payment.js
```diff
@@ -6,19 +6,45 @@ const validate = utils.common.validate;
 const toRippledAmount = utils.common.toRippledAmount;
 const Transaction = utils.common.core.Transaction;
 const ValidationError = utils.common.errors.ValidationError;
+// import type {Transaction} from '../../core/transaction';
```
clark800: Don't need this anymore?

darkdarkdragon:
@clark800 BTW, [here](https://github.com/darkdarkdragon/ripple-lib/blob/develop-flow-annotate-4/src/api/ledger/pathfind.js#L24), `destination_amount` that is returned from server gets overwritten with value from the function param (from the user). Is that ok?

src/api/transaction/utils.js
```diff
@@ -54,11 +58,13 @@ function prepareTransaction(transaction: any, remote: any, instructions: any,
 
   function prepareMaxLedgerVersion(callback_) {
     if (instructions.maxLedgerVersion !== undefined) {
-      txJSON.LastLedgerSequence = parseInt(instructions.maxLedgerVersion, 10);
+      txJSON.LastLedgerSequence =
+        typeof instructions.maxLedgerVersion === undefined ?
```
clark800: This is already in a branch where `instructions.maxLedgerVersion !== undefined`

clark800:
I think that's because the `ripple_path_find` command doesn't return the `destination_amount`, though `path_find` does. We used to use `ripple_path_find`, but now we're using `path_find`. I think the user and response values should be the same, but maybe the best option would be to add `destination_amount` only if it is not already in the response.

src/api/transaction/utils.js
darkdarkdragon: I added this to preserve old behavior - `parseInt` returns `NaN` if value isn't something like digit.
If it is not needed, I can just do assign here.

src/api/common/types.js
darkdarkdragon: maybe better to rename it to `RippledAmountIOU` and rename `SourceAmount` to `RippledAmount`

src/api/transaction/utils.js
clark800: I don't think it's possible, `typeof` always returns a string.

src/api/common/types.js
clark800: That sounds good

src/api/transaction/utils.js
darkdarkdragon: My error, it intended to be `typeof instructions.maxLedgerVersion === 'undefined'`
strange that flow didn't catch this :disappointed:

clark800: But that equality is guaranteed to always be false

darkdarkdragon: Yeah, it is me being stupid :disappointed:

darkdarkdragon:
@clark800 done

clark800:
LGTM

PR #584, opened 2015-10-06
Add Remote.closeCurrentPathFind function, so current pathfind can be …

darkdarkdragon:
…properly closed,

so new can be created without adding to queue

### DIFF
--- a/src/core/remote.js
+++ b/src/core/remote.js
@@ -1814,6 +1814,19 @@ Remote.prototype.findAccount = function(accountID) {
 };
 
 /**
+ * Closes current pathfind, if there is one.
+ * After that new pathfind can be created, without adding to queue.
+ *
+ * @return {void} -
+ */
+Remote.prototype.closeCurrentPathFind = function() {
+  if (this._cur_path_find !== null) {
+    this._cur_path_find.close();
+    this._cur_path_find = null;
+  }
+};
+
+/**
  * Create a pathfind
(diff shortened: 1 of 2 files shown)


clark800:
I think this should work. We should also throw an error if the user calls `createPathFind` without a callback with there is a `_cur_path_find`.

darkdarkdragon:
@clark800 done

src/core/remote.js
```diff
@@ -1822,6 +1835,10 @@ Remote.prototype.findAccount = function(accountID) {
  */
 Remote.prototype.createPathFind = function(options, callback) {
   if (this._cur_path_find !== null) {
+    if (callback === undefined) {
```
clark800: I just thought of maybe a better idea: I think we can just skip this whole queuing block if the `callback` is undefined. That would better preserve the old behavior.

darkdarkdragon: You mean is to create new `PathFind`, set it to `_cur_path_find` and return it?
The only problem with that, is what in `this._queued_path_finds` can be left hanging some requests.

clark800: Yeah, maybe that approach would get too complicated. I think we should change the error message though. "Only one streaming pathfind request at a time is supported"

darkdarkdragon:
@clark800 done

clark800:
LGTM

PR #587, opened 2015-10-08
Use ripple-binary-codec

clark800:


src/api/transaction/sign.js
```diff
@@ -17,20 +18,16 @@ const validate = utils.common.validate;
...
 function hashSerialization(serialized, prefix) {
-  return serialized.hash(prefix || HASH_TX_ID).to_hex();
-}
-
-function signingData(txJSON) {
-  return core.Transaction.from_json(txJSON).signingData().buffer;
+  const hexPrefix = prefix.toString(16).toUpperCase();
+  const toHash = hexPrefix + serialized;
+  return hashjs.sha512().update(new Buffer(toHash, 'hex')).digest('hex')
+    .toUpperCase().slice(0, 64);
```
geertweening: A sha512 in hex would be 128 characters long? And you're cutting it in half? A comment to this function would help with understanding

src/core/transaction.js
```diff
@@ -484,9 +479,10 @@ Transaction.prototype.hash = function(prefix_, asUINT256, serialized) {
     prefix = hashprefixes[prefix_];
   }
 
-  const hash = (serialized || this.serialize()).hash(prefix);
-
-  return asUINT256 ? hash : hash.to_hex();
+  const hexPrefix = prefix.toString(16).toUpperCase();
+  const toHash = hexPrefix + (serialized || this.serialize());
+  return hashjs.sha512().update(new Buffer(toHash, 'hex')).digest('hex')
```
geertweening: Same here, would appreciate some comments on what's going on exactly

geertweening:
Besides the two comments for clarification, LGTM

src/api/transaction/sign.js
clark800: Yeah, that's a ripple pattern: "For a hash function, rippled uses SHA-512 and then truncates the result to the first 256 bytes. This algorithm, informally called SHA-512Half, provides an output that has comparable security to SHA-256, but runs faster on 64-bit processors." https://ripple.com/build/ledger-format/

I'll factor out a the sha512hash function and add a comment based on this quote from the docs.

PR #588, opened 2015-10-09
fix Request.broadcast logic 

darkdarkdragon:
Add default timeout to Request.broadcas
Add default timeout to Remote.createPathFind
Handle 'slowDown' error in many places

### DIFF
--- a/src/core/remote.js
+++ b/src/core/remote.js
@@ -138,6 +138,9 @@ function Remote(options = {}) {
   if (typeof this.submission_timeout !== 'number') {
     throw new TypeError('submission_timeout must be a number');
   }
+  if (typeof this.pathfind_timeout !== 'number') {
+    throw new TypeError('pathfind_timeout must be a number');
+  }
   if (typeof this.automatic_resubmission !== 'boolean') {
     throw new TypeError('automatic_resubmission must be a boolean');
   }
@@ -197,6 +200,7 @@ Remote.DEFAULTS = {
   max_fee: 1000000, // 1 XRP
   max_attempts: 10,
   submission_timeout: 1000 * 20,
+  pathfind_timeout: 1000 * 10,
   automatic_resubmission: true,
   last_ledger_offset: 3,
   servers: [ ],
@@ -1852,8 +1856,15 @@ Remote.prototype.createPathFind = function(options, callback) {
   }
 
   if (callback) {
+    const updateTimeout = setTimeout(() => {
+      callback(new RippleError('tejTimeout'));
+      pathFind.close();
+      this._cur_path_find = null;
+    }, this.pathfind_timeout);
+
     pathFind.on('update', (data) => {
       if (data.full_reply && !data.closed) {
+        clearTimeout(updateTimeout);
         this._cur_path_find = null;
         callback(null, data);
         // "A client can only have one pathfinding request open at a time.
@@ -1869,7 +1880,10 @@ Remote.prototype.createPathFind = function(options, callback) {
         }
       }
     });
-    pathFind.on('error', callback);
+    pathFind.on('error', (error) => {
+      this._cur_path_find = null;
+      callback(error);
+    });
   }
 
   this._cur_path_find = pathFind;
--- a/src/core/request.js
+++ b/src/core/request.js
@@ -79,17 +79,23 @@ Request.prototype.request = function(servers, callback_) {
     // just in case
     this.emit = _.noop;
     this.cancel();
+    this.remote.removeListener('connected', doRequest);
   }, this._timeout);
-
-  function onResponse() {
-    clearTimeout(timeout);
-  }
 
   if (this.remote.isConnected()) {
     this.remote.on('connected', doRequest);
   }
 
-  this.once('response', onResponse);
+  function onRemoteError(error) {
+    self.emit('error', error);
+  }
+  this.remote.once('error', onRemoteError); // e.g. rate-limiting slowDown error
+
+  this.once('response', () => {
+    clearTimeout(timeout);
+    this.remote.removeListener('connected', doRequest);
+    this.remote.removeListener('error', onRemoteError);
+  });
 
   doRequest();
 
@@ -126,6 +132,7 @@ Request.prototype.broadcast = function(isResponseSuccess = isResponseNotError) {
     return this;
   }
 
+  this.on('error', function() {});
   let lastResponse = new Error('No servers available');
   const connectTimeouts = { };
   const emit = this.emit;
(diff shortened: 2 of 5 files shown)


src/core/server.js
```diff
@@ -800,7 +810,10 @@ Server.prototype._sendMessage = function(message) {
     if (this._remote.trace) {
       log.info(this.getServerID(), 'request:', message);
     }
-    this._ws.send(JSON.stringify(message));
+    this._ws.send(JSON.stringify(message), () => {
+      // sometimes gitves 'not opened'
```
clark800: typo

src/core/request.js
```diff
@@ -79,17 +79,21 @@ Request.prototype.request = function(servers, callback_) {
...
-    clearTimeout(timeout);
-  }
-
   if (this.remote.isConnected()) {
     this.remote.on('connected', doRequest);
   }
 
-  this.once('response', onResponse);
+  function onRemoteError() {}
+  this.remote.once('error', onRemoteError); // e.g. rate-limiting slowDown error
```
clark800: Is it safe to ignore all remote errors?

src/core/remote.js
```diff
@@ -1852,8 +1856,15 @@ Remote.prototype.createPathFind = function(options, callback) {
   }
 
   if (callback) {
+    const updateTimeout = setTimeout(() => {
```
clark800: Why does pathfinding need a special case? Is it possible to just pass the timeout error generated on the request level?

src/core/request.js
```diff
@@ -79,17 +79,21 @@ Request.prototype.request = function(servers, callback_) {
...
+
+  this.once('response', () => {
+    clearTimeout(timeout);
```
clark800: For streaming responses, could we reset the timeout rather than clearing it? Otherwise it seems that the stream could timeout.

src/core/remote.js
clark800: I guess because the request timeout is per-response and we could keep getting responses without getting a full_reply?

src/core/server.js
clark800: Shouldn't this generate an error? It looks like it's just silently ignoring the problem.

src/core/request.js
wltsmrz: What is this `slowDown

clark800: rippled has throttling, if you do too many requests too fast it will give you a slowDown message

src/core/server.js
darkdarkdragon: After error here server gets automatically reconnected.

src/core/remote.js
darkdarkdragon: Actually I've seen situations where server just stops sending updates - so we will wait indefinitely. I think this was the main reason of hangs that the Abiy saw.

src/core/request.js
darkdarkdragon: Sorry, didn't get it. Can you elaborate on that?

clark800: If a stream request gets a few responses, then stops sending updates, what will happen? It looks like the first response will clear the timeout and we will never get a timeout error.

src/core/server.js
clark800: How does that work? Is it a feature of the `ws` package? Does it automatically resend the message?

darkdarkdragon: I think what happens - is that `ws` internal state is updated (after connection dropped), but `Server's` isn't, so `Server` tries to send message, but `ws` gives error. But then `ws` emit `onerror` (or `onclose`, don't remeber) event, and `Server` standard reconnection starts.

Message gets lost. Do you think we should re-send it on reconnect?

src/core/request.js
darkdarkdragon: But on didn't streams return only one response as 'response' type, and others with specific types like 'path_find'?
And of first response `Request` will emit 'success' and will be done, right?

clark800: Hmmm, I didn't know that. So the "path_find" "responses" are still being processed by this Request object, but are not considered responses even though they are using the same request ID?

src/core/server.js
clark800: I'm not sure, but it seems like we should probably either resend or throw an error. I think it's usually dangerous to let things fail silently.

src/core/request.js
darkdarkdragon: They are not processed by `Request`. `Remote` has `switch` that routes special all other types of messages other than `response` to special handler [here](https://github.com/ripple/ripple-lib/blob/develop/src/core/remote.js#L499)
And `Server`, after receiving `response` and routing it to `Request`, [deletes](https://github.com/ripple/ripple-lib/blob/develop/src/core/server.js#L661) that `Request` object from own list, so if server will send another `response` message with the same id, `Server` will throw it as unexpected.

clark800: Ok, hopefully we can refactor this process later!

darkdarkdragon: Hm. You want to change all the core's logic? :smiley:

clark800: Yeah, I think it might be good to rewrite the connection/request/response logic from scratch actually.

src/core/server.js
darkdarkdragon: I've changed it to resend message

src/core/request.js
darkdarkdragon: Cool!

src/core/request.js
darkdarkdragon: > Is it safe to ignore all remote errors?
I've changed it to propagate to request.

darkdarkdragon:
@clark800 I've done changes

src/core/server.js
```diff
@@ -800,7 +810,16 @@ Server.prototype._sendMessage = function(message) {
     if (this._remote.trace) {
       log.info(this.getServerID(), 'request:', message);
     }
-    this._ws.send(JSON.stringify(message));
+    this._ws.send(JSON.stringify(message), (error) => {
+      // sometimes gives 'not opened'
+      // without callback it wil throw
+      if (error) {
+        // resend in case of error
+        this.once('connect', () => {
```
clark800: This worries me a bit. If there is any session-based message, it will probably produce unexpected behavior. Like if you try to close a pathfinding stream, but the message gets sent after a reconnect, the server might reply "there is no stream to close", which would be confusing if you didn't see any errors. This example is fairly innocuous, but I can imagine more problematic situations arising. But I guess this issue is baked into the whole system, so I suppose we can leave this for now.

clark800:
LGTM after squash

darkdarkdragon:
@clark800 done

PR #589, opened 2015-10-09
Fix Transaction.complete() for multisigned transactions

wltsmrz:
+ Don't check secret in complete() for multisigned transactions
+ Add LastLedgerSequence autofill test


clark800:
LGTM

PR #590, opened 2015-10-10
Update REFERENCE.md

FrRichard:
Changed the way you are setting the request callback for requestBookOffers .
Previous documentation didn't work, ie, the callback was never called (probably old way to set the callback).


clark800:
LGTM

PR #593, opened 2015-10-14
Remove rsign.js

clark800:


wltsmrz:
Shouldn't we remove most bin/ files? I doubt that this is the only one that isn't currently working

clark800:
@wltsmrz Yeah, looks like they are unnecessary now since we have simple libraries functions for each of them. Updated to remove "bin" directory.

wltsmrz:
LGTM

PR #598, opened 2015-10-15
Remove Currency class

clark800:


src/core/orderbook.js
```diff
@@ -1300,18 +1296,18 @@ OrderBook.prototype.toJSON =
 OrderBook.prototype.to_json = function() {
   const json = {
     taker_gets: {
-      currency: this._currencyGets.to_hex()
+      currency: this._currencyGets
```
darkdarkdragon: This will change output. It is OK?

src/core/amount.js
```diff
@@ -105,12 +106,12 @@ Amount.NaN = function() {
...
 };
 
-Amount.from_components_unsafe = function(value: Value, currency: Currency,
+Amount.from_components_unsafe = function(value: Value, currency: string,
   issuer: string, isNative: boolean
 ) {
   const result = new Amount(value);
   result._is_native = isNative;
-  result._currency = currency;
+  result._currency = normalizeCurrency(currency);
```
darkdarkdragon: I think we should remove `normalizeCurrency` from here - it is caller's responsibility to provide correct values.

src/core/currency.js
```diff
@@ -1,277 +1,58 @@
...
-
-  if (typeof j === 'number' && isFinite(j) && j >= 0) {
-    this._value = new BN(j);
-  }
-
-  this._update();
-  return this;
-};
+function isHexCurrency(currency) {
+  return /[A-Za-z0-9]{40}/.test(currency);
```
wltsmrz: A-F?

src/core/currency.js
```diff
@@ -1,277 +1,58 @@
...
-
-  // Json form:
-  //  '', 'XRP', '0': 0
-  //  3-letter code: ...
-  // XXX Should support hex, C++ doesn't currently allow it.
-
-  this._value = NaN;
-  this._update();
+function isISOCode(currency) {
+  return /^[A-Z0-9]{3}$/.test(currency);
```
wltsmrz: https://github.com/ripple/rippled/blob/43a480f02dca3bc4740b13f202d1b1815a8f5415/src/ripple/protocol/impl/UintTypes.cpp#L31-L37

src/core/orderbook.js
clark800: I don't see any problem with it; I'm not sure why it was hex in the first place.

src/core/amount.js
clark800: Ok

src/core/currency.js
clark800: Oops! Yeah, will fix.

src/core/currency.js
clark800: Yeah, I remember that. Those aren't actually ISO codes though, so if we support them we would have to rename the function. My preference would be to just leave it as is and try to discourage people from using non-ISO 3-character codes.

src/core/orderbook.js
darkdarkdragon: Then OK

clark800:
@darkdarkdragon @wltsmrz updated

darkdarkdragon:
LGTM

PR #601, opened 2015-10-16
Add Connection class

clark800:


src/api/common/connection.js
```diff
@@ -0,0 +1,125 @@
...
+        this.emit(data.type, data);
+      } else {
+        throw new Error('unrecognized message type: ' + data.type);
+      }
+    } catch (error) {
+      this.emit('badMessage', error);
+    }
+  }
+
+  state() {
```
darkdarkdragon: Maybe we should make this to be read-only property instead of function?

src/api/common/connection.js
```diff
@@ -0,0 +1,125 @@
...
+  _send(message) {
+    this._ws.send(message);
+  }
+
+  _sendWhenReady(message) {
+    return new Promise((resolve) => {
+      if (this.state() === WebSocket.OPEN) {
+        resolve(this._send(message));
+      } else {
+        this.once('open', () => resolve(this._send(message)));
```
darkdarkdragon: Can't find where `open` is emitted on this instance.

darkdarkdragon: And, if `this._send(message)` will throw error here, it won't be caught, and node will exit, right?

darkdarkdragon:
Do you want to add timeout-checking (on send) to some higher-level class?

src/api/common/connection.js
```diff
@@ -0,0 +1,125 @@
...
+      const data = JSON.parse(message);
+      if (data.type === 'response') {
+        if (!(Number.isInteger(data.id) && data.id >= 0)) {
+          throw new Error('valid id not found in response');
+        }
+        this.emit(data.id.toString(), data);
+      } else if (isStreamMessageType(data.type)) {
+        this.emit(data.type, data);
+      } else {
+        throw new Error('unrecognized message type: ' + data.type);
```
darkdarkdragon: `slowDown` error will be emitted as `badMessage`, without meaningful description.

src/api/common/connection.js
clark800: Sure

src/api/common/connection.js
clark800: Yeah, should be `this._ws.once('open', ...`

Promises capture exceptions, so if `send` throws an error, it will be caught by the `catch` in `request()`.

src/api/common/connection.js
clark800: Will add case for this

clark800:
I think I just forgot to add timeout functionality; I'll add that.

src/api/common/connection.js
darkdarkdragon: But here `send` will be executed in context of `open` event handler, not
inside of promise, no?
19 жовт. 2015 20:31 "Chris Clark" <notifications@github.com> пише:

clark800: Ah, you're right, will fix it

src/api/common/connection.js
```diff
@@ -0,0 +1,125 @@
+'use strict';
+const {EventEmitter} = require('events');
+const WebSocket = require('ws');
```
wltsmrz: Why do we not use ES6 import?

src/api/common/connection.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  disconnect() {
+    return new Promise((resolve) => {
+      if (this.state() === WebSocket.CLOSED) {
+        resolve();
+      } else if (this.state() === WebSocket.CLOSING) {
+        this._ws.once('close', resolve);
+      } else {
+        this._ws.removeAllListener('close');
```
wltsmrz:
```
> var em = new process.EventEmitter()
undefined
> em.removeAllListener
undefined
> em.removeAllListeners
[Function: removeAllListeners]
```

test/integration/connection-test.js
```diff
@@ -0,0 +1,48 @@
...
+    console.log(request);
+    console.log(JSON.stringify(response, null, 2));
+  }).catch((error) => {
+    console.log(request);
+    console.log(error);
+  });
+}
+
+function main() {
+  const connection = new Connection('wss://s1.ripple.com');
```
wltsmrz: Live test against public rippled network? Looks like this is the objective of /integration tests?

src/api/common/connection.js
clark800: I think there was something blocking us from using them, but I can't remember what it was. They might work now.

src/api/common/connection.js
clark800: Yeah, should be `removeAllListeners`

test/integration/connection-test.js
clark800: Yeah

clark800:
@darkdarkdragon @wltsmrz updated

wltsmrz:
LGTM

src/api/common/connection.js
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  _send(message) {
+    return new Promise((resolve, reject) => {
+      try {
+        this._ws.send(message);
+      } catch (error) {
+        reject(new DisconnectedError(error.message));
+      }
+      resolve();
```
darkdarkdragon: this will call both `reject` and `resolve` in case of error?

clark800: Oops, will fix

clark800:
@darkdarkdragon updated

src/api/common/connection.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  reconnect() {
+    return this.disconnect().then(() => this.connect());
+  }
+
+  _send(message) {
+    return new Promise((resolve, reject) => {
+      try {
+        this._ws.send(message);
```
darkdarkdragon: If we want to wait for message to be sent, maybe we should use callback from `_ws.send`?

clark800: Yeah, I think we have to; the `catch` won't even work this way because the error is returned as an event, not a synchronous exception,

darkdarkdragon: At least in `ws` module, errors are thrown as exceptions if there is no callback - I've hit that when I've worked on that `slowDown` error

clark800:
@darkdarkdragon updated. Note that the timer starts after the request is sent. If we ever have an issue with a timeout during the sending process, we could add a second timer for that.

src/api/common/connection.js
clark800: I think it's a case where it emits an "error" event, which node converts to an exception if it no listener is registered.

darkdarkdragon: I've actually hit [that](https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L218) line.

clark800: Oh yeah, looks like it does throw too. It also emits if you don't have a callback: https://github.com/websockets/ws/blob/master/lib/Sender.js#L123

But it looks like passing a callback should always work, so I think we're good now.

darkdarkdragon:
If we made request while still not connected, wait for it to timeout, and then connect - then that request will be fired on connection. It's not a problem in case of read-only requests, but if it submit - it will be unexpected for user. Not sure how much trouble that will cause, but still.

clark800:
The timer doesn't start until the request is actually sent, so it shouldn't be a problem right?

darkdarkdragon:
Yeah, you're right, missed that one.

darkdarkdragon:
LGTM

PR #603, opened 2015-10-19
Add basic auth

wltsmrz:


wltsmrz:
Halt

src/core/server.js
```diff
@@ -505,11 +513,13 @@ Server.prototype.connect = function() {
...
-        log.info(self.getServerID(), 'onerror:', e.data || e);
+        log.info(`${self.getServerID()} onerror: ${e.toString()}`);
       }
 
-      if (Server.TLS_ERRORS.indexOf(e.message) !== -1) {
-        // Unrecoverable
+      if (Server.TLS_ERRORS.includes(e.message)) {
+        throw e;
+      }
+      if (e.message.includes('unexpected')) {
```
wltsmrz: Not sure if this is going to be caught in the browser https://github.com/websockets/ws/blob/7debd827d39925650bba5a270811ed82616f0eb1/lib/WebSocket.js#L693

wltsmrz: Probably not

src/core/remote.js
```diff
@@ -149,6 +149,9 @@ function Remote(options = {}) {
   if (!Array.isArray(this.servers)) {
     throw new TypeError('servers must be an array');
   }
+  if (!(_.isUndefined(this.basic_auth) || _.isString(this.basic_auth))) {
```
sublimator: Why the check for undefined ?

src/core/server.js
```diff
@@ -442,23 +442,24 @@ Server.prototype.connect = function() {
...
-    } catch (error) {
-      throw new Error('"proxy" option is not supported in the browser');
-    }
-    const agent = new HttpsProxyAgent(opts);
+  const wsOptions = {};
 
-    this._ws = new WebSocket(this._opts.url, {agent: agent});
-  } else {
-    this._ws = new WebSocket(this._opts.url);
+  if (!_.isUndefined(this._remote.proxy)) {
```
sublimator: Shame lodash doesn't have `isDefined `

src/core/remote.js
wltsmrz: basic_auth is optional, so it must either be undefined or a string

src/index.js
```diff
@@ -1,4 +1,7 @@
 'use strict';
+
+require('babel-core/polyfill');
```
clark800: What's this for? Can we add a comment?

wltsmrz: Added comment

src/index.js
```diff
@@ -1,4 +1,8 @@
 'use strict';
+
+// Enable core-js polyfills
```
clark800: I still don't know what this is for. Like what functionality are we using that makes us need this, what does it provide? Why was it working without this before?

wltsmrz: Updated comment

src/index.js
```diff
@@ -1,4 +1,11 @@
 'use strict';
+
+/* eslint-disable max-len */
+// Enable core-js polyfills. This allows use of ES6/7 extensions listed here:
+// https://github.com/zloirock/core-js/blob/fb0890f32dabe8d4d88a4350d1b268446127132e/shim.js#L1-L103
```
clark800: Are you using one of these that we weren't using before?

clark800: I see: String.includes and Array.includes

clark800:
LGTM

PR #604, opened 2015-10-20
rename metaData to meta

shekenahglory:
from rippled's `ledger` command, `metaData` must be renamed to `meta` to conform with the response from the `tx` command

### DIFF
--- a/src/api/ledger/parse/ledger.js
+++ b/src/api/ledger/parse/ledger.js
@@ -5,6 +5,12 @@ const removeUndefined = require('./utils').removeUndefined;
 const parseTransaction = require('./transaction');
 import type {GetLedger} from '../types.js';
 
+function parseTransactionWrapper(tx) {
+  const transaction = _.assign({}, _.omit(tx, 'metaData'),
+    {meta: tx.metaData});
+  return parseTransaction(transaction);
+}
+
 function parseTransactions(transactions) {
   if (_.isEmpty(transactions)) {
     return {};
@@ -13,7 +19,7 @@ function parseTransactions(transactions) {
     return {transactionHashes: transactions};
   }
   return {
-    transactions: _.map(transactions, parseTransaction),
+    transactions: _.map(transactions, parseTransactionWrapper),
     rawTransactions: JSON.stringify(transactions)
   };
 }
--- a/test/api-test.js
+++ b/test/api-test.js
@@ -702,6 +702,16 @@ describe('RippleAPI', function() {
       _.partial(checkResult, responses.getLedger.header, 'getLedger'));
   });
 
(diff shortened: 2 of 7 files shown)


src/api/ledger/parse/ledger.js
```diff
@@ -5,6 +5,15 @@ const removeUndefined = require('./utils').removeUndefined;
 const parseTransaction = require('./transaction');
 import type {GetLedger} from '../types.js';
 
+function parseTransactionWrap(transaction) {
+  if (transaction.metaData) {
+    transaction.meta = transaction.metaData;
+    delete transaction.metData;
```
clark800: typo: "metData"

src/api/ledger/parse/ledger.js
```diff
@@ -13,8 +22,8 @@ function parseTransactions(transactions) {
     return {transactionHashes: transactions};
   }
   return {
-    transactions: _.map(transactions, parseTransaction),
-    rawTransactions: JSON.stringify(transactions)
+    rawTransactions: JSON.stringify(transactions),
```
clark800: The fact that the ordering matters here is concerning... I think we should just not mutate in the wrapper.

`_.assign({}, transaction, {meta: transaction.metaData})` should work

src/api/ledger/parse/ledger.js
```diff
@@ -5,6 +5,15 @@ const removeUndefined = require('./utils').removeUndefined;
 const parseTransaction = require('./transaction');
 import type {GetLedger} from '../types.js';
 
+function parseTransactionWrap(transaction) {
```
clark800: Can we call it `parseTransactionWrapper`?

shekenahglory:
fixed

src/api/ledger/parse/ledger.js
```diff
@@ -5,6 +5,11 @@ const removeUndefined = require('./utils').removeUndefined;
 const parseTransaction = require('./transaction');
 import type {GetLedger} from '../types.js';
 
+function parseTransactionWrapper(transaction) {
+  _.assign({}, transaction, {meta: transaction.metaData});
```
clark800: This has no effect; you have to pass the return value to parseTransaction. Also could you add a unit test to make sure it works?

shekenahglory:
I'm not sure of what would be a good way to test this, other than doing get ledger on one that has an 'AccountSet' transaction, since that is where it failed for me.  Do you have a better suggestion?

clark800:
Yeah, that would be a good way to do it. As long as the test case fails for the previous version of the code, it should be good.

shekenahglory:
added test for ledger with AccountSet tranasction

clark800:
LGTM

PR #606, opened 2015-10-21
Decouple core.Transaction from API

clark800:


src/api/transaction/utils.js
```diff
@@ -114,8 +101,47 @@ function prepareTransaction(txJSON: Object, remote: Remote,
   }));
 }
 
+function convertStringToHex(string: string) {
+  return string ? (new Buffer(string, 'utf8')).toString('hex').toUpperCase() :
```
sublimator: Does this work in the browser? Do we care either way?

src/api/transaction/utils.js
```diff
@@ -114,8 +101,47 @@ function prepareTransaction(txJSON: Object, remote: Remote,
...
+
+/**
+ * @param {Number|Date} timestamp (ms since unix epoch)
+ * @return {Number} seconds since ripple epoch ( 1/1/2000 GMT)
+ */
+function fromTimestamp(timestamp: number | Date): number {
+  const timestamp_ = timestamp instanceof Date ?
+                     timestamp.getTime() :
+                     timestamp;
+  return Math.round(timestamp_ / 1000) - 0x386D4380;
```
sublimator: Could extra a constant for `0x386D4380`

src/api/transaction/order.js
```diff
@@ -2,40 +2,45 @@
...
-
-  return transaction;
+  if (order.passive) {
+    txJSON.Flags |= offerFlags.Passive;
+  }
+  if (order.immediateOrCancel) {
+    txJSON.Flags |= offerFlags.ImmediateOrCancel;
+  }
+  if (order.fillOrKill) {
+    txJSON.Flags |= offerFlags.KillOrKill;
```
sublimator: `KillOrKill` 

That's a pretty bleak view :)

src/api/transaction/suspended-payment-execution.js
```diff
@@ -18,40 +17,36 @@ type SuspendedPaymentExecution = {
...
-    paymentSequence: payment.paymentSequence
-  });
+  const txJSON: Object = {
+    TransactionType: 'SuspendedPaymentFinish',
+    Account: account,
+    Owner: payment.owner,
+    OfferSequence: payment.paymentSequence
+  };
 
   if (payment.method) {
```
wltsmrz: I don't think we should ever rely on coercion to determine whether to set a property. Although `Method` will [probably start at 1](https://github.com/ripple/rippled/blob/6700e44793ff479bd81aa8a1008649ef229e1ffa/src/ripple/app/tx/impl/SusPay.cpp#L93-L95) this seems sloppy. The [schema](https://github.com/ripple/ripple-lib/blob/98422e4153e53b8006097c8f9cebbfa9d5f831d4/src/api/common/schemas/suspended-payment-execution.json#L14) seems right

src/api/transaction/utils.js
clark800: There is a webpack polyfill: https://webpack.github.io/docs/configuration.html

clark800:
@wltsmrz @sublimator updated

src/api/transaction/order.js
```diff
@@ -2,40 +2,45 @@
...
+  const txJSON: Object = {
+    TransactionType: 'OfferCreate',
+    Account: account,
+    TakerGets: takerGets,
+    TakerPays: takerPays,
+    Flags: 0
+  };
   if (order.direction === 'sell') {
-    transaction.setFlags('Sell');
+    txJSON.Flags |= offerFlags.Sell;
```
darkdarkdragon: [Here](https://github.com/ripple/ripple-lib/blob/develop/src/api/transaction/utils.js#L53) there is comment that says that we should convert flags to unsigned after doing bitwise operations. Do we need it?

clark800: I think `tfFullyCanonicalSig` is the only flag that needs the `>>> 0` because flags are 32 bits and that is the only one that is big enough to create a negative sign:
https://github.com/ripple/rippled/blob/4f9dba22c736cdd092dccd350b2386e6a726ec75/src/ripple/protocol/TxFlags.h#L46
> \> 0 | 0x80000000
> -2147483648
> \> 0 | 0x40000000
> 1073741824

sublimator: Yes

wltsmrz:
LGTM

PR #608, opened 2015-10-22
BREAKING CHANGE: getFee returns promise, express fee in XRP in instru…

clark800:
…ctions response, use rawRequest to start decoupling Remote from API


src/api/common/connection.js
```diff
@@ -62,6 +73,10 @@ class Connection extends EventEmitter {
         }
         this.emit(data.id.toString(), data);
       } else if (isStreamMessageType(data.type)) {
+        if (data.type === 'ledgerClosed') {
```
darkdarkdragon: I think this should be `ledger_closed`, and in `isStreamMessageType` too.

clark800: No, ripple-lib/core renames it from ledgerClosed to ledger_closed

https://ripple.com/build/rippled-apis/#subscribe
https://github.com/ripple/ripple-lib/blob/develop/src/core/remote.js#L503

darkdarkdragon: hm. unexpected :smile:

darkdarkdragon:
LGTM

PR #611, opened 2015-10-23
add timeout logic to Remote.getLedgerSequence

darkdarkdragon:


src/core/remote.js
```diff
@@ -539,8 +539,18 @@ Remote.prototype.getLedgerSequence = function(callback = function() {}) {
     // the "current" ledger is the one after the most recently closed ledger
     callback(null, this._ledger_current_index - 1);
   } else {
+    let timedOut = false;
+    const timeout = setTimeout(() => {
+      timedOut = true;
+      callback(new RippleError('tejTimeout',
```
wltsmrz: I think I've said before that tej* errors are not general-purpose; they are specifically transaction results generated from ripple-lib

src/core/remote.js
```diff
@@ -539,8 +539,18 @@ Remote.prototype.getLedgerSequence = function(callback = function() {}) {
     // the "current" ledger is the one after the most recently closed ledger
     callback(null, this._ledger_current_index - 1);
   } else {
+    let timedOut = false;
```
clark800: Do you think it would be simpler to `removeListener` on the `ledger_closed` event instead of using the `timedOut` variable?

src/core/remote.js
darkdarkdragon: OK, what would suit better here?

darkdarkdragon:
@clark800 updated

clark800:
LGTM

PR #613, opened 2015-10-27
Remove expired offers from orderbook

wltsmrz:


clark800:
I think Ivan will be creating a new repo for orderbooks today. Maybe we should merge this into the new repo since I'm about to make a pull request that deletes this file.

PR #621, opened 2015-10-30
Fix and reorganize schemas and switch to ajv validator

clark800:


clark800:
"objects" are just the components of other schemas that can be used for both "inputs" and "outputs"

geertweening:
LGTM

wltsmrz:
Sounds like they should be named "components"

PR #624, opened 2015-11-02
Better error message when using methods that require a connection when servers are not provided

clark800:


darkdarkdragon:
LGTM

PR #625, opened 2015-11-06
Organize JSON schemas for documentation generation

clark800:
The idea is to have one input schema and one output schema for each method so that we can more efficiently generate markdown tables for documentation. The next step will be to fill in all the necessary descriptions for each field in the JSON schemas.


darkdarkdragon:
LGTM

PR #629, opened 2015-11-16
Miscellaneous cleanup

clark800:
Lots of minor breaking changes


src/common/schemas/specifications/suspended-payment-cancellation.json
```diff
@@ -9,11 +9,11 @@
       "$ref": "address",
       "description": "The address of the owner of the suspended payment to cancel."
     },
-    "paymentSequence": {
+    "suspensionSequence": {
```
darkdarkdragon: Maybe `suspensionTxSequence`? 
For someone who new to ripple it is now obvious that sequence refers to transaction.

src/transaction/submit.js
```diff
@@ -27,7 +27,10 @@ function formatResponse(response) {
   if (isImmediateRejection(response.engine_result)) {
     throw new utils.common.errors.RippledError('Submit failed');
   }
-  return convertKeysFromSnakeCaseToCamelCase(response);
+  return {
+    resultCode: response.engine_result,
+    resultMessage: response.engine_result_message
+  };
```
darkdarkdragon: This would mean that there will be no `tx_json` field in `submit` response.
Right now I have `tx_json`, which is very convenient - I can see what this transaction it was, and it's hash.

src/common/schemas/specifications/suspended-payment-cancellation.json
clark800: I think the description clarifies it enough; we use the convention of "sequence" everywhere for "account sequence number".

src/transaction/submit.js
clark800: That's what you just submitted though right?

wltsmrz:
LGTM

src/transaction/submit.js
darkdarkdragon: Yeah. But, 
1. it is just convinient
2. where else do I get transaction hash?

clark800: The transaction hash is returned by `sign`. What is it convenient for? Shouldn't you already know what you submitted? None of the other methods redundantly return the request data.

darkdarkdragon: For debugging at least :smile: 
missed hash in `sign. then ok

darkdarkdragon:
LGTM

PR #631, opened 2015-11-17
Fix TODOs in schemas and update README for RippleAPI

clark800:


src/common/schemas/specifications/suspended-payment-execution.json
```diff
@@ -21,11 +21,11 @@
     },
     "digest": {
       "$ref": "hash256",
-      "description": "TODO"
+      "description": "The original `digest` from the suspended payment creation transaction."
     },
     "proof": {
       "type": "string",
-      "description": "This value will be hashed to produce the digest."
+      "description": "A value that produces the digest when hashed."
```
darkdarkdragon: I think you should fact that this must have length of 32 8-bit characters.

PR #636, opened 2015-11-19
Add documentation for API events

clark800:


src/common/schemas/output/ledger-closed.json
```diff
@@ -4,15 +4,48 @@
...
-    "ledgerVersion": {"$ref": "ledgerVersion"},
-    "ledgerTimestamp": {"type": "string", "format": "date-time"},
-    "reserveBase": {"type": "integer", "minimum": 0},
-    "reserveIncrement": {"type": "integer", "minimum": 0},
-    "transactionCount": {"type": "integer", "minimum": 0},
-    "validatedLedgerVersions": {"type": "string"}
+    "feeBase": {
+      "type": "integer",
+      "minimum": 0,
+      "description": "Base fee, in drops."
```
mDuo13: Does RippleAPI actually explain the concept of Drops anywhere else? This will be confusing otherwise.

src/common/schemas/output/ledger-closed.json
```diff
@@ -4,15 +4,48 @@
...
+    },
+    "feeReference": {
+      "type": "integer",
+      "minimum": 0,
+      "description": "Cost of the 'reference transaction' in 'fee units'."
```
mDuo13: This is also confusing, although sadly, I don't think we have a good explanation that's not overly complicated (because the product itself is just complicated here). Maybe not worth exposing in RippleAPI?

mDuo13:
Left some minor quibbles, but overall LGTM.

src/common/schemas/output/ledger-closed.json
clark800: I have to fix this to make it return XRP, but since it's a breaking change, I was going to leave it for a separate commit.

src/common/schemas/output/ledger-closed.json
clark800: It would be nice if we could just link to a separate page that explains fees in more detail.

PR #642, opened 2015-11-20
Doc introduction changes

mDuo13:
Rewrite of the introductory content to make the docs much more approachable with less context and foreknowledge.

### DIFF
--- a/docs/src/introduction.md.ejs
+++ b/docs/src/introduction.md.ejs
@@ -1,5 +1,12 @@
 # Introduction
 
-RippleAPI allows you to query and submit transactions to a node on the Ripple network.
+RippleAPI is the official client library to the Ripple Consensus Ledger. Currently, RippleAPI is only available in JavaScript. 
+Using RippleAPI, you can:
+
+* [Query transactions from the network](#gettransaction)
+* [Sign](#sign) transactions securely without connecting to any server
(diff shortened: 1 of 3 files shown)


docs/src/introduction.md.ejs
```diff
@@ -1,5 +1,11 @@
 # Introduction
 
-RippleAPI allows you to query and submit transactions to a node on the Ripple network.
+RippleAPI is the official client library to the Ripple Consensus Ledger. Currently, RippleAPI is only available in JavaScript. 
+Using RippleAPI, you can:
+
+* [Query transactions from the network](#gettransaction)
+* [Sign](#sign) transactions securely without connecting to any server
+* [Submit](#submit) transactions to the Ripple Consensus Ledger, including [Payments](#payment), [Orders](#order), [Settings changes](#settings), and [other types](#transaction-types)
+* [Generate a new Ripple Address](#generateaddress)
```
clark800: We should say "and more" or something so that it doesn't sound like this is all that it can do.

clark800:
The tests are failing because we added a rule that the `docs/index.md` file must always be up-to-date. You can regenerate it with `npm run docgen`.

mDuo13:
Ah, that explains it. I'll fix & push as soon as I get all the node modules installed.

clark800:
LGTM

PR #645, opened 2015-11-23
boost coverage to almost 100%

darkdarkdragon:


src/ledger/transactions.js
```diff
@@ -47,9 +47,6 @@ function parseAccountTxTransaction(tx) {
 }
 
 function counterpartyFilter(filters, tx: TransactionType) {
-  if (!filters.counterparty) {
-    return true;
```
clark800: Why is this removed?

test/fixtures/requests/index.js
```diff
@@ -8,18 +8,30 @@ module.exports = {
   },
   prepareOrderCancellation: require('./prepare-order-cancellation'),
   preparePayment: require('./prepare-payment'),
+  preparePaymentMinAmountXRP: require('./prepare-payment-min-xrp'),
```
clark800: Can we use the convention of namespacing by method name (here and in other fixtures index files)? e.g. `preparePayment.minAmounXRP`

test/api-test.js
```diff
@@ -31,11 +33,247 @@ function checkResult(expected, schemaName, response) {
...
+  });
+}
+
+
 describe('RippleAPI', function() {
   const instructions = {maxLedgerVersionOffset: 100};
   beforeEach(setupAPI.setup);
   afterEach(setupAPI.teardown);
 
+  describe('Connection', function() {
```
clark800: Maybe we can put the connection tests in a separate file?

src/ledger/transactions.js
darkdarkdragon: That check is done before this function called.
https://github.com/ripple/ripple-lib/blob/develop/src/ledger/transactions.js#L79

clark800: Ok

test/fixtures/requests/index.js
darkdarkdragon: OK

test/api-test.js
darkdarkdragon: OK

test/fixtures/responses/index.js
```diff
@@ -31,26 +34,37 @@ module.exports = {
...
+    withSettingsTx: require('./get-ledger-with-settings-tx'),
+    withStateAsHashes: require('./get-ledger-with-state-as-hashes')
   },
-  prepareOrderCancellation: require('./prepare-order-cancellation.json'),
   prepareOrder: {
     buy: require('./prepare-order.json'),
     sell: require('./prepare-order-sell.json'),
-    expiration: require('./prepare-order-expiration')
+    expiration: require('./prepare-order-expiration'),
+    cancellation: require('./prepare-order-cancellation.json'),
```
clark800: `prepareOrder` and `prepareOrderCancellation` should be separate namespaces because they are separate methods

test/fixtures/rippled/index.js
```diff
@@ -17,8 +18,11 @@ module.exports = {
   },
   account_offers: require('./account-offers'),
   account_tx: require('./account-tx'),
+  account_tx_one: require('./get-transactions-one'),
   gateway_balances: require('./gateway-balances'),
   book_offers: require('./book-offers'),
+  book_offers_1: require('./book-offers-1'),
```
clark800: Any more descriptive name than `1` and `2`?

src/transaction/trustline.js
```diff
@@ -8,8 +8,7 @@ import type {Instructions, Prepare} from './types.js';
 import type {TrustLineSpecification} from '../ledger/trustlines-types.js';
 
 function convertQuality(quality) {
-  return quality === undefined ? undefined :
```
clark800: Isn't the `undefined` check necessary?

test/api-test.js
```diff
@@ -31,11 +33,247 @@ function checkResult(expected, schemaName, response) {
...
+            assert.strictEqual(got, expect);
+            server.close();
+            done();
+          });
+        });
+
+        const options = {
+          proxy: 'ws://localhost:' + port,
+          authorization: 'authorization',
+          trustedCertificates: 'something'
```
clark800: This option is supposed to be an array of strings. What is this test case testing?

test/api-test.js
```diff
@@ -31,11 +33,247 @@ function checkResult(expected, schemaName, response) {
...
+        };
+        const connection =
+          new utils.common.Connection(this.api.connection._url, options);
+        connection.connect().catch(done);
+        connection.connect().catch(done);
```
clark800: Is this going to call `done` twice?

src/transaction/trustline.js
darkdarkdragon: It is always checked before
https://github.com/ripple/ripple-lib/blob/develop/src/transaction/trustline.js#L30-L34

clark800: Ok

test/api-test.js
darkdarkdragon: It really checks nothing, just adds to coverage. And I'm not sure we need to test something here.

clark800: Hmmm, is there a way to get that coverage with a test that makes more sense?

test/api-test.js
darkdarkdragon: Here `done` will be called only in case of error.
Normally it called earlier in 
`socket.on('data')` handler

test/api-test.js
darkdarkdragon: I'm can't think of anything that we can test here. This param just got passed into node's `tls.connect`.

clark800: It's not going to make it to `tls.connect` if there is a validation error though right?

clark800:
I merged this as https://github.com/ripple/ripple-lib/pull/649. Please address feedback in a separate refactoring pull request.

PR #647, opened 2015-11-23
BREAKING CHANGE: prepareOrderCancellation now takes orderCancellation specification instead of sequence

clark800:


darkdarkdragon:
LGTM

PR #651, opened 2015-11-24
Docs: suspay and other clarifications

mDuo13:
- SusPays are not available in prod right now, so I added a warning. (More might be warranted in other places, because people skim.)
- maxLedgerVersion is not the "only" way you can know a transaction won't be viable later (you could also cancel by using up the seq. num), so I rephrased.
- I felt that the Src/Dst Tag description was misleading, so I tweaked it.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -4,6 +4,7 @@
 
 - [Introduction](#introduction)
   - [Boilerplate](#boilerplate)
+  - [Offline functionality](#offline-functionality)
 - [Basic Types](#basic-types)
   - [Ripple Address](#ripple-address)
   - [Account Sequence Number](#account-sequence-number)
@@ -93,7 +94,7 @@ api.connect().then(() => {
 
 RippleAPI is designed to work in [NodeJS](https://nodejs.org) (version `0.12.0` or greater) using [Babel](https://babeljs.io/) for [ECMAScript 6](https://babeljs.io/docs/learn-es2015/) support.
 
-The code samples in this documentation are written in ES6, but `RippleAPI` will work with ES5 also. Regardless of whether you use ES5 or ES6, the methods that return promises will return [ES6-style promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise_).
+The code samples in this documentation are written in ES6, but `RippleAPI` will work with ES5 also. Regardless of whether you use ES5 or ES6, the methods that return promises will return [ES6-style promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).
 
 <aside class="notice">
 All the code snippets in this documentation assume that you have surrounded them with this boilerplate.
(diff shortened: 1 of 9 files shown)


docs/index.md
```diff
@@ -209,6 +211,8 @@ Type | Description
 [suspendedPaymentCancellation](#suspended-payment-cancellation) | A `suspendedPaymentCancellation` transaction unlocks the funds in a suspended payment and sends them back to the creator of the suspended payment, but it will only work after the suspended payment expires.
 [suspendedPaymentExecution](#suspended-payment-execution) | A `suspendedPaymentExecution` transaction unlocks the funds in a suspended payment and sends them to the destination of the suspended payment, but it will only work if the cryptographic condition is provided.
 
+The three "suspended payment" transaction types are not supported by the production Ripple peer-to-peer network at this time. They are available for testing purposes if you [connect](#boilerplate) to the [Ripple Test Net](https://ripple.com/build/ripple-test-net/) instead.
```
clark800: I think it would make more sense to link "connect" to the "connect" section, which links to "boilerplate".

docs/src/boilerplate.md.ejs
```diff
@@ -29,6 +29,8 @@ If you omit the "catch" section, errors may not be visible.
 
 ### Parameters
 
+The RippleAPI constructor takes an object with the following options:
```
clark800: Maybe add "optionally" since it can be omitted. And we could point out that if omitted, only offline operations like "sign" and "computeLedgerHash" will be allowed.

clark800:
> maxLedgerVersion is not the "only" way you can know a transaction won't be viable later (you could also cancel by using up the seq. num), so I rephrased.

I don't think that's true. I've never seen a rippled method to cancel a transaction by sequence number and it certainly doesn't exist in RippleAPI.

mDuo13:
Traditionally, you send an AccountSet transaction with no options:

https://ripple.com/build/transactions/#canceling-or-skipping-a-transaction

It was documented as "the canonical nop" in the wiki. 

I'm not sure RippleAPI supports it, but it should.

--Rome

On November 24, 2015 10:29:14 AM PST, Chris Clark <notifications@github.com> wrote:
>> maxLedgerVersion is not the "only" way you can know a transaction
>won't be viable later (you could also cancel by using up the seq. num),
>so I rephrased.
>
>I don't think that's true. I've never seen a rippled method to cancel a
>transaction by sequence number and it certainly doesn't exist in
>RippleAPI.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/ripple/ripple-lib/pull/651#issuecomment-159363645

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

clark800:
Ah, ok. The schemas will prevent that. If we need to support that I think it should be a separate method like `prepareTransactionCancellation`, but I'm not sure if it's necessary.

docs/index.md
```diff
@@ -132,6 +137,34 @@ Instead of using babel-node in production, we recommend using Babel to transpile
...
+To instantiate RippleAPI in offline mode, use the following boilerplate code:
+
+```javascript
+const {RippleAPI} = require('ripple-lib');
+
+const api = new RippleAPI();
+/* insert code here */
+```
+
+Methods that depend on the state of the Ripple Consensus Ledger are unavailable in offline mode. To prepare transactions offline, you **must** specify  the `fee` and `sequence` in the [transaction instructions](#transaction-instructions). The following methods should work offline:
```
clark800: It also requires `maxLedgerVersion` because it queries `getLedgerVersion` and adds 3 by default.

mDuo13:
Replaced by PR#654

PR #658, opened 2015-11-25
Fix support for proxy credentials in proxy URL and fix error when the…

clark800:
…re are more than 10 outstanding requests


geertweening:
LGTM

PR #662, opened 2015-12-02
getFee returns a string not float

lumberj:
> api.connect().then(() => api.getFee().then(fee => console.log(typeof fee)));
> string


clark800:
Thanks, LGTM

PR #663, opened 2015-12-04
http server example

darkdarkdragon:


clark800:
I think the tests won't be run by the CI server. Can you add the new tests to `scripts/ci.sh`? I think it can go in the `integrationtest` function.

darkdarkdragon:
done

package.json
```diff
@@ -34,6 +34,7 @@
     "babel-core": "^5.8.22",
     "babel-eslint": "^4.1.3",
     "babel-loader": "^5.3.2",
+    "bluebird": "^3.0.6",
```
clark800: Can we get rid of the bluebird dependency and just use ES6 promises?

package.json
```diff
@@ -46,6 +47,7 @@
     "gulp-rename": "^1.2.0",
     "gulp-uglify": "^1.1.0",
     "istanbul": "^0.3.5",
+    "jayson": "^1.2.2",
```
clark800: Looks like this is also used in the `src` directory, so I think it should be a non-dev dependency (which means we also need to shrinkwrap).

src/http.js
```diff
@@ -0,0 +1,66 @@
...
+
+const _ = require('lodash');
+const RippleAPI = require('./index').RippleAPI;
+const jayson = require('jayson');
+const Promise = require('bluebird');
+
+function createServer(options, httpPort) {
+  const rippleAPI = new RippleAPI(options);
+
+  const functions = _.filter(_.keys(RippleAPI.prototype), k => {
```
clark800: This should probably be called "methodNames" so that it doesn't sound like an array of references to functions.

clark800:
Can we add a "start" script to package.json to start the HTTP server? We can default to port 5990 since that's what ripple-rest used.

package.json
darkdarkdragon: ok

package.json
darkdarkdragon: BTW, I didn't expose http server in public api, so it can't be used by user directly, only as example

src/http.js
darkdarkdragon: ok

darkdarkdragon:
I'll do it

package.json
clark800: Oh, I was thinking we would expose it from index.js, but maybe add a comment that says "experimental". We might as well add "experimental" to RippleAPIBroadcast while we're at it too.

clark800:
I'm thinking about the interface and I think it would probably be more user-friendly if the request syntax matched the JSON schemas, which take an object rather than an array.

package.json
darkdarkdragon: Add that comment in index.js?

clark800:
It would require making a wrapper for each method individually, but it seems like it's worth it to make the interface more readable.

package.json
clark800: yeah

darkdarkdragon:
Right now those JSON schemas not exposed anywhere, and use even don't know what they exists.
Know request syntax is matched with api methods parameters, those can be easily checked in documentation. 
I think current syntax is intuitive and don't see need to change.

darkdarkdragon:
done

clark800:
Well the JSON schemas are "exposed" through the documentation because the tables of parameters are generated from them.

If you were auditing a log of HTTP requests, which would be more readable?

```
[
  "r3kmLJN5D28dHuH8vZNUZpMC43pEHpaocV", 
  {"regularKey": "rKVGAUZaquMiEvt3J51dUrFunNEv9rRTGf"},
  {"sequence": 123}
]
```

or

```
{
  "address": "r3kmLJN5D28dHuH8vZNUZpMC43pEHpaocV",
  "settings": {
    "regularKey": "rKVGAUZaquMiEvt3J51dUrFunNEv9rRTGf"
  },
  "instructions": {
    "sequence": 123
  }
}
```

I think the latter would be a lot more helpful. And also we may have to add HTTP request/response examples to the documentation, and I think having named parameters will make it look less cryptic. What do you think?

src/index.js
```diff
@@ -2,5 +2,8 @@
 
 module.exports = {
   RippleAPI: require('./api').RippleAPI,
-  RippleAPIBroadcast: require('./broadcast').RippleAPIBroadcast
+  // Broadcast api is experimental
+  RippleAPIBroadcast: require('./broadcast').RippleAPIBroadcast,
+  // HTTP server is hihgly experimental
```
clark800: typo: hihgly, i think we can just say "experimental"

darkdarkdragon:
> What do you think?
I'm looking from the perspective of programmer who will be writing http requests - he will be looking into RippleAPI documentation and into jayson documentation. And jayson implies that functions parameters will be passed as array. But I see that documentation has parameters names, so we can put parameters into object, but still, this will be somewhat unintuitive for user.

clark800:
I don't think we should assume that the user will be using `jayson` or any other specific library for making JSON RPC requests. JSON RPC allows both positional parameters and named parameters: https://en.wikipedia.org/wiki/JSON-RPC. What if we allowed both?

clark800:
Also, `jayson` supports named parameters: https://www.npmjs.com/package/jayson#named-parameters

clark800:
It looks like it might automagically work without wrapper functions, try changing the test cases to use objects.

darkdarkdragon:
just tried. not worked :disappointed:

darkdarkdragon:
will try to find what's wrong

src/http.js
```diff
@@ -0,0 +1,72 @@
...
+          callback({code: 99, message: err.message, data: {name: err.name}});
+        });
+    } catch (err) {
+      callback({code: 99, message: err.message, data: {name: err.name}});
+    }
+  }
+
+  const methods = {};
+  _.forEach(methodNames, fn => {
+    methods[fn] = jayson.Method((args, cb) => {
```
clark800: Ah, probably because this is using a single function argument. I think we would need to dynamically de-promisify while keeping the same method arguments.

src/common/schemas/input/compute-ledger-hash.json
```diff
@@ -5,7 +5,8 @@
   "properties": {
     "ledger": {
       "$ref": "getLedger",
-      "description": "The ledger header to hash."
+      "description": "The ledger header to hash.",
+      "position": 0
```
clark800: This is not a very robust solution because it can get out of sync with the actual method signatures. I think it would be better to use the same technique that jayson uses to get the method parameter ordering.

test/integration/http-integration-test.js
```diff
@@ -0,0 +1,254 @@
...
+  it('prepareSettings', function(done) {
+    const instructions = {
+      maxFee: '0.000012',
+      sequence: 23,
+      maxLedgerVersion: 8820051
+    };
+    const settings = apiRequests.prepareSettings;
+
+    this.client.request('prepareSettings',
+    [address, settings, instructions], '4',
```
clark800: Can we change all the requests to named format?

src/common/schemas/input/compute-ledger-hash.json
darkdarkdragon: I've made that way because we can't use jayson method - babel doesn't put
parameters that have default values into function definition - it checks
`arguments` value to get that parameters.
7 груд. 2015 21:21 "Chris Clark" <notifications@github.com> пише:

test/integration/http-integration-test.js
darkdarkdragon: Do you want to remove support for positional arguments?
7 груд. 2015 21:23 "Chris Clark" <notifications@github.com> пише:

src/common/schemas/input/compute-ledger-hash.json
clark800: Oh, that makes sense. But I think you can still partially use it because I think there is only ever one optional argument per method. You could use jayson's technique for the required arguments, then get all the parameters from the schema and if there is just one extra, that is the last argument, and if there is more than one extra, just throw an exception.

darkdarkdragon: Hmm, this will pose some restrictions on  code, but this is good solution,
I'll implement it.
7 груд. 2015 21:50 "Chris Clark" <notifications@github.com> пише:

test/integration/http-integration-test.js
clark800: I think it's probably best to keep support for positional arguments since that seems to be common for JSON RPC. But I think we should document and use named parameters exclusively.

darkdarkdragon: But without test we can't be sure they are working :smile:

clark800: That's true, I didn't notice that you had every test in both forms. It seems like there's a lot of redundancy though, could you make a helper method to reduce duplication?

darkdarkdragon:
done

test/integration/http-integration-test.js
```diff
@@ -0,0 +1,143 @@
...
+  function createTest(name, testFunc, id, params) {
+    createTestInternal(name + ' - positional params', name,
+      testFunc, id, makePositionalParams(params));
+    createTestInternal(name + ' - named params', name,
+      testFunc, id, makeNamedParams(params));
+  }
+
+  beforeEach(function(done) {
+    server = createHTTPServer(apiOptions, httpPort);
+    server.start().then(() => {
```
clark800: Doesn't mocha allow returning a promise here (and omitting "done")?

test/integration/http-integration-test.js
```diff
@@ -0,0 +1,143 @@
...
+  });
+
+  afterEach(function(done) {
+    server.stop().then(done, done);
+  });
+
+
+  createTest('getLedgerVersion', (result) => {
+    assert(_.isNumber(result.result));
+  }, '001', []);
```
clark800: How about using a property on the test suite to store an incrementing counter for the request ID?

test/integration/http-integration-test.js
```diff
@@ -0,0 +1,143 @@
...
+            return reject(err);
+          }
+          testFunc(result);
+          resolve();
+        });
+      });
+    });
+  }
+
+  function createTest(name, testFunc, id, params) {
```
clark800: I think it would be a little more clear if the params came before the testFunc.

test/integration/http-integration-test.js
```diff
@@ -0,0 +1,143 @@
...
+  this.timeout(TIMEOUT);
+
+  let server = null;
+  let client = null;
+
+  function createTestInternal(testName, methodName, testFunc, id, params) {
+    it(testName, function() {
+      return new Promise((resolve, reject) => {
+        client.request(methodName, params, id, (err, result) => {
+          if (err) {
```
clark800: Could shorten to `(err, result) => err ? reject(err) : resolve(testFunc(result));`

test/integration/http-integration-test.js
```diff
@@ -0,0 +1,143 @@
...
+    server = createHTTPServer(apiOptions, httpPort);
+    server.start().then(() => {
+      this.client = createClient();
+      client = this.client;
+      done();
+    }).catch(done);
+  });
+
+  afterEach(function(done) {
+    server.stop().then(done, done);
```
clark800: Might be able to return promise here too.

darkdarkdragon:
fixed

clark800:
LGTM

PR #666, opened 2015-12-04
Fix DeliverMin value when specifying minAmount

clark800:


src/transaction/payment.js
```diff
@@ -64,7 +64,7 @@ function createMaximalAmount(amount: Amount): Amount {
   const maxXRPValue = '100000000000';
   const maxIOUValue = '9999999999999999e80';
   const maxValue = amount.currency === 'XRP' ? maxXRPValue : maxIOUValue;
-  return _.assign(amount, {value: maxValue});
+  return _.assign({}, amount, {value: maxValue});
```
clark800: This was the bug.

darkdarkdragon:
LGTM

PR #667, opened 2015-12-07
BREAKING CHANGE: Fix types of XRP values in getServerInfo response

clark800:


src/common/serverinfo.js
```diff
@@ -49,6 +49,12 @@ function getServerInfo(connection: Connection): Promise<GetServerInfoResponse> {
       reserveIncXrp: 'reserveIncrementXRP',
       seq: 'ledgerVersion'
     });
+    info.validatedLedger.baseFeeXRP =
+      info.validatedLedger.baseFeeXRP.toString();
```
darkdarkdragon: Why do tests pass? And I don't see conversion from drop to XRP. What I'm missing?

clark800: It was expressed in XRP before, but as a number rather than a string.

darkdarkdragon: Argh, my bad. But it is still number in `get-server-info.json`, and test didn't catch this :disappointed:

clark800: Ah, nice catch, I can make the tests catch that case.

clark800:
@darkdarkdragon done

darkdarkdragon:
LGTM

PR #668, opened 2015-12-09
BREAKING CHANGE: Change error event format and fix crash due to error event on websocket

clark800:


docs/src/events.md.ejs
```diff
@@ -24,12 +24,22 @@ This event is emitted when there is an error on the connection to the server tha
...
+* the message that caused the error for `badMessage`
+* the error object emitted for `websocket`
+* the parsed response for rippled errors
 
 ### Example
 
 ```javascript
-api.on('error', (errorCode, errorMessage) => {
+api.on('error', (errorCode, errorMessage, data) => {
```
lumberj: Just a nit but you didn't log the `data`

clark800: Yeah, that was intentional, didn't think it was necessary

lumberj:
+2

lumberj:
I don't think this is a breaking change though, it just increases the arity of the event handler, but javascript does not enforce function arity

clark800:
The breaking change is that the second parameter of the badMessage event moved to the third parameter

lumberj:
Got it

PR #670, opened 2015-12-10
Remove Windows line endings in http example

lumberj:


clark800:
@darkdarkdragon Would you mind picking this up? Might be best to run a converter over all the files because they are scattered all over. It would also be nice if we had a check for this in the CI tests.

darkdarkdragon:
@clark800 ok.

clark800:
Replaced by https://github.com/ripple/ripple-lib/pull/674

PR #672, opened 2015-12-12
testing circleci standalone integration

darkdarkdragon:
still testing, do not merge


darkdarkdragon:
@clark800 done

circle.yml
```diff
@@ -1,7 +1,16 @@
...
+    - sudo dpkg -i rippled_0.30-1.deb
 test:
+  pre:
+    - sudo rippled -a --start --conf "$HOME/$CIRCLE_PROJECT_REPONAME/test/integration/rippled.cfg":
+        background: true
   override:
     - scripts/ci.sh "$CIRCLE_NODE_INDEX" "$CIRCLE_NODE_TOTAL":
         parallel: true
+  post:
+    - sudo killall -r rippled
```
clark800: What is the `-r` for?

circle.yml
```diff
@@ -1,7 +1,16 @@
 machine:
   node:
     version: 0.12.0
+dependencies:
+  pre:
+    - wget https://s3-us-west-2.amazonaws.com/ripple-debs/rippled_0.30-1.deb
+    - sudo dpkg -i rippled_0.30-1.deb
```
clark800: We should use the newer one, even if both of them work.

scripts/ci.sh
```diff
@@ -38,6 +38,7 @@ unittest() {
 
 integrationtest() {
   mocha test/integration/integration-test.js
+  # mocha test/integration/integration-strange-1-test.js
```
clark800: Can delete this now?

test/integration/integration-test.js
```diff
@@ -9,11 +9,14 @@ const requests = require('../fixtures/requests');
 const RippleAPI = require('../../src').RippleAPI;
 const {isValidAddress} = require('ripple-address-codec');
 const {isValidSecret} = require('../../src/common');
+const {payTo, ledgerAccept} = require('./utils');
 
 
-const TIMEOUT = 30000;   // how long before each test case times out
+const TIMEOUT = 10000;   // how long before each test case times out
```
clark800: Good idea to reduce this

test/integration/integration-test.js
```diff
@@ -323,30 +437,11 @@ describe('integration tests', function() {
...
-  }).then(prepared => {
-    return api.submit(api.sign(prepared.txJSON, secret).signedTransaction);
-  }).then(() => {
-    return acceptLedger(api);
-  });
-}
-
-describe.skip('integration tests - standalone rippled', function() {
+describe('integration tests - standalone rippled', function() {
   const instructions = {maxLedgerVersionOffset: 10, fee: '1'};
```
clark800: Oops, I forgot to delete the "fee" here, can you delete it? I think it should work now without it.

test/integration/utils.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  let id = null;
+  return api.preparePayment(from, paymentSpecification, {})
+    .then(data => api.sign(data.txJSON, secret))
+    .then(signed => {
+      id = signed.id;
+      api.submit(signed.signedTransaction);
+    })
+    .then(() => ledgerAccept(api))
```
clark800: It might be a little better to do `.then(id => {ledgerAccept(api); return id;})` so we can get rid of the `let`.

test/integration/integration-test.js
```diff
@@ -74,17 +78,110 @@ function setup(server = 'wss://s1.ripple.com') {
...
+function makeTrustLine(testcase, address, secret) {
+  const api = testcase.api;
+  const specification = {
+    currency: 'USD',
+    counterparty: masterAccount,
+    limit: '1341.1',
+    ripplingDisabled: true
+  };
+  const trust = api.prepareTrustline(address, specification, {})
+    .then(data => api.sign(data.txJSON, secret))
```
clark800: `sign` is synchronous, so you don't have to put it in a `then`, it can go into the next block

test/integration/integration-test.js
```diff
@@ -74,17 +78,110 @@ function setup(server = 'wss://s1.ripple.com') {
...
+    .then(signed => {
+      if (address === wallet.getAddress()) {
+        testcase.transactions.push(signed.id);
+      }
+      api.submit(signed.signedTransaction);
```
clark800: I think you need a `return` here

test/integration/utils.js
```diff
@@ -0,0 +1,56 @@
...
+    paymentSpecification.source.maxAmount.counterparty = counterparty;
+    paymentSpecification.destination.amount.counterparty = counterparty;
+  }
+
+  let id = null;
+  return api.preparePayment(from, paymentSpecification, {})
+    .then(data => api.sign(data.txJSON, secret))
+    .then(signed => {
+      id = signed.id;
+      api.submit(signed.signedTransaction);
```
clark800: Need a `return` here

circle.yml
darkdarkdragon: without that it wasn't working on my local ubuntu, don't know why
it means 'regular expression'

test/integration/integration-test.js
darkdarkdragon: It didn't work - results in `telINSUF_FEE_P`

darkdarkdragon:
@clark800 done

circle.yml
clark800: Maybe because rippled is run with `sudo`? Is it possible to run `rippled` without `sudo`?

darkdarkdragon: In my tests in wasn't the case.
`rippled` can be run without sudo, but this will require change to config to use different paths to store database and logs

darkdarkdragon:
@clark800 done

circle.yml
clark800: Hmmm, I thought I made the package set `/var/lib/rippled/db/` writable by anyone. Did you still see a problem? I think it would be better to run without `sudo` even if we have to change the config file.

darkdarkdragon: I've remembered - `killall` wants full name - /usr/bin/rippled

darkdarkdragon: > I thought I made the package set /var/lib/rippled/db/ writable by anyone. 
Wasn't aware of that. Strange behavior. 
And there is log file in /var/log dir, but not sure if we need it

clark800:
LGTM

PR #674, opened 2015-12-14
add check for windows eol

darkdarkdragon:


darkdarkdragon:
@clark800 done

scripts/checkeol.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash 
+
+function checkEol {
+  local X=$(git --no-pager diff --name-only -M100% --diff-filter=AM --relative $(git merge-base FETCH_HEAD origin/HEAD) FETCH_HEAD | grep "\.js$")
```
clark800: We should probably also check JSON files, and might as well check all file types, so we can get rid of the `grep`. Should use a better variable name than `X`.

scripts/checkeol.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash 
+
+function checkEol {
```
clark800: Can we call it `checkEOL`? I think it's more conventional to capitalize `EOL`.

scripts/checkeol.sh
darkdarkdragon: ok

scripts/checkeol.sh
darkdarkdragon: ok

scripts/ci.sh
```diff
@@ -47,6 +51,7 @@ doctest() {
 }
 
 oneNode() {
+  checkEol
```
clark800: Can you change to `checkEOL` here too?

darkdarkdragon:
done

clark800:
LGTM

PR #679, opened 2015-12-17
Add multisignature support

clark800:


src/transaction/sign.js
```diff
@@ -6,23 +6,38 @@ const binary = require('ripple-binary-codec');
...
   const tx = JSON.parse(txJSON);
+  if (tx.TxnSignature || tx.Signers) {
+    throw new utils.common.errors.ValidationError(
+      'txJSON must not contain "TxnSignature" or "Signers" properties');
+  }
+
   const keypair = keypairs.deriveKeypair(secret);
-  if (tx.SigningPubKey === undefined) {
-    tx.SigningPubKey = keypair.publicKey;
+  tx.SigningPubKey = options.signFor ? '' : keypair.publicKey;
```
darkdarkdragon: Looking at the name `signFor` I thought that this means 'address of transaction issuer, for whom we are signing this transaction', but actually this means 'address of account that we are signing by now'.
Maybe `signBy` will be better name?

sublimator: >  I thought that this means 'address of transaction issuer, for whom we are signing this transaction',

In rippled, that's definitely what it means in the 'sign_for' command

sublimator: proxy ?

clark800: @sublimator  "In rippled, that's definitely what it means in the 'sign_for' command" - the documentation and test results contradict this statement: https://github.com/ripple/ripple-dev-portal/compare/gh-pages...mDuo13:simplesigning#diff-546a0b8282c4e506a611d215a8bf65ddR354

Also, what does the "proxy ?" comment mean?

clark800: My understanding of the reason for "signFor" is that, for example, multiple accounts could have the same regular key and if something is signed by that regular key, we need to know which account it counts for. How about calling it `signAs`?

darkdarkdragon: `signAs` sounds good
17 груд. 2015 20:10 "Chris Clark" <notifications@github.com> пише:

clark800:
@darkdarkdragon done

darkdarkdragon:
LGTM

sublimator:
Ah, they must have changed the sign_for command, since removing multilevel
support.

sublimator:
sign_for used to require that you explicitly list the account you were
signing for, as the txn for account A could have a signer list including
account [B,C, D, E]. C could actually sign_for account B, if C was on B's
own signing list.

These days tx_json.Account is always the account you `sign_for` (in the
original sense)

signAs works :+1:

clark800:
Ok, thanks for the clarification

PR #681, opened 2015-12-23
do not merge - testing ci

darkdarkdragon:
fix for work in browser
add run unit tests and integration tests in PhantomJS


darkdarkdragon:
@clark800 this one is done

Gulpfile.js
```diff
@@ -35,6 +45,54 @@ function webpackConfig(extension, overrides) {
...
+      'net': 'null'
+    }],
+    entry: testFileName,
+    output: {
+      library: match[1].replace(/-/g, '_'),
+      path: './test-compiled-for-web/' + (path ? path : ''),
+      filename: match[1] + '-test.js'
+    }
+  };
+  var config = webpackConfig('rangeset-test.js', configOverrides);
```
clark800: `rangeset-test.js` is hard coded here

src/common/connection.js
```diff
@@ -148,10 +148,12 @@ class Connection extends EventEmitter {
       cert: this._certificate
     }, _.isUndefined);
     const websocketOptions = _.assign({}, options, optionsOverrides);
-    const websocket = new WebSocket(this._url, websocketOptions);
+    const websocket = new WebSocket(this._url, null, websocketOptions);
```
clark800: Is this a bug? If so we should put it in a separate commit.

src/common/utils.js
```diff
@@ -35,12 +35,13 @@ function toRippledAmount(amount: Amount): RippledAmount {
   };
 }
 
-const FINDSNAKE = /([a-zA-Z]_[a-zA-Z])/g;
 function convertKeysFromSnakeCaseToCamelCase(obj: any): any {
   if (typeof obj === 'object') {
     let newKey;
     return _.reduce(obj, (result, value, key) => {
       newKey = key;
+      // taking this out of function leads to error in PhantomJS
```
clark800: Do you understand how this can cause an error in PhantomJS?

src/common/connection.js
darkdarkdragon: Not sure that we can call this a bug. Native `WebSocket` constructor only has two arguments, and `ws` package allows to skip that second argument.

src/common/wswrapper.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+const {EventEmitter} = require('events');
+
+function unsused() {}
+
+class WSWrapper extends EventEmitter {
```
clark800: Can you add a comment here explaining what the wrapper is needed for?

src/common/utils.js
darkdarkdragon: Nope. That one was strange - regular expression was returning true for match for first call, but false for other calls, though string was matching that expression.
I have no idea why.

test/integration/wallet-web.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+function getAddress() {
+  return 'rQDhz2ZNXmhxzCYwxU6qAbdxsHA4HV45Y2';
+}
+
+function getSecret() {
+  return 'shK6YXzwYfnFVn3YZSaMh5zuAddKx';
```
clark800: We should avoid putting secrets in the source code so that people don't drain the funds and make us manually reset it a lot. You should be able to get it from the environment variables on CircleCI.

darkdarkdragon: That one is from `integration-test.js` that you've put there. Anyway we don't test against live ripple network anymore, so I was thinking about completely removing `wallet.js`, but left it for local testing.

Gulpfile.js
darkdarkdragon: changed to more neutral, but this doesn't matter because it's gets overridden.

src/common/wswrapper.js
darkdarkdragon: done

test/integration/wallet-web.js
clark800: I see, so it is not a real account on the live network, then that is fine.

test/broadcast-api-test.js
```diff
@@ -39,13 +41,16 @@ describe('RippleAPIBroadcast', function() {
     });
     const ledgerNext = _.assign({}, ledgerClosed);
     ledgerNext.ledger_index++;
-    this.mocks.forEach(mock => mock.socket.send(JSON.stringify(ledgerNext)));
+
+    this.api._apis.forEach(api => api.connection._send(JSON.stringify({
+      command: 'emit',
```
clark800: Should we call this "echo"?

Gulpfile.js
```diff
@@ -35,6 +45,54 @@ function webpackConfig(extension, overrides) {
...
+      'net': 'null'
+    }],
+    entry: testFileName,
+    output: {
+      library: match[1].replace(/-/g, '_'),
+      path: './test-compiled-for-web/' + (path ? path : ''),
+      filename: match[1] + '-test.js'
+    }
+  };
+  var config = webpackConfig('.js', configOverrides);
```
clark800: Might as well merge this line and the next

src/common/utils.js
clark800: It's a bit concerning because they should have the same semantics or else we might not be testing properly.

Gulpfile.js
darkdarkdragon: This was for debugging, will fix it.

test/broadcast-api-test.js
darkdarkdragon: Good name, will rename it.

Gulpfile.js
darkdarkdragon: fixed

test/broadcast-api-test.js
darkdarkdragon: fixed

src/common/utils.js
darkdarkdragon: Yes, they should. But in PhantomJS it has problems. I've moved it back so you could see errors by yourself.
And I don't think that we testing it wrong - I think it is problem inside PhantomJS. As they documentation states, PhantomJS in not just headless chrome, it is custom build with some version of V8, that, maybe, has some bugs, or linked with bugs.
We will be more sure when I will add Sauce Labs testing, so we can see test results in real browsers.

clark800:
Ok, you can fix the phantomjs issue, I think it's ready for merge

darkdarkdragon:
done

PR #686, opened 2016-01-12
switch from Babel 5 to Babel 6

darkdarkdragon:


clark800:
Looks like you deleted npm-shrinkwrap.json

darkdarkdragon:
@clark800 done

npm-shrinkwrap.json
```diff
@@ -23,155 +22,186 @@
...
           }
         },
         "extend": {
           "version": "3.0.0",
-          "from": "https://registry.npmjs.org/extend/-/extend-3.0.0.tgz",
           "resolved": "https://registry.npmjs.org/extend/-/extend-3.0.0.tgz"
         }
       }
     },
+    "is-my-json-valid": {
```
clark800: I don't think we depend on this anymore. Also "lru-cache".

src/common/connection.js
```diff
@@ -176,7 +176,8 @@ class Connection extends EventEmitter {
         this._ws.on('error', error =>
           this.emit('error', 'websocket', error.messsage, error));
         this._ws.on('message', this._onMessage.bind(this));
-        this._onUnexpectedCloseBound = this._onUnexpectedClose.bind(this);
+        this._onUnexpectedCloseBound = this._onUnexpectedClose.bind(this,
+          resolve, reject);
```
clark800: What is this change for? Isn't this going to make the promise double-fulfill?

test/api-test.js
```diff
@@ -325,6 +326,15 @@ describe('RippleAPI', function() {
     schemaValidator.schemaValidate('sign', result);
   });
 
+  it('sign - already signed', function() {
```
clark800: Thanks for adding coverage :)

darkdarkdragon: You're welcome! :smile:

src/common/connection.js
darkdarkdragon: Yep, you right, missed other case. Problem was when `close` event comes before `open` - that leaves first promise, returned from `connect` unfilled. 
I fixed that now.

darkdarkdragon:
@clark800 done now

npm-shrinkwrap.json
```diff
@@ -23,155 +22,154 @@
...
           "resolved": "https://registry.npmjs.org/core-js/-/core-js-1.2.6.tgz"
         }
       }
     },
     "bignumber.js": {
       "version": "2.1.2",
-      "from": "https://registry.npmjs.org/bignumber.js/-/bignumber.js-2.1.2.tgz",
       "resolved": "https://registry.npmjs.org/bignumber.js/-/bignumber.js-2.1.2.tgz"
     },
+    "create-hash": {
```
clark800: What about this?

src/common/connection.js
clark800: I don't think the problem is fixed. If this method succeeds, `resolve` will be called because of the last line. Then at some later time, if the connection is closed unexpectedly, resolve will get called again.

darkdarkdragon: Fix it here 
https://github.com/ripple/ripple-lib/pull/686/files#diff-c5fa987b0954acfe830ac6adda20f19fR106

clark800: Ok, I see. But I don't understand: how can we get a `close` event before an `open` event if we are starting in a non-open state?

darkdarkdragon: I don't know too :smile: I've whole evening yesterday trying to fix seemingly random in unit tests inside PhatomJS and finally found that this was the cause.

npm-shrinkwrap.json
darkdarkdragon: Removed

src/common/connection.js
clark800: Hmmm, maybe because it is in the state `closing` when we enter this method. Maybe it would be cleaner to check for the `closing` state and if so, wait for the `close` event before proceeding. To do this I think we would have to extract the `else` block to a separate function. I think it would be a bit simpler because the current code is getting pretty complex.

darkdarkdragon: it can't be in any state because whole `RippleAPI` object was just created.

clark800: Ok, well it sounds like we don't understand what is going on here, so future readers definitely are going to get confused. I think we should probably add a big comment for this.

darkdarkdragon: Hm. This is one is simple - after creating new `RippleAPI` object and calling `connect`, `close` event can came before `open`. Maybe it is not according to specification, but it happens in real life :smile: 
So code just checks for that situation.

clark800: "close event can came before open" - this is the type of thing that the reader would never guess and should be documented in a comment. Also, it should explain how the `_onUnexpectedCloseBound` is only _temporarily_ assigned here and gets reassigned in `_onOpen`.

darkdarkdragon: I've added comment

src/common/errors.js
```diff
@@ -1,7 +1,18 @@
...
+  function ExtendableBuiltin() {
+    cls.apply(this, arguments);
+  }
+  ExtendableBuiltin.prototype = Object.create(cls.prototype);
+  Object.setPrototypeOf(ExtendableBuiltin, cls);
+
+  return ExtendableBuiltin;
+}
+
+class RippleError extends extendableBuiltin(Error) {
```
lumberj: I think it would be cleaner to create a BaseError and extend that like this:

```javascript
function BaseError(message) {
  Error.captureStackTrace(this, this.constructor);

  Object.defineProperty(this, 'name', {
    get: () => this.constructor.name
  });
}
require('util').inherits(BaseError, Error);

class RippleError extends BaseError { /* ... */
```

lumberj: Or, just define RippleError using the ES5 class syntax

clark800: Hmm, I'm not sure which is better. Ivan's version looks more general and doesn't really seem more complicated.

clark800:
LGTM (will wait for @darkdarkdragon to respond to @lumberj 's suggestion)

src/common/errors.js
lumberj: @darkdarkdragon did you file an issue for Babel?

darkdarkdragon: @lumberj it's not thing that I've found first
https://phabricator.babeljs.io/T3083
in short - subclassing builtins wasn't supported in Babel 5 either, it just behaved differently

lumberj: This works fine:

```js
'use strict'

class MyError extends Error {
  constructor (message) {
    super()

    // Set this.message
    Object.defineProperty(this, 'message', {
      configurable: true,
      enumerable: false,
      value: message !== undefined ? String(message) : ''
    })

    // Set this.name
    Object.defineProperty(this, 'name', {
      configurable: true,
      enumerable: false,
      value: this.constructor.name
    })

    // Set this.stack
    Error.captureStackTrace(this, this.constructor)
  }
}


console.log((new MyError('foo') instanceof Error))
```


https://github.com/interledger/five-bells-shared/blob/master/errors/base-error.js

darkdarkdragon: It doesn't work
`console.log((new MyError('foo') instanceof MyError));`  - `false`

clark800: Note: five-bells-shared doesn't use Babel

lumberj: I ran this code locally using babel node

lumberj:
```
~ » babel-node --version
6.3.17
~ » babel-node test2.js
true
```

darkdarkdragon: Have you tried 
`console.log((new MyError('foo') instanceof MyError))`
?

lumberj: Yes. That returns true

clark800: Works for me too with babel-node v6.2.0. @darkdarkdragon what version are you using?

lumberj: Just updated to Babel 6.4.5 and tried again, got `true` for both

lumberj: Ah, I think I know what's happening. Babel 6 doesn't do anything by default. So, the code runs, because I'm running this on node v4. I need to try running this with the transform on

darkdarkdragon: 6.4.0

lumberj: I just checked out your PR and got this when I ran `npm install`

```
> ripple-lib@0.16.5 compile /Users/alan/Projects/ripple-lib
> babel -D --optional runtime -d dist/npm/ src/

module.js:339
    throw err;
    ^

Error: Cannot find module 'babel-core/lib/transformation/modules'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/alan/Projects/ripple-lib/node_modules/babel/lib/babel/index.js:6:24)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Module.require (module.js:366:17)
```

darkdarkdragon: strange. just checked - it worked ok

clark800: So do you want to make the change? I think it looks a bit simpler.

darkdarkdragon: I mean checking out fresh copy and running `npm install` worked

lumberj: Still trying to test this locally. npm is slow

lumberj: Ok. So got everything working, and as @darkdarkdragon it doesn't work as I wrote it originally. However, this works:

```
'use strict'

class MyError {
  constructor (message) {

    // Set this.message
    Object.defineProperty(this, 'message', {
      configurable: true,
      enumerable: false,
      value: message !== undefined ? String(message) : ''
    })

    // Set this.name
    Object.defineProperty(this, 'name', {
      configurable: true,
      enumerable: false,
      value: this.constructor.name
    })

    // Set this.stack
    Error.captureStackTrace(this, this.constructor)
  }
}

require('util').inherits(MyError, Error)



console.log((new MyError('foo') instanceof Error))
console.log((new MyError('foo') instanceof MyError))
```

```
true
true
```

lumberj: Extended the example, to double-check:

```js
'use strict'

class MyError {
  constructor (message) {

    // Set this.message
    Object.defineProperty(this, 'message', {
      configurable: true,
      enumerable: false,
      value: message !== undefined ? String(message) : ''
    })

    // Set this.name
    Object.defineProperty(this, 'name', {
      configurable: true,
      enumerable: false,
      value: this.constructor.name
    })

    // Set this.stack
    Error.captureStackTrace(this, this.constructor)
  }
}

require('util').inherits(MyError, Error)

class MyError2 extends MyError {
  constructor (message) {
    super(message)
  }
}



console.log((new MyError('foo') instanceof Error))
console.log((new MyError('foo') instanceof MyError))
console.log((new MyError2('foo') instanceof MyError))
console.log((new MyError2('foo') instanceof Error))
console.log((new MyError2('foo') instanceof MyError2))
```

```
true
true
true
true
true
```

lumberj: I don't think this works under nodev4 without babel

lumberj: Confirmed, this does not work under node v4. Is the reason we're doing this here for browser support?

darkdarkdragon: What variant doesn't work under v4?
`extendableBuiltin(Error)` was added for node 0.12 support
For browsers support, actually, there will be more hack :smile:

lumberj: The variant I posted above doesn't work under nodev4, but compiles successfully using Babel. I think we should just go with your variant for now until we can drop babel altogether :)

lumberj:
LGTM

src/common/errors.js
darkdarkdragon: I don't think we will be able to drop Babel, at least until we using Flow :smile:

clark800: @darkdarkdragon Wait, do you mean this is breaking browser support? Or just that it has to go through Babel before running in the browser?

darkdarkdragon: @clark800 no, it is not breaking browser support. I'm testing now in different browsers, and internet explorer has couple glitches in other places.

PR #690, opened 2016-01-28
add `create-hash` into npm-shrinkwrap.json

darkdarkdragon:
without that npm doesn't install `create-hash` in modules that depends on it


clark800:
This can't be the right solution: ripple-lib doesn't use create-hash. What's going on here?

darkdarkdragon:
`ripple-binary-codec` and `ripple-hashes` depends on `create-hash`
`create-hash` doesn't get installed because it is not listed as dependency in`npm-shrinkwrap.json` for corresponding modules.
so, I must put it under `ripple-hashes` and `ripple-binary-codec` - and it will be installed into the both location, or put it as direct dependency to `ripple-lib` and it will be installed only once

clark800:
It should be added to the packages that use it

darkdarkdragon:
ok

clark800:
Those packages don't use shrinkwrap though, so I'm confused why it's not working.

darkdarkdragon:
@clark800 can, you, please, restart CircleCI build without cache?

clark800:
Sure, running now

darkdarkdragon:
thanks

PR #691, opened 2016-01-30
add SauceLabs testing

darkdarkdragon:
add SauceLabs testing


darkdarkdragon:
@clark800 done

package.json
```diff
@@ -50,13 +50,16 @@
...
     "gulp-rename": "^1.2.0",
     "gulp-uglify": "^1.1.0",
+    "http-server": "^0.8.5",
     "isparta": "^4.0.0",
     "json-loader": "^0.5.2",
     "json-schema-to-markdown-table": "^0.4.0",
     "mocha": "^2.1.0",
     "mocha-junit-reporter": "^1.9.1",
     "mocha-phantomjs": "^4.0.1",
+    "mocha-sauce": "git://github.com/darkdarkdragon/mocha-sauce.git",
```
clark800: We can't depend on private github repos

src/common/errors.js
```diff
@@ -7,15 +7,23 @@ function extendableBuiltin(cls) {
...
+  /* eslint-enable */
   return ExtendableBuiltin;
 }
 
 class RippleError extends extendableBuiltin(Error) {
   constructor(message, data) {
     super(message);
-    this.name = this.constructor.name;
+
+    // hack for internet explorer
```
clark800: Does lodash have any methods that might help for these? If not, it would be nice to extract all the "browser hacks" to a single file at src/common/browser-hacks.js to keep them out of our real code.

clark800:
There's quite a bit of boilerplate, especially scripts/sauce-runner.js. Are there any libraries that do any of this?

darkdarkdragon:
It is the best way it can be done :smile: 
SauceLabs only has grunt plugin, and I'm not sure if it is maintained - in they example they making local clone of old version of mocha, so I didn't looked deeper.
Only mature (I think) solution available - it is sauce runner for karma, but karma works with broserify instead of webpack, and I wasn't able to compile our project with browserify - it doesn't handle flow annotations well.
If I move `sauce-runner.js` functionality into my mocha-sause repo and will publish it into npmjs it will look better?

clark800:
> it doesn't handle flow annotations well

The browser code has flow annotations stripped out by Babel already, so I don't understand. If we supported browserify would if significantly reduce the complexity?

darkdarkdragon:
I don't remember now it was problem with karma or with browserify itself, but they build process made in such way that it tries to parse code by itself, before feeding it to babel, and, obviously, it failed. 
> If we supported browserify would if significantly reduce the complexity?
I wasn't able to run it, so I don't know if it even works. I know that sauce-runner for karma exists, but I haven't tested it, and I don't know if it will handle our case.
And karma is complex beast by itself, and it add significant runtime complexity.

clark800:
Ok. Let me know when the other comments are handled :)

darkdarkdragon:
done. hacks are moved out, and sauce-runner much smaller now :smile:

test/saucerunner.html
```diff
@@ -0,0 +1,32 @@
...
+    <script src="/test/vendor/lodash.min.js"></script>
+  </head>
+  <body>
+    <div id="deb"></div>
+    <div id="mocha"></div>
+    <script src="/node_modules/mocha/mocha.js"></script>
+
+    <script src="/node_modules/mocha-sauce/client.js"></script>
+
+    <script src="/test-compiled-for-web/ripple-for-web-tests.js"></script>
```
clark800: This looks like something I missed in a previous pull request. Why is there a "ripple-for-web-tests.js"? Can't we just test the _exact_ file that is built for production web use?

darkdarkdragon: It is exactly same, byte-to-byte. Standard build creates files with version number in name, so it will require introducing some form of templating into html.

clark800: How about making the build script generate a symlink "latest.js" that points to the most recent build?

darkdarkdragon: I think this can be done. Do you thing this would be better?

clark800: Yeah, I think so. When there is a separate build it is harder to know for sure that it will always be identical. And the symlink could be useful for other situations too.

darkdarkdragon: ok

darkdarkdragon:
done

clark800:
LGTM

PR #693, opened 2016-02-10
Add deliveredAmount to payment outcome

lumberj:


src/ledger/parse/utils.js
```diff
@@ -65,7 +73,8 @@ function parseOutcome(tx: Object): ?Object {
     balanceChanges: balanceChanges,
     orderbookChanges: orderbookChanges,
     ledgerVersion: tx.ledger_index,
-    indexInLedger: tx.meta.TransactionIndex
+    indexInLedger: tx.meta.TransactionIndex,
+    deliveredAmount
```
clark800: I think we need a `removeUndefined` for this object now.

clark800:
Is this new field redundant with the amount in the balance changes? If not, I think the documentation should explain why.

lumberj:
Good point. This also should take into account partial payments. I'll
correct this implementation and update with an explanation.

PR #694, opened 2016-02-11
Add deliveredAmount to payment outcome

lumberj:
"It is impossible to reliably compute the delivered amount from the metadata due
to fixed precision. If the partial payment flag is not set and the transaction
succeeded, the delivered amount should always be considered to be the amount
specified in the transaction."

### DIFF
--- a/src/ledger/parse/utils.js
+++ b/src/ledger/parse/utils.js
@@ -38,14 +38,14 @@ function removeEmptyCounterparty(amount) {
 }
 
 function removeEmptyCounterpartyInBalanceChanges(balanceChanges) {
-  _.forEach(balanceChanges, (changes) => {
+  _.forEach(balanceChanges, changes => {
     _.forEach(changes, removeEmptyCounterparty);
   });
 }
 
 function removeEmptyCounterpartyInOrderbookChanges(orderbookChanges) {
-  _.forEach(orderbookChanges, (changes) => {
-    _.forEach(changes, (change) => {
+  _.forEach(orderbookChanges, changes => {
+    _.forEach(changes, change => {
       _.forEach(change, removeEmptyCounterparty);
     });
   });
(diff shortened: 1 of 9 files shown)


lumberj:
@clark800 added a check for partial payment and also added a more detailed description of why it is necessary to use `deliveredAmount` instead of `balanceChanges`

src/ledger/parse/utils.js
```diff
@@ -35,19 +38,37 @@ function removeEmptyCounterparty(amount) {
...
+function parseDeliveredAmount(tx: Object): Amount | void {
+  let deliveredAmount;
+
+  if (tx.TransactionType === 'Payment') {
+    if (tx.meta.delivered_amount) {
+      deliveredAmount = parseAmount(tx.meta.delivered_amount);
+    } else if (tx.Amount && !isPartialPayment(tx)) {
+      deliveredAmount = parseAmount(tx.Amount);
+    }
+  }
```
clark800: So it's possible that `deliveredAmount` will be undefined; do we understand when and why this case happens?

lumberj: It's caused by a rippled bug

lumberj: rippled should always provide delivered_amount if it's a partial payment. In the case that it's not a partial payment, you can use the Amount. I think there's a bug in the current patch version of rippled that sometimes leaves off delivered_amount, for account_tx

clark800: Ok, so how about we add a task to clean this up after rippled fixes the bug and add a comment to link to the task as an explanation?

lumberj: Added a comment

lumberj: https://github.com/ripple/ripple-lib/issues/696

lumberj: I'll add a link to the source after merge

clark800:
LGTM

PR #699, opened 2016-02-24
[FIX] fix connection error handling

darkdarkdragon:
if connection to server can't be established, reject Promise of `connect()` method caller,
instead of throwing error event on base object.

### DIFF
--- a/src/common/connection.js
+++ b/src/common/connection.js
@@ -97,6 +97,10 @@ class Connection extends EventEmitter {
   }
 
   _onUnexpectedClose(resolve = function() {}, reject = function() {}) {
+    if (this._onOpenErrorBound) {
+      this._ws.removeListener('error', this._onOpenErrorBound);
+      this._onOpenErrorBound = null;
+    }
     this._ws = null;
     this._isReady = false;
     this.connect().then(resolve, reject);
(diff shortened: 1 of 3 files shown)


lumberj:
Looks like we need to update eslintrc, http://eslint.org/docs/rules/space-after-keywords -- this rule was deprecated in eslint 2.0

lumberj:
@clark800 https://github.com/ripple/javascript-style-guide/pull/26

clark800:
Actually the problem is that the `es6` branch of the style guide already got updated for eslint 2.x, but ripple-lib is pinned at 1.x. I think we can just update to 2.x in the package.json to fix this issue.

src/common/connection.js
```diff
@@ -118,6 +126,10 @@ class Connection extends EventEmitter {
     });
   }
 
+  _onOpenError(reject, error) {
+    reject(new DisconnectedError(error && error.message));
```
clark800: I think we probably want `NotConnectedError` here; `DisconnectedError` is for when it was connected but got disconnected, whereas `NotConnectedError` is for when it just wasn't connected in the first place.

darkdarkdragon: You right, will fix it

darkdarkdragon:
@clark800 fixed

clark800:
LGTM

PR #704, opened 2016-03-15
[FIX] fix test hang in microsoft edge

darkdarkdragon:


lumberj:
LGTM

test/connection-test.js
```diff
@@ -117,7 +117,8 @@ describe('Connection', function() {
       }
     }
 
-    const connection = new utils.common.Connection('ws://127.0.0.1:321');
+    const connection =
+      new utils.common.Connection('ws://testripple.circleci.com:129');
```
clark800: I assume this is not a valid server, but might be a good thing to comment

PR #709, opened 2016-03-22
Reconnection fix

darkdarkdragon:
Alson add `connected` and `disconnected` events.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -61,6 +61,8 @@
 - [API Events](#api-events)
   - [ledger](#ledger)
   - [error](#error)
+  - [connected](#connected)
+  - [disconnected](#disconnected)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
@@ -89,6 +91,14 @@ const api = new RippleAPI({
 });
 api.on('error', (errorCode, errorMessage) => {
   console.log(errorCode + ': ' + errorMessage);
+});
+api.on('connected', () => {
+  console.log('connected');
+});
+api.on('disconnected', (code) => {
+  // code - [close code](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent) sent by the server
+  // will be 1000 if this was normal closure
+  console.log('disconnected, code:', code);
 });
 api.connect().then(() => {
   /* insert code here */
@@ -3591,5 +3601,37 @@ api.on('error', (errorCode, errorMessage, data) => {
 
 ```
 tooBusy: The server is too busy to help you now.
+```
+
+## connected
+
+This event is emitted after connection successfully opened.
+
+### Example
+
+```javascript
+api.on('connected', () => {
+  console.log('Connection is open now.');
+});
+```
+
+## disconnected
+
+This event is emitted when connection is closed.
+
+### Return Value
+
+The only parameter is a number containing the [close code](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent) send by the server.
+
+### Example
+
+```javascript
+api.on('disconnected', (code) => {
+  if (code !== 1000) {
+    console.log('Connection is closed due to error.');
+  } else {
+    console.log('Connection is closed normally.');
+  }
+});
 ```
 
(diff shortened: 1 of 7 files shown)


docs/index.md
```diff
@@ -90,6 +92,13 @@ const api = new RippleAPI({
 api.on('error', (errorCode, errorMessage) => {
   console.log(errorCode + ': ' + errorMessage);
 });
+api.on('connected', () => {
+  console.log('connected');
+});
+api.on('disconnected', (onError) => {
+  // onError will be true if disconnected due to error
```
lumberj: Is it possible to pass the error in here?

src/common/connection.js
```diff
@@ -95,14 +97,44 @@ class Connection extends EventEmitter {
...
+      this.emit('disconnected', true);
+      this._retryConnect();
+    }
+  }
+
+  _retryConnect() {
+    this._retry += 1;
+    const retryTimeout = (this._retry < 40)
+      // First, for 2 seconds: 20 times per second
+      ? (1000 / 20)
```
lumberj: Can you move this logic into a separate function? The nested ternary statements are confusing to follow

lumberj:
cc @clark800

src/api.js
```diff
@@ -1,5 +1,4 @@
 /* @flow */
-'use strict';
```
clark800: Is this the right thing to do? Do you know why it is causing this lint error now?

src/common/connection.js
clark800: Another option that might be simpler is to use a logarithmic function for this

docs/index.md
clark800: Also, `onError` is the name of a function, if it's a boolean it should be `isError` or something

src/api.js
clark800: The best option I have seen so far is to keep the "use strict" and add ` // eslint-disable-line strict` so that it doesn't create an error.

darkdarkdragon:
@clark800 @lumberj fixed

src/common/connection.js
darkdarkdragon: Using math functions here will make result not obvious.

src/common/connection.js
```diff
@@ -95,14 +99,50 @@ class Connection extends EventEmitter {
...
     this._isReady = false;
-    this.connect().then(resolve, reject);
+    if (_.isFunction(resolve)) {
+      // connection was closed before it was properly opened, so we must return
+      // error to connect's caller
+      this.connect().then(resolve, reject);
+    } else {
+      // if first parameter ws lib sends close code,
+      // but sometimes it forgots about it, so default to 1006 - CLOSE_ABNORMAL
+      this.emit('disconnected', resolve || 1006);
```
clark800: This is strange, `resolve` can be a number? It seems like it should be a type error if resolve is not a function.

clark800: Can we change it so that onUnexpectedCloseBound always has resolve and reject bound to functions? In the case that they aren't necessary it could just bind them to the empty function.

darkdarkdragon: yeah, but this will require adding another parameter, so we can check if we need to call this resolve function

clark800: You can just always call it in this case

darkdarkdragon: I need it not only for that, but to decide what to do next (line 109)

clark800: Ok. Being explicit with an additional boolean parameter is definitely preferable to encoding the information in this implicit way. Is there any way to distinguish these two cases using other properties of the class? If not, then an additional parameter sounds fine.

darkdarkdragon: you're right, I'll add additional parameter

darkdarkdragon:
@clark800 fixed. sauce labs tests hasn't started, but I think this is some infrastructure problem, I hope next time they will work. So I think you can merge, or try to restart circleci build.

darkdarkdragon:
@clark800 I can't tell fast what is wrong. I've run this tests inside sauce labs manually and they passed.

clark800:
I'll try running again. If it doesn't work let's try again tomorrow

darkdarkdragon:
ok

clark800:
LGTM

PR #711, opened 2016-03-24
[FIX] on connect return error if server doesn't have validated ledgers

darkdarkdragon:


clark800:
LGTM

lumberj:
Isn't this a bug in rippled?

lumberj:
I don't think we're ever testing with servers that have no ledgers

darkdarkdragon:
this happens when server just started and not synced yet. also, when you stop synced server and start it again for some time it is not synced too

lumberj:
> also, when you stop synced server and start it again for some time it is not synced too

Ok makes sense

PR #712, opened 2016-03-29
[FIX] on reconnect wait for server to be synced

darkdarkdragon:
on reconnect do not emit connected event till server_state is full

handle websocket errors in browsers

### DIFF
--- a/src/common/connection.js
+++ b/src/common/connection.js
@@ -6,7 +6,8 @@ const WebSocket = require('ws');
 const parseURL = require('url').parse;
 const RangeSet = require('./rangeset').RangeSet;
 const {RippledError, DisconnectedError, NotConnectedError,
-  TimeoutError, ResponseFormatError, ConnectionError} = require('./errors');
+  TimeoutError, ResponseFormatError, ConnectionError,
+  RippledNotInitializedError} = require('./errors');
 
 function isStreamMessageType(type) {
   return type === 'ledgerClosed' ||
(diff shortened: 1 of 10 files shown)


src/common/connection.js
```diff
@@ -162,15 +169,27 @@ class Connection extends EventEmitter {
       }
 
       this._updateLedgerVersions(data);
+      return this.request({command: 'server_info'});
+    }).then(serverInfo => {
+      const serverState = _.get(serverInfo, 'info.server_state');
+      if (serverState !== 'full') {
+        // rippled instance still syncing, so it can't be used
+        return this._disconnect(false).then(() => {
+          throw new NotConnectedError('Rippled is syncing');
```
clark800: How about `RippledNotSyncedError` to distinguish it?

clark800:
The connection class is getting more and more complicated, is there any refactoring that can be done to make it more manageable?

src/common/connection.js
darkdarkdragon: Added

darkdarkdragon:
yes, I was thinking about that. One thing I can think out - it is to move all connect/reconnect/error handling logic into base class or to separate ws wrapper.

clark800:
Yeah, I was thinking along the same lines: a `ws` wrapper that separates out the error handling and all the event handlers. Would you be up for working on that?

darkdarkdragon:
@clark800 fixed

darkdarkdragon:
@clark800 done

clark800:
LGTM

PR #721, opened 2016-06-16
[DOC] fix transferRate description

mDuo13:
The description field for the transferRate settings field incorrectly states that the value is in "billionths" (as it is represented in rippled). RippleAPI actually takes it as a decimal (same as Ripple-REST).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -498,7 +498,7 @@ signers | object | *Optional* Settings that determine what sets of accounts can 
 *signers.* weights[] | object | An association of an address and a weight.
 *signers.weights[].* address | [address](#ripple-address) | A Ripple account address
 *signers.weights[].* weight | integer | The weight that the signature of this account counts as towards the threshold.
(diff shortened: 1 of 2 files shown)


mDuo13:
Not sure why the CI checks failed. It doesn't seem to be related to my changes.

darkdarkdragon:
LGTM
@clark800 coverage generation fails with all the latest dependencies installed. If I remove `^` from versions in `package.json` coverage gets generated (but generated html report has some problems with formatting, report for coveralls I haven't checked)
I've tried to make coverage work with latest libs, but no luck. Will make another attempt later.

clark800:
The documentation test also would have failed here; the documentation sources weren't updated, so these changes will be deleted the next time somebody updates the docs.

mDuo13:
@clark800 The "documentation sources" _were_ updated in this PR: the change was to `src/common/schemas/objects/settings.json` only. The changes to `index.md` are the result of re-generating the documentation from source.

clark800:
@mDuo13 Ah, got it, thanks!

PR #724, opened 2016-07-14
fix missing deliveredAmount data from getLedger requests

shekenahglory:
ledger data from rippled includes meta.DeliveredAmount but not meta.delivered_amount, so it was not processed correctly on transactions coming from the 'getLedger' function vs. 'getTransaction'

### DIFF
--- a/src/ledger/parse/ledger.js
+++ b/src/ledger/parse/ledger.js
@@ -1,5 +1,5 @@
 /* @flow */
-'use strict';
+'use strict'; // eslint-disable-line
 const _ = require('lodash');
 const {removeUndefined, rippleTimeToISO8601} = require('./utils');
 const parseTransaction = require('./transaction');
@@ -8,6 +8,11 @@ import type {GetLedger} from '../types.js';
 function parseTransactionWrapper(ledgerVersion, tx) {
   const transaction = _.assign({}, _.omit(tx, 'metaData'),
     {meta: tx.metaData});
+
+  // normalize format from rippled
+  transaction.meta.delivered_amount = transaction.meta.DeliveredAmount;
+  delete transaction.meta.DeliveredAmount;
+
   const result = parseTransaction(transaction);
   if (!result.outcome.ledgerVersion) {
     result.outcome.ledgerVersion = ledgerVersion;
--- a/test/api-test.js
+++ b/test/api-test.js
@@ -1171,6 +1171,16 @@ describe('RippleAPI', function() {
       _.partial(checkResult, responses.getLedger.withSettingsTx, 'getLedger'));
   });
 
+  it('getLedger - with partial payment', function() {
+    const request = {
+      includeTransactions: true,
+      includeAllData: true,
+      ledgerVersion: 100000
+    };
+    return this.api.getLedger(request).then(
+      _.partial(checkResult, responses.getLedger.withPartial, 'getLedger'));
+  });
+
   it('getLedger - full, then computeLedgerHash', function() {
     const request = {
       includeTransactions: true,
--- a/test/fixtures/responses/get-ledger-with-partial-payment.json
+++ b/test/fixtures/responses/get-ledger-with-partial-payment.json
@@ -1,0 +1,391 @@
+{
+  "stateHash": "334EE5F2209538C3099E133D25725E5BFEB40A198EA7028E6317F13E95D533DF",
+  "closeTime": "2016-07-07T16:53:51.000Z",
+  "closeTimeResolution": 10,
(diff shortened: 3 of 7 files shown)


lumberj:
If `delivered_amount` is not available, that sounds like a bug.

> The delivered_amount field of transaction metadata is included in all successful Payment 

> transactions, and is formatted like a normal currency amount. However, the delivered amount is not available for transactions that meet both of the following criteria:
> 
> Is a partial payment, and
> Included in a validated ledger before 2014-01-20
> If both conditions are true, then delivered_amount contains the string value unavailable instead of an actual amount. If this happens, you can only figure out the actual delivered amount by reading the AffectedNodes in the transaction's metadata

https://ripple.com/build/transactions/#delivered-amount

src/ledger/parse/ledger.js
```diff
@@ -8,6 +8,11 @@ import type {GetLedger} from '../types.js';
 function parseTransactionWrapper(ledgerVersion, tx) {
   const transaction = _.assign({}, _.omit(tx, 'metaData'),
     {meta: tx.metaData});
+
+  // normalize format from rippled
+  transaction.meta.delivered_amount = transaction.meta.DeliveredAmount;
```
lumberj: This is not correct if it's a partial payment.

> The Amount of a Payment transaction indicates the amount to deliver to the Destination, so if the transaction was successful, then the destination received that much -- except if the transaction was a partial payment. (In that case, any positive amount up to Amount might have arrived.) Rather than choosing whether or not to trust the Amount field, you should use the delivered_amount field of the metadata to see how much actually reached its destination.

https://ripple.com/build/transactions/#delivered-amount

test/api-test.js
```diff
@@ -1171,6 +1171,19 @@ describe('RippleAPI', function() {
...
   });
 
+  it('getLedger - with partial payment', function() {
+    const request = {
+      includeTransactions: true,
+      includeAllData: true,
+      ledgerVersion: 100000
+    };
+    return this.api.getLedger(request).then(l => {
+      console.log(l.transactions);
```
lumberj: I think you left this debug log in by accident

shekenahglory:
rippled TX response includes both `DeliveredAmount` and `delivered_amount`, while ledger response includes only `DeliveredAmount`.  so the data is there, its just not named consistently.

lumberj:
> rippled TX response includes both DeliveredAmount and delivered_amount, while ledger response includes only DeliveredAmount. so the data is there, its just not named consistently.

My understanding is those two amounts are semantically different

lumberj:
What's a "ledger response"?

test/api-test.js
shekenahglory: yep

lumberj:
Do you mean using the `ledger` command and specifying `transactions: true` ?

https://ripple.com/build/rippled-apis/#ledger

shekenahglory:
https://ripple.com/build/websocket-tool/#ledger - rippleAPI's `getLedger` queries this data and formats it, but the transaction data is formatted slightly differently than what you get when you query rippled for a single transaction

lumberj:
Is that request supposed to return transactions?

lumberj:
Oh, the doc say "Admin required". I was wondering why it didn't return any

src/ledger/parse/ledger.js
shekenahglory: there is a step in processing later where rippleAPI's deliveredAmount gets determined from either rippled's 'meta.delivered_amount' or 'tx.Amount' based on whether or not the flag is present - but that was just to account for 'delivered_amount' not being present in older transactions.  as far as I know, 'meta.DeliveredAmount` should be exactly the same as `meta.delivered_amount`, just for some reason its not named the same from the `ledger` response.

shekenahglory:
So it looks like the rippled response on the `ledger` command still uses the old method of providing 'DeliveredAmount' on partial payment transactions and nothing otherwise - the workaround here and existing code deals with it correctly I believe, though hopefully it can be changed easily in rippled regardless.  I guess whether its fixed or not, this workaround would be necessary for older ledgers before 2014

src/ledger/parse/ledger.js
lumberj: Maybe we can add a comment indicating this is a bug in rippled. The rippled docs say this is deprecated.

> DeliveredAmount	String or Object	DEPRECATED. Replaced by delivered_amount. Omitted if not a partial payment.

Otherwise, LGTM

shekenahglory:
I'm going withdraw, discovered a related bug and got a better understanding of rippled responses

PR #725, opened 2016-07-18
update parseDeliveredAmount for cases where delivered_amount is not p…

shekenahglory:
following logic used in rippled for handling delivered amount

### DIFF
--- a/src/ledger/parse/ledger.js
+++ b/src/ledger/parse/ledger.js
@@ -1,13 +1,15 @@
 /* @flow */
-'use strict';
+'use strict'; // eslint-disable-line
 const _ = require('lodash');
 const {removeUndefined, rippleTimeToISO8601} = require('./utils');
 const parseTransaction = require('./transaction');
 import type {GetLedger} from '../types.js';
 
 function parseTransactionWrapper(ledgerVersion, tx) {
-  const transaction = _.assign({}, _.omit(tx, 'metaData'),
-    {meta: tx.metaData});
+  const transaction = _.assign({}, _.omit(tx, 'metaData'), {
+    meta: tx.metaData,
+    ledger_index: ledgerVersion
+  });
   const result = parseTransaction(transaction);
   if (!result.outcome.ledgerVersion) {
     result.outcome.ledgerVersion = ledgerVersion;
--- a/src/ledger/parse/utils.js
+++ b/src/ledger/parse/utils.js
@@ -1,5 +1,5 @@
 /* @flow */
-'use strict';
+'use strict'; // eslint-disable-line
 const _ = require('lodash');
 const transactionParser = require('ripple-lib-transactionparser');
 const utils = require('../utils');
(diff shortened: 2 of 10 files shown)


src/ledger/parse/utils.js
```diff
@@ -58,13 +58,32 @@ function isPartialPayment(tx) {
...
+    } else if (tx.meta.DeliveredAmount) {
+      deliveredAmount = parseAmount(tx.meta.DeliveredAmount);
+
+    // no partial payment flag
     } else if (tx.Amount && !isPartialPayment(tx)) {
       deliveredAmount = parseAmount(tx.Amount);
+
+    // partial payment after 4594094 without DeliveredAmount
+    } else if (tx.Amount && tx.ledger_index > 4594094) {
+      deliveredAmount = parseAmount(tx.Amount);
```
lumberj: If it's a partial payment, doesn't that mean the delivered amount is _not_ equal to `tx.Amount`?

shekenahglory: nope, partial payment flag means it can be successful even if its not the
full amount, but `DeliveredAmount` only shows up if it was less than the
full amount - so after 4594094 a tx with partial payment flag without
DeliveredAmount means it delivered the full amount.

lumberj:
LGTM

docs/index.md
```diff
@@ -839,6 +839,7 @@ outcome | object | The outcome of the transaction (what effects it had).
 *outcome.* ledgerVersion | integer | The ledger version that the transaction was validated in.
 *outcome.* indexInLedger | integer | The ordering index of the transaction in the ledger.
 *outcome.* deliveredAmount | [amount](#amount) | *Optional* For payment transactions, it is impossible to reliably compute the actual delivered amount from the balanceChanges due to fixed precision. If the payment is not a partial payment and the transaction succeeded, the deliveredAmount should always be considered to be the amount specified in the transaction.
+*outcome.* deliveredAmount | string | *Optional* The delivered amount is reported as 'unavailable' for partial payment transactions before 2014
```
clark800: `deliveredAmount` is already an optional field, so why do you have to say that it is "unavailable" as opposed to just not including it?

src/ledger/parse/utils.js
```diff
@@ -58,13 +58,32 @@ function isPartialPayment(tx) {
...
-  // TODO: Workaround for existing rippled bug where delivered_amount may not be
-  // provided for account_tx
-  if (tx.TransactionType === 'Payment') {
-    if (tx.meta.delivered_amount) {
+  if (tx.TransactionType === 'Payment' &&
+      tx.meta.TransactionResult === 'tesSUCCESS') {
+
+    if (tx.meta.delivered_amount &&
+        tx.meta.delivered_amount === 'unavailable') {
+      deliveredAmount = 'unavailable';
```
clark800: How about just returning in each branch instead of assigning to a variable?

src/ledger/parse/ledger.js
```diff
@@ -1,13 +1,15 @@
...
 const {removeUndefined, rippleTimeToISO8601} = require('./utils');
 const parseTransaction = require('./transaction');
 import type {GetLedger} from '../types.js';
 
 function parseTransactionWrapper(ledgerVersion, tx) {
-  const transaction = _.assign({}, _.omit(tx, 'metaData'),
-    {meta: tx.metaData});
+  const transaction = _.assign({}, _.omit(tx, 'metaData'), {
+    meta: tx.metaData,
+    ledger_index: ledgerVersion
```
clark800: Why is this necessary?

src/ledger/parse/utils.js
clark800: I think we need a comment for this, along with an explanation for what changed in ledger version 4594094

src/ledger/parse/utils.js
clark800: You could also break the else-if chain into just a sequence of if-returns which would make the comments less awkward

docs/index.md
shekenahglory: It has to be optional because it can only be present for successful Payment transactions, part of the reason they created the "delivered_amount" decorator because "DeliveredAmount" was not always present, so it was potentially confusing to understand exactly how much was transferred....this way it would always be present when *some* amount is transferred, even if it can't explictly state what that amount is.

clark800: > this way it would always be present when some amount is transferred, even if it can't explictly state what that amount is

What is the advantage of that if it's going to require a special case anyways?

shekenahglory: I don't think it matters much either way, since it will only be missing/unavailable on a few partialPayment transactions pre 2014 (once rippled team corrects their logic)

src/ledger/parse/ledger.js
shekenahglory: its another field that rippled includes in a 'tx' request, but not in a 'ledger' request - I need it at least for the ledger_index check in parseDeliveredAmount, I didn't check to see if it was expected and used elsewhere

docs/index.md
clark800: Ok, I think it's better to just omit it then, it will be less likely to produce bugs in the client code

clark800:
LGTM

PR #741, opened 2016-12-01
Fix eslint errors

wilsonianb:


wilsonianb:
This was done by:
- running:
```
$ node_modules/.bin/eslint -c ./eslintrc src/
```
- adding ` // eslint-disable-line strict` to `'use strict` to suppress `error  'use strict' is unnecessary inside of modules  strict`
- modifying code to address the remaining errors:
`error  use the spread operator instead of the '.apply()'       prefer-spread`
`error  Unexpected parentheses around single function argument  arrow-parens`

clark800:
LGTM

PR #745, opened 2017-02-17
Add transactions for payment channels

sharafian:
Adds methods to prepare transactions of the types: `PaymentChannelCreate`, `PaymentChannelFund`, and `PaymentChannelClaim`


src/common/schemas/specifications/payment-channel-claim.json
```diff
@@ -0,0 +1,38 @@
...
+    "renew": {
+      "type": "boolean",
+      "description": "Clear the channel's expiration time."
+    },
+    "close": {
+      "type": "boolean",
+      "description": "Request to close the channel."
+    }
+  },
+  "required": ["channel"],
```
wilsonianb: It could be good to capture these two rules:

- `renew` and `close` are mutually exclusive
- `signature` must be accompanied by `publicKey`

sharafian: Added some checks in https://github.com/ripple/ripple-lib/pull/745/files#diff-9efa2b0a32e1e12331fe1a8d45593513R36 this file, and tests. I was also wondering what happens when you don't specify the `cancelAfter` field on payment channel creation.

wilsonianb: Without `cancelAfter` the channel will remain open until there is either a PaymentChannelFund with an Expiration or a PaymentChannelClaim with the `tfClose` flag. At which point, the channel would only close immediately if the recipient used the `tfClose` flag. Otherwise, the channel would be set to expire either after the Expiration time or the settle delay.

wilsonianb:
I get this when trying to submit via ripple-lib:
```
Error: PaymentChannelCreate is not a valid name or ordinal for TransactionType
```
I'm guessing the new transactions need to be added here:
https://github.com/ripple/ripple-lib/blob/develop/src/ledger/transaction-types.js

sharafian:
I modeled my addition after the suspay classes, which have the type specified in their source files in `src/transactions/`. I'm using this branch on the testnet and I'm not getting that error. I'm using the `preparePaymentChannelCreate` function.

wilsonianb:
`preparePaymentChannelCreate` works for me.
I get the error when signing the transaction:
https://gist.github.com/wilsonianb/14b98f78998e3dcd425c6140cb5177e5#file-paychan-js-L22

wilsonianb:
It doesn't fix my error, but it looks like payment channel parse functions need to be added here:
https://github.com/ripple/ripple-lib/blob/develop/src/ledger/parse/transaction.js

wilsonianb:
Ok, it looks like these two files need to be updated as well as ripple-binary-codec
https://github.com/ripple/ripple-lib/blob/develop/src/common/schemas/output/get-transaction.json#L77
https://github.com/ripple/ripple-lib/blob/develop/src/common/schemas/objects/tx-type.json#L9

https://github.com/ripple/ripple-binary-codec/blob/master/src/enums/definitions.json#L1542

wilsonianb:
Doh, you're way ahead of me
https://github.com/ripple/ripple-binary-codec/pull/13

sharafian:
Oh, gotcha, I'll merge my binary-codec PR then. I must have had that linked locally.

wilsonianb:
I can successfully call `preparePaymentChannelCreate`, `preparePaymentChannelFund`, and `preparePaymentChannelClaim` and submit to the altnet.

I think the new transaction types need to be added here:
https://github.com/ripple/ripple-lib/blob/develop/src/ledger/parse/transaction.js
https://github.com/ripple/ripple-lib/blob/develop/src/common/schemas/output/get-transaction.json#L77
https://github.com/ripple/ripple-lib/blob/develop/src/common/schemas/objects/tx-type.json#L9

wilsonianb:
Can you add
```
docs/src/preparePaymentChannelCreate.md.ejs
docs/src/preparePaymentChannelFund.md.ejs
docs/src/preparePaymentChannelClaim.md.ejs
```
and run `npm run docgen`?

wilsonianb:
Looks like I missed a step. You'll need to include those three new `md.ejs` files in https://github.com/ripple/ripple-lib/blob/develop/docs/src/index.md.ejs
Then run `npm run docgen` and commit the modified `docs/index.md`.
Would you mind also rebasing on the latest `develop`?

sharafian:
Built docs and rebased.

wilsonianb:
Can you add the transaction specifications [here](https://github.com/ripple/ripple-lib/blob/develop/docs/src/specifications.md.ejs#L83) and rerun docgen?

wilsonianb:
What do you think about adding a `signPaymentChannelClaim` API method? It would take a private key, channel id, and claim amount and generate the signature for the claim that could later be included in a `PaymentChannelClaim` transaction.

sharafian:
That's a good idea, because it was a little tricky for me to implement in plugin-paychan. I could also add `verifyPaymentChannelClaim` or `validatePaymentChannelClaim`, to make sure that a claim is valid for a given channel ID, and maybe `getPaymentChannelId`, to get the ID of a channel from the source, destination, and sequence number. What files should these sorts of functions go into?

wilsonianb:
Good call on `verifyPaymentChannelClaim`
I was thinking it and `signPaymentChannelClaim` would be similar to the existing `sign` method, however I think they should probably go in `src/offline/`.
They'll also need to be added to the docs...
Let me think about `getPaymentChannelId`

PR #759, opened 2017-05-04
Switch to jsonschema

bachase:
Eliminates issues with using ripple-lib under restricted content security policy settings that block `ajv`.


bachase:
Forced push with eslint fixes.

PR #760, opened 2017-05-09
Docs: Require Node 6.9.0, update boilerplate to avoid destructuring assign

mDuo13:
Without the destructuring assignment, the boilerplate code runs on many Node versions (I think down to 4.x, but I don't know for sure). That said, it should definitely run on 6.9.x LTS which was Geert's suggestion for a reasonable minimum version to support, so I updated the docs to specify that version.

With that, you don't need Babel-Node to run code that uses RippleAPI.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -93,7 +93,7 @@ RippleAPI only provides access to *validated*, *immutable* transaction data.
 Use the following [boilerplate code](https://en.wikipedia.org/wiki/Boilerplate_code) to wrap your custom code using RippleAPI.
 
 ```javascript
-const {RippleAPI} = require('ripple-lib');
+const RippleAPI = require('ripple-lib').RippleAPI;
 
 const api = new RippleAPI({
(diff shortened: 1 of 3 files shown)


mDuo13:
I don't understand the build failure. It doesn't look related to what I changed. (I didn't forget to run docgen)

wilsonianb:
I think tests from forked repos always fail.

mDuo13:
Anything we can do to get this merged? I'm not familiar with the ripple-lib process

shekenahglory:
can you update the package.json as well? https://github.com/ripple/ripple-lib/blob/develop/package.json#L86

mDuo13:
I could change just that one line in the package.json, but I think we would also want to drop some of the babel dependencies too. I'm not familiar enough with the codebase and with ES6 in general to know which dependencies we can safely drop, though.

mDuo13:
Let's leave the package.json as-is so v6.9.0 is a "soft" requirement. The docs say you might be able to get it to work with previous versions of Node.js anyway, so leaving the old version makes it easier for people to get compatibility with prior versions.

PR #769, opened 2017-06-02
Adding the method signWithKeypair

amougel:
New method :
`signWithKeypair(txJSON: string, keypair: {privateKey: string, publicKey: string}, options: Object): {signedTransaction: string, id: string}`

Allow transaction signing for accounts that have not been created through a secret.

### DIFF
--- a/src/common/schemas/input/sign.json
+++ b/src/common/schemas/input/sign.json
@@ -10,7 +10,23 @@
     "secret": {
       "type": "string",
       "format": "secret",
-      "description": "The secret of the account that is initiating the transaction."
+      "description": "The secret of the account that is initiating the transaction. (This field is exclusive with keypair)."
+    },
+    "keypair": {
+      "type": "object",
+      "properties": {
+        "privateKey": {
+          "type": "privateKey",
+          "description": "The uppercase hexadecimal representation of the secp256k1 or Ed25519 private key."
+        },
+        "publicKey": {
+          "type": "publicKey",
+          "description": "The uppercase hexadecimal representation of the secp256k1 or Ed25519 public key."
+        }
+      },
+      "description": "The private and public key of the account that is initiating the transaction. (This field is exclusive with secret).",
+      "required": ["privateKey", "publicKey"],
+      "additionalProperties": false
     },
     "options": {
       "type": "object",
@@ -25,5 +41,15 @@
     }
   },
   "additionalProperties": false,
-  "required": ["txJSON", "secret"]
+  "required": ["txJSON"],
+  "oneOf": [
+    {
+      "required": ["secret"],
+      "not": {"required": ["keypair"]}
+    },
+    {
+      "required": ["keypair"],
+      "not": {"required": ["secret"]}
+    }
+  ]
 }
--- a/docs/index.md
+++ b/docs/index.md
@@ -3844,7 +3844,10 @@ return api.prepareCheckCash(address, checkCash).then(prepared =>
 
 ## sign
 
-`sign(txJSON: string, secret: string, options: Object): {signedTransaction: string, id: string}`
+```
+sign(txJSON: string, secret: string, options: Object): {signedTransaction: string, id: string}
+sign(txJSON: string, keypair: Object, options: Object): {signedTransaction: string, id: string}
+```
 
 Sign a prepared transaction. The signed transaction must subsequently be [submitted](#submit).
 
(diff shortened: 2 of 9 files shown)


wilsonianb:
It'd be great if we could just overload `sign` to take either a secret or keypair, in order to avoid code duplication between `sign` and `signWithKeypair`.

sublimator:
pretty easy to discriminate eh?

src/common/schemas/input/sign.json
```diff
@@ -10,7 +10,22 @@
...
+          "type": "string",
+          "description": "The uppercase hexadecimal representation of the secp256k1 or Ed25519 private key of the account that is initiating the transaction."
+        },
+        "publicKey": {
+          "type": "string",
+          "description": "The uppercase hexadecimal representation of the secp256k1 or Ed25519 public key of the account that is initiating the transaction."
+        }
+      },
+      "description": "The private and public key of the account that is initiating the transaction.(This field is exclusive with keypair).",
+      "required": ["privateKey", "publicKey"]
```
wilsonianb: add
```
"additionalProperties": false
```

src/common/schemas/input/sign.json
```diff
@@ -10,7 +10,22 @@
...
       "format": "secret",
-      "description": "The secret of the account that is initiating the transaction."
+      "description": "The secret of the account that is initiating the transaction.(This field is exclusive with keypair)."
+    },
+    "keypair": {
+      "type": "object",
+      "properties": {
+        "privateKey": {
+          "type": "string",
+          "description": "The uppercase hexadecimal representation of the secp256k1 or Ed25519 private key of the account that is initiating the transaction."
```
wilsonianb: I think this would be fine without `of the account that is initiating the transaction`
Same with `publicKey` below

src/common/schemas/input/sign.json
```diff
@@ -10,7 +10,22 @@
...
+        },
+        "publicKey": {
+          "type": "string",
+          "description": "The uppercase hexadecimal representation of the secp256k1 or Ed25519 public key of the account that is initiating the transaction."
+        }
+      },
+      "description": "The private and public key of the account that is initiating the transaction.(This field is exclusive with keypair).",
```
wilsonianb: `exclusive with keypair` --> `exclusive with secret`

intelliot:
@amougel This looks good to me; if you can merge or rebase on `develop`, that should enable CI to pass.

matejr:
What is the format of private and public key and how can one specify the type of signature scheme (ECDSA secp256k1/Ed25519)?

amougel:
@matejr The doc details it all. From memory the format is Hexadecimal.

sublimator:
All keys are 33 bytes encoded as hex

Ed25519
private: 0xED + 32 byte private key
public: 0xED + 32 byte public key

K256
private: 0x00 + 32 byte private key
public: public key point in compressed form, already 33 bytes, naturally
starting with either 0x02 or 0x03

A key is a naught, of course, of course, and no one can chalk to a horse,
of course! That is, of course, unless the horse is the famous Mister 0xED ??

matejr:
Documentation doesn't specify byte order, whether public key is compressed or not, size (in bytes) of private key and how to potentially deal with zeros in most significant bytes (can you e.g. omit leading zero bytes when you convert private key, represented as number, to hexadecimal string?), etc. It is still not clear to me how code distinguishes between secp256k1 and Ed25519 private keys.

src/transaction/sign.js
```diff
@@ -45,4 +42,17 @@ function sign(txJSON: string, secret: string, options: Object = {}
   }
 }
 
+function sign(txJSON: string, secret?: any, options?: Object, keypair?: Object
+): {signedTransaction: string; id: string} {
+  if(typeof(secret) === 'string') {
+    // we can't validate that the secret matches the account because
```
dncohen: I'd say "we don't..." rather than "can't".

Wondering aloud, is there any reason to validate that the publicKey and privateKey match each other?

intelliot: 👍  I think it would only be useful in a world where the ledger would only accept a signature from the secret that matches the account. However, there can be other keys that could sign for the account, I don't think there's a reason to validate here.

intelliot:
@matejr Thanks for the questions!

1. Keys are byte arrays which should have no byte order.
2. At 33 bytes, I believe the public key is compressed.
3. The private key is 33 bytes.
4. We would not omit leading zero bytes, but it should be feasible.
5. If the public key begins with `0xED` then it's an Ed25519 Key.

sublimator:
private keys have an extra byte prefixed, so if you are importing from some
3rd party source/lib, you’ll have to ...

they aren’t 33 bytes naturally in either scheme

darkmemo:
@amougel: Are you considering to rebase and resolve conflicts to get this PR merged?

amougel:
@darkmemo Yes, maybe not in the near future but eventually yes.

amougel:
@wilsonianb

FredKSchott:
Woah, the https://github.com/ripple/ripple-lib/pull/769/files#diff-12042bcf08631e33198bb47382ee4813 file got crazy-reformatted. Can you revert? (For the sake of PR diff readability, and anyone who already has open work on that test file that would have to deal with merge conflicts)

amougel:
@FredKSchott  amended the hicup

intelliot:
Thanks, @amougel !

PR #773, opened 2017-06-19
unref timer so it doesnt hang the node process

aMoniker:
I'm writing a CLI wrapper for ripple-lib, and this timer hangs the process for 20s after a successful `connect->request->response->disconnect` has finished and the process is otherwise ready to exit.


FredKSchott:
This would break web compatibility (timer.unref is Node only) so I'd hate to see this added if a better solution were available. The timeout timer should be getting cleared on successful request/response, so it sounds like that's the bug that needs to be fixed.

intelliot:
@aMoniker Are you still experiencing this problem? Can you show an example where the process hangs unnecessarily?

sharafian:
`clearTimeout` should work on both node and in the browser

intelliot:
It looks like `clearTimeout` is already used, so why is the node process hanging?

`timer.unref()` stops it from hanging, but [might not be best practice](https://www.bennadel.com/blog/3301-calling-timeout-unref-with-settimeout-does-not-appear-to-be-a-best-practice-in-node-js.htm).

intelliot:
See also: https://github.com/ripple/ripple-lib/issues/792

src/common/connection.js
```diff
@@ -441,6 +441,7 @@ class Connection extends EventEmitter {
       this._whenReady(this._send(message)).then(() => {
         const delay = timeout || this._timeout
         timer = setTimeout(() => _reject(new TimeoutError()), delay)
+        timer.unref()
```
FKSRipple: As long as we wrap this in a conditional, this should be fine on both Node.js and the browser:

```
       // In Node, tell Node not to leave the process open if this timeout is  keep the event loop active waiting only for this timer.
        if (timer.unref) {
          timer.unref();
        }
```

Also, can you confirm that the process doesn't exit before the response has returned (since this timeris no longer keeping it open)

aMoniker:
@FKSRipple Sure - I'll update and add that conditional today.

FKSRipple:
Thanks @aMoniker !

intelliot:
Released in 1.4.0

PR #794, opened 2017-09-06
Add signMessage and verifyMessage functions

intelliot:
- Sign an arbitrary string with a private key
- Verify a signature with a public key


src/offline/sign-message.js
```diff
@@ -0,0 +1,19 @@
...
+const binary = require('ripple-binary-codec')
+const {validate} = common
+
+function signMessage(message: string,
+  privateKey: string
+): string {
+  validate.signMessage({message, privateKey})
+
+  const signingData = binary.encode({
+    message: message
```
sublimator: This isn't going to work: `binary.encode` expects ledger entry, or transaction json, as spoken by rippled. It silently discards fields, so you're probably turning the message into a signingData of empty bytes.

sublimator: https://github.com/ripple/ripple-binary-codec/issues/10

sublimator: You probably just want to encode the message as utf-8 ?

sublimator: With some kind of `Hash Prefix` ?

intelliot: Thank you!

Not sure how to select a `Hash Prefix`. Perhaps just encoding as utf-8 and using its hex representation is sufficient?

sublimator: @JoelKatz is the chief crypto guy @ ripple :)

Tell him your use case.

sublimator: > Perhaps just encoding as utf-8 and using its hex representation is sufficient?

Perhaps, but I don't know what you're doing, and I'm not a crypto guy. 

Ask @justmoon or @JoelKatz

PR #802, opened 2017-11-03
Update bignumber.js, ripple-hashes, ripple-lib-transactionparser

### LINKED ISSUE #762: api.getOrderbook freezes on new iphones
See this thread:
https://www.xrpchat.com/topic/4477-apigetorderbook-freezing-on-iphone/

It only started happening a few days ago, no code changes, same phones, same ripplelib version (tested 0.17.1-7 - no effect).  So it seems like something to do with the rippled servers and not the library itself.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/45057795-api-getorderbook-freezes-on-new-iphones?utm_campaign=plugin&utm_content=tracker%2F323756&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F323756&utm_medium=issues&utm_source=github).
</bountysource-plugin>

intelliot:
Unit tests pass.

This works around a [bug in Safari 10.1](https://github.com/MikeMcl/bignumber.js/issues/120) that causes the browser to freeze (fixes #762).

This also removes `npm-shrinkwrap.json`: It makes adding/updating dependencies tricky. Let's try migrating to `yarn`, which will use the yarn lockfile now included here.


wilsonianb:
I agree that shrinkwrap complicates adding/updating dependencies. Can you elaborate on why we don't need the benefit of locked dependencies though?

intelliot:
Locked dependencies are a good thing. With good enough test coverage, we should be able to catch most of the problems caused by shifting dependencies. However, I've been reading about `yarn` and have committed a yarn lockfile here. By using `yarn install`, we'll get the same dependencies every time (until we upgrade them explicitly).

wilsonianb:
Should `npm install` be changed to `yarn install` in these files?
https://github.com/ripple/ripple-lib#running-tests
https://github.com/ripple/ripple-lib/blob/develop/scripts/publish#L10
https://github.com/ripple/ripple-lib/blob/develop/scripts/publish_rc#L10

intelliot:
Yes, good catch. I'll update those now.

intelliot:
@wilsonianb How does this look?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ripple-lib",
-  "version": "0.17.7",
+  "version": "0.17.8",
```
wilsonianb: Should HISTORY.md be updated since you're bumping the version?

HISTORY.md
```diff
@@ -1,9 +1,11 @@
 # ripple-lib Release History
 
-## UNRELEASED
+## 0.17.8 (UNRELEASED)
```
wilsonianb: Should this be included?
https://github.com/ripple/ripple-lib/commit/2a90f9b134e168937dceb7da283d63734eac9e7c

PR #806, opened 2017-11-10
Throw error when XRP-to-XRP payment has maxAmount (RLJS-591)

intelliot:
`maxAmount` ([SendMax](https://ripple.com/build/transactions/#payment)) must be omitted for XRP-to-XRP payments.


wilsonianb:
This needs to updated
https://github.com/ripple/ripple-lib/blob/develop/docs/samples/payment.js#L13
I'm also wondering if `destination.minAmount` makes sense for XRP-to-XRP (or any single IOU) payments

PR #810, opened 2017-11-16
fix and update flow to get typechecker running again

FredKSchott:
The repo's flow setup was broken, and incorrectly reporting zero errors. This PR updates flow and the config/setup, which also seemed out of date.

`yarn run typecheck` now fails with 49 errors.


package.json
```diff
@@ -28,20 +28,20 @@
...
     "babel-core": "^6.4.0",
     "babel-eslint": "^6.0.4",
     "babel-loader": "^6.2.1",
-    "babel-plugin-syntax-flow": "^6.3.13",
-    "babel-plugin-transform-flow-strip-types": "^6.4.0",
     "babel-preset-es2015": "^6.3.13",
+    "babel-preset-flow": "^6.23.0",
     "babel-preset-stage-1": "^6.3.13",
-    "babel-register": "^6.3.13",
+    "babel-babel-register": "^6.3.13",
```
intelliot: Typo - should be `babel-register`

FredKSchott: whoops! didn't even mean to touch that dep

FredKSchott:
@intelliot fixed, ptal

PR #811, opened 2017-11-16
fix flow type errors

FredKSchott:
This gets flow passing again on the codebase (a result of running #810)


src/common/utils.js
```diff
@@ -27,6 +27,7 @@ function toRippledAmount(amount: Amount): RippledAmount {
   if (amount.currency === 'XRP') {
     return xrpToDrops(amount.value)
   }
+  // $FlowFixMe: amount.issuer is not a Amount type property. Safe to remove?
```
intelliot: Good catch. `issuer` is actually the term that we use in rippled. We should add `issuer` to the `Amount` type, and keep using it here. (No hurry, but in a future release, we may deprecate `counterparty`.)

intelliot: (There are other areas in the code that will need to be updated to accommodate this)

src/transaction/settings.js
```diff
@@ -8,17 +8,36 @@ const validate = utils.common.validate
 const AccountFlagIndices = utils.common.constants.AccountFlagIndices
 const AccountFields = utils.common.constants.AccountFields
 import type {Instructions, Prepare} from './types.js'
-import type {Settings} from './settings-types.js'
+
+type Settings = {
+  passwordSpent?: boolean;
```
intelliot: I believe we prefer to use commas instead of semicolons here.

src/common/utils.js
FredKSchott: Say more about the other areas that need to be changed? They'll need to accommodate by  setting `issuer`? Read from `issuer`?

intelliot: @FredKSchott Right. For example, here are places where I think we'll need to [set issuer](https://github.com/ripple/ripple-lib/blob/develop/src/ledger/parse/amount.js#L17) or [read from issuer](https://github.com/ripple/ripple-lib/blob/develop/src/ledger/orderbook.js#L53). We can do this in a later PR though.

intelliot:
We can clean up counterparty/issuer in a later PR :) Merging.

FredKSchott:
👍 👍 👍

PR #813, opened 2017-11-25
Fix small lint errors

FredKSchott:
Ran `yarn lint` and found some errors, fixed them automatically with `eslint --fix`


FredKSchott:
Ha, TIL about that URL option

PR #815, opened 2017-12-02
convert common.js modules to new standard ESM module syntax

FredKSchott:
Node.js, Browser Vendors, and the rest of the JS ecosystem are moving towards a standard module syntax for JavaScript. It's "strict" by default, standard across environments, and easier for compilers to optimize for.

Why now? This work is driven by our switch from Flow to TypeScript, so that we can add powerful typing support to our new API request interface. TypeScript uses ESM exclusively, so this is the first step. See https://github.com/ripple/ripple-lib/issues/812 for more info. 

No changes to Babel were needed, it successfully compiles down to Node-style common.js modules for distribution. Tests continue to pass.

/cc @intelliot

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,56 +1,50 @@
 /* @flow */
-'use strict' // eslint-disable-line strict
 
-const _ = require('lodash')
-const EventEmitter = require('events').EventEmitter
-const common = require('./common')
-const server = require('./server/server')
+import * as _ from 'lodash'
+import events from 'events'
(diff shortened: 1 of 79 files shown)


intelliot:
What did you use to do this conversion?

FredKSchott:
A few regex find and replaces did most of the work, and then the `no-commonjs` eslint rule to confirm I got everything.

FredKSchott:
🤘

Issue #812, opened 2017-11-20
Proposal: Simplified Requests, Flexible Responses


FredKSchott:
Following up from #799:

> @intelliot: @FredKSchott I'm in favor of returning all API data to the consumer, as you said. Historically, RippleAPI was designed independent of rippled. However, moving forward, we'd like to unify their APIs and make them as similar as possible. Please feel free to take a stab at this!

So I've experimented a bit and have a proposal. Even with 100% proper typing, I'm nervous to make any big changes to the existing API methods and the interfaces/data that they already return. There are a lot of interconnected parts internally, so I'm not just worried about breaking external consumers with a lot of small, hard to detect/message changes.

This proposal aims to reduce the complexity & maintainance costs of having a complete secondary ripple-lib API on top of the standard Rippled API:

- **Unify the ripple-lib & rippled APIs without breaking existing users**
- **"Simplified API Request Interface":** 
  - Provide access to all request options, not just a filtered selection
  - Reduce complexity & high maintenance costs for the core team
- **"Flexible API Response Interface":** 
  - Provide access to all response data, not just a filtered selection
  - Continue to provide easy formatting

## Proposal: Simplified API Request Interface

All custom API methods (`getTransactions()`, `getTrustlines()`, etc) are deprecated in favor of just two new general-purpose API methods:

```ts
class RippleAPI:

  // Merge {...data, command} into a single request and send to the server.
  // Return the response object directly from the server.
  async request(command: string, data?: Object): Response;

  // Like request(), but requests multiple pages until the `limit.max` number
  // of `limit.collect` objects are returned from the response.
  async requestAll(command: string, data?: Object, limit?: {max: number, collect: string}): Response[];
```

### Examples

```diff
- return api.getAccountInfo('r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59');
  return api.request('account_info', {account: 'r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59'});

- return api.getOrders('r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59');
  return api.request('account_offers', {
    account: 'r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59',
    ledger_index: 'validated',
  });

- return api.getOrderbook('r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59', {
-   "base": {"currency": "XRP"},
[...]

PR #816, opened 2017-12-04
Convert from Flow to Typescript

FredKSchott:
TypeScript is very similar to Flow, in that both provide a flexible typing system for JavaScript. The syntax is almost identical, so it should feel perfectly comfortable already. But TypeScript is much more powerful, so with the same syntax you catch more errors and enforce clearer code. [This is a good, fair comparison of the two](https://github.com/niieani/typescript-vs-flowtype), and here are some examples in rippled-lib itself:

- [Simple bad typing not caught by Flow](https://github.com/FredKSchott/ripple-lib/compare/esm...FredKSchott:typescript#diff-a63cf6ab08f2ab8f6dfd154f50c27c6cR31)
- [Broken (or undocumented/deprecated) call to lodash](https://github.com/FredKSchott/ripple-lib/compare/esm...FredKSchott:typescript#diff-ac5fabe7c1af9ab4ec35af0e7a241af8R112)
- [Broken call to Error.captureStackTrace](https://github.com/FredKSchott/ripple-lib/compare/esm...FredKSchott:typescript#diff-3c825d30711b04c82190a4348b7800bbR18)
- You can search this PR for "TODO:" to see other bugs I highlighted but couldn't fix.

Additionally it's growing faster, adding new features, and is already fairly popular in the larger npm ecosystem. Lodash, for ex, publishes TS type definitions directly in their package, which is how I was able to catch that error above automatically. Also, type checking is a part of your build/compile, so they can't get out of date with the code.

This allows us to add function typing information to the new, main `request()` interface. That way all  consumers, even those using regular JavaScript with an IDE like VSCode, will get typing help when making API requests and handling API responses. See #812 for more information.

/cc @intelliot This touches a lot of code so please ask if anything is confusing / looks off / could use some documentation.


package.json
```diff
@@ -64,11 +60,11 @@
     "docgen": "node --harmony scripts/build_docs.js",
     "clean": "rm -rf dist/npm",
     "typecheck": "flow check",
-    "compile": "babel -D --optional runtime -d dist/npm/ src/",
-    "watch": "babel -w -D --optional runtime -d dist/npm/ src/",
+    "compile": "mkdir -p dist/npm/common && cp -r src/common/schemas dist/npm/common/ && tsc",
```
FredKSchott: typescript only handles typescript files, so we have to copy the `schemas` directory here ourselves. Normally you could just move this folder to a single place outside of the `src` directory and load them by relative path, but unfortunately that breaks because `dist/npm` & `src/` are at different levels of nesting (`../` vs `../../`).

This doesn't feel *too* complex but if we ever need to add more here we could move this into a build script or gulp file.

src/api.ts
```diff
@@ -107,50 +105,48 @@ class RippleAPI extends events.EventEmitter {
...
-}
 
-_.assign(RippleAPI.prototype, {
-  connect,
-  disconnect,
-  isConnected,
-  getServerInfo,
-  getFee,
-  getLedgerVersion,
+  connect = connect
```
FredKSchott: Adding these as class properties makes them a part of the interface, vs. `_.assign()` which was appending them to the prototype in a way that confused TypeScript.

src/broadcast.ts
```diff
@@ -3,11 +3,15 @@ import * as _ from 'lodash'
 import {RippleAPI} from './api'
 
 class RippleAPIBroadcast extends RippleAPI {
+
+  // TODO: Should this default to 0, or null/undefined?
```
intelliot: I think `0` is good here.

src/common/connection.ts
```diff
@@ -15,12 +13,49 @@ function isStreamMessageType(type) {
...
+class Connection extends EventEmitter {
+
+  private _url: string
+  private _trace: boolean
+  private _console?: Console
+  private _proxyURL?: string
+  private _proxyAuthorization?: string
+  private _authorization?: string
+  private _trustedCertificates?: string[]
+  // TODO: options.key never used?
```
intelliot: Looks like it's used in `_createWebSocket()`

intelliot:
Still looking at this. Hoping to try it out soon :)

Any potentially breaking changes?

AGallouin:
Any news on this ? I'm really looking forward to this !

sublimator:
They too busy celebrating? haha

image

src/broadcast.ts
sublimator:
```
> 1 > undefined
false
```

src/common/serverinfo.ts
```diff
@@ -60,16 +60,19 @@ function getServerInfo(connection: Connection): Promise<GetServerInfoResponse> {
   })
 }
 
+// TODO: This was originally annotated to return a number, but actually
+// returned a toString'ed number. Should this actually be returning a number?
```
intelliot: I believe this should be a string. We generally use strings for all currency amounts.

src/ledger/pathfind.ts
```diff
@@ -93,11 +95,13 @@ function filterSourceFundsLowPaths(pathfind: PathFind,
 ): RippledPathsResponse {
   if (pathfind.source.amount &&
       pathfind.destination.amount.value === undefined && paths.alternatives) {
-    paths.alternatives = _.filter(paths.alternatives, alt => {
-      return alt.source_amount &&
-        pathfind.source.amount &&
+    paths.alternatives = _.filter(paths.alternatives, alt =>
+        !!alt.source_amount &&
+        !!pathfind.source.amount &&
+        // TODO: Returns false when alt.source_amount is a string. Fix?
```
intelliot: How can `alt.source_amount` be a string?

src/server/server.ts
```diff
@@ -23,12 +21,14 @@ function getServerInfo(): Promise<GetServerInfoResponse> {
   return common.serverInfo.getServerInfo(this.connection)
 }
 
-function getFee(): Promise<number> {
+// TODO: getFee() was originally annotated to return a number, but actually
+// returned a toString'ed number. Should this actually be returning a number?
```
intelliot: This should return a string.

src/transaction/combine.ts
```diff
@@ -21,7 +19,9 @@ function compareSigners(a, b) {
 function combine(signedTransactions: Array<string>): Object {
   validate.combine({signedTransactions})
 
-  const txs = _.map(signedTransactions, binary.decode)
+  // TODO: signedTransactions is an array of strings in the documentation, but
+  // tests and this code handle it as an array of objects. Fix!
```
intelliot: Good catch 🤔

src/broadcast.ts
FredKSchott: @intelliot my concern is that 0 is a number but isn't a valid ledger number. It do mean that things like `(ledger.ledgerVersion > this.ledgerVersion)` will work by default, but it also means we have to document and explain that 0 = unset/unknown. It would be more explicit if `ledgerVersion = undefined` means it hasn't been looked up yet.

src/common/connection.ts
FredKSchott: Ha, can't believe I missed that

src/common/connection.ts
```diff
@@ -186,7 +212,8 @@ class Connection extends events.EventEmitter {
 
       this._retry = 0
       this._ws.on('error', error => {
-        if (process.browser && error && error.type === 'error') {
+        // TODO: "type" does not exist on official error type, safe to remove?
```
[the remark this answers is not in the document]
FredKSchott: sounds good, I'll just leave it as a TODO then

src/common/serverinfo.ts
FredKSchott: sounds good.

Looking at this more closely now, should this use `BigNumber` internally? `Number` could cause problems with `baseFeeXRP` being a float, right?

src/ledger/pathfind.ts
FredKSchott: ` alt.source_amount` is of type `RippledAmount`, which can be a string. https://ripple.com/build/rippled-apis/ seems to document a few places where source_amount is a string.

FredKSchott:
@intelliot Great catch re: webpack. I'll get that fixed for this PR.

Thanks for the review! No behavior should be changing here, except for a few small bug fixes that were explicitly bugs. Anything that could have been relied on as behavior was left as-is with a TODO, so no breaking changes expected.

Excited to do some more work on this over the holidays. Three branches can go into PR all together once this gets merged:

- [`istanbul update`](https://github.com/FredKSchott/ripple-lib/compare/typescript...FredKSchott:nyc-istanbul) Istanbul is now called "nyc", works with typescript automatically
- [`tslint`](https://github.com/FredKSchott/ripple-lib/compare/typescript...FredKSchott:nyc-istanbul) like eslint but for typescript, rules ported over from the eslintrc
- [beginning to flesh out new request api](https://github.com/FredKSchott/ripple-lib/compare/typescript...FredKSchott:request-interface): so far added `account_offers`, `book_offers`, and `account_info`. Doesn't need to be merged into develop just yet, but an early review would be good to test out & make sure it's going in the right direction for your team.

FredKSchott:
webpack & gulp updated for TS: https://github.com/FredKSchott/ripple-lib/compare/typescript...FredKSchott:webpack

It's currently in a separate branch, but I can add it directly into this PR if you'd prefer.

intelliot:
That looks great!

I attempted to [create a PR](https://github.com/FredKSchott/ripple-lib/pull/1) and merge that branch to this PR, but github won't allow it since I don't have write access to your fork 😅

FredKSchott:
Haha np, I'll merge it in now

src/broadcast.ts
intelliot: That's a good point. I'm fine with `undefined`, but make sure to update the if condition (`ledger.ledgerVersion > this.ledgerVersion`) too.

src/common/serverinfo.ts
intelliot: Yes, it would be correct to use `BigNumber` here internally.

src/ledger/pathfind.ts
intelliot: Looks like `source_amount` is only a string when the currency is XRP, in which case there is no need for pathfinding. So I think it's fine to return false in that case.

sublimator: >>> Looks like source_amount is only a string when the currency is XRP, in which case there is no need for pathfinding.

You can send XRP and use paths to send the recipient USD ... 

What you mean no need for path finding ?

intelliot:
Squashed and merged! I don't think there will be any difficulty rebasing on the squash commit, but let me know if so.

FredKSchott:
@intelliot woo! 🎉 

Just getting back & caught up from the holidays, I'll rebase those other branches and create the next batch of PRs

PR #833, opened 2018-01-23
Use BigNumber internally to compute fee

intelliot:
Resolves a TODO


intelliot:
Merged to `develop` via 51aaa75a05d8c1266e13b0c44f1d840988ffb80c

PR #834, opened 2018-01-26
Remove Flow preset

intelliot:
Just left over from when we were using Flow.

Thanks @tpkiddle

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015", "stage-1", "flow"]
+  "presets": ["es2015", "stage-1"]
 }


FredKSchott:
Actually, taking a deeper look at this:

Babel is no longer used at all (no longer a dependency either) so this file can go away completely. There are also 3 scripts that still reference it:

- `"compile-with-source-maps"` Sourcemaps are generated by default with our current `tsconfig`, so this can go away now.
- `"sauce": "babel-node scripts/sauce-runner.js"`: This file doesn't need babel-node
- `"start": "babel-node scripts/http.js"`: This file can be fixed to no longer need babel-node

@intelliot want me to create a PR to clean this up?

intelliot:
@FredKSchott That would be awesome!

intelliot:
Closed by #842

PR #836, opened 2018-01-26
Fix: isSameIssue() should check counterparty

### LINKED ISSUE #835: v0.18.0 returns different value in orderbook as in v0.17.9
Hello,

I am testing v0.18.0, but I find a difference on some of the returned orderbooks.
The script I am running:

```
const {RippleAPI} = require('ripple-lib');

const api = new RippleAPI({
  server: 'wss://s1.ripple.com' // Public rippled server hosted by Ripple, Inc.
});
api.on('error', (errorCode, errorMessage) => {
});

var address = "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"

var orderbook = {
  "counter": { currency: "USD", "counterparty": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B" },
  "base": { "currency": "USD", "counterparty": "rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq" }
}

var options = {
    "limit": 5
}

api.connect().then(() => {
  return api.getOrderbook(address, orderbook, options)
}).then(openorders => {
  console.log("ASK:", JSON.stringify(1/openorders.asks[0].properties.makerExchangeRate, null, 2))
  console.log("BID:", JSON.stringify(1/openorders.bids[0].properties.makerExchangeRate, null, 2))
  return api.disconnect();
}).catch(console.error);
```

Running under v0.17.9 returns:

```
>node getOrderbook.js
ASK: 1.0050251256281406
BID: 0.971722864639005

```

Running under v0.18.0 returns:

```
>node getOrderbook.js
ASK: 1.0050251256281406
BID: 1.0027103097025
```

The value for v0.17.9 is consistent with current values, v0.18.0 seems to be wrong. 
I am running node v9.2.0

n.b. Could this be autobridge values? Are they calculated correct? They don't seem to match with e.g. xrpcharts autobridge orders

intelliot:
Fixes #835

### DIFF
--- a/src/ledger/orderbook.ts
+++ b/src/ledger/orderbook.ts
@@ -52,7 +52,7 @@ function getBookOffers(connection: Connection, account: string,
 }
 
 function isSameIssue(a: Amount, b: Amount) {
-  return a.currency === b.currency && a.issuer === b.issuer
+  return a.currency === b.currency && a.counterparty === b.counterparty
 }
 
 function directionFilter(direction: string, order: OrderbookItem) {


intelliot:
We'll dig into this area of the code more closely later (at minimum, we should add a test), but I'm going to release this for now in order to fix the bug.

PR #837, opened 2018-01-27
Update README.md - How to build ripple-lib

intelliot:
This should make it easier to [build](https://github.com/ripple/ripple-lib/issues/835#issuecomment-360922057) ripple-lib.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,22 @@ $ yarn add ripple-lib
 
 Then see the [documentation](https://github.com/ripple/ripple-lib/blob/develop/docs/index.md) and [code samples](https://github.com/ripple/ripple-lib/tree/develop/docs/samples)
 
+## Development
+
+To build the library for Node.js:
+```
+$ yarn compile
+```
+
+The TypeScript compiler will [output](./tsconfig.json#L7) the resulting JS files in `./dist/npm/`.
+
+To build the library for the browser:
+```
+$ yarn build
+```
+
+Gulp will [output](./Gulpfile.js) the resulting JS files in `./build/`.
+
 ## Running tests
 
 1. Clone the repository
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ $ yarn build
 
 Gulp will [output](./Gulpfile.js) the resulting JS files in `./build/`.
 
+For more details, see the `scripts` in `package.json`.
+
 ## Running tests
 
 1. Clone the repository


jnr101:
I am used to import the api with 
`const {RippleAPI} = require('ripple-lib')`. 
Can you explain why this doesn't work after build? What works now is apparently 
`const {RippleAPI} = require('./dist/npm')` or 
`const RippleAPI = require('../../dist/npm').RippleAPI`, 
but when you download the api using "yarn add" or npm, it is different. What is the difference or what happens when the code is released to npm?

Thanks,
jnr

FredKSchott:
@intelliot fwiw, `compile` is not a very common name for building, and having both `compile` and `build` leads to a lot of confusion when its unclear which does what. 

Could `npm run build` build for both Node.js & the browser? There would be less confusion there (esp. for first-time users) and you could further split into `npm run build:node` and `npm run build:browser` if you still wanted to allow for single-env builds.

intelliot:
@jnr101 The difference is that we now `compile` the library into JavaScript and release that to npm. The code itself is in TypeScript, which provides type safety and other features while still easily compiling to JS. How do you download/use the code without yarn/npm?

jnr101:
I do the following steps:

```
git clone https://github.com/ripple/ripple-lib.git
cd ripple-lib
yarn install
yarn compile
```
I run in the same directory the following script:

```
const {RippleAPI} = require('ripple-lib');

const api = new RippleAPI({
  server: 'wss://s1.ripple.com' // Public rippled server hosted by Ripple, Inc.
});
```

It doesn't work, because 

`Error: Cannot find module 'ripple-lib'`

If I change the script as follows it works:

```
const {RippleAPI} = require('./dist/npm');

const api = new RippleAPI({
  server: 'wss://s1.ripple.com' // Public rippled server hosted by Ripple, Inc.
});
```

Deducting what happens when published to npm on the internet and thereafter installed locally, the `./dist/npm` directory content is placed in `./node_modules/ripple-lib` together with the `package.json`. I guess it is npm standard that I just didn't know of :-)

intelliot:
@jnr101 Ah, that's interesting. Did that same script, in the same directory, used to work with an older version of ripple-lib?

(I don't think it would)

jnr101:
hi @intelliot, yes, it's the same for the older versions.

intelliot:
@jnr101 Cool - that makes sense :)

Note that the typical way to use npm is to actually create your own project and add ripple-lib as a dependency. You can do this with [npm init](https://docs.npmjs.com/getting-started/using-a-package.json) or [yarn init](https://yarnpkg.com/lang/en/docs/creating-a-project/). Then you should be able to use `const {RippleAPI} = require('ripple-lib');` as you stated.

jnr101:
hi @intelliot,

so I first tried as you suggested with npm init, to create a package.json for my own project, but I don't get how to point to a dependancy for the ripple-lib version that I am building .. So I tried something else, which seems to work, not sure though if it is the typical/official way to do thism probably not, but here goes:

Imagine an existing project that makes use of ripple-lib, and you want to test a new - not yet released - version of ripple-lib against your existing project. Our project is the nodejs file `connectOnly.js` (with or without package.json and/or existing node_module installs)
```
const {RippleAPI} = require('ripple-lib');

const api = new RippleAPI({
  server: 'wss://s1.ripple.com' // Public rippled server hosted by Ripple, Inc.
});
```
The following steps would build and compile the newest ripple-lib version and would allow for no other changes in my existing code:
```
mkdir -p node_modules
cd node_modules
git clone https://github.com/ripple/ripple-lib.git
cd ripple-lib
# If you want another version uncomment and set version:
# git checkout 0.17.9
yarn install && yarn compile
cd ../..
```
and then the following works without changing the code:

`node connectOnly.js`

intelliot:
@jnr101 Instead of cloning the repository manually, `npm` or `yarn` could [do it for you](https://docs.npmjs.com/files/package.json#github-urls).

For example, the following in my `package.json` gives me the latest from `develop` (our default branch).
```
  "dependencies": {
    "ripple-lib": "ripple/ripple-lib"
  }
```

Unfortunately, this doesn't cause the `dist/npm/` files to be built, so you still need to do the following:
```
cd node_modules/ripple-lib
yarn install && yarn compile
cd ../..
```

@FredKSchott Any idea if this is the normal behavior? Not sure whether other projects tend to push (to github) the files that they build for npm.

FredKSchott:
@jnr101 what kind of use are you ultimately building for here? Just testing local changes?
 Distributing a custom version of ripple-lib yourself?

Check out `yarn link`, which lets you link a local directory into another package as a dependency. That will let you do what you need locally at least.

jnr101:
@FredKSchott It is about testing local changes. When a new version of ripple-lib is published I first test it in a test (PA) environment which mimics prod as much as possible.  Then I compare test output with prod output. I would normally not build ripple-lib myself. Except last time I ran into a bug. A fix was made, of which @intelliot asked me to test against to see if the problem was solved. As the patched ripple-lib was not published yet, I had to build it myself. This was a new cup of tea for me of which at the time I did not know much about, but - thanks for educating - I am starting to get how npm/yarn, node_modules and published items actually work :-)

Back to the question, would `yarn link` serve my purpose? To an extend it does. I can (at first sight) leave my existing code intact and use `yarn link` to temporarily point to another directory where I will have another version of ripple-lib built. Looking at what happens under the hood, a symlink is created and replaces the previous version of node_modules/ripple-lib directory. To get back to the previous version I must use `yarn unlink "ripple-lib" && yarn add ripple-lib`. So imo still not perfect, but the most clean solution up till now. I think I will use this in future cases.

Back to the original case - update of the README.md, not sure if this particular use-case should be added. I suspect the more advanced developers will have their own preferred way of dealing with this. As long as it is clear how to build the software, people will be able to find their way

PR #840, opened 2018-01-27
Change WebSocket dependency to be browser friendly

mikeb1rd:
Removed explicit dependency on ws, to instead use simple-websocket so that the connection works in node and the browser, and is now browserify friendly.

An example can be found here: https://github.com/mikeb1rd/ripple-lib-browser-client


FredKSchott:
@mikeb1rd ripple-lib already works in the browser, by replacing `ws` with `wswrapper.ts` (uses native browser WebSocket) during the browser build step. Is this not working for you?

mikeb1rd:
@FredKSchott If you use ripple-lib directly that's fine, but I was trying to use an npm module that has a dependency on ripple-lib (such as ilp-plugin-xrp-escrow or ilp-plugin-xrp-asym-client), and pull that into the browser using browserify. I'll have another think and see if using webpack works better.

FredKSchott:
We might just need to set the “browser” property in the package.json to
point to the bundle:
https://github.com/defunctzombie/package-browser-field-spec

intelliot:
@mikeb1rd See https://github.com/ripple/ripple-lib/pull/847 for an alternate solution; hopefully that works with your browserify setup?

Using your above example (ripple-lib-browser-client), I built three versions of `bundle.js`.

1. `bundle-simple-websocket.js` is generated using this PR. Filesize: 1.9 MB
2. `bundle-browser-in-package-json-npm-install.js` is using the `browser-in-package-json` branch (ad6d804) and using `npm install && npm build`. Filesize: 1.4 MB
3 `bundle-browser-in-package-json-yarn-install.js` is using the `browser-in-package-json` branch (ad6d804) and using `yarn install && yarn build`. Filesize: 1.3 MB

I think there are (minor?) differences between 2 and 3 because npm pulls in unpredictable versions of dependencies.

The files can be found here: https://gist.github.com/intelliot/cc4628e7b81d1527be690dcdac6bf7f9

Due to the large filesizes, it's hard to compare them unless you either clone or `Download ZIP`. Then you can use a diffing tool (I just use FileMerge).

intelliot:
To test it out, you can have a look at my fork of your example: https://github.com/intelliot/ripple-lib-browser-client/commit/89ce3e83fe6e3431b20a14dcabc0e40ed4b4600e

Other than using the `browser-in-package-json` branch of ripple-lib, there are no changes to your code.

mikeb1rd:
Closed in favour of #847

PR #841, opened 2018-01-27
Revert "Fix: isSameIssue() should check counterparty"

m3anid:
Reverts ripple/ripple-lib#836

### DIFF
--- a/src/ledger/orderbook.ts
+++ b/src/ledger/orderbook.ts
@@ -52,7 +52,7 @@ function getBookOffers(connection: Connection, account: string,
 }
 
 function isSameIssue(a: Amount, b: Amount) {
-  return a.currency === b.currency && a.counterparty === b.counterparty
+  return a.currency === b.currency && a.issuer === b.issuer
 }
 
 function directionFilter(direction: string, order: OrderbookItem) {


intelliot:
Why do you want to revert this commit?

FredKSchott:
Just catching up here, was #835 introduced in the move to TypeScript? If #836 fixed one bug and introduced another (this issue?), there might be a missing or added call to `renameCounterpartyToIssuer()`. Hard to say without more info about what's broken here, but I remember changing issuer to counterparty for a reason.

intelliot:
Good point. I'll take a closer look later today.

intelliot:
We may want to move to using the term `issuer` instead of `counterparty` for consistency with rippled. Unfortunately, this is a departure from what RippleAPI has done in the past.

For instance, a USD obligation between `A` and `B` can equally be thought of as `USD/A` or `USD/B`. `A` having `USD/B` and `B` having `USD/A` is just the same number with an opposite sign. There is a USD balance between `A` and `B` and it can be positive or negative. So `A` and `B` can both be issuers on the XRP Ledger, as long we don't restrict "issuer" to only refer to someone who _owes_ money to others.

I'm not sure if that really clarifies things, but it does seem to indicate that `issuer` is a fine term to use, as long as we expand what it means in our context. @mDuo13 what do you think?

mDuo13:
That's about right. I mean, a lot of times we use "issuer" to mean the one who has a negative balance, so in that sense you can only be an "issuer" for a currency if others set nonzero trust limits to you for that currency. But, of course, `rippled` uses the "issuer" field to refer to either side of that relationship, as you described.

intelliot:
We want to eventually move to using "issuer" in most contexts, but not for 1.0.0. Closing.

PR #843, opened 2018-01-31
Expose new request interface, implement first few request typings

FredKSchott:
*A part of https://github.com/ripple/ripple-lib/issues/812*
/cc @intelliot 

Now that the Flow->TypeScript migration is complete, we have support for the function type overloading that will support direct API requests (for all TypeScript projects and JavaScript projects via supported text editors like VS Code).

This PR intentionally only adds typing for a few commands for now. I want to get this reviewed first before going any further. Once this is merged the remaining requests/responses can be split up and parallelized.

The *existing* methods have been refactored to use the new interface internally, and the goal is still to not make any breaking changes to how they behave.

#### Summary 

- New `request()`/`requestAll()` methods (currently marked private until more baked)
- New type definitions for 5 selected rippled commands: 
  - `account_info`, `account_lines`, `account_offers`, `book_offers`, `gateway_balances`
- Refactored existing request methods to simplify & use those new request/response definitions
- Organized & cleaned up existing type definitions (removed duplicated / overlapping types)


src/ledger/accountinfo.ts
```diff
@@ -1,31 +1,12 @@
...
-  validated: boolean
-}
-
-type AccountInfoOptions = {
+type GetAccountInfoOptions = {
   ledgerVersion?: number
 }
 
-type AccountInfoResponse = {
+type GetAccountInfoFormattedResponse = {
```
intelliot: I see that some types were renamed, like `AccountInfoResponse` -> `GetAccountInfoFormattedResponse`, `TrustlinesOptions` -> `GetTrustlinesOptions`, `Trustline` -> `ParsedTrustline`. Could you explain a bit about how these types should be named?

FredKSchott: Definitely, a would love some feedback from you on these as well.

The main idea behind the type renaming/reorganization is to best support the rippled API as the main source of truth. But the current `AccountInfoResponse`, `TrustlinesOptions`, and `Trustline` types all describe special objects created and defined specifically by ripple-lib. So, for example, now that we're adding typing info for the actual standard "account_info" response format, `AccountInfoResponse` is a confusing name for the custom/specialized `getAccountInfo()` response  object. Same with the current `Trustline`, which is a special custom parsing of a trustline and not the standard API format for one.

What do you think? So far I haven't had followed any specific rules while renaming existing custom types that conflicted with standard formats. But happy to work out a set of rules to follow if that's a concern.

src/ledger/parse/account-trustline.ts
```diff
@@ -1,40 +1,24 @@
...
-    frozen: trustline.freeze || undefined,
-    authorized: trustline.authorized || undefined
+    frozen: trustline.freeze || undefined
   })
   // rippled doesn't provide the counterparty's qualities
   const counterparty = removeUndefined({
     limit: trustline.limit_peer,
     ripplingDisabled: trustline.no_ripple_peer || undefined,
-    frozen: trustline.freeze_peer || undefined,
-    authorized: trustline.peer_authorized || undefined
```
intelliot: I think we need to keep `authorized`. Unless there's a reason it was removed?

FredKSchott: I didn't see `peer_authorized` mentioned anywhere here, and assumed it was an old removed property / always undefined. https://ripple.com/build/rippled-apis/#account-lines

Happy to put it back in if that's not true (& create issue to document)

intelliot: I think it is missing from the documentation. I've opened a PR for the docs to add it in.

FredKSchott:
@intelliot Here are the general naming rules I've been following so far, lmk what you think:

- The goal is to re-organize and clean up existing types as we add new request/response types
- Big changes to existing type definitions is to be avoided, there's enough going on as it is
- During this transition we try to reuse existing types and avoid changing existing types names
- all general types (shared across files, used in more than one file, etc.) are moved into the `types/` directory.
- `types/commands/` stores all of the typings for rippled commands. Each file's format is roughly:
```
// account_offers.d.ts
export interface AccountOffersRequest { ...
export interface AccountOffersResponse { ...

// some_command.d.ts
export interface SomeCommandRequest { ...
export interface SomeCommandResponse { ...
```
- `types/objects/` stores all the typings for all shared rippled objects, like `RippledAmount`, `Memo`, Transaction types, etc.
  - If an existing type name conflicts with the new standard type names (ex: `AccountInfoResponse` conflicts with the new "account_info" response type) we prepend the "Formatted" prefix to convey that the type has been formatted by a ripple-lib API method, and isn't necessarily a standard rippled object.

src/ledger/parse/account-trustline.ts
FredKSchott: 👍 added back

intelliot:
Sounds good to me!

intelliot:
Aside from `authorized` 🙂 , are there any other changes/cleanups that could be breaking?

FredKSchott:
`authorized` is added back in, so nothing should be breaking/changed at this point. The only noticeable change (see tests) were that input validation for some API methods used to throw synchronously instead of rejecting the method promise. Now that we're using async functions, all throw errors reject the promise.

If a user had specific handling around function input validation, in some situations that may also need to change. It's minor, but may be worth considering when choosing the next version number.

src/common/types/objects/amounts.d.ts
```diff
@@ -0,0 +1,26 @@
...
+  currency: string,
+  issuer?: string,
+  counterparty?: string,
+}
+
+
+export type RippledAmount = string | Amount;
+
+/**
+ * Specification of which currency the account taking the offer would pay/receive, as an object with currency and issuer fields (omit issuer for XRP). Similar to currency amounts.
```
intelliot: nit: this comment is long for one line (could benefit from being hard wrapped)

src/ledger/orders.ts
```diff
@@ -1,37 +1,40 @@
...
-
-  return utils.ensureLedgerVersion.call(this, options).then(_options => {
-    const getter = _.partial(requestAccountOffers, this.connection, address,
-      _options.ledgerVersion)
-    return utils.getRecursive(getter, _options.limit).then(orders =>
-      _.sortBy(orders, order => order.properties.sequence))
+  // 2. Make Request
+  const responses = await this._requestAll('account_offers', {
+    account: address,
+    ledger_index: options.ledgerVersion || await this.getLedgerVersion(),
```
wilsonianb: Can we use `'validated'` instead of `await this.getLedgerVersion()`?

src/ledger/orderbook.ts
```diff
@@ -2,15 +2,20 @@ import * as _ from 'lodash'
...
-import {OrdersOptions, OrderSpecification} from './types'
-import {Amount, Issue} from '../common/types'
+import {OrderSpecification} from './types'
+import {Amount, Issue} from '../common/types/objects'
+import {RippleAPI} from '../api'
+import {OfferCreateTransaction} from '../common/types/objects'
+
+export type OrdersOptions = {
+  limit?: number,
+  ledgerVersion?: number
```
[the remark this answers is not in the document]
intelliot: Based on [the docs](https://ripple.com/build/rippleapi/#getorderbook), this is intended for specifying a historical ledger version. I'm not sure about closed/current - I'll take a closer look.

src/api.ts
```diff
@@ -106,6 +131,84 @@ class RippleAPI extends EventEmitter {
...
+        ...params,
+        limit: countRemaining,
+        marker
+      }
+      // NOTE: We have to generalize the `this._request()` function signature
+      // here until we add support for unknown commands (since command is some
+      // unknown string).
+      const singleResult = await (<Function>this._request)(command, repeatProps)
+      marker = singleResult.marker
+      count += singleResult[collectKey].length
```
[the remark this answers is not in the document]
FredKSchott: Good call, do we have a precedence for rejecting if the response object is malformed? 

As long as the API will always return this property for a response I think it's safe to reject if `singleResult[collectKey]` === `undefined`. But if that's not a safe assumption maybe we just stop collecting and return the responses collected so far. What do you guys think?

src/common/types/commands/book_offers.d.ts
```diff
@@ -0,0 +1,21 @@
...
+  taker_pays: TakerRequestAmount;
+  ledger_hash?: string;
+  ledger_index?: number | ("validated" | "closed" | "current");
+  limit?: number;
+  marker?: any;
+}
+
+export interface BookOffersResponse {
+  account: string;
+  offers: OfferCreateTransaction[];
```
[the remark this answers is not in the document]
FredKSchott: Good catch! Yea I think that's the right way to handle this

src/common/types/objects/transactions.d.ts
```diff
@@ -0,0 +1,14 @@
+import {RippledAmount} from './amounts'
+
+export interface OfferCreateTransaction {
```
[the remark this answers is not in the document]
FredKSchott: yea, these common fields should eventually go into some Transaction "base"/"common" type that the others can extend from. Let me add them to  `OfferCreateTransaction` directly for now, and then clean up when we touch the rest of the Transaction types (already looking to be the most complex part of typing).

src/ledger/orders.ts
FredKSchott: @wilsonianb My understanding is that `this.getLedgerVersion()` returns `_ledgerVersion`, which is always updated to the `ledger_index` of the last message. Would this be exactly similar to `validated`?

`"validated"` would be so much simpler, I'd love to make the change if we can.

FredKSchott:
Looks like the new types weren't linting, should be all fixed now.

@wilsonianb thanks for the feedback! Super helpful to get a second eye on those request/response type definitions. I left a few questions, PTAL when you can

src/ledger/orders.ts
```diff
@@ -1,37 +1,40 @@
...
+    limit: utils.clamp(options.limit, 10, 400) || undefined
```
intelliot: I think we should just pass `options.limit` here

src/ledger/trustlines.ts
```diff
@@ -1,53 +1,38 @@
...
+  // 1. Validate
+  validate.getTrustlines({address, options})
+  const ledgerVersion = await this.getLedgerVersion()
+  // 2. Make Request
+  const responses = await this._requestAll('account_lines', {
     account: address,
     ledger_index: ledgerVersion,
-    marker: marker,
-    limit: utils.clamp(limit, 10, 400),
+    limit: utils.clamp(options.limit, 10, 400),
```
intelliot: Should not `clamp`

FredKSchott:
@intelliot so I wanted to mimic the existing logic as close as possible, which meant continuing to use clamp in the existing API methods. But I totally agree it shouldn't be needed, which is why I was intentionally keeping it out of the new `_requestAll` method. The user will expect that request parameters passed through that function make it to the API.

Does that make sense? Happy to remove clamp from the existing logic if you don't think it's doing anything useful.

intelliot:
I think the existing logic works because it clamps later in the flow, allowing more than 400 results to ultimately be returned. (The limit is 400 for an individual request, but since ripple-lib makes multiple requests, you can get more than 400 items in the end.)

intelliot:
To clarify - this PR currently fails to mimic the existing logic because it doesn't replicate the behavior of `utils.getRecursive`. With the existing code, I am able to get multiple pages of results (by specifying `limit: 401` or higher), but with this PR, I can't go further than 1 page (400 results).

intelliot:
This revealed one of the (undoubtedly many) gaps in our unit test coverage. Here's a new test that catches the issue: https://github.com/ripple/ripple-lib/pull/850

FredKSchott:
@intelliot Ahhhh I see, so you can request `> 400` total items with the original `limit `option, and then the existing code clamps any one request `<= 400`. 

That's a bummer, I wanted to avoid manipulating the given request arguments as much as possible. But since we're already setting `limit` & `marker` across `requestAll` series, I guess that's fine for now. I'll move the logic into `requestAll`.

Longer term, if you said `rippled` is already implementing it's own clamping logic server-side, then we should defer to the server's preference and can remove the client-side clamping entirely.

intelliot:
Right! I've looked at the relevant rippled code and I believe we can remove the client-side clamping entirely. I'll call out the change in the release notes when that goes out.

Out of interest, is there a reason to use the do/while instead of recursion?

FredKSchott:
Just pushed the move the clamp inside `requestAll()`. Something I just realized: the current clamp lower bounds is 10, would you consider that a bug? If I passed `limit: 1`, I'd just expect to get a single result back. Can I drop that from 10 -> 1?

intelliot:
I think that's intentional based on [the docs for `account_lines`](https://ripple.com/build/rippled-apis/#account-lines):
> Must be within the inclusive range 10 to 400.

However, I don't think that's entirely accurate. Nothing bad happens if the value is beyond this range. I'll put in a PR for updating these docs and see what the rest of the team thinks.

In the current implementation, the result is truncated client-side, so if you pass `limit: 1`, you do get a single result back: https://github.com/ripple/ripple-lib/blob/0.18.2/src/ledger/utils.ts#L43

FredKSchott:
@intelliot ha! I didn't even realize that. I'll add that behavior back for now as well and then we can revaluate all this again when it's time to make `requestAll` public.

src/api.ts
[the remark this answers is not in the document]
intelliot: I'm not sure - I think either way should be fine, since this shouldn't actually happen.

src/ledger/orderbook.ts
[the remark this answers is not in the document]
intelliot: ripple-lib currently doesn't allow using not-validated ledger versions. This is something we could revisit in the future, but for now, this is intentional.

src/ledger/orders.ts
intelliot: I think `validated` would be the same. We can make the change in a later PR, though.

intelliot:
@wilsonianb Could you take another look?

src/api.ts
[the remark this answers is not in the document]
wilsonianb: Looking at some of the existing parse functions in `src/ledger/parse/`, we seem to be in the habit of *not* rejecting if the response object is malformed.
https://github.com/ripple/ripple-lib/blob/e311b74dac0a90e7a83b9126f0d64840a395e789/src/ledger/parse/payment-channel.ts#L46-L61
So I'd vote for returning the collected responses.

FredKSchott:
@wilsonianb thanks for the review! Addressed both 👍

src/api.ts
```diff
@@ -208,7 +208,8 @@ class RippleAPI extends EventEmitter {
       lastBatchLength = singleResult[collectKey].length
       results.push(singleResult)
     } while(!!marker && count < countTo && lastBatchLength !== 0)
-    return results
+    // NOTE: We slice out any additional items over the original requested num
+    return results.slice(0, countTo)
```
FredKSchott: follow up: this was reverted for now. Since `results` is an array of full response objects (where responses can contain multiple data objects).

PR #847, opened 2018-02-07
Add "browser" property to package.json

intelliot:
This should be an alternative to https://github.com/ripple/ripple-lib/pull/840

~~We need to run the `build` script for this file to exist, so how should we handle that?~~

~~Perhaps we can just make sure "./build/ripple-latest.js" is prebuilt and published to npm?~~

See also: https://github.com/ripple/ripple-lib/pull/844#issuecomment-363514078

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "clean": "rm -rf dist/npm",
     "compile": "mkdir -p dist/npm/common && cp -r src/common/schemas dist/npm/common/ && tsc",
     "watch": "tsc -w",
-    "prepublish": "npm run clean && npm run compile",
+    "prepublish": "npm run clean && npm run compile && npm run build",
     "test": "nyc mocha",
     "coveralls": "cat ./coverage/lcov.info | coveralls",
     "lint": "tslint -p ./",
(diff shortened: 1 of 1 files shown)


intelliot:
@mikeb1rd Could you verify that this solution works for you?

intelliot:
Note that you'll need to go inside `ripple-lib/` and run `yarn build` (or `npm build`), otherwise you'll get
```
Error: Cannot find module 'ripple-lib' from ...
```

FredKSchott:
@intelliot that's right, you would build these files on every publish to npm. `"prepublish": "npm run clean && npm run compile"` should take care of this for us already.

intelliot:
Okay, I have a working solution now. With bbfc53e, instead of using the webpack build, we are just substituting `wswrapper` instead of `ws`. This worked for me; @mikeb1rd could you give it a try?

package.json
```diff
@@ -64,7 +67,7 @@
     "clean": "rm -rf dist/npm",
     "compile": "mkdir -p dist/npm/common && cp -r src/common/schemas dist/npm/common/ && tsc",
     "watch": "tsc -w",
-    "prepublish": "npm run clean && npm run compile",
+    "prepublish": "npm run clean && npm run compile && npm run build",
```
mikeb1rd: This can be reverted now, as the built files are not used

mikeb1rd:
@intelliot Thanks for spending time on this. I've just added one comment, otherwise the solution works well.

PR #850, opened 2018-02-13
Add test of getTrustlines with multiple pages of results

intelliot:
We were not testing the pagination of `getTrustlines`, which should occur when more than 400 items are requested. This test passes on `develop` but fails with `0f0edbf0c875d0b076500cfdf35df2deba2e9f70` (branch tip of https://github.com/ripple/ripple-lib/pull/843).


test/fixtures/responses/index.js
```diff
@@ -1,5 +1,14 @@
 'use strict'; // eslint-disable-line strict
 
+function moreThan400Items() {
+  let items = [];
+  const oneItem = require('./get-trustlines-more-than-400-items.json');
+  for (let i = 0; i < 401; i++) {
+    items.push(oneItem);
+  }
+  return items;
```
hserang: In the words of James Brown, you can make it funky(tional) as follows:
```
function moreThan400Items(item) {
  return Array[401].fill(item);
}
```
then call it with the item to inject as an arg so it's generic, or require it as you see fit.

test/fixtures/responses/index.js
```diff
@@ -61,6 +70,7 @@ module.exports = {
   },
   getTrustlines: {
     filtered: require('./get-trustlines.json'),
+    moreThan400Items: moreThan400Items(),
```
hserang: given the pattern here, I would pass the json as an arg into the method:
```
moreThan400Items: moreThan400Items(require('./get-trustlines-more-than-400-items.json')),
```

test/fixtures/responses/get-trustlines-more-than-400-items.json
```diff
@@ -0,0 +1,13 @@
+{
```
hserang: Nit, this json fixture is only one item. Could this be renamed to something better aligned, (just brainstorming here) `trustline-response.json` or `trustline-item.json`?

intelliot:
@hserang Brilliant! I took all your suggestions (using the syntax `new Array(401)`). Could you have a second look?

test/fixtures/rippled/account-lines.js
```diff
@@ -321,3 +321,37 @@ module.exports.counterparty = function(request, options = {}) {
...
 };
+
+module.exports.manyItems = function(request, options = {}) {
+  _.defaults(options, {
+    ledger: BASE_LEDGER_INDEX
+  });
+
+  options.marker = 'MORE_TO_COME';
+
+  let lines = [];
```
hserang: Just a nit: let me know how you feel about this more functional approach:
Outside the `exports`, create a new method:
```
getLinesFromRequest(request) {
  const linesCount = Math.max(request.limit, 10);
  const trustLinesFixture = {
    account: 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B',
    balance: '0.3488146605801446',
    ...
  };
  
  return new Array(linesCount).fill(trustLinesFixture);
}
```

cont'd below

test/fixtures/rippled/account-lines.js
```diff
@@ -321,3 +321,37 @@ module.exports.counterparty = function(request, options = {}) {
...
+  return JSON.stringify({
+    id: request.id,
+    status: 'success',
+    type: 'response',
+    result: {
+      account: request.account,
+      marker: options.marker,
+      limit: request.limit,
+      ledger_index: options.ledger,
+      lines: lines
```
hserang: If you like the above suggestion, then you would call it here:
`lines: getTrustLinesFromRequest(request),`

test/fixtures/responses/index.js
```diff
@@ -61,6 +65,7 @@ module.exports = {
   },
   getTrustlines: {
     filtered: require('./get-trustlines.json'),
+    moreThan400Items: moreThan400Items(require('./trustline-item.json')),
```
hserang: One more refinement - how about this:
`moreThan400Items: buildList(require('./trustline-item.json'), 401)`

This keeps our builder generic and places all concerns in one place.

intelliot:
@hserang I've changed the test to return a realistic number of items - thanks for catching that. Have another look whenever!

test/api-test.js
```diff
@@ -995,6 +995,12 @@ describe('RippleAPI', function() {
         responses.getTrustlines.filtered, 'getTrustlines'));
   });
 
+  it('getTrustlines - more than 400 items', function() {
+    const options = {limit: 401};
```
hserang: Only asking, but do we need to set the `marker` in the options for the request? If so, should we have multiple tests for different markers?

intelliot: No, I think `marker` is just handled internally by ripple-lib. See [the parameters for `getTrustlines` `options`](https://ripple.com/build/rippleapi/#gettrustlines).

PR #851, opened 2018-02-13
Generate/publish *.ts.d

cryptcoin-junkey:
They are very useful in developing stable applications with TypeScript.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,7 +15,7 @@
     "removeComments": false,
     "preserveConstEnums": false,
     "suppressImplicitAnyIndexErrors": false,
-    "declaration": false,
+    "declaration": true,
     "sourceMap": true,
     "skipLibCheck": true
   },


intelliot:
@FredKSchott How are declaration files typically used? Is it automatically parsed by editors like VS Code?

FredKSchott:
Exactly, the definition files (`.d.ts`) are what allow your editors / editor plugins / build step to understand `ripple-lib` when it's a dependency and warn when it's used incorrectly.

darkmemo:
Afaik, additionally `types` property in `package.json` should point to main types declaration file.

```js
[...]
"main": "dist/npm",
"name": "ripple-lib",
"types": "dist/npm/index.d.ts"
[...]
```
See https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

Without this I always get _Could not find a declaration file for module 'ripple-lib'. '../ripple-lib/dist/npm/index.js' implicitly has an 'any' type..._ error. (I have `"noImplicitAny": true` in `tsconfig.json`.)

When I add `"types": "dist/npm/index.d.ts"` to `package.json` of `ripple-lib`, module and types are resolved correctly. (Using typescript version 2.7.2)

FredKSchott:
@darkmemo good catch!

PR #853, opened 2018-02-16
Add support for Checks

intelliot:
- See ripple/rippled#2245


src/common/schemas/specifications/check-cash.json
```diff
@@ -0,0 +1,26 @@
...
+      "$ref": "hash256",
+      "description": "The ID of the Check ledger object to cash, as a 64-character hexadecimal string."
+    },
+    "amount": {
+      "$ref": "laxAmount",
+      "description": "Redeem the Check for exactly this amount, if possible. The currency must match that of the sendMax of the corresponding CheckCreate transaction. You must provide either this field or deliverMin."
+    },
+    "deliverMin": {
+      "$ref": "laxAmount",
+      "description": "Redeem the Check for at least this amount as for as much as possible. The currency must match that of the sendMax of the corresponding CheckCreate transaction. You must provide either this field or amount."
```
scottschurr: "...amount _or_ for as much..."?

src/common/schemas/specifications/check-create.json
```diff
@@ -0,0 +1,31 @@
...
+  "link": "check-create",
+  "type": "object",
+  "properties": {
+    "destination": {
+      "$ref": "address",
+      "description": "Address of the account that can cash the check."
+    },
+    "sendMax": {
+      "$ref": "laxAmount",
+      "description": "Amount of source currency the check is allowed to debit the sender, including transfer fees on non-XRP currencies."
```
scottschurr: Nice description.

src/common/schemas/specifications/check-create.json
```diff
@@ -0,0 +1,31 @@
...
+    },
+    "destinationTag": {
+      "$ref": "tag",
+      "description": "Destination tag that identifies the reason for the check, or a hosted recipient to pay."
+    },
```
scottschurr: I'll mention is passing that a "sourceTag" is also an allowed field.  But that's allowed in any transaction, so you may be handling that elsewhere.

src/ledger/parse/check-cash.ts
```diff
@@ -0,0 +1,24 @@
...
+    checkID: tx.CheckID,
+
+    // (Optional) redeem the Check for exactly this amount, if possible
+    amount: parseAmount(tx.Amount).value,
+
+    // (Optional) redeem the Check for at least this amount and
+    // for as much as possible
+    deliverMin: parseAmount(tx.DeliverMin).value
+
+    // *must* include either Amount or DeliverMin, but not both
```
scottschurr: :+1:

wilsonianb:
>I'm wondering if ripple-lib has support for finding account object

There's [`getPaymentChannel`](https://github.com/ripple/ripple-lib/blob/develop/docs/index.md#getpaymentchannel), but that's by id as opposed to something like rippled's [`account_objects`](https://ripple.com/build/rippled-apis/#account-objects)

src/common/schemas/specifications/check-cash.json
intelliot: _and_

:)

FredKSchott:
> I'm wondering if ripple-lib has support for finding account object
> > There's getPaymentChannel, but that's by id as opposed to something like rippled's account_objects

Today you can call `api.connection.request({})` to make rippled requests like `{command: 'account_objects', ...}` directly, although it's kind of hidden. Work is being done to expose this more clearly via a new `api.request()` method.

src/ledger/parse/check-create.ts
```diff
@@ -0,0 +1,29 @@
+import * as assert from 'assert'
+import {parseTimestamp} from './utils'
+import {removeUndefined} from '../../common'
+import parseAmount from './amount'
+
+function parseCheckCreate(tx: any): Object {
```
FredKSchott: We're moving towards typing the output of these parsing functions so that they're easier to consume (also a great place to document the formatted type). 

So instead of returning a type `Object` you would return a type `FormattedCheckCreate`, something like this:

```ts
export type FormattedCheckCreate = {
  // account that can cash the check.
  destination: string,
  // amount the check is allowed to debit the sender, including transfer fees on non-XRP currencies.
  sendMax: Amount,
  // (Optional) identifies the reason for the check, or a hosted recipient.
  destinationTag?: string,
  // (Optional) time in seconds since the Ripple Epoch.
  expiration?: string, 
  // (Optional) 256-bit hash representing a specific reason or identifier.
  invoiceID?: string
}

function parseCheckCreate(tx: any): FormattedCheckCreate {
  // ...
}
```

Can you do this for the other two new parsers?

src/transaction/check-cancel.ts
```diff
@@ -0,0 +1,32 @@
...
+import {validate} from '../common'
+import {Instructions, Prepare} from './types'
+
+type CheckCancel = {
+  checkID: string
+}
+
+function createCheckCancelTransaction(account: string,
+  cancel: CheckCancel
+): Object {
```
FredKSchott: `Object` actually doesn't mean an arbitrary object like it did in Glow, it's closer to an empty object than an arbitrary one. Change to `any` or `object` (lower-case) to represent any arbitrary object, and then you should be able to remove the `: any` you had to add to the txJSON below.

src/transaction/check-cash.ts
```diff
@@ -0,0 +1,50 @@
...
+  checkCash: CheckCash
+): Object {
+  const txJSON: any = {
+    Account: account,
+    TransactionType: 'CheckCash',
+    CheckID: checkCash.checkID
+  }
+
+  if (checkCash.amount && checkCash.deliverMin) {
+    throw new ValidationError('"amount" and "deliverMin" properties on '
```
FredKSchott: We tend to do data validation as the first thing in the function. It's a good general rule so that you don't have to worry about what logic is safe to do pre-validation vs. post-validation (ex: if someone later had to add something to the txJSON definition above).

intelliot:
@FredKSchott Thank you! Have a second look?

intelliot:
@shekenahglory @wilsonianb Would love if you guys could review as well.

PR #856, opened 2018-02-25
Improve documentation of getTransactions params

### LINKED ISSUE #855: ValidationError in getTransactions
Getting transactions for specific account works find without "start" option but after including the "start" param I'm getting below error :

`instance.options is of prohibited type [object Object]`

-----------------------------------------------------------------

Works Fine :

`{minLedgerVersion: 5978794, maxLedgerVersion: 6963598, limit: 5}`

Getting Error : 
`
{minLedgerVersion: 5978794, maxLedgerVersion: 6963599, limit: 5, start: "0CD33DF995C6545038F58EB65F46671D0C3BCD2288EEF1CF350F597A3C5F110F"}`

Also tried to include other options and almost tried everything but can't figure out where is the problem .

Regards

intelliot:
- Resolve #855

### DIFF
--- a/src/common/schemas/specifications/payment.json
+++ b/src/common/schemas/specifications/payment.json
@@ -22,7 +22,7 @@
       "$ref": "hash256"
     },
     "allowPartialPayment": {
-      "description": "If true, this payment should proceed even if the whole amount cannot be delivered. This can occur due to a lack of liquidity or a lack of funds in the source account.",
+      "description": "If true, this payment should proceed even if the whole amount cannot be delivered due to a lack of liquidity or a lack of funds in the source account.",
       "type": "boolean"
     },
     "noDirectRipple": {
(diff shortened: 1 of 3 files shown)


src/common/schemas/specifications/payment.json
```diff
@@ -22,11 +22,11 @@
       "$ref": "hash256"
     },
     "allowPartialPayment": {
-      "description": "A boolean that, if set to true, indicates that this payment should go through even if the whole amount cannot be delivered because of a lack of liquidity or funds in the source account account",
+      "description": "If true, this payment should proceed even if the whole amount cannot be delivered. This can occur due to a lack of liquidity or a lack of funds in the source account.",
```
mDuo13: "This can occur" makes it sound almost like the `allowPartialPayment` flag could get turned on when this situation occurs. Suggested rephrase:

> If true, this payment should proceed even if the whole amount cannot be delivered due to a lack of liquidity or a lack of funds in the source account.

PR #857, opened 2018-02-25
Existing types cleanup + more api methods onto new request method

FredKSchott:
- This PR ended up tackling **a lot** of type clean up and organization
- getPaymentChannel() now uses this.request()
- getSettings() now uses this.request()
- getLedger() now uses this.request()
- existing transaction types (`ledger/types.ts`, `ledger/transaction-types.ts`, etc.) cleaned up, but not yet completely moved to new system

/cc @intelliot


src/common/types/objects/ledger_entries.ts
```diff
@@ -0,0 +1,56 @@
+import {SignerEntry} from './index'
+
+export type PayChannelLedgerEntry = {
```
intelliot: Why use `type` here instead of `interface`?

sublimator:
Might be a leftover from Flow syntax :)

intelliot:
Superseded by #860

src/common/types/objects/ledger_entries.ts
FredKSchott: @sublimator is right, that's a left over from Flow. Changing to interface...

FredKSchott:
@sublimator is right, that's a left over from Flow. Changing to interface...

PR #860, opened 2018-03-02
Existing types cleanup + more api methods onto new request method

intelliot:
This should be identical to https://github.com/ripple/ripple-lib/pull/857 but with merge conflicts resolved.

- This PR ended up tackling **a lot** of type clean up and organization
- getPaymentChannel() now uses this.request()
- getSettings() now uses this.request()
- getLedger() now uses this.request()
- existing transaction types (`ledger/types.ts`, `ledger/transaction-types.ts`, etc.) cleaned up, but not yet completely moved to new system


intelliot:
I learned how to push the merge commit to the original branch, so this PR is not needed. Use https://github.com/ripple/ripple-lib/pull/857 instead.

PR #865, opened 2018-03-08
Add request('account_objects', ...)

intelliot:
Use `request` for rippled `account_objects` command.

See https://ripple.com/build/rippled-apis/#account-objects


FredKSchott:
rebase on / reset Github PR "base" to master?

src/api.ts
```diff
@@ -85,6 +86,85 @@ function getCollectKeyFromCommand(command: string): string|undefined {
   }
 }
 
+// function validateRequest(
```
FredKSchott: remove?

src/common/types/objects/ledger_entries.ts
```diff
@@ -37,6 +37,49 @@ export interface AccountRootLedgerEntry {
...
+  SendMax: string | object,
+  Sequence: number,
+  DestinationNode: string,
+  DestinationTag: number,
+  Expiration: number,
+  InvoiceID: string,
+  SourceTag: number
+}
+
+export interface PayChannelLedgerEntry {
```
FredKSchott: merge with `PayChannelLedgerEntry ` already defined above

src/common/types/objects/ledger_entries.ts
```diff
@@ -37,6 +37,49 @@ export interface AccountRootLedgerEntry {
   Domain?: string
 }
 
+export interface AmendmentsLedgerEntry {
+  Amendments?: string[],
+  Majorities?: any[],
+  Flags: 0,
+  LedgerEntryType: 'Amendments'
```
FredKSchott: nit: move to first property in interface

src/api.ts
```diff
@@ -85,6 +86,85 @@ function getCollectKeyFromCommand(command: string): string|undefined {
...
+//   command: 'gateway_balances',
+//   params: GatewayBalancesRequest
+// ): void
+
+/**
+ * Validate the request.
+ *
+ * Throws if `params` are invalid for `command`.
+ */
+function validateRequest(command: string, params: any): void {
```
FredKSchott: This isn't necessary in the new `request()` world, it should be the responsibility of the API to tell the user what was wrong with their request. Otherwise we end up duplicating and having to maintain validation logic in two separate places.

(This did make sense in the old world & still makes sense for the legacy methods, where `getSettings()`/`getAccountInfo()` were all special and being defined by the SDK).

FredKSchott:
1 thing to keep in mind: I don't feel 100% ready to finalize the interface for `request()`. We're still in the process of figuring out the best interface here (see: recent discussions around requestAll) and giving this to users now means we won't be able to make changes later (or, we'll have to make changes very very carefully so as not to break their apps).

How do you feel about keeping `_request()`/`_requestAll()` private, and wrapping this new behavior in a simple legacy formatted API method like `getAccountObjects()`, or even just `getChecks()`. That way we could keep the validation logic in that method, and would be free to keep iterating on request/requestAll until we feel confident in how they look/perform.

wdyt?

intelliot:
Sounds good. I prefer `getAccountObjects()` as that's more aligned with the rippled API. If `getChecks()` is the better API, then we can propose that rippled add a `checks` command.

src/ledger/accountobjects.ts
```diff
@@ -0,0 +1,21 @@
+import {removeUndefined} from '../common'
+import {RippleAPI} from '../api'
+import {AccountObjectsResponse} from '../common/types/commands/account_objects'
+
+export default async function getAccountObjects(
+  this: RippleAPI, address: string, options: any = {}
```
FredKSchott: since we're pulling off specific properties in this function code below, define the type of the `options` object.

Ex:
```ts
interface GetAccountObjectsOptions {
  type: string,
  ledgerHash?: string,
  ledgerIndex?: string,
  limit?: number
}
```

src/ledger/accountobjects.ts
```diff
@@ -0,0 +1,21 @@
...
+): Promise<AccountObjectsResponse> {
+  // Intentionally omit local validation.
+  // Validation is performed by rippled.
```
FredKSchott: Do you want to validate the `options` object as we do in other functions? I'm okay if we don't, but I'd clarify the comment since we're talking about verifying the `options`/`GetAccountObjectsOptions` object, which is owned by ripple-lib and is never sent directly to rippled.

src/api.ts
```diff
@@ -85,6 +87,74 @@ function getCollectKeyFromCommand(command: string): string|undefined {
   }
 }
 
+/**
+ * Validate the request.
+ *
+ * Throws if `params` are invalid for `command`.
+ */
+function _validateRequest(command: string, params: any): void {
```
FredKSchott: Remove this function, unnecessary/redundant to the validation rippled is already doing (see last comment here: https://github.com/ripple/ripple-lib/pull/865#discussion_r174861916)

intelliot:
Superseded by https://github.com/ripple/ripple-lib/pull/881

PR #873, opened 2018-03-20
 more clean up for min/max adjustments

FredKSchott:
Builds on top of #870, ~should be okay to merge after that PR is merged in.~ #870 merged in, this is ready to review/merge. 

Uses some TS magic to make type checking easier, and also consolidates some destination/source amount logic to reduce repeated code.

### DIFF
--- a/src/transaction/payment.ts
+++ b/src/transaction/payment.ts
@@ -10,8 +10,8 @@ import {Amount, Adjustment, MaxAdjustment,
 
 
   export type Payment = {
-  source: Adjustment & MaxAdjustment,
-  destination: Adjustment & MinAdjustment,
+  source: Adjustment | MaxAdjustment,
+  destination: Adjustment | MinAdjustment,
   paths?: string,
   memos?: Array<Memo>,
   // A 256-bit hash that can be used to identify a particular payment
@@ -74,8 +74,8 @@ function createPaymentTransaction(address: string, paymentArgument: Payment
     throw new ValidationError('address must match payment.source.address')
   }
 
-  if ((payment.source.maxAmount && payment.destination.minAmount) ||
-      (payment.source.amount && payment.destination.amount)) {
+  if (((<MaxAdjustment>payment.source).maxAmount && (<MinAdjustment>payment.destination).minAmount) ||
+      ((<Adjustment>payment.source).amount && (<Adjustment>payment.destination).amount)) {
     throw new ValidationError('payment must specify either (source.maxAmount '
       + 'and destination.amount) or (source.amount and destination.minAmount)')
   }
@@ -86,9 +86,9 @@ function createPaymentTransaction(address: string, paymentArgument: Payment
   // send the whole source amount, so we set the destination amount to the
   // maximum possible amount. otherwise it's possible that the destination
   // cap could be hit before the source cap.
-  const amount = payment.destination.minAmount && !isXRPToXRPPayment(payment) ?
-    createMaximalAmount(payment.destination.minAmount) :
-    (payment.destination.amount || payment.destination.minAmount)
+  const amount = (<MinAdjustment>payment.destination).minAmount && !isXRPToXRPPayment(payment) ?
+    createMaximalAmount((<MinAdjustment>payment.destination).minAmount) :
+    ((<Adjustment>payment.destination).amount || (<MinAdjustment>payment.destination).minAmount)
 
   const txJSON: any = {
     TransactionType: 'Payment',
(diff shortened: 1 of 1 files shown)


src/transaction/payment.ts
```diff
@@ -30,11 +30,22 @@ import {Amount, Adjustment, MaxAdjustment,
   limitQuality?: boolean
 }
 
+function isMaxAdjustment(
+  source: Adjustment | MaxAdjustment): source is MaxAdjustment {
```
emschwartz: Nitpick: why is this on a new line?

src/transaction/payment.ts
```diff
@@ -9,7 +9,7 @@ import {Amount, Adjustment, MaxAdjustment,
   MinAdjustment, Memo} from '../common/types/objects'
 
 
-  export type Payment = {
+export type Payment = {
```
emschwartz: I think this should be an `interface` rather than a `type`

src/transaction/payment.ts
FredKSchott: exceeds 80 character max, enforced by our linting config :/

src/transaction/payment.ts
FredKSchott: +1, I can change that here

src/transaction/payment.ts
emschwartz: Ah, I see. Looks strange to me cause I'm used to the "standard" style

sublimator:
Worth considering if the test suite should be converted to TypeScript as well?

To handle cases like the following where accidentally specifying multiple properties:
```typescript
const source = {
  address: fromAddress,
  maxAmount: {value: amount, currency: "XRP"},
  amount: {value: amount, currency: "XRP"}
};
const destination = {
  address: toAddress,
  amount: {value: amount, currency: "XRP"},
};
```
You can use optional fields, specified as undefined:

```typescript
export interface Adjustment {
    address: string;
    amount: Amount;
    maxAmount?: undefined;
    minAmount?: undefined;
    tag?: number;
};
export interface MaxAdjustment  {
    address: string;
    amount?: undefined;
    maxAmount: Amount;
    tag?: number;
};
export interface MinAdjustment {
    address: string;
    amount?: undefined;
    minAmount: Amount;
    tag?: number;
};
```

Which yields:
```
      Type '{ address: string; amount: { value: string; currency: string; }; maxAmount: { value: string; curr...' is not assignable to type 'MaxAdjustment'.
        Types of property 'amount' are incompatible.
          Type '{ value: string; currency: string; }' is not assignable to type 'undefined'.
```

See also: [now recommended that you just use "as foo" for consistency](https://basarat.gitbooks.io/typescript/docs/types/type-assertion.html)
image

sublimator:
You can also do:
```
export declare type Payment = {
    paths?: string;
    memos?: Array<Memo>;
    invoiceID?: string;
    allowPartialPayment?: boolean;
    noDirectRipple?: boolean;
    limitQuality?: boolean;
} & ({
  source: Adjustment;
  destination: MinAdjustment;
} | {
  source: MaxAdjustment;
  destination: Adjustment;
} | {
    source: Adjustment,
    destination: Adjustment
});
```

Though the error messages are gonna be for C++ fans :)

sublimator:
Prior comment regarding: https://github.com/ripple/ripple-lib/blob/develop/src/transaction/payment.ts#L77-L81

sublimator:
You could even use the typescript compiler programmatically, and add test cases of strings which should work or fail ...

intelliot:
How does a3ae796 look?

sublimator:
Beginnings of TypeScript tests to build upon: https://github.com/sublimator/ripple-lib/commit/28e172ab115f4b6a4cba90f95a52f082a2f4276f

FredKSchott:
+1 for converting the tests to TS as well

---

@intelliot 👍 

---

@sublimator judging by the code, it sounds like `{source: Adjustment, destination: Adjustment}` isn't actually valid either, which would simplify your example above 👌

If this function was user-facing I think your more complex but most strict type definition would be worth the gross error messages. But since it wouldn't simplify the `createPaymentTransaction()` implementation (I don't think?) and the user never actually interacts with this type directly, I'm inclined to value the more simple definition.

sublimator:
> it sounds like {source: Adjustment, destination: Adjustment} isn't actually valid either, which would simplify your example above 👌

@FredKSchott 

I never use this library, just happened to be chatting to @intelliot on Google chat recently and this issue came up. This lib gives me headaches having to mentally map back and forth between  ripple terms.  But why wouldn't it be a valid combination?  Are you just going by the error message I linked?

What about: https://github.com/ripple/ripple-lib/blob/develop/src/transaction/payment.ts#L98-L100

? 

> the user never actually interacts with this type directly
`this type` = Payment ?

api.preparePayment ? 

> Though the error messages are gonna be for C++ fans :)

^^^ 
lol, I don't think I was seriously suggesting it :)

FredKSchott:
> `api.preparePayment`

Whoops, of course! Okay, well then I could definitely see the more strict type being useful, if it could catch bad user input before runtime. 

> Why wouldn't it be a valid combination?

I was going by this validation check in the source code itself: https://github.com/ripple/ripple-lib/blob/develop/src/transaction/payment.ts#L91, which throws a `ValidationError` if that's encountered.

> This lib gives me headaches having to mentally map back and forth between ripple terms.

Agreed! [The work we've been doing since way back in November](https://github.com/ripple/ripple-lib/issues/812) has been to move us towards a world where ripple-lib uses the exact same concepts/terms/types/commands you're already using with rippled.

PR #881, opened 2018-04-02
Add getAccountObjects

intelliot:
Add `getAccountObjects` to implement the rippled `account_objects` command.

See https://ripple.com/build/rippled-apis/#account-objects


src/common/types/objects/ledger_entries.ts
```diff
@@ -1,5 +1,47 @@
...
+  PreviousTxnID: string,
+  PreviousTxnLgrSeq: number,
+  AccountTxnID?: string,
+  RegularKey?: string,
+  EmailHash?: string,
+  MessageKey?: string
+  TickSize?: number,
+  TransferRate?: number,
+  Domain?: string
+}
```
mDuo13: A quick glance at the code tells me it may still be possible to get (and set!) `WalletLocator` and `WalletSize` fields with the AccountSet transaction. They've been marked deprecated for a long time, but still, may be better to have them here so you don't error out if you encounter one?

test/api-test.js
```diff
@@ -1149,6 +1149,41 @@ describe('RippleAPI', function () {
...
+  it('request account_objects - invalid options', function () {
+    // Intentionally no local validation of these options
+    return this.api._request('account_objects', {
+      account: address,
+      invalid: 'options'
+    }).then(response =>
+      checkResult(responses.getAccountObjects, 'AccountObjectsResponse', response));
+  });
+
+  xit('request account_objects - multiple pages', function () {
```
mDuo13: should this say "it" instead of "xit"?

PR #883, opened 2018-04-10
Upgrade https-proxy-agent to version 2

kizzx2:
https-proxy-agent <= 2.1.1 (including 1.0.0) has a potential vulernability as documented in https://hackerone.com/reports/319532

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@types/lodash": "^4.14.85",
     "@types/ws": "^3.2.0",
     "bignumber.js": "^4.1.0",
-    "https-proxy-agent": "^2.0.0",
+    "https-proxy-agent": "2.2.1",
     "jsonschema": "1.2.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "@types/lodash": "^4.14.85",
     "@types/ws": "^3.2.0",
     "bignumber.js": "^4.1.0",
-    "https-proxy-agent": "^1.0.0",
+    "https-proxy-agent": "^2.0.0",
```
intelliot: Looks like the vuln was fixed in 2.2.0. Can we make this "2.2.1" (no `^`) to make sure we get the version we expect?

Due to https://github.com/ripple/ripple-lib/pull/882 we'd like to use specific versions to reduce unpredictability.

kizzx2: Sure, done

PR #886, opened 2018-04-24
getOrderbook - return raw order data

### LINKED ISSUE #799: getOrderbook doesn't return all the book's info
Some fields returned by rippled are not returned by api.getOrderbook().
For example the "owner_funds" field which is fundamental to build autobridged books and the "flags" field.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49634790-getorderbook-doesn-t-return-all-the-book-s-info?utm_campaign=plugin&utm_content=tracker%2F323756&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F323756&utm_medium=issues&utm_source=github).
</bountysource-plugin>

intelliot:
Include the full `BookOffer` data under `data`

Fix #799

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2272,6 +2272,7 @@ bids[] | object | An order in the order book.
 *bids[].properties.* maker | [address](#address) | The address of the account that submitted the order.
 *bids[].properties.* sequence | [sequence](#account-sequence-number) | The account sequence number of the transaction that created this order.
 *bids[].properties.* makerExchangeRate | [value](#value) | The exchange rate from the point of view of the account that submitted the order (also known as "quality").
+*bids[].data.* \* | object | 
 *bids[].* state | object | *Optional* The state of the order.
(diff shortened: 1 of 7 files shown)


sublimator:
@FredKSchott 
> spend some of my sprint
Is  that some of the fabled 20% time at Google?

intelliot:
@FredKSchott I don't think we should transparently make multiple requests behind-the-scenes. Pagination exists to prevent overloading the rippled server. If we wanted to request _all_ at once, we'd make a change in rippled. However, rippled already supports this: developers already get all of the results when they're on an 'admin' connection.

Yes, we should finalize `request()`. Here's one proposal for doing it: https://github.com/ripple/ripple-lib/pull/887

FredKSchott:
@intelliot +1, I like the approach in #887 even better. I'll try to take a look this week

PR #892, opened 2018-05-10
Allow specifying amounts in drops

intelliot:
### Breaking Changes

+ Amounts in drops (recommended) and XRP are checked for validity. Some
  methods may now throw a `BigNumber Error` or `ValidationError` if the amount
  is invalid. This may include methods that previously did not throw.
+ Note that 1 drop is equivalent to 0.000001 XRP and 1 XRP is equivalent to 1,000,000 drops.

### Other Changes

+ Allow specifying amounts in drops for consistency with the `rippled` APIs.
+ Export `xrpToDrops()` and `dropsToXrp()` functions.


sublimator:
Dayumn, not sure what I think of this :)

tuloski:
This is very dangerous. I'm sure someone (myself included) will make some mess forgetting that XRP will be in drops and not anymore in XRP.

src/common/utils.ts
```diff
@@ -13,18 +13,73 @@ function isValidSecret(secret: string): boolean {
   }
 }
 
-function dropsToXrp(drops: string): string {
-  return (new BigNumber(drops)).dividedBy(1000000.0).toString()
+function dropsToXrp(drops: string | BigNumber): string {
+  if (typeof drops === 'string' && !drops.match(/^-?[0-9.]+$/)) {
```
sharafian: Shouldn't drops support integer only? Something like `^(0|-?[1-9][0-9]*)$`

intelliot:
@tuloski Existing code continues to work as-is. So XRP amounts are exactly the same as before:
```
{
  "currency": "XRP",
  "value": "2000"
}
```

This change enables a new "currency" called "drops":
```
{
  "currency": "drops",
  "value": "2000000000"
}
```

For convenience, you can use `xrpToDrops()` and `dropsToXrp()` to convert between XRP and drops.

The main reasons for this addition are:
1. All rippled APIs require XRP amounts to be specified in drops.
2. Other assets like Bitcoin and Ethereum work similarly (satoshis and wei, respectively).

test/api-test.js
```diff
@@ -51,6 +52,206 @@ describe('RippleAPI', function () {
...
+        this.api.xrpToDrops('1e-7')
+      }, /invalid value/)
+
+      assert.throws(() => {
+        this.api.xrpToDrops('2,0')
+      }, /invalid value/)
+
+      assert.throws(() => {
+        this.api.xrpToDrops('.')
+      }, /BigNumber Error: new BigNumber\(\) not a number: \./)
```
wilsonianb: I'd be in favor of modifying the initial `invalid value` checks in both `xrpToDrops` and `dropsToXrp` to catch anything that we currently expect a `BigNumber Error` for.
https://github.com/ripple/ripple-lib/pull/892/commits/70368ab44aa2791a27c7afca7bcebf3162661769#diff-077594e9fc0011267e9f099a04f7a2b1R17
https://github.com/ripple/ripple-lib/pull/892/commits/70368ab44aa2791a27c7afca7bcebf3162661769#diff-077594e9fc0011267e9f099a04f7a2b1R41

intelliot: Hmm… I don't know all of the ways that a `BigNumber Error` can occur, and the bignumber.js code is difficult for me to understand.

wilsonianb: I think changing the regexes to `/^-?[0-9]*\.?[0-9]+$/` would at least let us change all the unittests that currently expect `BigNumber Error` to instead expect `invalid value`.

intelliot: Since I want to allow numbers ending in a decimal point (e.g. `2.`), I opted for `/^-?[0-9]*\.?[0-9]*$/` and a separate check for `.`.

intelliot:
@sharafian For using drops with the Paychan stuff, how about something like: https://github.com/ripple/ripple-lib/pull/898

PR #894, opened 2018-05-12
use any instead of object

FredKSchott:
In TS, the `object` type is just an empty `{}` object. Instead, use `any` to represent an object as a flexible, undefined object type.

Without this change, you get errors when using an untyped request/response.

(this PR also cleaned up the request() jsdoc a bit as well)


FredKSchott:
Superseded by #896

PR #895, opened 2018-05-12
add "server_info" rippled request typing

FredKSchott:
- Converts `getServerInfo` & `getFee` to use the new `api.request()` interface internally
- Adds typing info for the `server_info` request/response formats


src/common/serverinfo.ts
```diff
@@ -61,18 +70,15 @@ function getServerInfo(connection: Connection): Promise<GetServerInfoResponse> {
...
-
-function getFee(connection: Connection, cushion: number): Promise<string> {
-  return getServerInfo(connection).then(serverInfo => {
-    return computeFeeFromServerInfo(cushion, serverInfo)
-  })
+async function getFee(this: RippleAPI, cushion?: number): Promise<string> {
+  if (cushion === undefined) {
+    cushion = this._feeCushion
+  }
+  if (cushion === undefined) {
```
intelliot: Any reason to do it this way rather than the previous code (`cushion = this._feeCushion || 1.2`)?

FredKSchott: yea, if `this._feeCushion` is set to `0`, it would get overwritten to `1.2`.

I'm unfamiliar with the concept though, so maybe that's not something anyone would do :) but it's good to be safe

intelliot: Transactions probably won't succeed with a feeCushion < 1.0 since the result may be less than the base fee. But it's still a good idea to allow developers do what they want - it's possible they have a reason for it.

PR #896, opened 2018-05-12
Update some API methods to use api.request() internally

FredKSchott:
- Use `any` instead of `object` (`object` is an empty object, while any is an untyped, flexible object)
- Convert `getTransaction()` to use `api.request()`
- Convert `submit()` to use `api.request()`
- Convert `prepareSequence()` to use `api.request()`


intelliot:
I thought that `{}` is an empty object, `Object` only has functionality that is common to all objects, and `object` is any thing that is not a primitive type.

Based on:
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#object-type
- https://blog.mariusschulz.com/2017/02/24/typescript-2-2-the-object-type

FredKSchott:
Yea, it's weird, I think any object can be an `object` type, but accessing properties off of an `object` types isn't allowed. You can see for yourself here: https://www.typescriptlang.org/play/#src=const%20bob%3A%20object%20%3D%20%7B%7D%3B%0D%0Abob.foo%20%3D%2012%3B

PR #897, opened 2018-05-17
Add more tests of getFee

intelliot:
A couple small unit tests that show how getFee uses the feeCushion


docs/src/getFee.md.ejs
```diff
@@ -4,13 +4,15 @@
 
 Returns the estimated transaction fee for the rippled server the RippleAPI instance is connected to.
 
+This will use the [feeCushion parameter](#parameters) provided to the RippleAPI constructor, or the default value of 1.2.
```
wilsonianb: Maybe change this to `1.2`.

PR #902, opened 2018-05-24
Maximum fee values

intelliot:
- Cap the value returned by getFee() at '1' XRP
- Default the maxFee to '2' XRP (can be overridden in transaction instructions)


src/common/serverinfo.ts
```diff
@@ -70,15 +61,31 @@ function getServerInfo(this: RippleAPI): Promise<GetServerInfoResponse> {
   })
 }
 
-async function getFee(this: RippleAPI, cushion?: number): Promise<string> {
+async function getFee(
+  this: RippleAPI,
+  cushion?: number,
+  maximum: number | BigNumber | null = 1
```
wilsonianb: `getFee`'s arguments (including the new default maximum of 1) are not documented
https://github.com/ripple/ripple-lib/blob/51197a1c6848283abca8b34ff4f9a4f0ffe1deaa/docs/index.md#getfee

src/transaction/utils.ts
```diff
@@ -63,25 +64,35 @@ function prepareTransaction(txJSON: any, api: RippleAPI,
...
+      if (fee.greaterThan('2') && instructions.allowHighFee !== true) {
+        const errorMessage = `Fee of ${fee.toString(10)} XRP exceeds ` +
+          'soft limit of 2 XRP. To use this fee, set ' +
+          '`instructions.allowHighFee = true`.'
+        throw new ValidationError(errorMessage)
+      }
       txJSON.Fee = scaleValue(common.xrpToDrops(instructions.fee), multiplier)
       return Promise.resolve(txJSON)
     }
     const cushion = api._feeCushion
```
wilsonianb: This isn't new, but we set `cushion` to `api._feeCushion` both here and inside `getFee`.

src/transaction/types.ts
```diff
@@ -56,7 +57,8 @@ export type KeyPair = {
 }
 
 export type SignOptions = {
-  signAs: string
+  signAs: string,
+  allowHighFee?: boolean
```
wilsonianb: I think `docgen` needs to be run again to include this.

intelliot:
To specify a maxFee higher than 2, should we require allowHighFee?

intelliot:
The justification for not requiring it is that the fee won't _always_ be high. You could have `maxFee: '3'` and only pay a 12 drop fee (if `load_factor` is 1).

- When preparing a transaction and having ripple-lib provide the `Fee`, `maxFee` applies. Most users can leave this alone: the default of `'2'` is appropriate. Users who want to save on fees can set `maxFee` to a lower value e.g. `'0.5'`.
- In the rare case that an advanced user wants to allow a higher fee, they can set `maxFee` to a higher value e.g. `'3'`.
- If `load_factor` is not escalating the fee appropriately, users can manually set `Fee`. However, to use a fee higher than 2 XRP, `allowHighFee` must be set. This prevents math errors and faulty fee escalation algorithms from unintentionally using an unreasonably high fee.
- `allowHighFee` has to be set in both `prepare*` and `sign` because some apps use ripple-lib to prepare, but then use a separate signing tool; and others generate raw `txJSON`, and use ripple-lib only to sign.

src/transaction/utils.ts
intelliot: I think this is ok.

src/common/serverinfo.ts
intelliot: Done

src/transaction/types.ts
intelliot: Done

src/transaction/utils.ts
```diff
@@ -63,25 +64,35 @@ function prepareTransaction(txJSON: any, api: RippleAPI,
...
-    return api.getFee(cushion).then(fee => {
+    return api.getFee(cushion, null).then(fee => {
```
wilsonianb: Do we need to add the `maximum` argument to `getFee` if we're not using it?

HISTORY.md
```diff
@@ -5,6 +5,18 @@
...
++ The value returned by `getFee()` is capped at 1 XRP. You can ignore this cap by setting `maximum` to `null`.
++ When using `prepare*` methods, the `maxFee` parameter now has a default value of `'2'`.
+
+#### Overview of fee limits
+
+Most users of ripple-lib do not need to make any code changes to accommodate the new soft limits on fees. These limits are designed to protect against the most severe cases where an unintentionally high fee may be used.
+
++ When having ripple-lib provide the fee with a `prepare*` method, a default `maxFee` of `'2'` applies. You can prepare more economical transactions by setting `maxFee` lower, or support high-priority transactions by setting `maxFee` higher.
++ To set a specific fee higher than 2 XRP, set `allowHighFee: true`. Such a high fee is not typical, but ripple-lib permits it with this override.
++ When using `sign` with a Fee higher than 2 XRP, a `ValidationError` is thrown unless `allowHighFee` is set.
```
wilsonianb: This (unlikely) scenario seems unintuitive:
1. call `prepare*` with `maxFee` > 2 XRP and `allowHighFee` not set
2. calculated fee is > 2 XRP (and possibly capped at `maxFee`)
3. call to `sign` using fee returned by `prepare*` fails without setting `allowHighFee`

intelliot: Good point. So when someone calls `prepare*` with `maxFee` > 2 XRP, we should require `allowHighFee`?

src/transaction/utils.ts
intelliot: We are using it. If the argument is undefined, it will default to `1`, but with `null`, no maximum is imposed.

wilsonianb: That should have said: Do we need `getFee` to enforce a maximum (which is configurable via the `maximum` argument`) if we're going to bypass it?

src/transaction/utils.ts
```diff
@@ -63,25 +64,35 @@ function prepareTransaction(txJSON: any, api: RippleAPI,
...
               new Buffer(txJSON.Fulfillment, 'hex').length / 16)))
         const feeDrops = common.xrpToDrops(fee)
-        if (instructions.maxFee !== undefined) {
-          const maxFeeDrops = common.xrpToDrops(instructions.maxFee)
-          const normalFee = scaleValue(feeDrops, multiplier, extraFee)
-          txJSON.Fee = BigNumber.min(normalFee, maxFeeDrops).toString()
-        } else {
-          txJSON.Fee = scaleValue(feeDrops, multiplier, extraFee)
+
+        let maxFee = instructions.maxFee
```
FredKSchott: `const maxFee = instructions.maxFee || '2'` would allow you to define this as a const.

src/transaction/utils.ts
intelliot: I think so. This is not the only place `getFee` is used; it's also a public API.

HISTORY.md
wilsonianb: What if `maxFee` were an optional RippleAPI constructor [parameter](https://github.com/ripple/ripple-lib/blob/develop/docs/index.md#parameters) instead of an optional [transaction instruction](https://github.com/ripple/ripple-lib/blob/develop/docs/index.md#transaction-instructions) (similar to `feeCushion`)?
It could default to 2 XRP, and no calculated or specified fee would be allowed to exceed it. (We wouldn't have `allowHighFee`.)

sublimator:
Haha ripple lib was like that in 2012/2013

HISTORY.md
intelliot: This is a good idea. I'll push a proposed change for this shortly.

src/common/schemas/objects/instructions.json
```diff
@@ -14,7 +14,7 @@
       "$ref": "value"
     },
     "maxFee": {
-      "description": "The maximum fee to pay for the transaction. See [Transaction Fees](#transaction-fees) for more information.",
+      "description": "The maximum fee to pay for this transaction. If this exceeds the `maxFeeXRP` set in the RippleAPI constructor, `maxFeeXRP` will be used instead. See [Transaction Fees](#transaction-fees) for more information.",
```
wilsonianb: What do you think about removing `maxFee`?

intelliot: I think there's a use case for some transactions being higher priority than others. For example, if an existing user has set `maxFee` to `0.5` to ensure more economical transactions, I don't want to break their integration.

intelliot: After a quick chat, we've decided to deprecate `maxFee`. It will be removed entirely in a future release of ripple-lib (perhaps 2.0.0).

src/api.ts
```diff
@@ -105,6 +107,7 @@ class RippleAPI extends EventEmitter {
     super()
     validate.apiOptions(options)
     this._feeCushion = options.feeCushion || 1.2
```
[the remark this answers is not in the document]
intelliot: Not a bad idea, but I think people are paying a lot more due to `load_factor` increases. I don't have evidence of this, though.

A few drops doesn't make much difference in comparison, and it will ensure that transactions have some priority over "super economy" txs paying just the base fee.

PR #912, opened 2018-06-26
Round XRP fee to 6 decimal places

### LINKED ISSUE #911: fee format when using maxFee
Not sure if this is actually a bug as the field `maxFee` in the Transaction Instructions is now deprecated.
But it is surely a backwards compatibility issue.

I have seen that if you specify a `maxFee` value and don't make use of the `maxFeeXRP` field in the constructor (like in the old days) you get random validation errors as the calculated fee is not properly formatted:

`[ValidationError(xrpToDrops: value '0.0064895625' has too many decimal places.)]`

intelliot:
Fix #911

### DIFF
--- a/test/api-test.js
+++ b/test/api-test.js
@@ -1826,10 +1826,6 @@ describe('RippleAPI', function () {
       data: { loadFactor: 5407.96875 }
     }));
 
-    const localInstructions = _.defaults({
-      maxFee: '4'
-    }, instructions);
-
     const expectedResponse = {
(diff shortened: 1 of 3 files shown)


test/api-test.js
```diff
@@ -1820,6 +1820,30 @@ describe('RippleAPI', function () {
...
 
+  it('fee - calculated fee does not use more than 6 decimal places', function () {
+    this.api.connection._send(JSON.stringify({
+      command: 'config',
+      data: { loadFactor: 5407.96875 }
+    }));
+
+    const localInstructions = _.defaults({
+      maxFee: '4'
+    }, instructions);
```
wilsonianb: I think we could just use `instructions`

PR #922, opened 2018-07-27
Export createPaymentTransaction

soichisumi:
I want to create signed transactions locally by using ripple-lib.
To create signed transaction locally, it is needed to give txJson and some instructions to preparePayment() (discussed at https://github.com/ripple/ripple-lib/issues/905).

It is useful to create txJSON if createPaymentTransaction is exported.
This PR to export createPaymentTransaction and add a test for createPaymentTransaction.

If the course of an implementation of this PR, I'll add more test to createPaymentTransaction

Any comments are welcomed.


yoyoyousei:
this is a PR for https://github.com/ripple/ripple-lib/issues/913

intelliot:
I think you can use preparePayment. Can you show an example where you need to access createPaymentTransaction directly?

src/api.ts
```diff
@@ -28,6 +28,7 @@ import getAccountInfo from './ledger/accountinfo'
 import getAccountObjects from './ledger/accountobjects'
 import getPaymentChannel from './ledger/payment-channel'
 import preparePayment from './transaction/payment'
+import createPaymentTransaction from './transaction/payment'
```
intelliot: This uses the default export. You'll need to use something like `{createPaymentTransaction}`

test/api-test.js
```diff
@@ -435,13 +435,19 @@ describe('RippleAPI', function () {
...
-
+      
       return this.api.preparePayment(
         address, requests.preparePayment.normal, localInstructions).then(
           _.partial(checkResult, expectedResponse, 'prepare'));
     });
   });
 
+  it('createPaymentTransaction - normal', function(){
+    return this.api.createPaymentTransaction(address, requests.preparePayment.normal).then((res)=>{
```
intelliot: `createPaymentTransaction` should not return a promise, so `then` should not be used here.

test/api-test.js
```diff
@@ -435,13 +435,19 @@ describe('RippleAPI', function () {
...
+      assert.deepStrictEqual(res.txJSON,  "{\"TransactionType\":\"Payment\",\"Account\":\"r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59\",\"Destination\":\"rpZc4mVfWUif9CRoHRKKcmhu1nx2xktxBo\",\"Amount\":{\"currency\":\"USD\",\"issuer\":\"rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM\",\"value\":\"0.01\"},\"Flags\":2147483648,\"SendMax\":{\"currency\":\"USD\",\"issuer\":\"rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM\",\"value\":\"0.01\"},\"LastLedgerSequence\":8819954,\"Sequence\":23,\"Fee\":\"12\"}")
```
intelliot: `deepStrictEqual` doesn't exist. Also, `createPaymentTransaction` returns a txJSON object, not an object containing a string called `txJSON`. So you'll need to do something like:
```
  it('createPaymentTransaction - normal', function(){
    const res = this.api.createPaymentTransaction(address, requests.preparePayment.normal)
    assert.deepEqual(res, JSON.parse("{\"TransactionType\":\"Payment\",\"Account\":\"r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59\",\"Destination\":\"rpZc4mVfWUif9CRoHRKKcmhu1nx2xktxBo\",\"Amount\":{\"currency\":\"USD\",\"issuer\":\"rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM\",\"value\":\"0.01\"},\"Flags\":2147483648,\"SendMax\":{\"currency\":\"USD\",\"issuer\":\"rMH4UxPrbuMa1spCBR98hLLyNJp4d8p4tM\",\"value\":\"0.01\"},\"LastLedgerSequence\":8819954,\"Sequence\":23,\"Fee\":\"12\"}"))
  })
```
Notice that `createPaymentTransaction` doesn't set `LastLedgerSequence`, `Sequence`, or `Fee`.

intelliot:
@yoyoyousei Any thoughts on @wilsonianb 's comment above?

I think this still might be useful if we document clearly that fee and sequence will need to be determined out-of-band, and set manually.

yoyoyousei:
Thank you for reviewing!
Now I have time, I'll read your review comments

yoyoyousei:
It is understandable returning txJson which is not guaranteed that fee and sequence are contained is not good (because developers may publish invalid txJson).

But, I think this change is useful for offline usage.
And changing createPaymentTransaction to return valid txJson in offline is difficult because providing information about the fee, sequence, etc requires asynchronous process(retrieving current sequence or fee from rippled).

So, I want to add createPaymentTransaction and its documentation. @wilsonianb @intelliot

yoyoyousei:
Or, as another implementation,
adding another function which executes createPaymentTransaction and sets sequence and fee given by argument, looks good for @wilsonianb ?

Ex: 
```
function createPaymentTransactionOffline(address, payment, instructions){
   if(instructions.fee || instructions.maxLedgerVersion){ // todo: validate precisely
      return {txJson: "", message:"error"};
   }
  const txJSON = createPaymentTransaction(address, payment)
  txJSON.LastLedgerSequence = instructions.maxLedgerVersion;
  txJSON.Fee = instructions.fee
  return {txJson: txJSON, message: "ok"}
}

```

yoyoyousei:
In the meantime, I fixed commented issues.

yoyoyousei:
Do `createPaymentTransaction` have to return complete txJson?

intelliot:
@soichisumi I think it would be good for it to return complete txJSON, yes. Can you bring it up-to-date with the latest changes on `develop`?

intelliot:
Essentially, this PR aims to provide a synchronous variant of `preparePayment`.

> I'm using ripple-lib in iOS app and JavascriptCore is used to call javascript functions.
> But as long as I examined, JavascriptCore seems not to handle promise-wrapped values.

https://github.com/ripple/ripple-lib/issues/913#issuecomment-408579436

FKSRipple:
@soichisumi are you still interested in seeing this merged? If so, I'll review and then can help with any changes.

FKSRipple:
@soichisumi closing the PR for lack of activity, but don't hesitate to comment if you're still interested in getting this merged and I can re-open.

Issue #915, opened 2018-07-02
Export schemaValidator as part of API and isValidAddress as part of schemaValidator


movitto:
Greetings! Related to #770, we're looking for a quick way to validate addresses & secrets. ripple-lib has this functionality built in via the mechanisms in $subject, but they are not exposed publically.

Yes ripple-keypairs has this functionality (and ripple-address-codec?) but those projects do not incorporate a 'gulpfile' so that they can be bundled for browser access! Having less experience with nodejs tools than in other languages, it was just easier for me to modify the exports of ripple-lib to expose the aforementioned methods than to figure out everything that's needed to build ripple-keypairs into a single package.

I can send a PR exposing the methods if desired, else please advise how to access this functionality from a browser environment, it'd be great to be using stock ripple libs w/out mods!!!

Thanks for the great library and all of ripple's efforts!

movitto:
In case anyone was wondering, to expose schemaValidator, I simply modified src/api.ts to remove 'schemaValidator' from the PRIVATE map:

https://github.com/ripple/ripple-lib/blob/develop/src/api.ts#L103

```
static _PRIVATE = {
    validate: validate,
    RangeSet,
    ledgerUtils
}
```

And added 'schemaValidator' to the 'RippleAPI' class properties below:

https://github.com/ripple/ripple-lib/blob/develop/src/api.ts#L308

`schemaValidator = schemaValidator`

---

Simply build according to the build instructions on the ripple website (**npm run build**) and you're good to go!

intelliot:
Would you like to create a PR?

movitto:
PR sent, thanks for taking the time to consider this.

wilsonianb:
As a follow up, should the exposed methods be added to the docs?

movitto:
@wilsonianb PR #927 sent

intelliot:
This is done. Docs: https://xrpl.org/rippleapi-reference.html#schemavalidator

PR #923, opened 2018-07-27
Expose validation methods in public api

movitto:
This comment addresses issue #915 so as to expose the schema validator object and various validation methods in it via the public API. This patch contains no functional changes, it simple removes the target logic from the resticted 'private' methods and adds them to the public api registry. Please advise if there is anything else that needs TBD to see this in. Thank you.


src/common/schema-validator.ts
```diff
@@ -4,6 +4,7 @@ const {Validator} = require('jsonschema')
 import {ValidationError} from './errors'
 import {isValidAddress} from 'ripple-address-codec'
 import {isValidSecret} from './utils'
```
intelliot: Due to line 7, I think you'll need to remove this line.

Currently, the build fails with:
```
src/common/schema-validator.ts(6,9): error TS2300: Duplicate identifier 'isValidSecret'.
```

movitto:
@intelliot doh! Missed that one. It's now fixed. Thanks for the review

src/api.ts
```diff
@@ -109,8 +109,7 @@ class RippleAPI extends EventEmitter {
   static _PRIVATE = {
     validate: validate,
     RangeSet,
-    ledgerUtils,
-    schemaValidator
```
intelliot: Since `schemaValidator` is no longer under `_PRIVATE`, you'll need to update [the reference in api-test.js](https://github.com/ripple/ripple-lib/blob/28796d37cb9af2c62d41f1f171aac82eafb480db/test/api-test.js#L16) in order for the unit tests to pass.

src/api.ts
```diff
@@ -329,6 +328,7 @@ class RippleAPI extends EventEmitter {
   dropsToXrp = dropsToXrp
   iso8601ToRippleTime = iso8601ToRippleTime
   txFlags = txFlags
+  schemaValidator = schemaValidator
```
intelliot: Since this doesn't rely on any options or connect to a rippled server, I would suggest making this `static` so that it can be accessed as `RippleAPI.schemaValidator`.
```
static schemaValidator = schemaValidator
```

src/api.ts
intelliot: Probably good to update the reference in broadcast-api-test.js too.

movitto:
@intelliot the patch was updated  to fix the tests and to mark schemaValidator as static. I ran the tests locally post-commit to verify the results.

intelliot:
I'm going to merge this since there have been no objections in over a week. If anyone spots any problems, please open a new issue or PR!

PR #924, opened 2018-08-06
Prevent 'amount' from being misinterpreted

intelliot:
The `amount` field should almost never be used.

With partial payments, the field can show an amount that is significantly more than the amount that the transaction actually delivered. This change removes the amount field from parsed payment transactions returned by `getTransaction()` and `getTransactions()`.

There's a new option (`includeRawTransaction` and `includeRawTransactions`) for getting the full raw transaction data. Also, advanced users can always use `request()` to get unaltered responses from rippled.

Per the [Partial Payments](https://developers.ripple.com/partial-payments.html) page, if delivered_amount is unavailable*, then you can only determine the actual delivered amount by reading the AffectedNodes in the transaction's metadata; so the `amount` field is still not useful in this case.

\* This only applies to partial payments included in a validated ledger before 2014-01-20


intelliot:
Great feedback.

> At the very least, I do think the amount from the original transaction can be useful in some cases

I haven't seen a real case where `Amount` was needed to debug. Also, auditing (verifying signatures?) is not needed for most use cases. However, for those rare situations, I'll add an option for the user to get the raw transaction data as well. Furthermore, users will still get the raw transaction data when using `request` with `tx`, `account_tx`, etc.

> Also, the description of the amount field here is for the "prepare" helper, which describes constructing a transaction before submitting it. You should maybe add a note to describe the field in the response.

It's for the object itself, so the description appears for both the request and the response, even though it's only relevant for the response. But I've decided to clean things up a bit by just omitting `amount`. It's a breaking change, but that'll be OK if we include this change in 1.0.0 and call it out in the release notes.

> I'm not actually clear how ripple-lib translates the `destination.minAmount` and `destination.amount` fields into a Payment transaction when it says they're mutually exclusive, but in the actual API you always need an `amount`.

Here's how the logic looks.

When using `destination.minAmount`:
1. If it's an XRP-to-XRP payment, then `Amount` is `destination.minAmount`.
2. Otherwise: if the destination currency is XRP, `Amount` is `100000000000`; if the destination currency is an IOU, `Amount` is `9999999999999999e80`*. The transaction's `DeliverMin` is `destination.minAmount`. Also, the partial payment flag is enabled.

* These are the maximum amounts. This ensures that all of the source funds are sent. The `source.maxAmount` or `source.amount` are set as the transaction's `SendMax`.

When using `destination.amount`:
1. In all cases, `Amount` is `destination.amount`.

intelliot:
I think we can make this change in ripple-lib now, regardless of what changes are later made in the rippled API.

src/ledger/parse/payment.ts
```diff
@@ -27,12 +28,22 @@ function parsePayment(tx: any): Object {
...
+    amountInfo?: string,
+    tag: number | undefined
+  } = {
     address: tx.Destination,
-    amount: removeGenericCounterparty(parseAmount(tx.Amount), tx.Destination),
     tag: tx.DestinationTag
   }
 
+  // Omit 'amount' for partial payments to prevent misuse
+  if (tx.Amount && !utils.isPartialPayment(tx)) {
```
wilsonianb: What do you think about including the partial payment amount if the full amount was actually delivered?

test/fixtures/responses/get-ledger-pre2014-with-partial.json
```diff
@@ -25,12 +25,7 @@
           }
         },
         "destination": {
-          "address": "rNNuQMuExCiEjeZ4h9JJnj5PSWypdMXDj4",
-          "amount": {
-            "currency": "USD",
-            "value": "10",
-            "counterparty": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
-          }
```
wilsonianb: I don't like how we lose the destination currency/counterparty.
What do you think about keeping the amount and either omitting or renaming `"value"`?

docs/index.md
```diff
@@ -1960,7 +1963,7 @@ source | object | Properties of the source of the payment.
 *source.* maxAmount | [laxAmount](#amount) | The maximum amount to send. (This field is exclusive with source.amount)
 destination | object | Properties of the destination of the payment.
 *destination.* address | [address](#address) | The address to receive at.
-*destination.* amount | [laxAmount](#amount) | An exact amount to deliver to the recipient. If the counterparty is not specified, amounts with any counterparty may be used. (This field is exclusive with destination.minAmount).
+*destination.* amount | [laxAmount](#amount) | *Optional* An exact amount to deliver to the recipient. If the counterparty is not specified, amounts with any counterparty may be used. (This field is exclusive with destination.minAmount.) When parsing a partial payment, this field may be hidden; for the amount that the transaction delivered, see 'deliveredAmount'.
```
wilsonianb: I understand why the `When parsing a partial payment...` is showing up here, but I think it's a little confusing since it's not relevant to `getPaths`

test/fixtures/responses/get-ledger-pre2014-with-partial.json
intelliot: I think the destination currency/counterparty can still be seen in deliveredAmount, or if that's not present, then under balanceChanges. I'll test this out.

intelliot: I've also been considering just renaming amount to provisionalAmount or requestedAmount, but then there's still the risk that someone would misinterpret it.

src/ledger/parse/payment.ts
intelliot: I would prefer that users instead look at deliveredAmount, always.

I've been considering going the route of actually omitting `amount` in *all* cases... 🤔

docs/index.md
intelliot: Yeah, I might need to create separate schemas for a transaction request vs. a transaction response.

test/fixtures/responses/get-ledger-pre2014-with-partial.json
mDuo13: I mean, really, the problem is that people tend to assume that "success" means "it did what you think it did".

You'd think that people would learn that's now how computers work, but it keeps happening.

mDuo13:
Oh, to clarify earlier, when I said auditing I wasn't not talking strictly about verifying signatures. The scenario I had in mind was more like, "Some suspicious behavior occurred or maybe even a theft happened, and we're trying to figure out exactly who moved what money when."

For example, if you're investigating how someone exploited a partial payment weakness in some integration logic that wasn't written with this version of ripple-lib, it'll be much more helpful if you know the transaction's amount field.

Just having an option to request the specific amount covers that scenario, though, so :+1: on that.

test/fixtures/responses/get-ledger-pre2014-with-partial.json
intelliot: Good point. I guess we can say that the transaction was applied to the ledger, but you have to parse the outcome to find out what it really did.

intelliot:
This PR has been updated to include a new json schema (getPayment) for payment specifications that do not include `destination.amount`. It's now ready for review.

src/ledger/parse/payment.ts
```diff
@@ -28,10 +29,15 @@ function parsePayment(tx: any): Object {
     tag: tx.SourceTag
   }
 
-  const destination = {
+  const destination: {
+    address: string,
+    amount?: Amount,
+    amountInfo?: string,
```
wilsonianb: What is `amountInfo`?

test/fixtures/responses/get-transactions-include-raw-transactions.json
```diff
@@ -0,0 +1,190 @@
...
...D89AC1F737AE3B6BB5D976702B171CC7B4DE5CA3D444D5B8D6B4\",\"LedgerIndex\":\"F3E119AAA87AF3607CF87F5523BB8278A83BCB4142833288305D767DD30C392A\",\"PreviousFields\":{\"Balance\":\"912694201420\"},\"FinalFields\":{\"Flags\":0,\"Sequence\":59,\"OwnerCount\":10,\"Balance\":\"912695302618\",\"Account\":\"r9tGqzZgKxVFvzKFdUqXAqTzazWBUia8Qr\"}}},{\"ModifiedNode\":{\"LedgerEntryType\":\"RippleState\",\"PreviousTxnLgrSeq\":348433,\"PreviousTxnID\":\"8F571C346688D89AC1F737AE3B6BB5D976702B171CC7B4DE5CA3D444D5B8D6B4\",\"LedgerIndex\":\"FA1255C2E0407F1945BCF9351257C7C5C28B0F5F09BB81C08D35A03E9F0136BC\",\"PreviousFields\":{\"Balance\":{\"value\":\"-5.5551658883365\",\"currency\":\"USD\",\"issuer\":\"rrrrrrrrrrrrrrrrrrrrBZbvji\"}},\"FinalFields\":{\"Flags\":131072,\"LowNode\":\"000000000000000C\",\"HighNode\":\"0000000000000000\",\"Balance\":{\"value\":\"-5.5541638883365\",\"currency\":\"USD\",\"issuer\":\"rrrrrrrrrrrrrrrrrrrrBZbvji\"},\"LowLimit\":{\"value\":\"0\",\"currency\":\"USD\",\"issuer\":\"rpZc4mVfWUif9CRoHRKKcmhu1nx2xktxBo\"},\"HighLimit\":{\"value\":\"1000\",\"currency\":\"USD\",\"issuer\":\"r9tGqzZgKxVFvzKFdUqXAqTzazWBUia8Qr\"}}}}],\"TransactionResult\":\"tesSUCCESS\"},\"validated\":true}"
```
wilsonianb: Including the `rawTransaction` in each parsed transaction contrasts with [getLedger](https://github.com/ripple/ripple-lib/blob/develop/docs/index.md#getledger), which groups all of the `rawTransactions` separate from the parsed transactions.

test/fixtures/responses/get-ledger-with-partial-payment.json
```diff
@@ -94,12 +89,7 @@
...
-          "address": "rNNuQMuExCiEjeZ4h9JJnj5PSWypdMXDj4",
-          "amount": {
-            "currency": "USD",
-            "value": "10",
-            "counterparty": "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
-          }
+          "address": "rNNuQMuExCiEjeZ4h9JJnj5PSWypdMXDj4"
         },
         "allowPartialPayment": true
       },
```
wilsonianb: I'm wondering why this transaction doesn't have a `deliveredAmount`

src/ledger/pathfind.ts
```diff
@@ -23,7 +23,7 @@ function addParams(request: PathFindRequest, result: RippledPathsResponse
 function requestPathFind(connection: Connection, pathfind: PathFind
 ): Promise<RippledPathsResponse> {
   const destinationAmount: Amount = _.assign(
-    {value: '-1'},
+    {value: dropsToXrp('-1')},
```
wilsonianb: What do you think about moving this change to https://github.com/ripple/ripple-lib/pull/930?

src/ledger/parse/payment.ts
intelliot: Leftovers - good catch! Removing

src/ledger/pathfind.ts
intelliot: Good call

test/fixtures/responses/get-transactions-include-raw-transactions.json
intelliot: Since `getTransactions` returns an array, we can't easily add a top-level property with all of the rawTransactions. I think this is ok. If it's a big problem, we could remove this entirely; then the way to get the data would be `request('account_tx', ...)` instead.

test/fixtures/responses/get-ledger-with-partial-payment.json
intelliot: Fixed

src/ledger/parse/transaction.ts
```diff
@@ -42,10 +42,11 @@ function parseTransactionType(type) {
   return mapping[type] || null
 }
 
-function parseTransaction(tx: any): any {
+function parseTransaction(tx: any, includeRawTransaction: boolean): any {
   const type = parseTransactionType(tx.TransactionType)
+  const _parsePayment = payment => parsePayment(payment, includeRawTransaction)
```
wilsonianb: It looks like `includeRawTransaction` only happens for payments.

intelliot: I think only payments have the confusing `amount` field, but it's probably not a bad idea to include the raw transaction for all transaction types.

docs/src/rippledAPIs.md.ejs
```diff
@@ -5,7 +5,7 @@ ripple-lib relies on [rippled APIs](https://ripple.com/build/rippled-apis/) for
 * Use `hasNextPage()` to determine whether a response has more pages. This is true when the response includes a [`marker` field](https://ripple.com/build/rippled-apis/#markers-and-pagination).
 * Use `requestNextPage()` to request the next page of data.
 
-When using rippled APIs, [specify XRP amounts in drops](https://ripple.com/build/rippled-apis/#specifying-currency-amounts). 1 XRP = 1000000 drops.
+When using rippled APIs, [specify XRP amounts in drops and timestamps as the number of seconds since the "Ripple Epoch"](https://developers.ripple.com/basic-data-types.html#specifying-currency-amounts).
```
wilsonianb: Consider linking the timestamps portion to https://developers.ripple.com/basic-data-types.html#specifying-time

test/fixtures/responses/get-transactions-include-raw-transactions.json
wilsonianb: Another option could be to change getLedger to include the `rawTransaction` in each parsed transaction (call `parseTransaction(transaction, true)`) instead of grouping them in the `rawTransactions` field

docs/index.md
```diff
@@ -1153,7 +1154,7 @@ id | [transactionHash](#transaction-id) | A hash of the transaction that can be
 address | [address](#address) | The address of the account that initiated the transaction.
 sequence | [sequence](#account-sequence-number) | The account sequence number of the transaction for the account that initiated it.
 type | [transactionType](#transaction-types) | The type of the transaction.
-specification | object | A specification that would produce the same outcome as this transaction. The structure of the specification depends on the value of the `type` field (see [Transaction Types](#transaction-types) for details). *Note:* This is **not** necessarily the same as the original specification.
+specification | object | A specification that would produce the same outcome as this transaction, but for payment transactions, without the `destination.amount`. The structure of the specification depends on the value of the `type` field (see [Transaction Types](#transaction-types) for details). *Note:* This is **not** necessarily the same as the original specification.
```
mDuo13: The rephrased sentence is a little tough to understand. Here's my suggested change:

> A specification that would produce the same outcome as this transaction. *Exception:* For payment transactions, this omits the `destination.amount` field, to prevent misunderstanding.

docs/index.md
```diff
@@ -1962,7 +1953,7 @@ source | object | Properties of the source of the payment.
 *source.* maxAmount | [laxAmount](#amount) | The maximum amount to send. (This field is exclusive with source.amount)
 destination | object | Properties of the destination of the payment.
 *destination.* address | [address](#address) | The address to receive at.
-*destination.* amount | [laxAmount](#amount) | An exact amount to deliver to the recipient. If the counterparty is not specified, amounts with any counterparty may be used. (This field is exclusive with destination.minAmount).
+*destination.* amount | [laxAmount](#amount) | An exact amount to deliver to the recipient. If the counterparty is not specified, amounts with any counterparty may be used. (This field is exclusive with destination.minAmount.)
```
mDuo13: Nit (slightly unrelated, but it's in the same line): Instead of saying "This field is exclusive" say "This field cannot be used with"

test/fixtures/responses/get-transactions-include-raw-transactions.json
intelliot: That's a good idea for consistency, although I'm not sure that it's worth the trouble of breaking existing users of getLedger. I'll open a separate PR for it though.

intelliot: I hit a snag here: the existing getLedger parser renames metaData to meta.

https://github.com/ripple/ripple-lib/commit/aeb56dee17aef1e43d61a7d6b907d3082060e598

I don't know why. @shekenahglory ?

wilsonianb: :thinking: here's the pr https://github.com/ripple/ripple-lib/pull/604

intelliot: Ah. Is that an inconsistency that could be fixed in the rippled API?

PR #926, opened 2018-08-08
Document xrpToDrops, dropsToXrp, iso8601ToRippleTime, schemaValidator…

intelliot:
- [xrpToDrops](#xrptodrops)
  - [dropsToXrp](#dropstoxrp)
  - [iso8601ToRippleTime](#iso8601torippletime)
  - [txFlags](#txflags)
  - [schemaValidator](#schemavalidator)
  - [schemaValidate](#schemavalidate)
  - [isValidAddress](#isvalidaddress)
  - [isValidSecret](#isvalidsecret)
  - [deriveKeypair](#derivekeypair)
  - [deriveAddress](#deriveaddress)

These docs are far from perfect, but this should help to clarify how these methods can be used.

PR #927, opened 2018-08-10
Add API docs for exposed deriveAddress and deriveKeypair methods

movitto:


movitto:
Apologies, I didn't see #926 before submitting this. Closing

intelliot:
@movitto Thanks. What do you think of moving deriveAddress and deriveKeypair off of the schemaValidator object? Since they are not schema validators, it doesn't seem like the right place for them. I reverted the original merge (see https://github.com/ripple/ripple-lib/commit/f5e1a4a588e555db600afa0905ef827c75194ff6). What do you think of adding it back but putting the methods directly on the RippleAPI object?

movitto:
@intelliot no worries. Yes that seems reasonable, I can look into implementing this sometime next week

PR #930, opened 2018-08-15
Fix getPaths

intelliot:
When pathfinding with a source amount, rippled can return paths where the source amount in the returned path is not equal to the requested source amount. These are known as "insufficient source funds paths."

ripple-lib removes these paths from the response, but the filter did not handle the case where the amount is specified in drops, rather than an IOU with a `value` field.

The pathfinding request asks to find paths that deliver as much as possible of the destination amount by spending up to X source amount. In the returned path, if the source amount is less than X, that means there is not enough liquidity to spend it all. This means a payment that was meant to convert the entire amount of the `send max` would fail. It may be useful to know how much of the `send max` would be spent, so this may be something to revisit in the future. Of course, advanced users can already use `request()` with `ripple_path_find` to get the raw results.

### DIFF
--- a/src/ledger/pathfind.ts
+++ b/src/ledger/pathfind.ts
@@ -1,7 +1,13 @@
 import * as _ from 'lodash'
 import BigNumber from 'bignumber.js'
 import {getXRPBalance, renameCounterpartyToIssuer} from './utils'
-import {validate, toRippledAmount, errors, xrpToDrops} from '../common'
+import {
+  validate,
+  toRippledAmount,
+  errors,
+  xrpToDrops,
+  dropsToXrp
+} from '../common'
 import {Connection} from '../common'
 import parsePathfind from './parse/pathfind'
 import {RippledAmount, Amount} from '../common/types/objects'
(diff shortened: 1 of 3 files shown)


test/api-test.js
```diff
@@ -2292,6 +2292,45 @@ describe('RippleAPI', function () {
...
+              value: this.api.dropsToXrp(1000000),
+              currency: 'XRP'
+          }
+      },
+      destination: {
+          address: 'rhpJkBfZGQyT1xeDbwtKEuSrSXw3QZSAy5',
+          amount: {
+              counterparty: 'rGpGaj4sxEZGenW1prqER25EUi7x4fqK9u',
+              currency: 'EUR'
+          }
```
wilsonianb: nit: tab size

src/ledger/pathfind.ts
```diff
@@ -95,13 +95,25 @@ function filterSourceFundsLowPaths(pathfind: PathFind,
...
+      } else {
+        // No alt.source_amount
+        return false
+      }
+      if (pathfind.source.amount.currency === 'XRP') {
+        pathfindSourceAmountValue =
+          new BigNumber(xrpToDrops(pathfind.source.amount.value))
+      } else {
+        pathfindSourceAmountValue = new BigNumber(pathfind.source.amount.value)
+      }
```
wilsonianb: This could be changed to
```
const pathfindSourceAmountValue = new BigNumber(
  pathfind.source.amount.currency === 'XRP' ?
  xrpToDrops(pathfind.source.amount.value) :
  pathfind.source.amount.value)
```

PR #931, opened 2018-08-15
Rename json schemas

intelliot:
With the exception of these 5, all of the other schemas have filenames that match their titles (aside from transforming camelCase titles to snake-case filenames).

By making these names consistent, we can more easily parse and dereference the schemas programmatically.


intelliot:
Merging since there have been no comments in 8 days

PR #932, opened 2018-08-16
Expose validation methods

movitto:
This PR exposes the deriveKeypair and deriveAddress methods to the public ripple-lib API. The schemaValidator.isValidAddress method is also exposed.

Included are tests and docs pertaining to the new exposed methods


movitto:
@intelliot @wilsonianb PR incorporates feedback, verified locally. Any chance for review?

docs/index.md
```diff
@@ -5105,6 +5107,48 @@ return api.generateAddress();
...
+
+Derive a public and private key from a seed.
+
+### Parameters
+
+This method takes one parameter, the seed from which to derive the public and private key.
+
+### Return Value
+
+The method return an object containing the public and private components of the keypair corresponding to the seed.
```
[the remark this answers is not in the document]
intelliot: Note: these changes should be made in the .ejs file. Use yarn docgen to generate index.md

src/common/utils.ts
```diff
@@ -1,6 +1,6 @@
 import * as _ from 'lodash'
 import BigNumber from 'bignumber.js'
-import {deriveKeypair} from 'ripple-keypairs'
+import {deriveKeypair,deriveAddress} from 'ripple-keypairs'
```
intelliot: add space after `,`

movitto:
@intelliot incorporated feedback

docs/src/deriveAddress.md.ejs
```diff
@@ -0,0 +1,19 @@
+## deriveAddress
+
+`deriveAddress(publicKey: string) -> string
```
intelliot: Notation: Use `: ` instead of ` -> ` before the return value type

docs/src/deriveAddress.md.ejs
```diff
@@ -0,0 +1,19 @@
...
+
+Derive a XRP Ledger address from a public key.
```
intelliot: Nit: `a` -> `an`

docs/src/deriveAddress.md.ejs
```diff
@@ -0,0 +1,19 @@
...
+
+### Parameters
+
+This method takes one parameter, the public key from which to derive the address.
+
+### Return Value
+
+The method returns a string corresponding to the address derived from the public key.
```
intelliot: Nit: `The` -> `This` (for consistency with the prior sentence)

docs/src/deriveKeypair.md.ejs
```diff
@@ -0,0 +1,21 @@
+## deriveKeypair
+
+`deriveKeypair(seed: string) -> {privateKey: string, publicKey: string}`
```
intelliot: Same as above

docs/src/deriveKeypair.md.ejs
```diff
@@ -0,0 +1,21 @@
...
+
+Derive a public and private key from a seed.
+
+### Parameters
+
+This method takes one parameter, the seed from which to derive the public and private key.
+
+### Return Value
+
+The method returns an object containing the public and private components of the keypair corresponding to the seed.
```
intelliot: Same as above

test/api-test.js
```diff
@@ -270,6 +270,34 @@ describe('RippleAPI', function () {
       }, /dropsToXrp:\ invalid\ value\ '\.\.\.'\,\ should\ be\ a\ number\ matching\ \(\^\-\?\[0\-9\]\*\.\?\[0\-9\]\*\$\)\./)
     })
   })
+
+  describe('deriveKeypair', function () {
+    it('returns keypair for secret', function () {
+      var keypair = this.api.deriveKeypair("snsakdSrZSLkYpCXxfRkS4Sh96PMK");
+      assert(keypair.privateKey == "008850736302221AFD59FF9CA1A29D4975F491D726249302EE48A3078A8934D335");
+      assert(keypair.publicKey == "035332FBA71D705BD5D97014A833BE2BBB25BEFCD3506198E14AFEA241B98C2D06");
```
intelliot: Use assert.equal so that failures are reported better:
```
      assert.equal(keypair.privateKey, '008850736302221AFD59FF9CA1A29D4975F491D726249302EE48A3078A8934D335');
      assert.equal(keypair.publicKey, '035332FBA71D705BD5D97014A833BE2BBB25BEFCD3506198E14AFEA241B98C2D06');
```

test/api-test.js
```diff
@@ -270,6 +270,34 @@ describe('RippleAPI', function () {
...
+    })
+
+    it("throws with an invalid secret", function (){
+      assert.throws(() => {
+        this.api.deriveKeypair("...");
+      }, "Non-base58 character")
```
intelliot: You must use a regex here for the comparison to actually work. If you change this to `"Non-base59 character"`, the test still passes, despite the `9`.

Here's a regex for this one:
```
/^Error\: Non\-base58 character$/
```

See https://nodejs.org/api/assert.html#assert_assert_throws_block_error_message

> Due to the confusing notation, it is recommended not to use a string as the second argument. This might lead to difficult-to-spot errors.

test/api-test.js
```diff
@@ -270,6 +270,34 @@ describe('RippleAPI', function () {
...
+    })
+  })
+
+  describe('deriveAddress', function () {
+    it('returns address for public key', function () {
+      var address = this.api.deriveAddress("035332FBA71D705BD5D97014A833BE2BBB25BEFCD3506198E14AFEA241B98C2D06");
+      assert(address == "rLczgQHxPhWtjkaQqn3Q6UM8AbRbbRvs5K");
```
intelliot: Same as above

movitto:
@intelliot incorporated all feedback except to the **a** to **an** change. **a** is used when the following word begins with a constant whereas **an** is used when the following word begins with a vowel. In that sentence the following words are **XRP** and **public**, both of which begin with constants.

Let me know if there is anything else. Thanks.

intelliot:
The X in XRP is usually pronounced "ex" which starts with a vowel sound. 

It’s not a big deal to me though.

movitto:
@intelliot gotchya, grammer was never really my subject :laughing: 

Updated.

movitto:
> A few things:
> -This doesn't appear to expose isValidAddress

Must've missed this one, we'll expose this and isValidSecret as part of a reworked patchset once we iron out the details below.

> -Assuming #830 also gets added, would we want to have two versions of deriveAddress (one that takes a public key, and one that takes a seed/secret)?

Hrm... any suggestions as to rectifying this? With JS the version that is defined second will override the first. Perhaps we can define these as deriveAddressFromKey and deriveAddressFromSecret? We can then add the 'deriveAddress' method as a wrapper which detects is if the user specified a secret or key (using isValidSecret, etc) and then dispatching to the correct corresponding method.


> -It seems like deriveKeypair and deriveAddress should be listed under https://github.com/ripple/ripple-lib/blob/develop/docs/index.md#offline-functionality
> which begs the question: does the implementation belong in src/offline/?

We're easy with whatever (conceptually, haven't looked into what it would take to make this happen). If this is the preferred approach, we can look into moving this there when we rework the PR.

Just let us know about the naming above and we'll start updating this 

Thanks.

intelliot:
> Perhaps we can define these as deriveAddressFromKey and deriveAddressFromSecret? We can then add the 'deriveAddress' method as a wrapper which detects is if the user specified a secret or key (using isValidSecret, etc) and then dispatching to the correct corresponding method.

I'm OK with this, but I also think there may be less opportunity for confusion if there is just "one" way to do things. It's also a little simpler. In this case, that would mean that the way to derive an address from a secret is:

1. Derive the public key from the secret.
2. Derive the address from the public key.

Example usage:
```
const address = api.deriveAddress(api.deriveKeypair(secret).publicKey)
```

(And we would scrap #830)



> does the implementation belong in src/offline/?

Yes, I think it should be in `src/offline/`.

movitto:
@intelliot @wilsonianb sorry for the belated response. Updated the patchset to incorporate all feedback.

Based of the original changes, the new patchset:
- exposes isValidAddress and isValidSecret to the public API
- Adds docs and tests for these API methods
- Move the deriveAddress and deriveKeypair methods to the 'offline' module
- Updates the docs to include these two methods in the 'offline' section

Let us know if there is anything else!

movitto:
@intelliot fixed in latest revision

intelliot:
Merged with 9eec9877. Thanks, @movitto !

movitto:
:+1:

PR #936, opened 2018-08-24
Use `object` instead of `Object`

intelliot:
per https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html


FredKSchott:
Good catch!

PR #943, opened 2018-09-07
added remaining LedgerEntry types

jcperkins12:
I was looking at the types included in the API and realized that the OfferLedgerEntry type was missing. There was a todo comment indicating that there were several more that were missing as well. 

This commit adds the missing types as indicated by the online documentation (DirectoryNode, OfferDirectoryNode, Escrow, FeeSettings, LedgerHashes, Offer, RippleState) and then joins them with the existing ones under type 'LedgerEntry'. I then removed the 'any' type from the union because all entries found in the documentation were included per the todo comment.

This is my first attempt at a pull request, so please let me know if there is a better way to structure things.

### DIFF
--- a/src/common/types/objects/ledger_entries.ts
+++ b/src/common/types/objects/ledger_entries.ts
@@ -1,4 +1,5 @@
 import {SignerEntry} from './index'
+import { Amount, RippledAmount } from './amounts'
 
 export interface AccountRootLedgerEntry {
   LedgerEntryType: 'AccountRoot',
@@ -44,6 +45,71 @@ export interface CheckLedgerEntry {
   SourceTag: number
 }
 
+export interface DirectoryNodeLedgerEntry {
+  LedgerEntryType: 'DirectoryNode';
+  Flags: number;
+  RootIndex: number;
+  Indexes: string[];
+  IndexNext?: number;
+  IndexPrevious?: number;
+}
+
+export interface OfferDirectoryNodeLedgerEntry extends DirectoryNodeLedgerEntry {
+  Owner?: string;
+  TakerPaysCurrency?: string;
+  TakerPaysIssuer?: string;
+  TakerGetsCurrency?: string;
+  TakerGetsIssuer?: string;
+};
+
+export interface EscrowLedgerEntry {
+  LedgerEntryType: 'Escrow';
+  Account: string;
+  Destination: string;
+  Amount: string;
+  Condition?: string;
+  CancelAfter?: number;
+  FinishAfter?: number;
+  Flags: number;
+  SourceTag?: number;
+  DestinationTag?: number;
+  OwnerNode: string;
+  DestinationNode?: string;
+  PreviousTxnID: string;
+  PreviousTxnLgrSeq: number;
+}
+
+export interface FeeSettingsLedgerEntry {
+  LedgerEntryType: 'FeeSettings';
+  BaseFee: string;
+  ReferenceFeeUnits: number;
+  ReserveBase: number;
+  ReserveIncrement: number;
+  Flags: number;
+};
+
+export interface LedgerHashesLedgerEntry {
+  LedgerEntryType: 'LedgerHashes';
+  LastLedgerSequence: number;
+  Hashes: string[];
+  Flags: number;
+}
+
+export interface OfferLedgerEntry {
+  LedgerEntryType: 'Offer';
+  Flags: number;
+  Account: string;
+  Sequence: number;
+  TakerPays: RippledAmount;
+  TakerGets: RippledAmount;
+  BookDirectory: string;
+  BookNode: string;
+  OwnerNode: string;
+  PreviousTxnID: string;
+  PreviousTxnLgrSeq: number;
+  Expiration?: number;
+}
+
 export interface PayChannelLedgerEntry {
   LedgerEntryType: 'PayChannel',
   Sequence: number,
@@ -63,6 +129,23 @@ export interface PayChannelLedgerEntry {
   index: string
 }
 
+export interface RippleStateLedgerEntry {
+  LedgerEntryType: 'RippleState';
+  Flags: number;
+  Balance: Amount;
+  LowLimit: Amount;
+  HighLimit: Amount;
+  PreviousTxnID: string;
+  PreviousTxnLgrSeq: number;
+  LowNode?: string;
+  HighNode?: string;
+  LowQualityIn?: number;
+  LowQualityOut?: number;
+  HighQualityIn?: number;
+  HighQualityOut?: number;
+
+}
+
 export interface SignerListLedgerEntry {
   LedgerEntryType: 'SignerList',
   OwnerNode: string,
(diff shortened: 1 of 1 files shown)


src/common/types/objects/ledger_entries.ts
```diff
@@ -73,11 +156,16 @@ export interface SignerListLedgerEntry {
...
   AmendmentsLedgerEntry |
+  DirectoryNodeLedgerEntry |
+  OfferDirectoryNodeLedgerEntry |
+  EscrowLedgerEntry |
+  FeeSettingsLedgerEntry |
+  LedgerHashesLedgerEntry |
+  OfferLedgerEntry |
   PayChannelLedgerEntry |
-  SignerListLedgerEntry |
-  any
```
intelliot: Do we want to add `CheckLedgerEntry` here?

jcperkins12: Good catch Elliot. I added `CheckLedgerEntry` between `AmendmentsLedgerEntry` and `DirectoryNodeLedgerEntry` to keep things in same order as the the documentation (alphabetical, if I remember right). 

Will it be possible to condense all of these commits into one for clarity, or is that something I should do on my side first?

intelliot: I believe I can use GitHub’s squash option to merge this PR as one commit. But if you prefer to do it on your side, that’s fine too. So - either way :)

wilsonianb: fwiw, `DepositPreauth` is also on the way in rippled 1.1.0 (pending amendment vote)
https://github.com/ripple/rippled/blob/7b0367730cf11b05395f24b537247d7341ea106a/src/ripple/protocol/impl/LedgerFormats.cpp#L172-L178

src/common/types/objects/ledger_entries.ts
```diff
@@ -44,6 +45,72 @@ export interface CheckLedgerEntry {
...
+  RootIndex: number,
+  Indexes: string[],
+  IndexNext?: number,
+  IndexPrevious?: number
+}
+
+export interface OfferDirectoryNodeLedgerEntry
+       extends DirectoryNodeLedgerEntry {
+  Owner?: string,
+  TakerPaysCurrency?: string,
```
wilsonianb: I think all `TakerPays`/`TakerGets` fields are required in offer directory nodes.
Same for `Owner` in owner directories.

src/common/types/objects/ledger_entries.ts
```diff
@@ -44,6 +45,72 @@ export interface CheckLedgerEntry {
   SourceTag: number
 }
 
+export interface DirectoryNodeLedgerEntry {
+  LedgerEntryType: 'DirectoryNode',
+  Flags: number,
+  RootIndex: number,
```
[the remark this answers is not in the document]
jcperkins12: Good catch. I'll make this change and several others when I get some time tonight or tomorrow. 

FYI, the code example at  https://developers.ripple.com/directorynode.html is a string and looks correct, but the table below the example code has the RootIndex property listed as a number.

src/common/types/objects/ledger_entries.ts
```diff
@@ -44,6 +45,72 @@ export interface CheckLedgerEntry {
...
+  Indexes: string[],
+  IndexNext?: number,
+  IndexPrevious?: number
+}
+
+export interface OfferDirectoryNodeLedgerEntry
+       extends DirectoryNodeLedgerEntry {
+  Owner?: string,
```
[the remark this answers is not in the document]
jcperkins12: Removed `Owner` from OfferDirectoryNode and created a new `OwnerDirectoryNodeLedgerEntry` with the Owner property required.
`export interface OwnerDirectoryNodeLedgerEntry
  extends DirectoryNodeLedgerEntry {
  Owner: string,
}`

Made TakerGets/TakerPays properties required on OfferDirectoryNodeLedgerEntry and added the `ExchangeRate?: number` as a depreciated/optional property.
`export interface OfferDirectoryNodeLedgerEntry
       extends DirectoryNodeLedgerEntry {
  TakerPaysCurrency: string,
  TakerPaysIssuer: string,
  TakerGetsCurrency: string,
  TakerGetsIssuer: string,
  ExchangeRate?: number  // DEPRECIATED
}`

src/common/types/objects/ledger_entries.ts
```diff
@@ -44,6 +45,72 @@ export interface CheckLedgerEntry {
...
+  TakerPaysCurrency?: string,
+  TakerPaysIssuer?: string,
+  TakerGetsCurrency?: string,
+  TakerGetsIssuer?: string
```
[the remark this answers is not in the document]
jcperkins12: added

src/common/types/objects/ledger_entries.ts
```diff
@@ -44,6 +45,72 @@ export interface CheckLedgerEntry {
...
+  BaseFee: string,
+  ReferenceFeeUnits: number,
+  ReserveBase: number,
+  ReserveIncrement: number,
+  Flags: number
+}
+
+export interface LedgerHashesLedgerEntry {
+  LedgerEntryType: 'LedgerHashes',
+  LastLedgerSequence: number,
```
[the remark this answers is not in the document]
jcperkins12: The documentation at https://developers.ripple.com/ledgerhashes.html lists `LastLegderSequence` as a required property, but the rippled code you referenced in another comment indicates that it is an optional property. 

Rippled src code is considered the authority, correct?

src/common/types/objects/ledger_entries.ts
jcperkins12: @intelliot I'd say lets use the squash option.

@wilsonianb Thanks for the input. I went ahead and added the DepositPreauth legder_entry. Even if it doesn't get voted in right away, I saw no harm considering the types don't affect runtime performance. All of the fields were easy to interpret due to their use in the other entries except for `Authorize` which I had to pick up from the SFields.cpp file. Can you confirm for me that I interpreted `Authorize` as a string (account) correctly. I've been learning cpp because of the rippled code, and have made my way through most of it, but I am far from proficient. 

https://github.com/ripple/rippled/blob/7b0367730cf11b05395f24b537247d7341ea106a/src/ripple/protocol/impl/SField.cpp#L223

```Javascript
export interface DepositPreauthLedgerEntry {
  LedgerEntryType: 'DepositPreauth',
  Account: string,
  Authorize: string,
  OwnerNode: string,
  PreviousTxnID: string,
  PreviousTxnLgrSeq: number
}
```

wilsonianb: Nice digging. Yes, `Authorize` is a string.

src/common/types/objects/ledger_entries.ts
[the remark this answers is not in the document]
wilsonianb: It sounds like every `LedgerHashes` entry probably has `LastLedgerSequence`, so I don't think the docs are wrong.
But you're correct to assume the rippled source is authoritative, so I'd vote for marking it as optional here for now.

mDuo13:
For what it's worth, the docs for DepositPreauth are here (just waiting for the release so I can merge & deploy them): https://github.com/ripple/ripple-dev-portal/pull/448/files#diff-5f76f116dfb0a34aa59ae29d8a2a393b

mDuo13:
Also, I'm not an expert on how this code is used but the list of objects & fields looks good to me.

FredKSchott:
Looks great, awesome stuff @jcperkins12 !

Issue #939, opened 2018-08-29
RippleAPI server should support unix socket


smsunarto:
ripple-lib seems to only work with `ws://` and `wss://`

It should support `ws+unix://` too.

intelliot:
What is `ws+unix://` and how is it different from `ws://`?

smsunarto:
@intelliot it's for websocket through unix socket 

ex: `ws+unix:///tmp/server.sock`

node.js `ws` library already supports it https://github.com/websockets/ws/blob/master/doc/ws.md so I think it would be sensical for ripple-lib to also support it by default.

smsunarto:
@intelliot I've confirmed that ws+unix:// actually works out of the box with ripple-lib, you just need to update your regex validator here https://github.com/ripple/ripple-lib/blob/develop/src/common/schemas/input/api-options.json

PR #944, opened 2018-09-12
Update server regex to accommodate UDS

smsunarto:
#939


intelliot:
UDS = Unix domain sockets

I'm guessing that these wouldn't work in the browser?

Is there a tutorial or documentation that shows how to set up a UDS?

I'm curious how to actually use it, since I've only used ws or wss to connect to rippled, even when running on the same machine.

intelliot:
Thank you. I don't have a lot of experience with docker but that sounds reasonable to me.

wilsonianb:
Run `npm run docgen` to update docs
https://github.com/ripple/ripple-lib/blob/develop/docs/index.md#parameters

PR #950, opened 2018-09-23
Allow to remove signerList

ihomp:
As SignerEntries always specified it returns temMALFORMED when SignerQuorum is 0. (removing signerList)

to reproduce, submit 
```
settings.signers = {
  threshold: 0
};
```

### DIFF
--- a/src/transaction/settings.ts
+++ b/src/transaction/settings.ts
@@ -91,12 +91,15 @@ function createSettingsTransactionWithoutMemos(
   }
 
   if (settings.signers !== undefined) {
-    return {
+    const setSignerList = {
       TransactionType: 'SignerListSet',
       Account: account,
-      SignerQuorum: settings.signers.threshold,
-      SignerEntries: _.map(settings.signers.weights, formatSignerEntry)
+      SignerQuorum: settings.signers.threshold
+    };
+    if (settings.signers.weights !== undefined) {
+      setSignerList.SignerEntries = _.map(settings.signers.weights, formatSignerEntry);
     }
+    return setSignerList;
   }
 
   const txJSON: any = {


src/transaction/settings.ts
```diff
@@ -91,12 +91,15 @@ function createSettingsTransactionWithoutMemos(
   }
 
   if (settings.signers !== undefined) {
-    return {
+    const setSignerList = {
       TransactionType: 'SignerListSet',
       Account: account,
-      SignerQuorum: settings.signers.threshold,
-      SignerEntries: _.map(settings.signers.weights, formatSignerEntry)
```
intelliot: For TypeScript you'll need to add some type info for `setSignerList`...

src/transaction/settings.ts
```diff
@@ -91,12 +91,15 @@ function createSettingsTransactionWithoutMemos(
...
+      SignerQuorum: settings.signers.threshold
+    };
+    if (settings.signers.weights !== undefined) {
+      setSignerList.SignerEntries = _.map(settings.signers.weights, formatSignerEntry);
```
intelliot: Currently line  produces an error:
```
src/transaction/settings.ts:100:21 - error TS2339: Property 'SignerEntries' does not exist on type '{ TransactionType: string; Account: string; SignerQuorum: number; }'.
```

intelliot:
@ihomp Can you fix the two problems I mentioned above?

intelliot:
@ihomp Would you be able to update this PR?

ihomp:
Hi, sorry, don't have time to look into it.
Just wanted to show where is the problem... @intelliot 
would be nice if that would be fixed in feature realises.

PR #951, opened 2018-09-25
book_offers returns offers type OfferLedgerEntry

jcperkins12:
discovered in [PR#460](https://github.com/ripple/ripple-dev-portal/pull/460)

the rippled-api book_offers functions returns offers with a base type of `OfferLedgerEntry` not `OfferCreateTransaction`.

It appears that there is enough crossover between the two types that I haven't found anything that breaks after switching the type, but any code that is developed in the future would have expected offers to have fields such as `TransactionType: 'OfferCreate` when the offer really had the field 'LedgerEntryType: 'Offer'`.

This corrects the getOrderBook response by indicating that the `data` property of each order included in the bids/asks array of the response is now the correct type

This does not impact the getOrder, getTransaction, getTransactions, or getAccountObjects responses which either do not refer to the `BookOffer` type or clean the offer with the `parseOrderbookOrder` function.

### DIFF
--- a/src/common/types/commands/account_objects.ts
+++ b/src/common/types/commands/account_objects.ts
@@ -1,4 +1,9 @@
-import {CheckLedgerEntry, RippleStateLedgerEntry, OfferLedgerEntry, SignerListLedgerEntry, EscrowLedgerEntry, PayChannelLedgerEntry, DepositPreauthLedgerEntry} from '../objects'
+import {
+  CheckLedgerEntry, RippleStateLedgerEntry,
+  OfferLedgerEntry, SignerListLedgerEntry,
+  EscrowLedgerEntry, PayChannelLedgerEntry,
+  DepositPreauthLedgerEntry
+} from '../objects'
 
 export interface GetAccountObjectsOptions {
   type?: string | (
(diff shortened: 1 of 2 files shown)


jcperkins12:
while inspecting the impact of the type change, I changed the type of the `account_objects` property of `AccountObjectsResponse` to reflect the allowed types for the field. 

In addition to the `CheckLedgerEntry` type that was already allowed, I removed the ambiguous `object` type and added 
   ```
   RippleStateLedgerEntry |
   OfferLedgerEntry |
   SignerListLedgerEntry |
   EscrowLedgerEntry |
   PayChannelLedgerEntry |
   DepositPreauthLedgerEntry
   ```

Change made in second commit

src/common/types/commands/account_objects.ts
```diff
@@ -44,7 +49,16 @@ export interface AccountObjectsResponse {
   account: string,
 
   // Array of objects owned by this account.
-  account_objects: CheckLedgerEntry | object,
+  // from the getAccountObjects section of the dev center
+  account_objects: Array<
```
FredKSchott: good find. If we find multiple uses for this Union Type, we can refactor it into it's own type as well

PR #958, opened 2018-10-18
Add DepositPreauth

intelliot:
- Add documentation of `prepareTransaction`, the preferred method for preparing `DepositPreauth` transactions.
  - ripple-lib was already capable of signing `DepositPreauth` transactions thanks to ripple-binary-codec v0.1.15. However, this release was not tagged on GitHub, so I incremented the library's version to v0.2.0 and published it to npm and GitHub.
  - I'm pinning ripple-lib to use ripple-binary-codec v0.2.0 so that we can be explicit about using updates of this dependency in the future.
- Add support for parsing `DepositPreauth` transactions.
- Allow ripple-lib to parse unrecognized transaction types, so that ripple-lib is still usable when new transaction types are added in the future.
  - For unrecognized transaction types, no `specification` will be provided, but `rawTransaction` will be included.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -58,6 +58,7 @@
   - [getAccountObjects](#getaccountobjects)
   - [getPaymentChannel](#getpaymentchannel)
   - [getLedger](#getledger)
+  - [prepareTransaction](#preparetransaction)
   - [preparePayment](#preparepayment)
   - [prepareTrustline](#preparetrustline)
   - [prepareOrder](#prepareorder)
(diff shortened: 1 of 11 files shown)


test/api-test.js
```diff
@@ -883,6 +883,30 @@ describe('RippleAPI', function () {
...
+        instructions: {
+          fee: '0.000012',
+          sequence: 23,
+          maxLedgerVersion: 8820051
+        }
+      }
+      return checkResult(expected, 'prepare', response)
+    })
+  })
+
```
shekenahglory: Would it be worthwhile to include a test for `Unauthorize`?

intelliot: Good call. Added in cd1be9d

intelliot:
@mDuo13 I exposed a method called `parseAccountFlags()` which takes a `Flags` number and returns an object naming the flags and indicating whether they are true or false.

docs/index.md
```diff
@@ -4103,6 +4105,109 @@ return api.getLedger()
...
+
+Parse an `AccountRoot` object's [`Flags`](https://developers.ripple.com/accountroot.html#accountroot-flags).
+
+### Parameters
+
+This method takes one parameter, the `Flags` number to parse.
+
+### Return Value
+
+This method returns an object with containing a key for each parsed flag. Each flag has a boolean value of `true` or `false`.
```
mDuo13: The example implies that flags that are `false` are omitted from the return value, or at least that flags whose value is their default value are omitted. That should be explicit. Alternatively, it would be OK (maybe more useful?) if all known flags were included (even if most of them are `false`).

docs/index.md
```diff
@@ -4103,6 +4105,109 @@ return api.getLedger()
...
 
+## parseAccountFlags
+
+`parseAccountFlags(Flags: number): object`
+
+Parse an `AccountRoot` object's [`Flags`](https://developers.ripple.com/accountroot.html#accountroot-flags).
+
+### Parameters
+
+This method takes one parameter, the `Flags` number to parse.
```
mDuo13: Maybe worth a note that this only works for AccountRoot `Flags` since the flags have different mappings on other types of objects or in transactions like AccountSet.

docs/index.md
```diff
@@ -4103,6 +4105,109 @@ return api.getLedger()
...
+
+This method works with any of [the transaction types supported by rippled](https://developers.ripple.com/transaction-types.html).
+
+Notably, this is the preferred method for preparing a `DepositPreauth` transaction (added in rippled 1.1.0).
+
+### Parameters
+
+Name | Type | Description
+---- | ---- | -----------
+transaction | [transaction](https://developers.ripple.com/transaction-formats.html) | The specification (JSON) of the transaction to prepare. Set `Account` to the address of the account that is creating the transaction. Common fields like `Fee`, `Flags`, and `Sequence` may be omitted; `prepareTransaction` will set them automatically.
```
mDuo13: In the rippled API documentation, we use the term "Auto-fillable fields" for fields whose values can be automatically provided, and "Common fields" for fields that apply to all transactions (but whose values cannot be automatically provided, such as `Account`). Let's try to match that usage here.

docs/index.md
```diff
@@ -4955,6 +5060,10 @@ sign(txJSON: string, keypair: object, options: object): {signedTransaction: stri
 
 Sign a prepared transaction. The signed transaction must subsequently be [submitted](#submit).
 
+This method can sign any of [the transaction types supported by ripple-binary-codec](https://github.com/ripple/ripple-binary-codec/blob/cfcde79c19c359e9a0466d7bc3dc9a3aef47bb99/src/enums/definitions.json#L1637). When a new transaction type is added, it will be unrecognized until ripple-binary-codec is updated. An error will be thrown:
```
mDuo13: For style reasons, avoid future tense and passive voice except when talking about future events. (In this paragraph, there are two instances of future tense, one of which is appropriate and one of which isn't.) Here's a suggested rephrase:

> When a new transaction type is added to the XRP Ledger, it will be unrecognized until ripple-binary-codec is updated. If you try to sign an unrecognized transaction type, this method throws an error similar to the following:

intelliot:
This is ready for re-review

docs/index.md
```diff
@@ -4164,8 +4155,8 @@ Notably, this is the preferred method for preparing a `DepositPreauth` transacti
 
 Name | Type | Description
 ---- | ---- | -----------
-transaction | [transaction](https://developers.ripple.com/transaction-formats.html) | The specification (JSON) of the transaction to prepare. Set `Account` to the address of the account that is creating the transaction. Common fields like `Fee`, `Flags`, and `Sequence` may be omitted; `prepareTransaction` will set them automatically.
-instructions | [instructions](#transaction-instructions) | *Optional* Instructions for executing the transaction
+transaction | [transaction](https://developers.ripple.com/transaction-formats.html) | The specification (JSON) of the transaction to prepare. Set `Account` to the address of the account that is creating the transaction. Auto-fillable fields like `Fee`, `Flags`, and `Sequence` may be omitted; `prepareTransaction` will set them automatically.
```
mDuo13: Nit: I suggest changing the last sentence just a little more for conciseness, active voice, etc.:

> You may omit auto-fillable fields like `Fee`, `Flags`, and `Sequence` to have them set automatically.

PR #962, opened 2018-10-29
Add comments for accountRootFlags

intelliot:
Note where the account root flags are documented on the developers site

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -1,16 +1,51 @@
 
 import {txFlagIndices} from './txflags'
 
+// Ordering from https://developers.ripple.com/accountroot.html
 const accountRootFlags = {
-  PasswordSpent: 0x00010000, // password set fee is spent
-  RequireDestTag: 0x00020000, // require a DestinationTag for payments
-  RequireAuth: 0x00040000, // require authorization to hold IOUs
-  DepositAuth: 0x01000000, // require account to auth deposits
-  DisallowXRP: 0x00080000, // disallow sending XRP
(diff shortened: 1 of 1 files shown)


src/common/constants.ts
```diff
@@ -1,16 +1,51 @@
...
-  DisallowXRP: 0x00080000, // disallow sending XRP
-  DisableMaster: 0x00100000, // force regular key
-  NoFreeze: 0x00200000, // permanently disallowed freezing trustlines
-  GlobalFreeze: 0x00400000, // trustlines globally frozen
-  DefaultRipple: 0x00800000
+
+  // lsfDefaultRipple:
+  // Enable rippling on trust lines by default.
+  // Required for issuing addresses; discouraged for others.
+  DefaultRipple: 0x00800000,
```
FredKSchott: If it doesn't matter to you, using JSDoc syntax comment syntax will cause VSCode/other code editors to inline the documentation (this is true for variable definitions, and I believe object properties as well).

```diff
-  // lsfDefaultRipple:
-  // Enable rippling on trust lines by default.
-  // Required for issuing addresses; discouraged for others.
-  DefaultRipple: 0x00800000,
+  /**
+   * Enable rippling on trust lines by default.
+   * Required for issuing addresses; discouraged for others.
+   */
+  DefaultRipple: 0x00800000,
```

intelliot: Nice! I would like to keep `lsfDefaultRipple` somewhere if possible. The initial motivation for these comments is that I tried to search the codebase for `lsfDepositAuth` to see where it was implemented/used.

PR #964, opened 2018-10-30
Support node 10

intelliot:
Add support for node v10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ node_js:
   - 6
   - 8
   - 9
+  - 10
 script:
   - yarn compile
   - yarn test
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -3,6 +3,7 @@ node_js:
   - 6
   - 8
   - 9
```
FredKSchott: Node 9 is no longer supported by the core team (not covered by LTS). Fine to keep here just to be notified if/when we do lose support, but I wouldn't go out of our way to keep supporting it if something does come up.

README.md
```diff
@@ -16,11 +16,14 @@ A JavaScript API for interacting with the XRP Ledger
 
 See also: [RippleAPI Beginners Guide](https://ripple.com/build/rippleapi-beginners-guide/)
 
-You can use `npm`, but we recommend using `yarn` for the added assurance provided by `yarn.lock`.
+### Requirements
 
-+ [Yarn Installation Instructions](https://yarnpkg.com/en/docs/install)
++ **[Node v10](https://nodejs.org/)** is recommended. Other versions may work but are not frequently tested.
```
FredKSchott: nice! very clear

PR #970, opened 2018-11-22
Fix getOrderbook

intelliot:
- Modifies `getOrderbook` so that orders are sorted correctly

### DIFF
--- a/test/mock-rippled.js
+++ b/test/mock-rippled.js
@@ -11,6 +11,7 @@ const accountLinesResponse = require('./fixtures/rippled/account-lines');
 const accountObjectsResponse = require('./fixtures/rippled/account-objects');
 const fullLedger = require('./fixtures/rippled/ledger-full-38129.json');
 const { getFreePort } = require('./utils/net-utils');
+const fs = require('fs');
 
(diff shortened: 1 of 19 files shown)


FredKSchott:
Sorry I'm late to seeing this, but belated 👍 

Really excited to see this actually, the `api.request()` method gets to stay flexible & true to the real API interface, but helper methods can still be used to format and/or handle the request/response where user-friendliness is needed.

Rojo123abc (approved):

test/mock-rippled.js
```diff
@@ -479,6 +482,30 @@ module.exports = function createMockRippled(port) {
...
+            "status": "success",
+            "result": JSON.parse(res)
+          });
+          conn.send(response);
+          return;
+        }
+      }
+
       assert(false, 'Unrecognized order book: ' + JSON.stringify(request));
     }
```
Rojo123abc, line 510: Send null booleane

Rojo123abc: Fubar

Rojo123abc (approved):

Issue #977, opened 2018-11-30
Submit Response Object does not include tx_json


alexchiriac:
Hi! I've been working with the library for a couple months, but this is actually my first attempt at providing a contribution! Here is the issue I'm facing:

When submitting a transaction, the response object returned by the library is in the following format
`src/transaction/submit.ts`:

```
{
    resultCode: response.engine_result,
    resultMessage: response.engine_result_message
}
```

My use case is that the server receives a signed transaction object but then has to return the transaction hash and other data about the pre-signed tx_json object, which is currently not supported by ripple-lib.

I only propose a non-breaking-change of adding just another field in the response object with the tx_json object provided by the rippled server, such as the following:

```
 {
    resultCode: response.engine_result,
    resultMessage: response.engine_result_message,
    transactionInput: response.tx_json
  }
```
*please tell me if you'd like another name for the `transactionInput` field and if you'd also want the `tx_blob` string included as well.

I have already worked on the fix. If the change is welcome then I can contribute and submit the PR in the following days!

intelliot:
Yes, that sounds like a great contribution!

Personally, I would like to see the following:

1. In the response, add all of the fields returned by rippled:
    - engine_result
    - engine_result_code
    - engine_result_message
    - tx_blob
    - tx_json

    (https://developers.ripple.com/submit.html)

    To prevent confusion for developers using both APIs, I would use the same names as `rippled`.

2. Mark resultCode and resultMessage as deprecated, to be removed in a future version of ripple-lib (most likely 2.0.0).

alexchiriac:
Great! I'm on it!

PR #978, opened 2018-12-03
Update Submit Response

alexchiriac:
Hi!

I've updated the submit response as suggested in [#977](https://github.com/ripple/ripple-lib/issues/977). It includes the full object returned by the rippled server and now contains the following:

- resultCode
- resultMessage
- engine_result
- engine_result_code
- engine_result_message
- tx_blob
- tx_json

I've updated the related schemas, fixed the tests to correspond to the changes and updated the documentation with the `docgen` script.

Please tell me if I missed anything or if something is not quite right! I appreciate any feedback!


alexchiriac:
Thanks! Yeah, definitely!

PR #989, opened 2019-03-07
fix: add amount to FormattedPaymentChannel type

kincaidoneil:
The `amount` field is returned from `getPaymentChannel(<channelId>)` and is [documented](https://github.com/ripple/ripple-lib/blob/develop/docs/index.md#getpaymentchannel), but wasn't included in the type for `FormattedPaymentChannel`.


kincaidoneil:
Oh, my bad! Fixed.

kincaidoneil:
@intelliot Should I add `package-lock.json` to the .gitignore so it doesn't get accidentally committed in the future?

intelliot:
I think that’s a good idea! 

In the future, if we decide to maintain a package-lock file, we can remove it from gitignore at the same time we commit the new file.

kincaidoneil:
@intelliot Updated!

intelliot:
Perfect - thank you!

PR #990, opened 2019-03-13
prepareTransaction should not overwrite Sequence

intelliot:
Also cleans up some code and fixes some type errors

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -5491,7 +5491,7 @@ engine_result | string | Code indicating the preliminary result of the transacti
 engine_result_code | integer | Numeric code indicating the preliminary result of the transaction, directly correlated to `engine_result`
 engine_result_message | string | Human-readable explanation of the transaction's preliminary result.
 tx_blob | string | The complete transaction in hex string format.
-tx_json | [tx](https://ripple.com/build/transactions/) | The complete transaction in JSON format.
+tx_json | [tx-json](https://ripple.com/build/transactions/) | The complete transaction in JSON format.
 
 ### Example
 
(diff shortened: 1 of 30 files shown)


docs/index.md
```diff
@@ -5491,7 +5491,7 @@ engine_result | string | Code indicating the preliminary result of the transacti
 engine_result_code | integer | Numeric code indicating the preliminary result of the transaction, directly correlated to `engine_result`
 engine_result_message | string | Human-readable explanation of the transaction's preliminary result.
 tx_blob | string | The complete transaction in hex string format.
-tx_json | [tx](https://ripple.com/build/transactions/) | The complete transaction in JSON format.
+tx_json | [tx-json](https://ripple.com/build/transactions/) | The complete transaction in JSON format.
```
mDuo13: Should update this link to https://developers.ripple.com/transaction-formats.html

src/transaction/utils.ts
```diff
@@ -12,6 +12,15 @@ export type ApiMemo = {
   MemoFormat?: string
 }
 
+export type TransactionJSON = {
+  Account: string,
+  TransactionType: string,
+  Memos?: {Memo: ApiMemo}[],
+  Flags?: number,
+  Fulfillment?: string,
```
mDuo13: It's weird to me that `Fulfillment` (an EscrowFinish-specifie field) is here, but the other [common fields](https://developers.ripple.com/transaction-common-fields.html) are not.

Is it because the required transaction cost is higher if `Fulfillment` is provided?

intelliot: Yes, exactly

PR #994, opened 2019-04-08
Add "strict": true to tsconfig.json

intelliot:
- Remove process.browser

@adrianhopebailie @kincaidoneil I don't know why this doesn't catch the error seen in #993 but regardless, I would appreciate if you could have a look at this PR too!

### DIFF
--- a/src/common/serverinfo.ts
+++ b/src/common/serverinfo.ts
@@ -1,7 +1,7 @@
 import * as _ from 'lodash'
 import {convertKeysFromSnakeCaseToCamelCase} from './utils'
 import BigNumber from 'bignumber.js'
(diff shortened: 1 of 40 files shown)


src/common/serverinfo.ts
```diff
@@ -1,7 +1,7 @@
 import * as _ from 'lodash'
 import {convertKeysFromSnakeCaseToCamelCase} from './utils'
 import BigNumber from 'bignumber.js'
-import {RippleAPI} from '../index'
+import {RippleAPI} from '..'
```
FredKSchott: fwiw I prefer `../index` vs. `..`, but that's just personal preference

PR #996, opened 2019-04-09
Add deliveredAmount as optional field for type Outcome

Daniel-MCO:
This PR is created according to issue https://github.com/ripple/ripple-lib/issues/995

**The gist**
Adding back the `deliveredAmount` field to the type `Outcome` based on the `getTransaction` api call from https://developers.ripple.com/rippleapi-reference.html#gettransaction,

### DIFF
--- a/src/transaction/types.ts
+++ b/src/transaction/types.ts
@@ -82,6 +82,11 @@ export type Outcome = {
     }[]
   },
   orderbookChanges: object,
+  deliveredAmount?: {
+    currency: string;
+    value: string;
+    counterparty?: string;
+  };
   timestamp?: string
 }
 
--- a/src/transaction/types.ts
+++ b/src/transaction/types.ts
@@ -83,10 +83,10 @@ export type Outcome = {
   },
   orderbookChanges: object,
   deliveredAmount?: {
-    currency: string;
-    value: string;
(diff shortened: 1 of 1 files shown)


src/transaction/types.ts
```diff
@@ -82,6 +82,11 @@ export type Outcome = {
     }[]
   },
   orderbookChanges: object,
+  deliveredAmount?: {
+    currency: string;
+    value: string;
+    counterparty?: string;
+  };
```
intelliot: I think these should be commas rather than semicolons

Daniel-MCO: Sorry on missing the comma conventions here. I didn't know that Typescript supports both commas and semicolons as delimiters for `type` and `interface` declarations.

We've been using `semicolons` personally and it appears this has been the case too from their official documentation https://www.typescriptlang.org/docs/handbook/advanced-types.html

Definitely interesting to see different conventions being used.

intelliot: Interesting!

intelliot:
Thanks! Adding a snippet that uses the `deliveredAmount` field: https://github.com/ripple/ripple-lib/pull/998/files#diff-d2afa978a427b83c13dd6f17477391f5R13

PR #998, opened 2019-04-12
Use TS project references

intelliot:
Add getTransaction example snippet


intelliot:
In the future we will migrate the tests to TypeScript and use a project reference for that

intelliot:
Replaced by #1002

PR #1000, opened 2019-04-12
Prevent prepareTransaction from overwriting values in txJSON

### LINKED ISSUE #997: prepareTransaction should not overwrite `Fee`
Similar to #990, the prepareTransaction() method should respect the `Fee` parameter of the input txJSON if provided.

Example:
```js
  const resp = await api.prepareTransaction({
    "TransactionType": "AccountSet",
    "Account": from_address,
    "Fee": "10"
  }) // No-op tx
  console.log("resp:", resp)
```

Output:

```js
resp: { txJSON:
   '{"TransactionType":"AccountSet","Account":"rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn","Fee":"12","Flags":2147483648,"LastLedgerSequence":46447424,"Sequence":376}',
  instructions:
   { fee: '0.000012', sequence: 376, maxLedgerVersion: 46447424 } }
```

Note that the `Fee` is `"12"` , not `"10"` as specified.

intelliot:
Fix #997

### DIFF
--- a/src/transaction/utils.ts
+++ b/src/transaction/utils.ts
@@ -68,11 +68,11 @@ function prepareTransaction(txJSON: TransactionJSON, api: RippleAPI,
     //   instructions.maxLedgerVersionOffset
     if (txJSON.LastLedgerSequence && instructions.maxLedgerVersion) {
       return Promise.reject(new ValidationError('`LastLedgerSequence` in txJSON and `maxLedgerVersion`' +
-        ' in Instructions cannot both be set'))
+        ' in `instructions` cannot both be set'))
     }
     if (txJSON.LastLedgerSequence && instructions.maxLedgerVersionOffset) {
       return Promise.reject(new ValidationError('`LastLedgerSequence` in txJSON and `maxLedgerVersionOffset`' +
-        ' in Instructions cannot both be set'))
+        ' in `instructions` cannot both be set'))
     }
     if (txJSON.LastLedgerSequence) {
       return Promise.resolve(txJSON)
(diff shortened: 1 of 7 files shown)


src/transaction/utils.ts
```diff
@@ -51,11 +51,32 @@ function prepareTransaction(txJSON: TransactionJSON, api: RippleAPI,
 ): Promise<Prepare> {
   common.validate.instructions(instructions)
   common.validate.tx_json(txJSON)
+  const disallowedFieldsInTxJSON = ['maxLedgerVersion', 'maxLedgerVersionOffset', 'fee', 'sequence']
+  const badFields = disallowedFieldsInTxJSON.filter(field => txJSON[field])
+  if (badFields.length) {
+    return Promise.reject(new ValidationError('txJSON additionalProperty "' + badFields[0] +
```
jwbusch: nit: this only returns the first invalid error, could include all of the bad fields.

intelliot: 👍 Thanks! I think it'll be rare for someone to use multiple bad fields and if they do, it's still OK to point them out one by one.

src/transaction/utils.ts
```diff
@@ -51,11 +51,32 @@ function prepareTransaction(txJSON: TransactionJSON, api: RippleAPI,
...
 
-  function prepareMaxLedgerVersion(): Promise<object> {
+  function prepareMaxLedgerVersion(): Promise<TransactionJSON> {
+    // Up to one of the following is allowed:
+    //   txJSON.LastLedgerSequence
+    //   instructions.maxLedgerVersion
+    //   instructions.maxLedgerVersionOffset
+    if (txJSON.LastLedgerSequence && instructions.maxLedgerVersion) {
+      return Promise.reject(new ValidationError('`LastLedgerSequence` in txJSON and `maxLedgerVersion`' +
+        ' in Instructions cannot both be set'))
```
mDuo13: nit: I suggest writing \`instructions` rather than Instructions, to be consistent with how the other parts are written

src/transaction/utils.ts
```diff
@@ -51,11 +51,32 @@ function prepareTransaction(txJSON: TransactionJSON, api: RippleAPI,
...
+    }
+    if (txJSON.LastLedgerSequence && instructions.maxLedgerVersionOffset) {
+      return Promise.reject(new ValidationError('`LastLedgerSequence` in txJSON and `maxLedgerVersionOffset`' +
+        ' in Instructions cannot both be set'))
```
mDuo13: \`instructions` here too

PR #1005, opened 2019-04-18
Update webpack and ts-loader

intelliot:
Includes the fixes in #1006 and goes further by:
- Updating webpack and ts-loader
- Removing uglify since webpack seems to automatically do equally good minification
- Modifying the gulp tasks to output the errors/warnings from webpack
- Removing gulp tasks that I don't think we use anymore (?)

TODO: Looks like gulp has a new API pattern - we may want to export our tasks instead of using [task()](https://gulpjs.com/docs/en/api/task).

Details:

- Building ripple-lib for the browser was not working because we set `"noEmitOnError": true` and webpack/ts-loader was encountering an error when packaging `https-proxy-agent` for the browser.
- `https-proxy-agent` can only be used in Node.js, so this PR sets `browser["https-proxy-agent"] = false` in package.json.
- When using TS project references, we need to set `composite: true` in tsconfig.json, but when using webpack, in order to use `declaration: false`, we need `composite` and `declarationMap` to both be `false`.
- Prior to this PR, we were already on webpack@3.12.0; since webpack >= v2.0.0, importing of JSON files should work by default. Furthermore, the build fails if we attempt to use json-loader as previously specified in our webpack config (I'm not sure why this is, but we get parse errors).
- We cannot update `uglifyjs-webpack-plugin` to its latest version because it drops support for ES6. We could switch to `terser-webpack-plugin` (a fork that adds support for ES6), but it seems that the new version of webpack (v4) automatically performs minimization with `mode: 'production'`.
- I don't know what the use case for the non-minified version of ripple-lib was; if you're using it in production, you should use the minified version. So I think we can minify `ripple-{version}.js` and get rid of `ripple-{version}-min.js`. (Or vice-versa - let me know what you'd prefer.)

This was all very difficult to figure out and debug because gulp doesn't output the errors/warnings from webpack, at least not when using it in this way (perhaps we should be using [`webpack-stream`](https://webpack.js.org/guides/integrations#gulp)?). But I've put in a workaround here that just `console.log`s webpack's errors and warnings when webpack is done.

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -6,24 +6,24 @@ const fs = require('fs');
 const path = require('path');
 const assert = require('assert');
 const gulp = require('gulp');
-const rename = require('gulp-rename');
+// const rename = require('gulp-rename');
 const webpack = require('webpack');
-const bump = require('gulp-bump');
-const argv = require('yargs').argv;
-const UglifyJsPlugin = require('uglifyjs-webpack-plugin')
+// const bump = require('gulp-bump');
+// const argv = require('yargs').argv;
+// const TerserPlugin = require('terser-webpack-plugin');
 
 var pkg = require('./package.json');
 
-var uglifyOptions = {
-  mangle: {
-    reserved: ['_', 'RippleError', 'RippledError', 'UnexpectedError',
-    'LedgerVersionError', 'ConnectionError', 'NotConnectedError',
-    'DisconnectedError', 'TimeoutError', 'ResponseFormatError',
-    'ValidationError', 'NotFoundError', 'MissingLedgerHistoryError',
-    'PendingLedgerVersionError'
-    ]
-  }
-};
+// var terserOptions = {
+//   mangle: {
+//     reserved: ['_', 'RippleError', 'RippledError', 'UnexpectedError',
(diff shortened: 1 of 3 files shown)


mDuo13:
I think the use case for the non-minified version is to help with debugging dependent code during development. If you're using ripple-lib wrong, you might get a confusing error message or strange behavior; if you get a traceback or something from a non-minified version, you can maybe see what part of ripple-lib didn't like your call, which can help figure out what you're doing wrong. (I've used it that way in the past.)

intelliot:
@mDuo13 Good point, but why not use the debug version for that? (`ripple-{version}-debug.js`)

mDuo13:
¯\_(ツ)_/¯

I think it wasn't there the first time I did it? Or maybe I was paranoid about the debug version coming with a performance hit?

intelliot:
Based on Gulpfile.js, we have been building `ripple-{version}-debug.js` for at least 4 years. I'm not clear on what the difference is between the debug version and the normal version. I would recommend using the minified version in production, to reduce download/load times.

FKSRipple:
@intelliot asked me to take this work over, see #1061 for latest

PR #1021, opened 2019-07-11
Fix #971 - Allow to remove signer list

### LINKED ISSUE #971: Allow to remove signerList
See https://github.com/ripple/ripple-lib/pull/950

namratahonnavar:
When the SignerQuorum is set to 0(signer list is removed), SignerEntries field can be deleted/omitted. Otherwise it returns a temMALFORMED error.


src/transaction/settings.ts
```diff
@@ -89,12 +90,16 @@ function createSettingsTransactionWithoutMemos(
...
+    const setSignerList = {
       TransactionType: 'SignerListSet',
       Account: account,
-      SignerQuorum: settings.signers.threshold,
-      SignerEntries: settings.signers.weights.map(formatSignerEntry)
+      SignerEntries: [],
+      SignerQuorum: settings.signers.threshold
+    };
+    if (settings.signers.weights !== undefined) {
+      setSignerList.SignerEntries = _.map(settings.signers.weights, formatSignerEntry);
```
intelliot: Is it necessary to use lodash here? We would actually like to remove the lodash dependency entirely. It adds a lot to the size of the library (especially the browser version) and there are always other ways to accomplish the same thing.

test/fixtures/requests/index.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
   },
   prepareSettings: {
     domain: require('./prepare-settings'),
+    noSignerEntries: require('./prepare-settings-no-signer-entires'),
```
intelliot: entires -> entries

namratahonnavar: oops. Will fix the typo

namratahonnavar: Thanks for taking the time to review the code.

src/transaction/utils.ts
```diff
@@ -58,6 +58,11 @@ function prepareTransaction(txJSON: TransactionJSON, api: RippleAPI,
       '" exists in instance when not allowed'))
   }
 
+  // To remove the signer list, SignerEntries field should be omitted.
+  if (txJSON['SignerQuorum'] === 0) {
+    delete txJSON.SignerEntries;
```
intelliot: It looks like the code here ignores and deletes the value anyway...

intelliot: It looks like the code here ignores and deletes the value anyway...

test/fixtures/requests/prepare-settings-no-signer-entries.json
```diff
@@ -0,0 +1,11 @@
+{
+  "signers": {
+    "threshold": 0,
+    "weights": [
```
intelliot: It's odd that the `weights` need to be specified when we are trying to delete the list of signers.

Notice that in the rippled API, the SignerEntries field is omitted when deleting: https://xrpl.org/signerlistset.html

I think we can make the `weights` field optional: https://github.com/ripple/ripple-lib/blob/develop/src/common/schemas/objects/settings-plus-memos.json#L90

Then, it probably makes sense to add validation to ensure that if `threshold` is 0, then `weights` must not exist; if threshold >= 1, then `weights` is required.

test/fixtures/requests/prepare-settings-no-signer-entries.json
```diff
@@ -0,0 +1,11 @@
...
+      {
+        "address": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
```
intelliot: Since we're deleting the list, why do we need to specify an address here at all?

test/fixtures/requests/prepare-settings-no-signer-entries.json
namratahonnavar: Thank you for your input. I agree with you on making the weights field optional based on the threshold value.

test/fixtures/requests/prepare-settings-no-signer-entries.json
namratahonnavar: I was getting validation error when I left out the Weight and the address fields. As you suggested above, let me try making the `weights` field optional based on threshold value.

intelliot: I think you can make "weight" optional by removing it from the "required" array in `settings-plus-memos.json`: https://github.com/ripple/ripple-lib/blob/develop/src/common/schemas/objects/#L90

namratahonnavar: @intelliot Thanks for your input. Instead of making the 'weight' field optional,
I made the 'weights' field optional(it has both weight and address) and updated the corresponding tests as well. 
I tested the following scenarios against the test net server:
1. When threshold: 0, and no weights are specified - The transaction is submitted successfully.
2. When threshold: 0 and optional 'weights' are passed - The transaction is submitted successfully.
3. When threshold > 0 and no 'weights' are passed - The transaction fails with a Rippled error.
4. When threshold > 0 and either weight or address is not specified - The transaction fails with 
a ValidationError.
Please let me know what you think about this approach.

PR #1026, opened 2019-08-08
Sign method - verify accurate encoding

intelliot:
When signing a transaction, we should decode the signedTransaction blob and check the integrity (accuracy) of the data.

### DIFF
--- a/src/transaction/sign.ts
+++ b/src/transaction/sign.ts
@@ -1,3 +1,4 @@
+import * as isEqual from '../common/js/lodash.isequal'
 import * as utils from './utils'
 import keypairs = require('ripple-keypairs')
 import binary = require('ripple-binary-codec')
@@ -32,14 +33,7 @@ function signWithKeypair(
     )
   }
 
-  const fee = new BigNumber(tx.Fee)
-  const maxFeeDrops = xrpToDrops(api._maxFeeXRP)
-  if (fee.greaterThan(maxFeeDrops)) {
-    throw new utils.common.errors.ValidationError(
-      `"Fee" should not exceed "${maxFeeDrops}". ` +
-      'To use a higher fee, set `maxFeeXRP` in the RippleAPI constructor.'
-    )
-  }
+  checkFee(api, tx.Fee)
 
   tx.SigningPubKey = options.signAs ? '' : keypair.publicKey
 
@@ -55,9 +49,58 @@ function signWithKeypair(
   }
 
   const serialized = binary.encode(tx)
+
+  // Decode the serialized transaction:
+  const decoded = binary.decode(serialized)
+
+  // ...And ensure it is equal to the original tx, except:
+  // - It must have a TxnSignature or txSigners (multisign).
+  if (!decoded.TxnSignature && !tx.Signers) {
+    throw new utils.common.errors.ValidationError(
+      'Serialized signed transaction is missing "TxnSignature" property'
+    )
+  }
+  // - We know that the original tx did not have TxnSignature, so we should delete it:
+  delete decoded.TxnSignature
+  // - We know that the original tx did not have Signers, so we should delete it:
+  delete decoded.Signers
+
+  // - If SigningPubKey was not in the original tx, then we should delete it:
+  const parsedTxJSON = JSON.parse(txJSON)
+  if (!parsedTxJSON.SigningPubKey) {
+    delete decoded.SigningPubKey
+  }
+
+  // - We know that the original tx did not have Signers, so if it exists, we should delete it:
+  delete decoded.Signers
+
+  if (!isEqual(decoded, parsedTxJSON)) {
+    const error = new utils.common.errors.ValidationError(
+      'Serialized transaction does not match original txJSON'
+    )
+    error.data = {
+      decoded,
+      parsedTxJSON
+    }
+    throw error
+  }
+
+  checkFee(api, decoded.Fee)
+
   return {
     signedTransaction: serialized,
     id: computeBinaryTransactionHash(serialized)
+  }
+}
+
+function checkFee(api: RippleAPI, txFee: string): void {
+  const fee = new BigNumber(txFee)
+  const maxFeeDrops = xrpToDrops(api._maxFeeXRP)
+  if (fee.greaterThan(maxFeeDrops)) {
+    throw new utils.common.errors.ValidationError(
+      `"Fee" should not exceed "${maxFeeDrops}". ` +
+      'To use a higher fee, set `maxFeeXRP` in the RippleAPI constructor.'
+    )
   }
 }
 
--- a/src/transaction/sign.ts
+++ b/src/transaction/sign.ts
@@ -1,7 +1,7 @@
 import * as isEqual from '../common/js/lodash.isequal'
 import * as utils from './utils'
 import keypairs = require('ripple-keypairs')
-import binary = require('ripple-binary-codec')
+import binaryCodec = require('ripple-binary-codec')
 import {computeBinaryTransactionHash} from 'ripple-hashes'
 import {SignOptions, KeyPair} from './types'
 import {BigNumber} from 'bignumber.js'
@@ -11,8 +11,8 @@ const validate = utils.common.validate
 
 function computeSignature(tx: object, privateKey: string, signAs?: string) {
   const signingData = signAs
-    ? binary.encodeForMultisigning(tx, signAs)
-    : binary.encodeForSigning(tx)
+    ? binaryCodec.encodeForMultisigning(tx, signAs)
+    : binaryCodec.encodeForSigning(tx)
   return keypairs.sign(signingData, privateKey)
 }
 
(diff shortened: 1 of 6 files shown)


src/transaction/sign.ts
```diff
@@ -55,12 +49,61 @@ function signWithKeypair(
   }
 
   const serialized = binary.encode(tx)
+
+  // Decode the serialized transaction:
+  const decoded = binary.decode(serialized)
+
+  // ...And ensure it is equal to the original tx, except:
+  // - It must have a TxnSignature or txSigners (multisign).
```
mDuo13: What about flags? The "standard" way of serializing a transaction usually adds tfFullyCanonicalSig even if the user didn't specify it. I could see that feature accidentally tripping this check.

intelliot: That's a great question. I think that if the user didn't specify `tfFullyCanonicalSig` then ripple-lib's `sign` method does not add it. However, the `prepare*` methods -- including `prepareTransaction` -- do: https://github.com/ripple/ripple-lib/blob/develop/src/transaction/utils.ts#L67

mDuo13: I thought if you left `Flags` off entirely it _did_ provide it.

intelliot:
As far as I'm currently aware, the only case that this will catch occurs when an amount of XRP drops contains a decimal point; and this will be caught ahead of time with ripple-binary-codec 0.2.2. (This PR also updates ripple-lib to use ripple-binary-codec 0.2.2.) This PR incorporates the secondary check as "defense in depth". Here's how I tested it:

1. Created unit tests.
2. I verified that the unit tests pass with the secondary check alone (using the older ripple-binary-codec).
3. I removed the secondary check and verified that the tests fail.
4. I updated ripple-binary-codec and verified that the tests pass (with a small change to the tests: the error message that the test expects).
5. I added the secondary check back and verified that the tests still pass.

I also verified that the decoded transaction's Fee is checked (to ensure it's not higher than maxFeeXRP) by temporarily commenting out the first Fee check (which checks the original txJSON prior to signing/encoding).

src/transaction/sign.ts
intelliot: - I've verified that signing a transaction with ripple-lib's `sign` method (which also serializes it) does not add tfFullyCanonicalSig, regardless of whether the user specified Flags or not. But note that the `prepare*` methods do add it.
- I've added a test that ensures that no Flags are added: https://github.com/ripple/ripple-lib/pull/1026/commits/c5714deddcebd496e9f248e7fee5ab1f697d2320

ximinez: I think that despite the `prepare*` methods, the `sign` method should also add `Flags` if none are specified, mainly because our documentation says we do: https://xrpl.org/transaction-malleability.html

> To further protect users, Ripple has configured its code to enable the `tfFullyCanonicalSig` flag by default where possible. Ripple strongly encourages third-party implementations of XRP Ledger software to generate only fully-canonical signatures, and enable tfFullyCanonicalSig on transactions by default.

src/transaction/sign.ts
```diff
@@ -1,3 +1,4 @@
+import * as isEqual from '../common/js/lodash.isequal'
 import * as utils from './utils'
 import keypairs = require('ripple-keypairs')
 import binary = require('ripple-binary-codec')
```
keefertaylor: optional: Consider rebasing this import to binaryCodec because it makes the code below easier to read IMO:

```javascript
const decoded = binary.decode(serialized)
```
vs
```javascript
const decoded = binaryCodec.decode(serialized)
```

src/transaction/sign.ts
```diff
@@ -55,12 +49,61 @@ function signWithKeypair(
...
   const serialized = binary.encode(tx)
+
+  // Decode the serialized transaction:
+  const decoded = binary.decode(serialized)
+
+  // ...And ensure it is equal to the original tx, except:
+  // - It must have a TxnSignature or txSigners (multisign).
+  if (!decoded.TxnSignature && !tx.Signers) {
+    throw new utils.common.errors.ValidationError(
+      'Serialized signed transaction is missing "TxnSignature" property'
```
keefertaylor: Consider rephrasing to line up with the check you're actually performing:
"Serialized transaction must have a TxnSignature or Signers property."

src/transaction/sign.ts
keefertaylor: nit: Refer to this property the same way you refer to it in code: 'Signers' vs txSigners, or at the very least 'tx.Signers'.

Or, if you think it should be called txSigners, you could refactor the code to this property name.

src/transaction/sign.ts
```diff
@@ -55,12 +49,61 @@ function signWithKeypair(
...
+  delete decoded.Signers
+
+  // - If SigningPubKey was not in the original tx, then we should delete it:
+  const parsedTxJSON = JSON.parse(txJSON)
+  if (!parsedTxJSON.SigningPubKey) {
+    delete decoded.SigningPubKey
+  }
+
+  // - We know that the original tx did not have Signers, so if it exists, we should delete it:
+  delete decoded.Signers
```
keefertaylor: You've already deleted this on line 66.

src/transaction/sign.ts
```diff
@@ -55,12 +49,61 @@ function signWithKeypair(
...
+
+  checkFee(api, decoded.Fee)
+
   return {
     signedTransaction: serialized,
     id: computeBinaryTransactionHash(serialized)
   }
 }
 
+function checkFee(api: RippleAPI, txFee: string): void {
```
keefertaylor: Nit: Please document this function.

test/api-test.js
```diff
@@ -2608,6 +2608,115 @@ describe('RippleAPI', function () {
     assert.deepEqual(signature, responses.sign.signAs);
   });
 
+  it('sign - succeeds - prepared payment', async function () {
```
keefertaylor: I think there's a few code paths that aren't fully unit tested. Can you add tests for them or explain why they're infeasible.
- Test missing signature / signers param (sign.ts:58)
- Test the new call to `checkFee()` works correctly (sign.ts:88). 

Aside, it would be great to get code coverage reporting hooked up to a CI on this library to make this easy for reviewers to visualize. Are there blockers to doing so (If not, I'm happy to help out).

keefertaylor:
Some initial (mostly stylistic) comments.

src/transaction/sign.ts
intelliot: @ximinez I think that's beyond the scope of this PR, but I've created a new PR that does that: https://github.com/ripple/ripple-lib/pull/1028

test/api-test.js
intelliot: It _should_ be impossible for the signature / signers param to be missing; the check is there "just in case".

The same goes for the new call to checkFee; the fee is already checked prior to that, and the decoded transaction must be equal to the original (aside from the fields added during signing). I'll remove this redundant check.

[Code coverage reporting](https://github.com/ripple/ripple-lib/blob/develop/.nycrc) and [CI](https://github.com/ripple/ripple-lib/blob/develop/circle.yml) have been set up for this library at various times in the past several years, but they are currently broken (or in need of updates). There are no blockers. Your help is welcome!

src/transaction/sign.ts
keefertaylor: Not done

src/transaction/sign.ts
keefertaylor: Not done

src/transaction/sign.ts
keefertaylor: not done

test/api-test.js
keefertaylor: Acknowledged. Lets sync offline about getting code coverage and CI set up.

intelliot:
Thanks to @mDuo13 for taking a look! I've also done some additional manual testing myself, and everything looks good.

jnr101:
Hi,

I am using values with standard amount of digits after the '.' like e.g. "3.141000". It turns out that this fails the verification. It was easy for me to fix, but if you feel like values with zeros should be accepted, then you might want to change it.

The following code will show the issue:

```typescript
const {RippleAPI} = require('ripple-lib')

api = new RippleAPI({ server: 'wss://s1.ripple.com' });

const address = "rrrrrrrrrrrrrrrrrNAMEtxvNvQ"
const secret = "shNs3GgHxoNPRWL6chhwH5s5A97T5"

const order = {
  "direction": "sell",
  "quantity": {
    "currency": "USD",
    "counterparty": 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B',
    "value": "3.140000"
  },
  "totalPrice": {
    "currency": "XRP",
    "value": "31415"
  }
};

(async () => {
  await api.connect()
  const accountInfo = await api.getAccountInfo(address)
  const prepared = await api.prepareOrder(address, order, { "sequence": accountInfo.sequence } )
  const {signedTransaction} = api.sign(prepared.txJSON, secret);
  api.disconnect()
})().catch(err => {
  console.error(err)
})
```

intelliot:
@jnr101 Thanks for pointing that out! I put some thought into it and decided to keep the current behavior because it's risky (and potentially unexpected) for the encoded transaction to contain different values than were originally provided to the `sign` method. In the case of trailing zeros, you are probably doing the right thing by trimming those off prior to asking ripple-lib to sign.

jnr101:
Hi @intelliot, sounds good to me. I have run the tests and looks good 👍

yxxyun:
Screenshot_20191004-004846_Utoken
after this update,  is this right behavior ?

intelliot:
@yxxyun Correct. The underlying transaction format is in drops and thus does not allow more than 6 decimal places for XRP amounts. It would be risky for the library to silently/automatically do rounding or truncation, so apps/clients must do that prior to passing the amount to ripple-lib. Depending on the context/use case, different rounding policies are possible. Hope this helps!

yxxyun:
use ripple-lib to create offers,  the input are direction, quantity and price, not taker pay takerget, in this case,  I didn't input XRP amount,  how can I deal it?

intelliot:
@yxxyun Could you create a new issue describing the problem?

Note that you can specify `TakerGets` and `TakerPays` by using [OfferCreate JSON](https://xrpl.org/offercreate.html) with ripple-lib's [prepareTransaction method](https://xrpl.org/rippleapi-reference.html#preparetransaction).

PR #1028, opened 2019-08-09
sign method: Add tfFullyCanonicalSig

intelliot:
- tfFullyCanonicalSig will be enabled regardless of whether
  the original transaction contained Flags or not.

With this PR, the `sign` method will add tfFullyCanonicalSig unless the new `setFullyCanonicalSig` option is set to `false`.

I think this is technically a breaking change and should be called out as such in HISTORY.md.

### DIFF
--- a/src/common/schemas/input/sign.json
+++ b/src/common/schemas/input/sign.json
@@ -35,6 +35,10 @@
         "signAs": {
           "$ref": "address",
           "description": "The account that the signature should count for in multisigning."
+        },
+        "setFullyCanonicalSig": {
+          "type": "boolean",
+          "description": "If `false`, the sign method will not set the tfFullyCanonicalSig flag on the transaction. The default is `true`. [New in ripple-lib {VERSION}.]"
         }
       },
       "additionalProperties": false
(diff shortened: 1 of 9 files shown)


mDuo13:
Multi-signing introduces some additional ways that transactions can be [malleable](https://xrpl.org/transaction-malleability.html) but I think better behavior would be to always add tfFullyCanonicalSig when providing any kind of signature unless the user very explicitly opts out.

Actually, I think the best behavior would be to disqualify all non-canonical signatures entirely and then ignore the tfFullyCanonicalSig flag, but that would need an amendment.

src/common/schemas/input/sign.json
```diff
@@ -35,6 +35,10 @@
         "signAs": {
           "$ref": "address",
           "description": "The account that the signature should count for in multisigning."
+        },
+        "setFullyCanonicalSig": {
+          "type": "boolean",
+          "description": "If `false`, the sign method will not set the tfFullyCanonicalSig flag on the transaction. The default is `true`. [New in ripple-lib {VERSION}.]"
```
mDuo13: Is `{VERSION}` going to get changed when this PR is released? Maybe it should be marked with a TODO or something just to make that clear.

intelliot: Yes.

intelliot:
With the RequireFullyCanonicalSig amendment (activated July 3, 2020) I don't think this improvement is needed any longer.

PR #1029, opened 2019-08-20
Adding Fail_hard option in submit Command

AGallouin:
Hi,

I havn't seen any way to use the fail_hard option listed here https://xrpl.org/submit.html, so I'm adding it.
Should I add Something else to make it work ?

Thanks


intelliot:
The [fail_hard option is likely to be improved](https://github.com/ripple/rippled/issues/2847) soon, so I would like for that to be settled first.

Thank you for submitting this!

intelliot:
This is expected to be unblocked by https://github.com/ripple/rippled/pull/3160

intelliot:
This is unblocked and fixed by rippled 1.5.0

Issue #904, opened 2018-05-28
Address format including destination tag


intelliot:
For exchanges and hosted wallets, it would be useful to have a "packed" address format that includes and encodes the destination tag.

See https://github.com/ripple/ripple-address-codec/issues/7

MarkusTeufelberger:
There used to be this: https://wiki.ripple.com/Ripple_URIs

sublimator:
Nail it up and get the money!

MarkusTeufelberger:
What money?

sublimator:
It's an expression from the building industry. Quit the hand wringing and just get it done type thing ... Nail it up and get the money!

sublimator:
I was helping a guy who sent to a gateway with a destination tag with a single trailing digit wrong in  a September 2018 transaction ... 

Upon looking at around one and half million XRP payment transactions, yet without knowing for sure, it looks like the gateway has received millions of ill gotten XRP due to messed up destination tags ... 

That's tragic ... 

You'd be amazed at the amount of varying length slices of 123456 there was ...

sublimator:
@intelliot

intelliot:
Don’t people notice fairly quickly when they aren’t credited for their deposit?

We’ll suggest a new packed address format, though it’ll likely be a while before people actually adopt it.

nfcpasses:
In this case the funds became in dispute

mDuo13:
https://github.com/xrp-community/standards-drafts/issues/6 is a proposal for this

intelliot:
This need is met by [the X-address format (supported by ripple-address-codec)](https://github.com/ripple/ripple-address-codec).

PR #1041, opened 2019-09-28
Add support for the X-address format

intelliot:
Developers can install this version with one of the following commands, using npm or yarn respectively:
```
npm install ripple-lib@beta

yarn add ripple-lib@beta
```


intelliot:
This closes #904

FredKSchott:
@intelliot random aside based on your comment in the OP: Pika CI is something that I built so that you get an "install this PR" for every PR, instead of having to manually manage and move around a beta tag yourself. I'd recommend it for the repo: https://www.pika.dev/cicd/

docs/index.md
```diff
@@ -538,7 +550,7 @@ signers | object | *Optional* Settings that determine what sets of accounts can
 *signers.* threshold | integer | A target number for the signer weights. A multi-signature from this list is valid only if the sum weights of the signatures provided is equal or greater than this value. To delete the signers setting, use the value `0`.
 *signers.* weights | array | *Optional* Weights of signatures for each signer.
 *signers.* weights[] | object | An association of an address and a weight.
-*signers.weights[].* address | [address](#address) | A Ripple account address
+*signers.weights[].* address | [address](#address) |
```
keefertaylor: I'd consider writing 'either a classic or an X-address' here

docs/index.md
```diff
@@ -3906,7 +3918,7 @@ signers | object | *Optional* Settings that determine what sets of accounts can
...
```
keefertaylor: ditto

src/ledger/accountinfo.ts
```diff
@@ -34,6 +35,22 @@ export default async function getAccountInfo(
...
+    const {
+      classicAddress,
+      tag
+    } = xAddressToClassicAddress(address)
+    if (tag !== false) {
+      return Promise.reject('getAccountInfo does not support the use of a tag. Use an address without a tag.')
+    }
+    address = classicAddress
+  } catch (_) {
+    // `address` is already a classic address
```
keefertaylor: This syntax feels awkward. 

Your error case (the `Promise.reject` should get called here. Rather than an error case being an accepted part of execution.

src/ledger/settings.ts
```diff
@@ -38,6 +39,22 @@ export async function getSettings(
 ): Promise<FormattedSettings> {
   // 1. Validate
   validate.getSettings({address, options})
+
+  // Only support retrieving settings without a tag,
+  // since settings do not distinguish by tag.
+  try {
```
keefertaylor: This code gets repeated many times. You should refactor it into a helper.

src/api.ts
```diff
@@ -173,6 +174,24 @@ class RippleAPI extends EventEmitter {
   async request(command: string, params: any):
     Promise<any>
   async request(command: string, params: any = {}): Promise<any> {
+    if (params.account) {
```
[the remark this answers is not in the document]
keefertaylor: +1

src/offline/generate-address.ts
```diff
@@ -1,19 +1,39 @@
...
+function generateAddressAPI(options?: GenerateAddressOptions): GeneratedAddress {
   validate.generateAddress({options})
   try {
     const secret = keypairs.generateSeed(options)
     const keypair = keypairs.deriveKeypair(secret)
-    const address = keypairs.deriveAddress(keypair.publicKey)
-    return {secret, address}
+    const classicAddress = keypairs.deriveAddress(keypair.publicKey)
+    return {
+      xAddress: classicAddressToXAddress(classicAddress, false, options && options.test),
```
keefertaylor: Consider whether you want to return classicAddresses here? If we want to push folks towards X-Address maybe we leave this function as is, mark as deprecated, and add a new API for X-address only. 

Pushing more options onto clients lets them be lazy about upgrading and make the programming model more confusing.

src/ledger/accountinfo.ts
intelliot: This needs to be an accepted part of execution because the use of a classic address is allowed and supported.

keefertaylor: I would encourage you to read about why this is generally accepted as an anti pattern: https://web.archive.org/web/20140430044213/http://c2.com/cgi-bin/wiki?DontUseExceptionsForFlowControl. 

It’s relatively easy to break this into sub functions that check if it is an x-address or not.

intelliot: The reliable way to check whether it's a valid X-address is to treat it as one; if that fails, then [it is not a valid X-address](https://github.com/ripple/ripple-address-codec/blob/master/src/index.ts#L113). By the way, rippled accepts classic addresses but it also accepts other kinds of account identifiers too.

keefertaylor: You can still do this without using exceptions as a control flow handler though:
```javascript
let transactionRetrievalAddress = address;
if (isValidXAddress(address)) {
    transactionRetrievaAddress = decodeXAddress(xAddress);
}
```

You can even flip this on it's head and code everything to an X-address. It kinda depends on whether you'd like to represent your variables as X-addresses or classic addresses. You'd need two new functions: `isValidAddress`, `xAddressHasTag`, but both are trivial to implement. That would look like this:
```javascript
let xAddress = address;
if (isValidClassicAddress(address)) {
   xAddress = encodeXAddress(address, false)
} else if (xAddressHasTag(address)) {
   throw new Exception("Oops, we don't support tags.")
}
```

My understanding is we'd like to move towards X-Address only, so the second seems like a reasonable choice to me. You are closer to the metal on those sorts of strategic decisions in Ripple lib however, so I dee

intelliot:
```
let transactionRetrievalAddress = address;
if (isValidXAddress(address)) {
    transactionRetrievaAddress = decodeXAddress(xAddress);
}
```
That's fine, but now we're decoding the address two times; `isValidXAddress` decodes the X-address in order to determine whether it's valid.

We need to have the classic address in the end because that's what rippled supports.

src/offline/generate-address.ts
intelliot: Good idea -- done

src/api.ts
```diff
@@ -288,6 +290,16 @@ class RippleAPI extends EventEmitter {
     return results
   }
 
+  // @deprecated Use X-addresses instead
+  generateAddress(options?: GenerateAddressOptions): GeneratedAddress {
+    options.includeClassicAddress = true
```
keefertaylor: Generally, you want to avoid parameter mutability as well. Modifying this variable here will modify it in the enclosing scope as well, which might not be something the caller is expecting to happen. 

Do you think that a caller will ever want to manually set this option? If so, then make a copy of it before  modifying it. If not, then I would just remove this as an option on the `GenerateAddressOptions` object and pass an extra parameter into the generateAddressAPI.

src/transaction/utils.ts
```diff
@@ -56,6 +57,37 @@ function scaleValue(value, multiplier, extra = 0) {
...
+
+/**
+ * Given an address (account), get the classic account and tag.
+ *
+ * @param Account The address to parse.
+ * @returns {ClassicAccountAndTag}
+ *          The classic account and tag.
+ */
+function getClassicAccountAndTag(Account: string): {classicAccount: string, tag: number | false | undefined} {
+  try {
```
keefertaylor: Don't use try / catch to do flow control.

src/api.ts
FKSRipple: What about `return generateAddressAPI({...options, includeClassicAddress: true})`?

FKSRipple: Also, if options can be undefined you should give it a default value of empty object:

```
  generateAddress(options?: GenerateAddressOptions= {}): GeneratedAddress {
```

src/transaction/utils.ts
intelliot: What alternative would you suggest? In this case, keep in mind that we want to decode the X-address only once.

src/common/index.ts
```diff
@@ -2,13 +2,36 @@ import * as constants from './constants'
 import * as errors from './errors'
 import * as validate from './validate'
 import * as serverInfo from './serverinfo'
+import {xAddressToClassicAddress} from 'ripple-address-codec'
+
+export function ensureClassicAddress(account: string | undefined): string {
```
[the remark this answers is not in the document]
intelliot: I've cleaned this up so that we don't check for `undefined` here. The issue with the recommendation is that it would decode the X-address four times (see [the implementation of `isValidXAddress` here](https://github.com/ripple/ripple-address-codec/blob/master/src/index.ts#L112)) when it only needs to be done once.

src/transaction/utils.ts
keefertaylor: This is a strong anti pattern and there should be compelling evidence that this deviation is warranted by things that are out of our control. 

```javascript 
if (isXAddress) {
  // handle x address
} else {
  // handle classic address
}
```

I don't think that decoding is a big deal for the code clarity it buys us.  If this really bugs you, then you can replace the top `if` with 'starts with X || starts with T'.

src/api.ts
keefertaylor: +1 on these changes

src/common/index.ts
```diff
@@ -2,13 +2,34 @@ import * as constants from './constants'
...
+  // Except for special cases, X-addresses used for requests
+  // must not have an embedded tag. In other words,
+  // `tag` should be `false`.
+  try {
+    const {
+      classicAddress,
+      tag
+    } = xAddressToClassicAddress(account)
+    if (tag !== false) {
+      throw new Error('This command does not support the use of a tag. Use an address without a tag.')
```
FKSRipple: one tricky thing: this will throw inside of the try block, meaning it will be caught below and then immediately swallowed. If that's intentional, I'd say just `return account;` here.

src/transaction/payment.ts
```diff
@@ -90,7 +91,22 @@ function createPaymentTransaction(address: string, paymentArgument: Payment
...
-    throw new ValidationError('address must match payment.source.address')
+    try {
+      const {
+        classicAddress,
+        tag
+      } = xAddressToClassicAddress(address)
+      if (classicAddress !== payment.source.address ||
+        (tag !== false && tag !== payment.source.tag) ||
+        (payment.source.tag !== undefined && payment.source.tag !== tag)) {
+          throw new ValidationError('address must match payment.source; if either contains a tag, it must match.')
```
FKSRipple: ditto here, will be caught immediately. and then swallowed. What you could do is re-throw any validation errors:

```
    } catch (err) {
      if (err instanceof ValidationError) {
        throw err;
    }
    /...
```

But it might be easier to just move this part out of the try block entirely:

```
let addr;
try {
      addr = xAddressToClassicAddress(address)
} catch { ... }
// move the rest of the logic down here, now that the addr variable is set
```

src/transaction/utils.ts
```diff
@@ -68,56 +100,106 @@ function prepareTransaction(txJSON: TransactionJSON, api: RippleAPI,
...
+      newTxJSON.SourceTag = sourceTag
+    }
+  }
+
+  // Destination:
+  if (typeof txJSON.Destination === 'string') {
+    const {classicAccount: destinationAccount, tag: destinationTag} = getClassicAccountAndTag(txJSON.Destination)
+    newTxJSON.Destination = destinationAccount
+    if (destinationTag !== undefined) {
+      const transactionTypesWithDestinationTagField = ['Payment', 'CheckCreate', 'EscrowCreate', 'PaymentChannelCreate']
```
FKSRipple: nit: you could move this up to the top of a file:

```
const TRANSACTION_TYPES_WITH_DESTINATION_TAG_FIELD = ['Payment', 'CheckCreate', 'EscrowCreate', 'PaymentChannelCreate']
```

src/common/index.ts
keefertaylor: +1 this will be much simpler to spot these errors using a traditional if / else statement rather than trying to enumerate all cases. 

You are also assuming that anything that xAddressToClassicAddress threw was an error indicating that it was not a classic address, which while true today, isn't guaranteed to hold true in this library as it's code from a dependency.

src/transaction/payment.ts
```diff
@@ -85,29 +85,57 @@ function createMaximalAmount(amount: Amount): Amount {
   return _.assign({}, amount, {value: maxValue})
 }
 
+function normalizeAddress(address: string, tag: number | undefined, verification?: string):
```
keefertaylor: Please document this function - it has a fine name but the functionality is non-obvious

src/transaction/payment.ts
```diff
@@ -85,29 +85,57 @@ function createMaximalAmount(amount: Amount): Amount {
...
+      classicAddress: address,
+      tag
+    }
+  }
+
+  if (!verification) {
+    return addressClassic
+  }
+
+  let verificationClassic
```
keefertaylor: This is more or less the same logic logic as above. Maybe it makes sense to factor this to a helper? I'm not positive what verification is so it's a bit tough to say.

src/transaction/payment.ts
```diff
@@ -85,29 +85,57 @@ function createMaximalAmount(amount: Amount): Amount {
   return _.assign({}, amount, {value: maxValue})
 }
 
+function normalizeAddress(address: string, tag: number | undefined, verification?: string):
+    {classicAddress: string, tag?: number} {
+  let addressClassic
```
keefertaylor: small nit: you call everything else `classicAddress` and this is `addressClassic`. I'd value consistency unless you think that there's something I'm missing.

intelliot: This one is meant to be more like `address (classic)` or a reference to its origin (the function's `address` parameter). I'm fine with naming it classicAddress though.

keefertaylor:
LGTM!

intelliot:
Would like @FKSRipple to take another look when he has a chance

PR #1046, opened 2019-10-09
Improve getTransaction() error when tx has not been validated yet

intelliot:
The current error message, `ledger_index and LedgerSequence not found in tx`, is not helpful. The new message is `Transaction has not been validated yet; try again later` which should be accurate.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
-export {RippleAPI} from './api'
+export * from './api'
 
 export {
     FormattedTransactionType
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -1,4 +1,4 @@
-export {RippleAPI} from './api'
+export * from './api'
```
FredKSchott: was this intentional? Seems like a bigger change than the PR title conveys

intelliot: I think there was some other stuff in ripple-lib that I want to access, but you're right, that should go in a separate PR.

PR #1048, opened 2019-10-13
Update Typescript

FredKSchott:
- Updated typescript to latest
- Fixed up some typescript-y things (ex: `import from require()` is no longer really a thing)
- Removed a custom lodash.isequal implementation with the actual one from npm


FredKSchott:
I don't have merge permissions, so merge away!

PR #1049, opened 2019-10-14
replace tslint (since deprecated) with eslint

FredKSchott:


src/common/utils.ts
```diff
@@ -17,7 +17,7 @@ function dropsToXrp(drops: string | BigNumber): string {
   if (typeof drops === 'string') {
     if (!drops.match(/^-?[0-9]*\.?[0-9]*$/)) {
       throw new ValidationError(`dropsToXrp: invalid value '${drops}',` +
-        ` should be a number matching (^-?[0-9]*\.?[0-9]*$).`)
+        ` should be a number matching (^-?[0-9]*.?[0-9]*$).`)
```
intelliot: I know this is just the text of the error, but I think we want the `\` to be part of the regex here since `.` matches any character (but `\.` only matches `.`). In other words, `123.456` matches but `123a456` should not.

src/common/utils.ts
```diff
@@ -51,7 +51,7 @@ function xrpToDrops(xrp: string | BigNumber): string {
   if (typeof xrp === 'string') {
     if (!xrp.match(/^-?[0-9]*\.?[0-9]*$/)) {
       throw new ValidationError(`xrpToDrops: invalid value '${xrp}',` +
-        ` should be a number matching (^-?[0-9]*\.?[0-9]*$).`)
+        ` should be a number matching (^-?[0-9]*.?[0-9]*$).`)
```
intelliot: Same comment as above

src/common/utils.ts
FredKSchott: that's a great point, however the linter caught that `'(^-?[0-9]*\.?[0-9]*$)'` was being shown to the user as `'(^-?[0-9]*.?[0-9]*$)'` already. I'll change this to `'(^-?[0-9]*\\.?[0-9]*$)'` instead to make sure that the backslash isn't escaped (linting ftw!)

package.json
```diff
@@ -61,7 +62,7 @@
     "watch": "tsc -w",
     "prepublish": "npm run clean && npm run compile && npm run build",
     "test": "TS_NODE_PROJECT=src/tsconfig.json nyc mocha --exit",
-    "lint": "tslint -p ./",
+    "lint": "eslint src/**/*.ts",
```
keefertaylor: I believe this will exclude the test directory. You probably want `eslint **/*.ts`.

keefertaylor:
Thanks for the PR and helping us maintain this library!

I don't have a great mental map of the rules from  TsLint to EsLint, but the new config looks sane. Ripple is going to roll out their own standardized config soon (using EsLint) so I'm happy to take this for now and fix when we're ready.

PR #1050, opened 2019-10-16
api-test.js -> api-test.ts

FredKSchott:
> Note: this contains some other commits that should get removed from the diff once #1048 & #1049 get merged.

This PR moves the main api-test file from JavaScript from TS, and sets the foundation for the rest of the tests to make the change as well.


FKSRipple:
Brought in to #1051

PR #1051, opened 2019-10-16
Move tests to TypeScript

FredKSchott:
This moves the test files from JS to TS. We fall back to a few "any" types here just to not block the main move from JS to TS, and then can come back to resolve those after if they're worth it (some uses of any actually make sense, for example testing bad input that TS would disallow but a JS user might do anyway).

### DIFF
--- a/test/api-test.ts
+++ b/test/api-test.ts
@@ -1,28 +1,21 @@
-/* eslint-disable max-nested-callbacks */
-'use strict'; // eslint-disable-line
-const _ = require('lodash');
-const assert = require('assert-diff');
-const setupAPI = require('./setup-api');
-const RippleAPI = require('ripple-api').RippleAPI;
-const validate = RippleAPI._PRIVATE.validate;
-const fixtures = require('./fixtures');
-const requests = fixtures.requests;
-const responses = fixtures.responses;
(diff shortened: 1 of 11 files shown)


test/api-test.ts
```diff
@@ -4606,17 +4599,17 @@ describe('RippleAPI', function () {
...
+        if (marker !== undefined) {
           reject(new Error());
+          return;
         }
+        resolve({ marker: 'A', results: [1] });
       });
     }
     return utils.getRecursive(getter, 10).then(() => {
-      assert(false, 'Should throw Error');
+      assert(false, 'Should th row Error');
```
intelliot: typo?

tsconfig-base.json
```diff
@@ -13,10 +13,12 @@
...
     "noUnusedLocals": true,
-    "noUnusedParameters": true,
     "removeComments": true,
     "preserveConstEnums": false,
     "esModuleInterop": true,
-    "suppressImplicitAnyIndexErrors": false
+    "suppressImplicitAnyIndexErrors": false,
+    "resolveJsonModule": true,
+    "preserveSymlinks": true
+
```
intelliot: extra newline?

PR #1052, opened 2019-10-16
Convert the rest of the test helpers to TS

FredKSchott:
> ~~Note: this contains some other commits that should get removed from the diff once #1048 #1049 #1050 #1051 get merged.~~

This is good from a compilation standpoint, but I'm not sure how to run the integration tests that rely on these. @intelliot will need your help on that, if you can give some guidance.


test/integration/integration-test.ts
```diff
@@ -71,7 +68,7 @@ function testTransaction(testcase, type, lastClosedLedgerVersion, prepared,
   });
 }
 
-function setup(server = 'wss://s1.ripple.com') {
+function setup(this: any, server = 'wss://s1.ripple.com') {
```
keefertaylor: I'm not familiar with this syntax? 

`this` isn't referring to anything. If we want iVars, should we refactor into a class? If we want a global var, then we should define that and drop the use of `this`.

test/mock-rippled.ts
```diff
@@ -49,8 +48,17 @@ function createLedgerResponse(request, response) {
   return JSON.stringify(newResponse);
 }
 
-module.exports = function createMockRippled(port) {
-  const mock = new WebSocketServer({ port: port });
+interface CustomWebSocketServer extends WebSocketServer, EventEmitter2 {
+  expectedRequests: any
```
keefertaylor: Why create an interface for this? It's basically untyped anyway. 

I also don't know why it needs to extend anything.

test/mock-rippled.ts
```diff
@@ -1,17 +1,16 @@
...
+import {Server as WebSocketServer} from 'ws';
+import {EventEmitter2} from 'eventemitter2';
+import fixtures from './fixtures/rippled';
+import addresses from './fixtures/addresses.json';
+import hashes from './fixtures/hashes.json';
+import transactionsResponse from './fixtures/rippled/account-tx';
+import accountLinesResponse from './fixtures/rippled/account-lines';
+import accountObjectsResponse from './fixtures/rippled/account-objects';
+import fullLedger from './fixtures/rippled/ledger-full-38129.json';
+import { getFreePort } from './utils/net-utils';
```
keefertaylor: nit: sometimes you use `{ foo }` and sometimes `{foo}`. 

Could you change these to be consistent? I don't care about which, but we should be consistent. 

Please correct here and in other files.

test/integration/integration-test.ts
FKSRipple: yup, this is a TypeScript-ism. More info here: https://github.com/microsoft/TypeScript/wiki/'this'-in-TypeScript#specify-type-of-this-in-function-signature

I'd like to go back and replace/refactor a lot of these "any" types, but in the interest of not letting this PR sit too long I just added type definitions for the code as it existed in this PR, and then will do a follow up for any refactoring.

test/mock-rippled.ts
FKSRipple: +1, cleaned this up and added a comment

PR #1059, opened 2019-10-25
Add example of reliable transaction submission

intelliot:
Here is one example approach

To run the example:
```
yarn run compile:snippets && node ./snippets/dist/reliableTransactionSubmission
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,10 @@
     "test": "TS_NODE_PROJECT=src/tsconfig.json nyc mocha --exit",
     "lint": "eslint src/**/*.ts 'test/*-test.{ts,js}'",
     "perf": "./scripts/perf_test.sh",
-    "start": "node scripts/http.js"
+    "start": "node scripts/http.js",
+    "compile:snippets": "tsc -p snippets/tsconfig.json",
+    "start:snippet": "npm run compile:snippets && node ./snippets/dist/start.js",
+    "inspect:snippet": "npm run compile:snippets && node inspect ./snippets/dist/start.js"
   },
   "repository": {
     "type": "git",
--- a/snippets/src/reliableTransactionSubmission.ts
+++ b/snippets/src/reliableTransactionSubmission.ts
@@ -10,6 +10,7 @@ import https = require('https')
  * 1) Transaction preparation:
  *    - How do we decide which account sequence and LastLedgerSequence numbers to use?
  *      (To prevent unintentional duplicate transactions, an {account, account_sequence} pair can be used as a transaction's idempotency key)
+ *    - How do we decide how much to pay for the transaction fee? (If our transactions have been failing due to low fee, we should consider increasing this value)
  * 2) Transaction status retrieval. Options include:
  *    - Poll for transaction status:
  *      - On a regular interval (e.g. every 3-5 seconds), or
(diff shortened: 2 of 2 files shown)


intelliot:
Sorry, I forgot to add package.json. Will do that when I get back to my desk later

keefertaylor:
What's the background here? Is this meant to be production code? or just a sample?

snippets/src/reliableTransactionSubmission.ts
```diff
@@ -0,0 +1,200 @@
...
+    }, {
+      sequence
+    })
+    const signed = api.sign(preparedPayment.txJSON, payment.source.secret)
+    finalResults.push({
+      id: signed.id
+    })
+    const result = await api.submit(signed.signedTransaction)
+
+    // Most of the time we'll get 'tesSUCCESS' or (after many submissions) 'terQUEUED'
```
keefertaylor: Why do we get different responses back?

snippets/src/reliableTransactionSubmission.ts
```diff
@@ -0,0 +1,200 @@
...
+          value: payment.amount_drops,
+          currency: 'drops'
+        }
+      }
+    }, {
+      sequence
+    })
+    const signed = api.sign(preparedPayment.txJSON, payment.source.secret)
+    finalResults.push({
+      id: signed.id
```
keefertaylor: this is really a txHash right?

snippets/src/reliableTransactionSubmission.ts
```diff
@@ -0,0 +1,200 @@
...
+        try {
+          status = await api.getTransaction(signed.id)
+        } catch (e) {
+          // Typical error when the tx hasn't been validated yet:
+          if (e.name !== 'MissingLedgerHistoryError') {
+            console.log(e)
+          }
+
+          if (event.ledger_index > preparedPayment.instructions.maxLedgerVersion + 3) {
+            // Assumptions:
```
keefertaylor: It's probably worthwhile to state these assumptions at the top rather than burying them down here. They all sound reasonable however.

intelliot:
This is just a sample

snippets/src/reliableTransactionSubmission.ts
intelliot: It's theoretically possible to get any of [the transaction result codes](https://xrpl.org/transaction-results.html) back.

When implementing transaction submission at scale, one of the factors we should consider is how quickly we want to process transactions. If we want to process a lot of transactions very quickly, it's possible that we could fill the open ledger. The open ledger may also be full due to timing, or heavy transaction volume (from other users) on the network in general. At these times, our transaction may be queued and we'll get the `terQUEUED` result back.

snippets/src/reliableTransactionSubmission.ts
intelliot: yes

keefertaylor: Got it. I'd consider renaming it, but no strong opinions since you have more contextual knowledge.

snippets/src/reliableTransactionSubmission.ts
keefertaylor: ah! got it. Makes sense, thanks for the explanation.

keefertaylor:
LGTM!

RobertEmmettJones (approved):

package.json
```diff
@@ -64,7 +64,10 @@
     "test": "TS_NODE_PROJECT=src/tsconfig.json nyc mocha --exit",
     "lint": "eslint src/**/*.ts 'test/*-test.{ts,js}'",
     "perf": "./scripts/perf_test.sh",
-    "start": "node scripts/http.js"
```
RobertEmmettJones, line 67: 8reliable-tx-submission

RobertEmmettJones (approved):

package.json
RobertEmmettJones: package.json

RobertEmmettJones (approved):
Approved

PR #1060, opened 2019-10-25
fix broken tests

FKSRipple:
Tests were broken in some past PRs.  

@intelliot lets get either Travis or Circle set back up for the repo, so that we can get tests running per-PR. Let me know if I can help with that (I don't have settings access).

Fixes:
1. Connection timeout wasn't being cleared properly / in all cases.
2. Tests were testing outdated error messages


intelliot:
Thanks. I think Travis is the way to go. Will look into it later

PR #1061, opened 2019-10-25
Update Webpack

FKSRipple:
supercedes #1005

### Gulp + Webpack -> Webpack

Gulp is powerful but ultimately flawed (streams don't propagate errors well as well, see #1005). Webpack's CLI can actually do everything that we need, without the extra layer of gulp-plexity. This PR primarily moves our webpack config out of the custom Gulp file and into a more standard webpack.config.js with better logging & error messages.

Builds are also now 3x faster, down from ~30 seconds to ~10 seconds.

### Add Analysis Support

To test that the new webpack build was working properly, I added an optional bundle analysis step. Pass an `--analyze` flag to the build command to generate a visual, explorable analysis. This matches what we already see on develop, but I'll spin off some other tickets to try and reduce the size of this (duplicate bignum libraries, elliptic libraries, etc).

Screen Shot 2019-10-25 at 9 20 39 AM

### Updates Webpack, related deps, and removes a bunch of now-unused deps

All webpack deps are up-to-date, and a bunch of now-unnecessary deps were removed.

### Cleans up webpack build artifacts

I couldn't see why we needed so many builds, so I cleaned this up to just create a `ripple-lib.development.js` and `ripple-lib.production.min.js` and remove the duplicates. This is a naming scheme made popular by React and some popular webpack build tooling (which lets you add extra debug logs to the dev build) but I'm happy to bikeshed on whatever names we go with, or to just leave them as is (`ripple-latest.js` and `ripple-latest.min.js`). @intelliot let me know your thoughts.

### Cleans up build-related commands

Happy to bikeshed on this as well, but as a contributor I always get "build" and "compile" confused. This cleans up the naming scheme to just "build" & "build:*" for specific parts of the build. Similarly,  I'm happy to bikeshed any naming here or even walk this out if people disagree with the direction (although based on #1059 we're already using this naming scheme)


FKSRipple:
@intelliot I'd love your review on the "Cleans up webpack build artifacts" section specifically, since it affects the names of the final build artifacts.

intelliot:
I’m fine either way. The names were in place before I started on the project. I’d lean toward keeping the current names but only due to inertia.

FKSRipple:
Woo! This improves our `yarn install` time considerably as well.

Issue #1070, opened 2019-11-01
Typescript target changed


tylerlevine:
Hello,

I'm the lead dev for [BitGoJS](https://github.com/BitGo/BitGoJS), where we depend on ripple-lib extensively for our support for XRP.

Recently, a vulnerability in https-proxy-agent was disclosed (see #1054). We're now updating to ripple-lib@1.4.0. to pick up the fix, but in commit https://github.com/ripple/ripple-lib/commit/fadfd4e06c1ebbb71ed71efe6adbf566ddbd61e5 the typescript compile target was changed from es6 to es2017. This causes tsc to emit native async code, which breaks in node 6, which we are still interested in supporting. (see https://github.com/BitGo/BitGoJS/pull/585 and https://cloud.drone.io/BitGo/BitGoJS/2830/2/4)

Is there any possibility of re-targeting es6? How come the target was changed from es6 to es2017 without a major version bump?

FKSRipple:
Since Node v6 is no longer supported by Node.js (End-of-life'd back in April) we didn't consider it when updating. I'm sure you have your reasons, but I'd *highly* recommend updating to Node v8/v10 asap, to keep getting security updates and fixes from the Node.js team. They stopped  backporting major system security fixes when v6 hit end-of-life. Amazon has more info: https://aws.amazon.com/blogs/developer/node-js-6-is-approaching-end-of-life-upgrade-your-aws-lambda-functions-to-the-node-js-10-lts/

@intelliot I think it would make sense to either:
- tag a special v1.4.0-node6 version just for BitGo/others which is the current code but compiled for older Node
- compile down to es6 for everyone. The async transform that we're adding back is heavy, so everyone gets more code and slower code, but it would also fix the node v6 use case.

intelliot:
I’m leaning toward the latter (compile down to es6 for everyone) since we’re not intending a major version bump at the moment. I assume the size/performance difference won’t make a difference to most apps especially since we’ve gotten this far with it. Any concerns?

FKSRipple:
```
// ES2017
417,922 bytes (741 KB on disk) for 123 items
// ES6
430,071 bytes (745 KB on disk) for 124 items
```

Huh., async/over-transpiled code makes up much less of this library than I had expected. Given that, I'm fine reverting this change back to ES6 for a little while longer, to give any Node v6 consumers a little more time to upgrade.

There is a perf cost as well that I haven't measured, but that I'd guess is a few ms added to each request.

[...]

PR #1071, opened 2019-11-01
Change typescript compile target to es6

tylerlevine:
See #1070

### DIFF
--- a/tsconfig-base.json
+++ b/tsconfig-base.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "pretty": true,
-    "target": "es2017",
+    "target": "es6",
     "module": "commonjs",
     "moduleResolution": "node",
 


FKSRipple:
There are significant performance/size costs to transpiling async code in es6, so this won't be something that we handle in the main branch. Let's continue to discuss in #1070 and see how we can get you a working snapshot for Node v6.

intelliot:
Yes, I'll work on this now. Thanks!

intelliot:
@tylerlevine I don't understand the `NodeJS Version Support Policy` [here](https://github.com/BitGo/BitGoJS#nodejs-version-support-policy). The first line says, "We only support Long-Term Support versions of Node starting with Node 8.9.0 (LTS)." But later, it says "our continuous integration setup runs the full test suite on the latest release of the following versions of node: 6 [...]" Is that not a contradiction?

tylerlevine:
Sure, I can explain a bit more. We established that policy on 7/19 of this year because we had (and still have) some users running unsupported node versions. We can't force people to upgrade, but new users should be aware which versions are supported and which are not.

We don't officially support node 6, but we want to be aware of breakage on that platform, so we still run tests there. We recommend to all customers that they should update, but we'd like to avoid breaking these users where possible. In my opinion, targeting es6 instead of es2017 is not so cumbersome that it's worth forcing these users into a tough decision: upgrade to a supported node version (not trivial for some), or use an older version of ripple-lib which has known vulnerabilities. With this one line change, we are allowing more time for these users to upgrade, and I think that's worthwhile.

Anyway, thanks for merging this PR.

PR #1072, opened 2019-11-02
start running x-address tests

FKSRipple:
Looks like this test file stopped being run at some point during the development of #1041, since we'd made the change to TS and this file was still in JS format.

This PR just does the work to get it running again, converted to TS, and a few small bugs fixed, **but many tests in the file are failing.** @intelliot I'm happy to take a look at fixing them, but figured I'd ask first since you'd have more context.


test/x-address-api-test.ts
```diff
@@ -3371,7 +3366,7 @@ describe('X-address Usage', function () {
             taker: address
           })
         ]
-      ).then((directOfferResults, reverseOfferResults) => {
+      ).then(([directOfferResults, reverseOfferResults]) => {
```
keefertaylor: How did this API  move from two parameters to an array of one parameter? Was there an underlying change somewhere that went in without this file being updated? Did this ever work?

FKSRipple: Yea, this was a bug caught by TS. Promises never return more than one value, so `(directOfferResults, reverseOfferResults)` was a typo where the author meant to pass a single array of the two values.

Looks like this never worked, and isn't currently being run in the test suite. 

I'll continue with the work to fix the tests, now that we have them running and failing in this branch.

keefertaylor: SGTM - thanks for the explanation.

FKSRipple:
Update: So it looks like `test/x-address-api-test.js` is mainly copy-pasted  from `api-test.ts`, I'd guess to get around the fact that the non-x addresses are built into the fixtures themselves. This is a 4000+ line test file, so I think copy-pasting probably isn't the best way to get what we want here.

Is everyone okay with me moving the x-address specific things into the main `api-test.ts` file, so that we're at least back to tests passing? (remember, `test/x-address-api-test.js` isn't currently running).

@intelliot what's the deprecation plan/timeline like for address -> xaddress? Quick enough that we can just replace them all in our fixture files?

intelliot:
I don’t expect classic addresses to be deprecated anytime soon. If it happens, it’ll be more than a year from now. (There’s a large existing ecosystem and it’ll take time for everyone to update.) We need a way to basically do all of the tests with X-addresses and make sure everything works as expected. Copy pasting and modifying the file worked well initially, but I’m definitely open to alternative approaches. We just need to make sure classic address and X-address use are both well-tested.

FKSRipple:
Yea I tried to keep it just for the short term, but enough tests were broken over the 4000 lines, in ways that I couldn't quickly solve, that I figured it wasn't worth spending the time until we'd had a chance to talk.

@intelliot are we sure that we need every test to run on each type of address? Could we get away with moving a few fixtures over to x-address style, and trusting that if those worked then the library was handling them properly throughout?

Otherwise, I'd probably recommend that we look into cleaning up the fixtures system to support a larger set of requests & responses, with support for testing the two different types of addresses.

intelliot:
I’m not certain. I know there were quite a few cases that the tests were able to catch, so it was definitely useful to have many of these tests. I would lean toward improving the fixtures system, but open to other options.

FKSRipple:
I'm still curious how the tests ever ran on your machine, given the issues found in https://github.com/ripple/ripple-lib/pull/1072/commits/067fd756fb64d9278b30cec1b3fa03034748dad6. I still haven't been able to get the tests running on my machine (~30 or so broken tests, iirc).

Okay, so I'd recommend merging this PR to get a few of the tests that you added running (directly covering the added methods) and then in a follow-up PR I can look into bringing X-Address coverage into the indirectly-related request/response tests. Does that work?

intelliot:
@FKSRipple If you checkout the `master` branch right now (or `1.4.0`) and run `yarn test`, doesn't it run the tests? Seems to work for me. 🤷‍♂️ I haven't looked into it too closely, but the tests seemed to pass back then.

FKSRipple:
Closing in favor of a PR that will tackle this + #1081 together

RareData:
@intelliot regarding:
> I don’t expect classic addresses to be deprecated anytime soon. If it happens, it’ll be more than a year from now. (There’s a large existing ecosystem and it’ll take time for everyone to update.) [...]

For things like hardware wallets, do you envision the devices will have to encode all classical addresses to X-addresses before presenting things like `destination` and `issuer`?

Issue #1063, opened 2019-10-25
Update dependencies


FKSRipple:
Some lowhanging fruit here, which should help with the version mismatches in #1062 

```
Package            Current Wanted  Latest  Package Type    URL                                                    
@types/node        11.13.0 11.13.0 12.11.7 devDependencies https://github.com/DefinitelyTyped/DefinitelyTyped.git 
@types/ws          3.2.1   3.2.1   6.0.3   dependencies    https://github.com/DefinitelyTyped/DefinitelyTyped.git 
assert-diff        1.2.6   1.2.6   2.0.3   devDependencies https://github.com/pihvi/assert-diff#readme            
bignumber.js       4.1.0   4.1.0   9.0.0   dependencies    https://github.com/MikeMcl/bignumber.js#readme         
doctoc             0.15.0  0.15.0  1.4.0   devDependencies https://github.com/thlorenz/doctoc#readme              
eventemitter2      0.4.14  0.4.14  5.0.1   devDependencies https://github.com/hij1nx/EventEmitter2#readme         
jsonschema         1.2.2   1.2.2   1.2.4   dependencies    https://github.com/tdegrunt/jsonschema#readme          
mocha              6.2.0   6.2.0   6.2.2   devDependencies https://mochajs.org/                                   
ripple-keypairs    0.10.2  0.10.2  0.11.0  dependencies    https://github.com/ripple/ripple-keypairs#readme       
source-map-support 0.5.12  0.5.12  0.5.13  devDependencies https://github.com/evanw/node-source-map-support#readme
ws                 3.3.3   3.3.3   7.2.0   dependencies    https://github.com/websockets/ws                      
```

FKSRipple:
Done in #1073 #1074 #1075

PR #1073, opened 2019-11-02
update ws dependency

FKSRipple:
This is a part of #1063, tests & TypeScript are both passing

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -140,6 +140,12 @@ class RippleAPI extends EventEmitter {
         this.emit('connected')
       })
       this.connection.on('disconnected', code => {
+        // This is a backwards-compatible fix for this change in the ws library: 
+        //   https://github.com/websockets/ws/issues/1257
+        // TODO: Remove in next major, breaking version
+        if (code === 1005) {
+          code = 1000;
+        }
         this.emit('disconnected', code)
       })
     } else {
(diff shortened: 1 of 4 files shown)


src/api.ts
```diff
@@ -140,6 +140,12 @@ class RippleAPI extends EventEmitter {
         this.emit('connected')
       })
       this.connection.on('disconnected', code => {
+        // This is a backwards-compatible fix for this change in the ws library: 
+        //   https://github.com/websockets/ws/issues/1257
+        // TODO: Remove in next major, breaking version
+        if (code === 1005) {
```
keefertaylor: Don't mutate parameter inputs. Create a new variable, `normalizedCode`. This helps prevent bugs when future programmers don't realize `code` was mutated. 

I know this example is super simple but it's best practice.

----

Optionally, you can do this in a one liner:
```javascript
const nomralizedCode = code !== 1005 ? code : 1000
```

FKSRipple: Big +1 to the sentiment. I was balancing that against the desire to make sure that this would be easy to remove in the next major version (deleting lines vs. refactoring lines). 

Given that this is pretty self contained, do you still feel strongly? Happy to make the change if so, just wanted to give the missing context :)

keefertaylor: I do feel strongly - I've watched people get burned on this too many times.

FKSRipple: 👍 sgtm, made the change

PR #1077, opened 2019-11-05
export some types those could use in third project

yin-june:
1. export some useful types
2. update the interface `FormattedSubmitResponse `
3. format style

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,8 +1,8 @@
-export {RippleAPI} from './api'
+export * from './api'
 
-export {
-    FormattedTransactionType
-} from './transaction/types'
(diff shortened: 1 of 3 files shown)


FKSRipple:
From a maintenance perspective, this change adds a lot more to our "public" interface which will cause more breaking changes in the future. With this PR, any change in export from `./api`, `./transaction/types`, `./broadcast` & `./common` would have a chance to break users.

If you want to accept the risk that things may change internally at any time, you can require these files yourself via `require('ripple-lib/dist/npm/broadcast')`, etc.

Or, if you think anything deserves to be in the public API that isn't today, we can talk about adding them individually.

PR #1083, opened 2019-11-08
Add support for WalletLocator

RareData:
Being able to fetch the WalletLocator field opens up new use-cases. The field can store for example a transaction hash, referencing a transaction with memos containing useful data.


FKSRipple:
lgtm, but can you update tests to passing? Will leave for @intelliot or @keefertaylor to approve

RareData:
Let me know if you want me to add additional tests.

keefertaylor:
This looks reasonable to me, but we're un-deprecating a deprecated field. Do we know why this field was deprecated to begin with? Is there background here that I'm not familiar with?

RareData:
I'd argue that it's unused, but not deprecated. The XRPL supports it, so I see no reason neglecting to implement support for it in ripple-lib. To me, it's equivalent to `tickSize`, which is also a field ripple-lib currently lacks support for.

RareData:
However, it'd be good to know if there's any plan to drop support for the WalletLocator field in rippled.

nbougalis:
There's no plan to remove the field currently (turns out that adding features is the easy part, removing them is what's hard!). If you want to add support for and document this, that's fine.

PR #1085, opened 2019-11-10
Fix Node.js Testing Versions

FKSRipple:
Updated our testing matrix based on: https://nodejs.org/en/about/releases/

- Node 11 is no longer supported (not LTS)
- Node 12 added (active LTS)
- Node 13 added ("current" release)


keefertaylor:
Thank you!

PR #1086, opened 2019-11-10
Add RippleAPI test runner, with support for multiple address testing

FKSRipple:
Screen Shot 2019-11-10 at 1 27 12 PM

### Background 
"test/api-test.ts" had hit 4000+ lines of test code and 300+ individual tests. Additionally, a new address format was added which forced us to copy-paste/duplicate the test file to test both the old forms of address. This added a non-trivial maintenance cost to maintain two near-identical files going forward.

### RippleAPI Test Runner
This test runner allows us to split our tests by RippleAPI method, and automatically load, validate, and run them. Each tests accepts arguments to test with, which allows us to re-run tests across different data (ex: different address styles).

This PR moves a few methods over to the new system, but leaves the rest for a future PR (wanted to get feedback on the direction before going any further).

### Additional benefits:
  - Throw errors when we detect the absence of tests.
  - Type the API object under test and catch typing issues (currently untyped in tests).
  - Starts to break up the 4000 line  test file that we had, which was harder to navigate and grok what was being tested.
  - Sets the stage for more cleanup, like moving test-specific fixtures closer to their tests.


package.json
```diff
@@ -44,7 +44,6 @@
     "mocha": "6.2.0",
     "mocha-junit-reporter": "^1.9.1",
     "nyc": "^14.1.1",
-    "source-map-support": "0.5.12",
```
FKSRipple: confirmed that this is not needed / is already connected automatically by ts-node

src/common/utils.ts
```diff
@@ -13,7 +13,7 @@ function isValidSecret(secret: string): boolean {
   }
 }
 
-function dropsToXrp(drops: string | BigNumber): string {
+function dropsToXrp(drops: BigNumber.Value): string {
```
FKSRipple: New tests that used a number argument caught this type mismatch. A number is valid in the implementation, although we could still intentionally limit this to string in the typings. 

I'd lean towards making this change, since If a user has a number we're unnecessarily asking them to do more work with something like `String(number)` before passing it to us.

keefertaylor: I'm confused. 

You've dropped the string argument, which makes this method *less* flexible (not more, as your comment seems to say). 

If you do this, you can remove the `if` statement on line 17. 

Also, this method will be a lot more simple if you just use BigInt, which I think is supported as of Node 10.4. Do you really need to support Node 8?

keefertaylor: Also, what is `BigNumber.Value` vs `BigNumber`? I'm unfamiliar with the notation.

test/api/getLedger/index.ts
```diff
@@ -0,0 +1,78 @@
+import assert from 'assert-diff'
+import { assertResult, TestSuite } from '../utils'
+import responses from '../../fixtures/responses'
+const { getLedger: RESPONSE_FIXTURES } = responses
+
+export default {
+  'simple test': async api => {
```
keefertaylor: Would these test cases become more indexable if you used an enum rather than a magic string?  What about an array that you could just iterate over?

test/api/getLedger/index.ts
```diff
@@ -0,0 +1,78 @@
+import assert from 'assert-diff'
+import { assertResult, TestSuite } from '../utils'
+import responses from '../../fixtures/responses'
+const { getLedger: RESPONSE_FIXTURES } = responses
+
+export default {
```
keefertaylor: It wasn't immediately obvious what this file is or where `api` was getting passed in from. Do you think it would be clearer to add a link to the TestRunner here?

test/api/utils.ts
```diff
@@ -0,0 +1,87 @@
+import _ from 'lodash'
+import { RippleAPI } from 'ripple-api'
+import assert from 'assert-diff'
+const { schemaValidator } = RippleAPI._PRIVATE
+
+export type TestFn = (
```
keefertaylor: How does this typing work? Some of your tests take in a single `typescript` argument. I would think that typescript would yell at you that you're missing the `address` parameter.

Should we mark `address: string` as optional `address?: string?`

src/common/utils.ts
FKSRipple: Sorry, I should have clarified the change in my comment:
`string | BigNumberInstance`: What we have today. Accepts a string or an already-created BigNumber object, but not a number.
`BigNumber.Value`: An alias to `string | number | BigNumberInstance`, which is what we accept today but PLUS a number. If you're using JavaScript you can pass a number today, and BigNumber supports this, but our types today artificially limit this. This change removes that artificial limit.

test/api/getLedger/index.ts
FKSRipple: For sure, some better docs/comments here would make sense.

test/api/getLedger/index.ts
FKSRipple: Can you explain a bit more about what you mean by more indexable? I'd say that this was designed primarily for the user experience flow and was designed to match what it feels like to write tests in mocha (string -> test function). My worry is that an array would feel odd as a test writer/maintainer.

test/api/utils.ts
FKSRipple: By `typescript` argument do you mean the `api` argument? If so, you can think of it like this:

1. We tell TypeScript that this kind function is called with `api` & `address` arguments.
2. When we implement this kind of function, we can choose to ignore arguments if we don't need them. They'll still get passed, and you couldn't expect a *different* kind of argument (a number vs. a string) but ignoring them is fine.

src/common/utils.ts
keefertaylor: Ah, got it. Thanks for explaining.

test/api/utils.ts
keefertaylor: I understand that javascript can ignore arguments. According to this though, you'd need to mark an argument as optional to get it to compile: https://www.typescriptlang.org/docs/handbook/functions.html#optional-and-default-parameters

test/api/getLedger/index.ts
keefertaylor: I guess I assumed you may share these magic strings across tests, which might  make them better as an emum.

As a programmer, I'd still expect this class to be an array (of objects which contained a title, and a test function) but I'm fine letting this go in if you think it's idiomatic for javascript.

test/api/utils.ts
FKSRipple: That's right, but It only works in the other direction. Think of it this way:

- I am a function that takes `(a, b, c)`. If you pass me a function of type `(a, b)` I will get mad, since you're not giving me the `c` arg I expected.
- I am a function that takes `(a, b, c)`. If you pass me a function of type `(a, b, c, d)` that is fine, since you're passing me all the args that I ask for. You're passing me a `d` arg that I didn't ask for, but since you matched my `a,b,c` type signature I don't care enough to throw an error.

test/api/getLedger/index.ts
FKSRipple: Sounds good 👍 Happy to revisit later as well if we think there's a better way

PR #1087, opened 2019-11-10
Add a prettier config

FKSRipple:
Does anyone mind if we add a .prettierrc file to the repo? If you have the either the VSCode plugin setup or the CLI tool installed, you can just run `> Format Document` and get automatic formatting. You can also configure it to auto-format on save. 

I use this in all of my projects at this point and love it. I'd recommend we set it up for the project (`yarn run format` to format the whole repo), but first just adding this config file only for the people who want it would be a good first step.

demo

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,9 +1,9 @@
 {
-    "trailingComma": "es5",
-    "printWidth": 80,
-    "tabWidth": 2,
-    "semi": false,
-    "singleQuote": true,
-    "parser": "typescript",
-    "quoteProps": "consistent"
+  "parser": "typescript",
+  "printWidth": 80,
+  "tabWidth": 2,
+  "semi": false,
+  "singleQuote": true,
+  "trailingComma": "none",
+  "quoteProps": "consistent"
   }
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,9 @@
+{
+    "trailingComma": "es5",
+    "printWidth": 80,
+    "tabWidth": 2,
+    "semi": false,
+    "singleQuote": true,
+    "parser": "typescript",
+    "quoteProps": "consistent"
+  }


keefertaylor:
This is great. 

@Stormtv wanted to standardize these across Ripple and Xpring. Adding him here to review. 

If we're all happy with this prettier config then I can adopt globally across the Xpring SDK.

.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
+    "trailingComma": "es5",
+    "printWidth": 80,
+    "tabWidth": 2,
```
Stormtv: These are defaults is there a reason of explicitly defining them?

Stormtv:
Looks good, on the projects I am working on we integrate prettier into our linter.

https://prettier.io/docs/en/integrating-with-linters.html

Such as with eslint

`"rules": { "prettier/prettier": "error" }`

`"plugins": [ "prettier", "@typescript-eslint"]`

We then setup our VSCode to auto lint + prettier on save
.vscode -> settings.json
```
{
  "eslint.alwaysShowStatus": true,
  "eslint.autoFixOnSave": true,
  "editor.tabSize": 2,
  "[javascript]": {
      "editor.defaultFormatter": "esbenp.prettier-vscode",
      "editor.formatOnSave": true
  },
  "[typescript]": {
      "editor.defaultFormatter": "esbenp.prettier-vscode",
      "editor.formatOnSave": true
  },
  "eslint.validate": [
      "javascript",
      {
          "autoFix": true,
          "language": "typescript"
      }
  ]
}
```

My person style preferences are to keep trailingComma set to none and quoteProps set to as-needed.

.prettierrc
FKSRipple: No reason other than just being explicit, when I created the file I tried my best to describe the code base as it exists today.

FKSRipple:
@Stormtv +1 on setting `trailingComma` to "none", I took another look and that's what the repo is currently using most.

 We'll need quoteProps as consistent for #1086 so that prettier doesn't remove test name strings, even if it can. But if people feel strongly about this we could probably fool around with putting a special .prettierrc into the test dir. I'll make sure that we circle back to that in the larger "turn this on for everything automatically" PR.

PR #1090, opened 2019-11-11
Add support for tick size

RareData:
This PR adds support for the tick size field introduced by the TickSize amendement (https://xrpl.org/known-amendments.html#ticksize)

The implementation follows the documentation in https://xrpl.org/accountset.html#accountset-fields, meaning that valid values are 3 to 15 inclusive and 0 is used to disable the custom tick size.


mDuo13:
Looks pretty good to me. :+1: 

Thanks for contributing!

PR #1096, opened 2019-11-15
TransactionJSON conflict fix

magmel48:


FKSRipple:
@magmel48 can you explain this change a bit? Why are you moving TransactionJSON? Was there a conflict in src/transaction/types.ts?

The idea that breaking changes can happen between minor versions of TS means that we should probably change from a `^` to a `~` so that we don't pull in new minor versions unintentionally.

magmel48:
@FKSRipple the issue is described here https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#local-and-imported-type-declarations-now-conflict: `Local and Imported Type Declarations Now Conflict`.

So, any user that would like to use the library in TS 3.7 version and higher will not be able to get his code compiled if he decide to add this library into dependencies of his project.

Development is evolving process usually, so, in some time you will need to upgrade your TS version to come developer's needs anyway. The change is simple, does not break anything, but provides better experience for users of TS 3.7.

So, I would love to see it at least reviewed.

magmel48:
Going to remove typescript upgrade.

FKSRipple:
@magmel thanks for the explanation and links. The changes make sense, LGTM!

PR #1097, opened 2019-11-16
Bump ripple-lib-transactionparser to 0.8.1

r0bertz:


FKSRipple:
Thanks @r0bertz! I reverted the main package version change since we don't change that on a PR-by-PR basis, but will merge in the dependency change

PR #1099, opened 2019-11-18
move majority of tests over to new api test runner

FKSRipple:
~3000 lines of test moved out of `test/api-test.ts` and into the more structured api test runners!

Most of the time went to updating the tests to use async/await + our test helpers. The remaining ~1000 lines of tests can be moved out in a second wave.


src/common/serverinfo.ts
```diff
@@ -29,7 +29,8 @@ export type GetServerInfoResponse = {
     reserveIncrementXRP: string,
     ledgerVersion: number
   },
-  validationQuorum: number
+  validationQuorum: number,
+  networkLedger?: string
```
FKSRipple: Looks like we were testing for this value but hadn't covered in in our type definitions. Adding it here as an optional property.

test/api/index.ts
```diff
@@ -33,7 +33,12 @@ describe('RippleAPI [Test Runner]', function() {
   const filteredTestSuites = allTestSuites.filter(({ isMissing }) => !isMissing)
 
   // Run all the tests:
-  for (const { name: suiteName, tests, isMissing } of filteredTestSuites) {
+  for (const {
+    name: suiteName,
+    tests,
+    isMissing,
+    config
```
FKSRipple: we added support for an `export config = {}` in the test files, to tell the test runner how they should be run. We needed this because one of the test files tested the mocked server in a way that always returned a non-X style address, even if you passed an X-style address. Skipped these for now while we finish up this migration, and then can revisit how we handle mocked responses.

test/api/utils.ts
```diff
@@ -67,22 +71,31 @@ export function assertResultMatch(
...
 
 /**
- * Check that the promise rejects with an expected error instance.
+ * Check that the promise rejects with an expected error.
  */
 export async function assertRejects(
   promise: PromiseLike<any>,
-  instanceOf: any
+  instanceOf: any,
+  message?: string | RegExp
```
FKSRipple: Adds a check that error message matches expected.

test/api/utils.ts
```diff
@@ -67,22 +71,31 @@ export function assertResultMatch(
...
+    if (typeof message === 'string') {
+      assert.strictEqual(error.message, message);
+    } else if (message instanceof RegExp) {
+      assert(message.test(error.message));
+    }
   }
 }
 
 export function getAllPublicMethods(api: RippleAPI) {
-  return Object.keys(api).filter(key => !key.startsWith('_'))
```
FKSRipple: it turns out, `Object.keys()` on an object didn't return methods on its prototype (the RippleAPI base class that the instance inherits from)

FKSRipple:
breaking this into smaller PRs

PR #1107, opened 2019-12-10
[package.json] Require node 8 and yarn

intelliot:
This library should be used with node 8 and yarn

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,6 @@
   "readmeFilename": "README.md",
   "engines": {
     "node": ">=8",
-    "yarn": "^1.19.2"
+    "yarn": "^1.15.2"
   }
 }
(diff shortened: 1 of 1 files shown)


intelliot:
@tylerlevine I guess this would remove node 6 support, right? If that's the case I'm ok with leaving this alone for now. https://github.com/ripple/ripple-lib/issues/1070#issue-516270141

tylerlevine:
> @tylerlevine I guess this would remove node 6 support, right? If that's the case I'm ok with leaving this alone for now. [#1070 (comment)](https://github.com/ripple/ripple-lib/issues/1070#issue-516270141)

Not directly, we can still use `--ignore-engines` to get an install working there. We actually already do this for other reasons for node 6, so this shouldn't affect us.

In other news, we should be dropping node 6 completely by the end of the year.

PR #1109, opened 2019-12-12
Fix tests

intelliot:


intelliot:
Merging this so that tests pass, but others should feel free to fix this in a more elegant manner (and open a new PR).

PR #1111, opened 2019-12-14
Update README

intelliot:
Thanks to @hbergren for pointing out the removed npm run script


intelliot:
I think this is low-risk but feel free to review/comment even after I merge. We can always open a new PR for improvements!

PR #1112, opened 2019-12-14
Update webpack

intelliot:


intelliot:
I think this is low-risk but feel free to review/comment even after I merge.

FKSRipple:
If the frontend build still works, then lgtm!

PR #1114, opened 2019-12-16
Cleanup the connection trace logic

FKSRipple:
Changes trace from a `boolean` to a `boolean | function`, where the user can provide a function for logging the request/response (vs. having to monkey-patch the _console message in our tests).

The public API docs only mention the current boolean behavior, do we want to also mention that a function works as an option? It's fairly advanced usage, but could still be good to mention.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -82,7 +82,6 @@ export interface APIOptions extends ConnectionOptions {
   server?: string,
   feeCushion?: number,
   maxFeeXRP?: string,
-  trace?: boolean,
   proxy?: string,
   timeout?: number
 }
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -8,7 +8,7 @@ import {RippledError, DisconnectedError, NotConnectedError,
   RippledNotInitializedError} from './errors'
 
 export interface ConnectionOptions {
-  trace?: boolean
+  trace?: boolean | ((id: string, msg: string) => void)
   proxy?: string
   proxyAuthorization?: string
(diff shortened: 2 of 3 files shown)


src/api.ts
```diff
@@ -82,7 +82,6 @@ export interface APIOptions extends ConnectionOptions {
   server?: string,
   feeCushion?: number,
   maxFeeXRP?: string,
-  trace?: boolean,
```
FKSRipple: not needed, since APIOptions extends ConnectionOptions

src/common/connection.ts
```diff
@@ -8,7 +8,7 @@ import {RippledError, DisconnectedError, NotConnectedError,
   RippledNotInitializedError} from './errors'
 
 export interface ConnectionOptions {
-  trace?: boolean
+  trace?: boolean | ((id: string, msg: string) => void)
```
keefertaylor: super nit (feel free to ignore): I prefer to not use abbreviations in code. It makes things harder to reason about. 

I.E. if this is `msg`, why do we call it `proxyAuthorization` instead of `proxyAuth`? There's no clear rule about when to use or not use abbreviations, so it's easier to just never use them by default.

FKSRipple: Personally I feel the same way, although I think some ultra-common generic terms get a pass: `id`, `err`, `msg`, etc. are all fine by me.

I do see `message` used elsewhere in the codebase, so let's make the switch for consistency's sake, if nothing else.

Issue #1110, opened 2019-12-13
Connection Logic Audit


FKSRipple:
As a part of our work on https://github.com/ripple/xrp-api resiliency, I took a look this week at the connection logic inside of ripple-lib, specifically inside the `Connection` class in the `common/connection.ts` file. This file controls all of the connection/reconnection/disconnect/retry/heartbeat/timeout logic that implements the base of RippleAPI's websocket request->response interface, in addition to some basic ledger information tracking. It hasn't seen a meaningful refactoring since it was first created [4 years ago](https://github.com/ripple/ripple-lib/commit/15eb4c290abe144f16f3c64da8e9677fd20b2436) and is probably overdue for one.

Why overdue? The connect logic works fine for the first connection and simple unit tests, but begins to break down when network flakiness is introduced. Scenarios attempted:

- **Connect to Testnet, then remove network connection:** Attempts to reconnect indefinitely, never returns an error to the user that anything is wrong with the reconnect attempt
- **Connect to Testnet without network, then add network connection:** Would refuse to connect if first connection failed, even if network was fine.

Additionally, internal digging identified a few different implementation issues:
- `_disconnect(calledByUser)`: calledByUser has been overloaded to mean a ton of different things, where some not-called-by-user flows still needed the called-by-user behavior.
- `_retryConnect()` would hang unless you first disconnected. `_retry` counter was not always reset as expected
- Events didn't stick to a single format/structure
- errors weren't always propagated through the reconnect flow
- errors weren't always propagated through the request flow
- "onX" event handlers had to be bound and then juggled across async flows/callbacks

## Full interface audit

```ts
class Connection extends EventEmitter {
    
    //////////////
    // PUBLIC INTERFACE
    //////////////

    // GOOD! We don't want to change the public interface at this point
    connect(): Promise<void>;
    disconnect(): Promise<void>;
    reconnect(): Promise<void>;
    getLedgerVersion(): Promise<number>;
    hasLedgerVersions(lowLedgerVersion: any, highLedgerVersion: any): Promise<boolean>;
    hasLedgerVersion(ledgerVersion: any): Promise<boolean>;
    isConnected(): boolean;
    getFeeBase(): Promise<number>;
    getFeeRef(): Promise<number>;
    request(request: any, timeout?: number): Promise<any>;

    // 1. IMPROVE: Add public isReady() to read state, and return boolean if ready for requests
    private _isReady;


    //////////////
    // PRIVATE INTERFACE
    //////////////

[...]

PR #1115, opened 2019-12-16
Cleanup the connection config state

FKSRipple:
This moves all of the configuration state into a single `_config` property. The _config object is created from the user-provided config after the constructor cleans it (basically just sets some defaults).

This is nice because it gets all the user-provided configuration into a single property, which should scale nicely / not pollute the global `Connection` object with config values. However the is probably the most subjective of everything suggested in #1110, so if you prefer it as-is let me know.

### DIFF
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -7,6 +7,9 @@ import {RippledError, DisconnectedError, NotConnectedError,
   TimeoutError, ResponseFormatError, ConnectionError,
   RippledNotInitializedError} from './errors'
 
+/**
+ * ConnectionOptions is the configuration for the configuration object.
+ */
 export interface ConnectionOptions {
   trace?: boolean | ((id: string, message: string) => void)
   proxy?: string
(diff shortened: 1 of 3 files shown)


src/common/connection.ts
```diff
@@ -44,28 +46,24 @@ class Connection extends EventEmitter {
   private _onUnexpectedCloseBound: null|((...args: any[]) => void) = null
   private _fee_base: null|number = null
   private _fee_ref: null|number = null
-  private _connectionTimeout: number
 
   private _trace: (id: string, msg: string) => void = () => {}
+  private _config: ConnectionOptions
 
-  constructor(url, options: ConnectionOptions = {}) {
+  constructor(url?: string, options: ConnectionUserOptions = {}) {
```
keefertaylor: Why change `url` to optional?

FKSRipple: Thanks for pointing this out. `url` was untyped before, so it hasn't been changed from required to optional, just from untyped to typed.

As for why it's optional, I was confused by this too. But apparently this is documented behavior to get access to some of the helper methods / signing methods that don't necessarily need a URL to work. (It would be nice to revisit this: what methods are those, and can we pull them off of the Connection object and onto RippleAPI or even just turn them into stateless functions?)

keefertaylor: Makes sense, thanks!

PR #1119, opened 2019-12-17
Add LedgerHistory to Connection

FKSRipple:
Implements a change suggested in #1110.

- Moved ledger logic off of the Connection class and into it's own `Ledger` class.
- Updated `_whenReady(promise)` -> `_waitForReady()`. The old method appeared to not be working as expected: it was meant to run a promise after the connection was ready, but a promise would always start immediately at creation time.  Instead we just remove the promise argument, and resolve once the connection is ready. Then, the caller can wait for this to resolve before moving on to start the thing that needed to wait.
- There's a little bit of implicit behavior that I didn't get rid of: we assume that `waitForReady` will set some ledger data, so that methods like `getBaseFee()` will always return a number and never return null. It would be considered a bug if this promise was ever broken, but we could go one step further to throw an error OR add "null" as a possible return value so that the user had to handle this error case just in case. LMK your thoughts if you have any.

### DIFF
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -37,41 +37,53 @@ export interface ConnectionOptions {
 export type ConnectionUserOptions = Partial<ConnectionOptions>
 
 /**
- * Ledger is used to store and reference ledger information that has been
- * captured by the Connection class.
+ * LedgerHistory is used to store and reference ledger information that has been
+ * captured by the Connection class over time.
  */
-class Ledger {
+class LedgerHistory {
   private availableVersions = new RangeSet()
   latestVersion: null | number = null
   feeBase: null | number = null
   feeRef: null | number = null
 
-  hasVersions(lowVersion: number, highVersion: number): boolean {
-    return this.availableVersions.containsRange(lowVersion, highVersion)
-  }
-
+  /**
+   * Returns true if the given version exists.
+   */
   hasVersion(version: number): boolean {
     return this.availableVersions.containsValue(version)
   }
 
-  update(data: {
+  /**
+   * Returns true if the given range of versions exist (inclusive).
+   */
+  hasVersions(lowVersion: number, highVersion: number): boolean {
+    return this.availableVersions.containsRange(lowVersion, highVersion)
+  }
+
+  /**
+   * Update LedgerHistory with a new ledger response object. The "responseData"
+   * format lets you pass in any valid rippled ledger response data, regardless
+   * of whether ledger history data exists or not. If relevant ledger data
+   * is found, we'll update our history (ex: from a "ledgerClosed" event).
+   */
+  update(responseData: {
     ledger_index?: string
     validated_ledgers?: string
     fee_base?: string
     fee_ref?: string
   }) {
-    this.latestVersion = Number(data.ledger_index)
-    if (data.validated_ledgers) {
+    this.latestVersion = Number(responseData.ledger_index)
+    if (responseData.validated_ledgers) {
       this.availableVersions.reset()
-      this.availableVersions.parseAndAddRanges(data.validated_ledgers)
+      this.availableVersions.parseAndAddRanges(responseData.validated_ledgers)
     } else {
       this.availableVersions.addValue(this.latestVersion)
     }
-    if (data.fee_base) {
-      this.feeBase = Number(data.fee_base)
+    if (responseData.fee_base) {
+      this.feeBase = Number(responseData.fee_base)
     }
-    if (data.fee_ref) {
-      this.feeRef = Number(data.fee_ref)
+    if (responseData.fee_ref) {
+      this.feeRef = Number(responseData.fee_ref)
     }
   }
 }
@@ -90,13 +102,13 @@ class Connection extends EventEmitter {
 
   private _trace: (id: string, message: string) => void = () => {}
   private _config: ConnectionOptions
-  private _ledger: Ledger
+  private _ledger: LedgerHistory
 
   constructor(url?: string, options: ConnectionUserOptions = {}) {
     super()
     this.setMaxListeners(Infinity)
     this._url = url
-    this._ledger = new Ledger()
+    this._ledger = new LedgerHistory()
     this._config = {
       timeout: 20 * 1000,
       connectionTimeout: 2 * 1000,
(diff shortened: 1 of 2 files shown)


src/common/connection.ts
```diff
@@ -36,29 +36,67 @@ export interface ConnectionOptions {
  */
 export type ConnectionUserOptions = Partial<ConnectionOptions>
 
+/**
+ * Ledger is used to store and reference ledger information that has been
+ * captured by the Connection class.
+ */
+class Ledger {
```
keefertaylor: This class names feels like a misnomer to me.  If this is about a connection, call it a connection? Is it a client of a ledger? A connection to a ledger? 

Calling it a `Ledger` makes it sound like a model object, and though I am having trouble reasoning about what this actually is, it's not a model? `LedgerMetadata` perhaps?

src/common/connection.ts
```diff
@@ -36,29 +36,67 @@ export interface ConnectionOptions {
...
+  private availableVersions = new RangeSet()
+  latestVersion: null | number = null
+  feeBase: null | number = null
+  feeRef: null | number = null
+
+  hasVersions(lowVersion: number, highVersion: number): boolean {
```
keefertaylor: Please doc, it's unclear what a version is. 

Are low / high inclusive or exclusive?

src/common/connection.ts
```diff
@@ -36,29 +36,67 @@ export interface ConnectionOptions {
...
+    return this.availableVersions.containsRange(lowVersion, highVersion)
+  }
+
+  hasVersion(version: number): boolean {
+    return this.availableVersions.containsValue(version)
+  }
+
+  update(data: {
```
keefertaylor: Please doc.

src/common/connection.ts
```diff
@@ -36,29 +36,67 @@ export interface ConnectionOptions {
...
+    ledger_index?: string
+    validated_ledgers?: string
```
keefertaylor: Nit: this class is mixing camel case and snake case.

src/common/connection.ts
```diff
@@ -432,44 +453,53 @@ class Connection extends EventEmitter {
...
       } else {
-        this.once('connected', () => promise.then(resolve, reject))
+        this.once('connected', () => resolve())
       }
     })
   }
 
-  getLedgerVersion(): Promise<number> {
-    return this._whenReady(Promise.resolve(this._ledgerVersion!))
+  async getLedgerVersion(): Promise<number> {
```
keefertaylor: possible nit: missing an access modifier here. Is this on purpose?

src/common/connection.ts
```diff
@@ -432,44 +453,53 @@ class Connection extends EventEmitter {
...
-  hasLedgerVersion(ledgerVersion): Promise<boolean> {
-    return this.hasLedgerVersions(ledgerVersion, ledgerVersion)
+  async getFeeRef(): Promise<number> {
+    await this._waitForReady()
+    return this._ledger.feeRef!
   }
 
-  getFeeBase(): Promise<number> {
-    return this._whenReady(Promise.resolve(Number(this._fee_base)))
+  async hasLedgerVersions(
```
keefertaylor: Please doc. Are low / high inclusive or exclusives?

src/common/connection.ts
```diff
@@ -543,7 +573,7 @@ class Connection extends EventEmitter {
       // JSON.stringify automatically removes keys with value of 'undefined'
       const message = JSON.stringify(Object.assign({}, request, {id}))
 
-      this._whenReady(this._send(message))
+      this._send(message)
```
keefertaylor: There's no guarantee here that this class is ready. Shouldn't you `await _whenReady()` above here?

src/common/connection.ts
intelliot: perhaps `LedgerHistory`

src/common/connection.ts
intelliot: should be inclusive, I believe

src/common/connection.ts
intelliot: I think we can use snake case here if it is coming directly from rippled: https://xrpl.org/subscribe.html#ledger-stream

src/common/connection.ts
FKSRipple: 👍 will go with `LedgerHistory`

src/common/connection.ts
FKSRipple: yea, the idea is that this just takes any Rippled response object, and checks it for these properties. will clarify this in docs or argument name

src/common/connection.ts
FKSRipple: +1, good catch!

src/common/connection.ts
FKSRipple: Which access modifier were you expecting? Since it's already a public method, we don't want to mark it `private` now and force a breaking change unnecessarily.

src/common/connection.ts
FKSRipple: Done, lmk if that clarifies whats going on

src/common/connection.ts
FKSRipple: Ah wait no, this is right. (it's been so long since I originally wrote this, I had to go back and investigate :) 

whenReady wasn't working as expected, so before this PR this was essentially a no-op (not waiting for anything). It's a bit weird, but what was basically happening was:

1.  `this._whenReady` won't resolve until the first "subscribe" request completes. So if you await it inside the request function itself, you've essentially bricked the request method.
2. Luckily (unluckily?) the way we were calling `this._whenReady` before this PR wasn't actually doing any waiting. Since `this._send(message)` was passed as an argument, JS would have to run the method first (making the actual request) and then pass the result as the `_whenReady()` argument. So whenReady was always being called after the request.
3. This PR fixes that bug in other places, but just removes the no-op call here since it bricks the current connection flow. If this is new functionality that we actually do want, I'd like to add it in a separate PR, given how much is already in flight at the moment.

```
// Called second, with result from this._send() as argument.
this._whenReady(
  // Called first.
  this._send(message)
);
```

keefertaylor: Oh interesting. Thank you for the explanation.

src/common/connection.ts
keefertaylor: I feel like my `eslint` config always yells at me to make access modifies explicit.  I prefer to make things explicit, though if this library doesn't then I defer to the existing style.

src/common/connection.ts
keefertaylor: Sounds good. Thank you for humoring my pedantic nature :)

src/common/connection.ts
FKSRipple: Oh like adding "public"? I've actually never used that before. Not against it, but yea lets hold off to a future PR if you're interested in making that change. This library has only used them in 1-2 places and I see our eslint passing clean.

PR #1120, opened 2019-12-19
Add support for AccountDelete

intelliot:
https://xrpl.org/accountdelete.html

### DIFF
--- a/src/common/schemas/specifications/account-delete.json
+++ b/src/common/schemas/specifications/account-delete.json
@@ -1,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "accountDelete",
+  "link": "account-delete",
+  "type": "object",
+  "properties": {
+    "destination": {
+      "$ref": "address",
+      "description": "Address of an account to receive any leftover XRP after deleting the sending account. Must be a funded account in the ledger, and must not be the sending account."
+    },
+    "destinationTag": {
+      "$ref": "tag",
+      "description": "(Optional) Arbitrary destination tag that identifies a hosted recipient or other information for the recipient of the deleted account's leftover XRP."
+    },
+    "destinationXAddress": {
+      "$ref": "address",
+      "description": "X-address of an account to receive any leftover XRP after deleting the sending account. Must be a funded account in the ledger, and must not be the sending account."
+    }
+  },
+  "required": ["destination, destinationXAddress"],
+  "additionalProperties": false
+}
--- a/src/common/schemas/specifications/account-delete.json
+++ b/src/common/schemas/specifications/account-delete.json
@@ -17,6 +17,13 @@
       "description": "X-address of an account to receive any leftover XRP after deleting the sending account. Must be a funded account in the ledger, and must not be the sending account."
     }
   },
-  "required": ["destination, destinationXAddress"],
+  "anyOf": [
+    {
+      "required": ["destination"]
+    },
+    {
+      "required": ["destinationXAddress"]
+    }
+  ],
   "additionalProperties": false
 }
--- a/src/ledger/parse/account-delete.ts
+++ b/src/ledger/parse/account-delete.ts
@@ -1,0 +1,30 @@
+import * as assert from 'assert'
+import {removeUndefined} from '../../common'
+import {classicAddressToXAddress} from 'ripple-address-codec'
+
+export type FormattedAccountDelete = {
+  // account (address) of an account to receive any leftover XRP after deleting the sending account.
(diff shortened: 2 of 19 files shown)


src/common/schemas/specifications/account-delete.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "accountDelete",
+  "link": "account-delete",
+  "type": "object",
+  "properties": {
+    "destination": {
```
keefertaylor: So are these properties mutually exclusive? 

Should we only be using X-Addresses?

intelliot: Great catch! I’m fine with only using X-addresses here. Will update the PR later today

keefertaylor: A product decision is to whether we want to support X-Addresses only. 

If you want both APIs, that's fine. But I think you then need a `oneof` in your `required` property below.

intelliot: Good point, it should be fine to support both classic addresses and X-addresses. Updated.

src/ledger/parse/account-delete.ts
```diff
@@ -11,6 +11,10 @@ export type FormattedAccountDelete = {
   // for the recipient of the deleted account's leftover XRP. NB: Ensure that the hosted recipient is
   // able to account for AccountDelete transactions; if not, your balance may not be properly credited.
   destinationTag?: number
+
+  // X-address of an account to receive any leftover XRP after deleting the sending account.
+  // Must be a funded account in the ledger, and must not be the sending account.
+  destinationXAddress: string
```
hbergren: Does this mean you have to supply both the normal address and X Address? Ideally it's either/or:

`normal_destination_address (+ optional_tag) || x_address`

right?

intelliot: right, it should be either/or

intelliot: This particular file is `src/ledger/parse/account-delete.ts` which is used to parse an existing `AccountDelete` transaction from the ledger. When you retrieve an `AccountDelete` tx and it's parsed by ripple-lib, you should always get both the classic address and the X-address. So I think this code is correct

keefertaylor: I'm not going to block this PR from going in, but that feels like a *really* weird API that isn't intuitive to use. I'd recommend just choosing one or the other (probably X-Addresses) or creating two sets of APIs. 

My rationale is that giving users too much information, or too many options, leads to confusion and greater chances of mistakes and / or bugs.

intelliot: 👍 Let's try to move to X-address-only in ripple-lib v2.0

test/api/getTransaction/index.ts
```diff
@@ -354,6 +354,16 @@ export default <TestSuite>{
     )
   },
 
+  'AccountDelete': async (api, address) => {
+    const hash = 'EC2AB14028DC84DE525470AB4DAAA46358B50A8662C63804BFF38244731C0CB9'
```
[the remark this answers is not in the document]
intelliot: Yes

test/fixtures/rippled/tx/account-delete.json
```diff
@@ -0,0 +1,66 @@
...
+      "DeliveredAmount": "9995000000",
+      "TransactionIndex": 0,
+      "TransactionResult": "tesSUCCESS",
+      "delivered_amount": "9995000000"
+    },
+    "validated": true
+  },
+  "status": "success",
+  "type": "response"
+}
```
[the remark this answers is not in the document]
intelliot: In this case, I created/submitted an AccountDelete tx on Devnet, and then retrieved it.

src/ledger/parse/transaction.ts
```diff
@@ -1,27 +1,31 @@
...
+import parsePaymentChannelClaim from './payment-channel-claim'
 import parsePaymentChannelCreate from './payment-channel-create'
 import parsePaymentChannelFund from './payment-channel-fund'
-import parsePaymentChannelClaim from './payment-channel-claim'
-import parseFeeUpdate from './fee-update'
-import parseAmendment from './amendment'
+import parseTrustline from './trustline'
+
+import parseAmendment from './amendment' // pseudo-transaction
+import parseFeeUpdate from './fee-update' // pseudo-transaction
```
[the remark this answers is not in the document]
intelliot: Yes, this was alphabetized to make it easier to find things, and the pseudo-transactions were separated out

PR #1127, opened 2019-12-23
fix AccountObjectsResponse structure

you21979:
Does not match the structure returned by the API.

```:sample code
import {RippleAPI} from '.'

const main = async() => {
  const api = new RippleAPI({server: 'wss://s1.ripple.com'})
  await api.connect()
  const object = await api.getAccountObjects('rDbWJ9C7uExThZYAwV8m6LsZ5YSX3sa6US')
  for(const accountObject of object.account_objects){
    if(accountObject.LedgerEntryType === 'SignerList'){
      console.log(accountObject.SignerEntries[0].SignerEntry.Account)
    }
  }
}
main()
```

### DIFF
--- a/src/common/types/objects/ledger_entries.ts
+++ b/src/common/types/objects/ledger_entries.ts
@@ -166,7 +166,7 @@ export interface SignerListLedgerEntry {
   LedgerEntryType: 'SignerList'
   OwnerNode: string
   SignerQuorum: number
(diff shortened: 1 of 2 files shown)


FKSRipple:
thanks for catching this @you21979! I made a small change to move the fix into the `SignerEntry` type directly, but otherwise this looks great.

PR #1140, opened 2019-12-27
Refactor tests to support the new connection logic

FKSRipple:
**Note: Tests are expected to be failing for this PR! Fixes to the ~3 broken tests are in the next PR #1141. Still good to review!**

The Connection refactoring kept the public API as-is, but since some of the tests were using private/internal APIs some changes were still needed to the tests. This PR moves some tests to test public APIs instead of private APIs. This keeps these "overall test cleanups" separate from the new logic PR to help reviewers.

This doesn't convert all private API tests to public APIs, but just enough to keep our test coverage good for the new Connection logic. If that's something we care about, we can do a full pass to update the rest of the private tests in a follow-up PR.


test/api/getFee/index.ts
```diff
@@ -19,12 +19,10 @@ export default <TestSuite>{
   },
 
   'getFee - high load_factor': async (api, address) => {
-    api.connection._send(
```
FKSRipple: This is the main test change in this PR: `_send` (private) -> `request` (public).

test/connection-test.ts
```diff
@@ -245,34 +231,16 @@ describe('Connection', function() {
...
       done()
     })
-
     setTimeout(() => {
       this.api.connection._ws.close()
     }, 1)
   })
 
   describe('reconnection test', function() {
-    beforeEach(function() {
```
FKSRipple: no longer needed

test/connection-test.ts
```diff
@@ -323,11 +289,11 @@ describe('Connection', function() {
                   ' instead)'
               )
             )
-          } else if (reconnectsCount !== num * 2) {
```
FKSRipple: It seems like we were double-emitting 'reconnecting' events.

test/mock-rippled.ts
```diff
@@ -119,12 +119,26 @@ export function createMockRippled(port) {
   mock.on('request_config', function(request, conn) {
     assert.strictEqual(request.command, 'config')
     conn.config = _.assign(conn.config, request.data)
+    conn.send(
```
FKSRipple: a response is expected for every request, otherwise the Connection rejects with a timeout error.

test/connection-test.ts
```diff
@@ -573,7 +540,6 @@ describe('Connection', function() {
 
   it('should try to reconnect on empty subscribe response on reconnect', function(done) {
```
FKSRipple: @intelliot would love some guidance here. Based on the old & new logic, an "empty subscribe response" would cause a connection failure, and not a second reconnect. But then this test seems to be testing that only a single reconnect happens, and `{returnEmptySubscribeRequest: 3}` doesn't seem to be doing anything (only `returnEmptySubscribeRequest: 1` seems to have an effect in mock-rippled).

I think this test is safe to remove, since we're already testing the reconnect logic and what happens when the server doesn't have validated ledgers.

intelliot: I would be fine with removing this test.

PR #1141, opened 2019-12-27
Refactor the Connection class

FKSRipple:
This PR refactors the `Connection` class to improve resiliency to disconnects, retried connections, and error propagation. See https://github.com/ripple/ripple-lib/issues/1110 for more background.

#### The work has been separated into 4 main areas of improvement:
- (DONE IN #1119) `LedgerHistory`: Manage the connection ledger fees, data, and updates.
- `ConnectionManager`: Manages multiple connection attempts using a single WebSocket instance, and handling any connection errors. 
- `RequestManager`: Manage the requests, and resolve the original request Promise once the WebSocket response comes in after-the-fact.
- Retry logic: Add an exponential backoff to reconnect attempts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
   "dependencies": {
     "@types/lodash": "^4.14.136",
     "@types/ws": "^6.0.3",
-    "backo": "^1.1.0",
     "bignumber.js": "^9.0.0",
     "https-proxy-agent": "^3.0.0",
     "jsonschema": "1.2.2",
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "dependencies": {
     "@types/lodash": "^4.14.136",
     "@types/ws": "^6.0.3",
+    "backo": "^1.1.0",
     "bignumber.js": "^9.0.0",
     "https-proxy-agent": "^3.0.0",
     "jsonschema": "1.2.2",
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -10,11 +10,13 @@ import {
   TimeoutError,
   ResponseFormatError,
   ConnectionError,
-  RippledNotInitializedError
+  RippledNotInitializedError,
+  RippleError
 } from './errors'
+const ExponentialBackoff = require('backo')
 
 /**
- * ConnectionOptions is the configuration for the configuration object.
+ * ConnectionOptions is the configuration for the Connection class.
  */
 export interface ConnectionOptions {
   trace?: boolean | ((id: string, message: string) => void)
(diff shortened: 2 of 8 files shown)


src/common/connection.ts
```diff
@@ -32,10 +34,84 @@ export interface ConnectionOptions {
...
+ * WebSocket spec allows 4xxx codes for app/library specific codes.
+ * See: https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent
+ **/
+const INTENTIONAL_DISCONNECT_CODE = 4000
+
+/**
+ * Create a new websocket given your URL and optional proxy/certificate
+ * configuration.
+ */
+function createWebSocket(url: string, config: ConnectionOptions): WebSocket {
```
FKSRipple: Moved off of the Connection class (no changes to the logic itself besides updating the references to use function arguments instead of `this`).

src/server/server.ts
```diff
@@ -9,12 +9,14 @@ function getLedgerVersion(this: RippleAPI): Promise<number> {
...
+async function connect(this: RippleAPI): Promise<void> {
   return this.connection.connect()
 }
 
-function disconnect(this: RippleAPI): Promise<void> {
-  return this.connection.disconnect()
+async function disconnect(this: RippleAPI): Promise<void> {
+  // backwards compatibility: connection.disconnect() can return a number, but
+  // this method returns nothing. SO we await but don't return any result.
+  await this.connection.disconnect()
```
keefertaylor: This looks funky to me but I'm not a JS expert like you :)

implicitly this is adding a new statement that says:
```
return new Promise<Void>(function(res, rej) { res() });
```

As a non-JS programmer, I'd prefer you just make this explicit. If not, why not just return the number itself?

package.json
```diff
@@ -19,6 +19,7 @@
   "dependencies": {
     "@types/lodash": "^4.14.136",
     "@types/ws": "^6.0.3",
+    "backo": "^1.1.0",
```
intelliot: I think this is only ~50 LOC. Instead of pulling in a new dependency here, can we just handle this logic internally?

If we want to keep `src/common/connection.ts` focused, this can be in a separate file.

src/common/connection.ts
```diff
@@ -76,27 +152,147 @@ class Ledger {
...
+      this.reject(data.id, error)
+      return
+    }
+    this.resolve(data.id, data.result)
+  }
+}
+
+/**
+ * The main Connection class. Responsible for connecting to & managing
+ * an active WebSocket connection to a Ripple node.
```
intelliot: Ripple node -> XRPL node

FKSRipple: another reason for doing that major release that we talked about: we could finally rename from ripple-lib to xrpl-lib :)

package.json
FKSRipple: fine by me! A little extra work, but it's an old package so re-writing it in TypeScript will be good for maintenance.

src/server/server.ts
FKSRipple: We can make this change later 👍

PR #1160, opened 2020-01-09
Remove deprecated ledger fields

intelliot:
Per ripple/rippled#3214

The following fields are deprecated and may be removed without further notice: accepted, hash (use ledger_hash instead), seqNum (use ledger_index instead), totalCoins (use total_coins instead).


FKSRipple:
Can you remove from relevant code / type definitions as well? Ex: I see `seqNum` used in:

- `src/offline/ledgerhash.ts`
- `src/ledger/parse/ledger.ts`
- `src/common/types/objects/ledger.ts`

intelliot:
> Can you remove from relevant code / type definitions as well? Ex: I see `seqNum` used in:
> 
> * `src/offline/ledgerhash.ts`
> * `src/ledger/parse/ledger.ts`
> * `src/common/types/objects/ledger.ts`

Done

intelliot:
This might technically be a breaking change, but these fields are directly redundant with the snake_case versions and have been deprecated since ~2015. So I think this would be fine to merge

src/ledger/parse/ledger.ts
```diff
@@ -67,19 +66,19 @@ function parseState(state) {
  * @throws RangeError: Invalid time value (rippleTimeToISO8601)
  */
 export function parseLedger(ledger: Ledger): FormattedLedger {
-  const ledgerVersion = parseInt(ledger.ledger_index || ledger.seqNum, 10)
```
natenichols, line 70: Any particular reason why we're using `parseInt(..., 10)`, could we just use Number(...) if parsing in base 10?

intelliot: No reason. I suspect `Number()` would work equally well, but I'm not 100% sure. It has been `parseInt()` from [the beginning of time](https://github.com/ripple/ripple-lib/pull/504/files#diff-4f915b23ac0d38c14331690bd88e1b4af5588dacf2f87fd658748bebb21742f1R39) and I think we can leave it that way unless there is some reason to change it.

natenichols: I'm perfectly happy with `parseInt()`, mostly just curious.

natenichols (approved):

PR #1166, opened 2020-01-13
Protects Connection _onUnexpectedClose fn from possibly null _ws property

nickewansmith:
This PR applies to master branch as the Connection class is heavily refactored in the develop branch.

This protects Connection `_onUnexpectedClose` from the possibility of `this._ws` being null which would cause an uncaught TypeError when removing listeners on the property.

### DIFF
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -134,12 +134,14 @@ class Connection extends EventEmitter {
   }
 
   _onUnexpectedClose(beforeOpen, resolve, reject, code) {
-    if (this._onOpenErrorBound) {
+    if (this._onOpenErrorBound && this._ws) {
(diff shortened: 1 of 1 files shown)


FKSRipple:
@intelliot given that develop is ready to go out, I'm fine doing whatever you'd prefer here (this bug is already fixed in `develop`)

intelliot:
I’ll publish a release and then @nickewansmith can confirm whether it’s already fixed and potentially close this PR

intelliot:
This should be fixed in ripple-lib 1.6.1 which was just released. Please confirm and then feel free to close this PR

nickewansmith:
I believe it is fixed, thanks.

PR #1169, opened 2020-01-13
Improve documentation for generateXAddress()

hbergren:
It takes an optional property for including the classic address in the return payload.


intelliot:
Merging this, but we can continue to improve it later on too

PR #1179, opened 2020-01-15
Catch possible error in reconnect() on _heartbeat(), emit reconnect error

nickewansmith:
This addresses a possible uncaught error on run of `this._heartbeat` and subsequently `this.reconnect()` as the reconnection promise could throw.

Emits an error event if `this.reconnect()` does throw as seen in other reconnect failures.


nickewansmith:
@FKSRipple ,  not sure of your release schedule. I'm wondering when the next release would be?

intelliot:
We don't have a particular schedule and can do a release anytime one is warranted. It tends to be every 1-3 weeks but I can go ahead and do a release later today.

intelliot:
ripple-lib 1.6.2 released ✅

PR #1186, opened 2020-01-24
Assign event listener to socket close event on open before attempting post-open logic

nickewansmith:
This commit adds the Connection `_ws.close` event listener post `_ws.open` before executing any post `_ws.open` logic, i.e. `Connection._subscribeToLedger`.  This prevents a reconnection error loop that occurs if `Connection._ws` is never cleaned up by the unreachable `_ws.close` event listener. Also, this ensures that a possible disconnect() promise rejection is not unhandled if any `_ws.open` logic in `Connection.connect()` throws.

### DIFF
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -521,7 +521,9 @@ export class Connection extends EventEmitter {
         this.emit('connected')
       } catch (error) {
         this._connectionManager.rejectAllAwaiting(error)
-        await this.disconnect().catch(() => {}) // Ignore this error
+        await this.disconnect().catch(() => {
+          this.emit('error', 'disconnect', error.message, error)
+        })
       }
     })
     return this._connectionManager.awaitConnection()
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -489,18 +489,6 @@ export class Connection extends EventEmitter {
       this._ws.on('error', error =>
         this.emit('error', 'websocket', error.message, error)
       )
-      // Finalize the connection and resolve all awaiting connect() requests
-      try {
-        this._retryConnectionBackoff.reset()
-        await this._subscribeToLedger()
-        this._startHeartbeatInterval()
-        this._connectionManager.resolveAllAwaiting()
-        this.emit('connected')
-      } catch (error) {
-        this._connectionManager.rejectAllAwaiting(error)
-        this.disconnect()
-        return
-      }
       // Handle a closed connection: reconnect if it was unexpected
       this._ws.once('close', code => {
         this._clearHeartbeatInterval()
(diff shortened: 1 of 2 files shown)


src/common/connection.ts
```diff
@@ -524,6 +512,17 @@ export class Connection extends EventEmitter {
...
+      // Finalize the connection and resolve all awaiting connect() requests
+      try {
+        this._retryConnectionBackoff.reset()
+        await this._subscribeToLedger()
+        this._startHeartbeatInterval()
+        this._connectionManager.resolveAllAwaiting()
+        this.emit('connected')
+      } catch (error) {
+        this._connectionManager.rejectAllAwaiting(error)
+        await this.disconnect().catch(() => {}) // Ignore this error
```
hbergren: Why do we ignore this error?

nickewansmith: I was mimicking code around ~407, but I think I was wrong. What should be done with the error? Can we emit the error, similar to an error on reconnect?

```await this.disconnect().catch((error) => {
      this.emit('error', 'disconnect', error.message, error)
})```

We should probably await and catch here as it's inside of the listener.

Comment (nickewansmith) [reply src/common/connection.ts]: And yes, what you summarized above is the problem.

Comment (FredKSchott) [reply src/common/connection.ts]: So the reason that it would be okay to ignore the error here is that we are already rejecting the `connect()` promise with the error on the line above: `this._connectionManager.rejectAllAwaiting(error)`

Ultimately, the problem here is: in a world where we have two errors, which do we return to the user when we can only reject one error? We could look into creating a new `MultipleError` type, but for now it feels like the failure to subscribe is more important than a failure to disconnect, which would make no sense to the user in the context of a failed `connect()`.

Comment (nickewansmith) [reply src/common/connection.ts]: Ok, it kind of seemed that way. I'll remove the emit of disconnect error then, as before.

File: test/connection-test.ts [hunk 2]
```diff
@@ -590,6 +590,26 @@ describe('Connection', function() {
...
   )
 
+  it('should clean up websocket connection if error after websocket is opened', async function() {
+    await this.api.disconnect();
+    // fail on connection
+    this.api.connection._subscribeToLedger = async () => {
+      throw new Error('error on _subscribeToLedger')
+    }
+    try {
+      await this.api.connect();
```
FredKSchott: can you throw an "expected to fail" error here, and then check for the correct error below? Something like:

```
try {
  await this.api.connect();
  throw new Error('expected connect() to reject, but it resolved');
} catch (err) {
  assert (err.message === 'error on _subscribeToLedger');
}

this.api.connection._subscribeToLedger = // ...
```

test/connection-test.ts
```diff
@@ -590,6 +590,26 @@ describe('Connection', function() {
...
+    } catch (err) {
+      // _ws.close event listener should have cleaned up the socket when disconnect _ws.close is run on connection error
+      // do not fail on connection anymore
+      this.api.connection._subscribeToLedger = async () => {}
+      await new Promise((resolve, reject) => {
+        setTimeout(async () => {
```
FredKSchott: Why the setTimeout? Can you remove it?

test/connection-test.ts
nickewansmith: Thx, will add.

test/connection-test.ts
nickewansmith: This connect() needs some time for the disconnect to run _ws.close() and clean up the connection or we will still get `RippleError: Websocket connection never cleaned up` in the CI tests. In a real world scenario, if that happened it would be ok, as the connection would eventually get cleaned up, and the error would technically be correct if it was not.

nickewansmith: Actually, figured it out, updated, thx.

Issue #1191, opened 2020-01-28
SSL Issue connecting to devnet.rippletest.net


okanaganrusty:
Cannot connect to s.devnet.rippletest.net with an `openssl` command or with the `ripple-lib` library without defining the environment variable (NODE_TLS_REJECT_UNAUTHORIZED=0).

```bash
$ echo | openssl s_client -connect s.devnet.rippletest.net:51233 | openssl x509 -noout
depth=0 OU = Domain Control Validated, OU = Gandi Standard Wildcard SSL, CN = *.devnet.rippletest.net
verify error:num=20:unable to get local issuer certificate
verify return:1
depth=0 OU = Domain Control Validated, OU = Gandi Standard Wildcard SSL, CN = *.devnet.rippletest.net
verify error:num=21:unable to verify the first certificate
verify return:1
DONE
```

Code to reproduce the error pasted below

```javascript
const RippleAPI = require('.').RippleAPI;

const api = new RippleAPI({
  server: 'wss://s.devnet.rippletest.net:51233' // Public rippled server hosted by Ripple, Inc.
});
api.on('error', (errorCode, errorMessage) => {
  console.log(errorCode + ': ' + errorMessage);
});
api.on('connected', () => {
  console.log('connected');
});
api.on('disconnected', (code) => {
  // code - [close code](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent) sent by the server
  // will be 1000 if this was normal closure
  console.log('disconnected, code:', code);
});
api.connect().then(() => {
  /* insert code here */
}).then(() => {
  return api.disconnect();
}).catch(console.error);
```

Error message produced
```
NotConnectedError: unable to verify the first certificate
    code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE'
```

okanaganrusty:
After downloading http://crt.usertrust.com/GandiStandardSSLCA2.crt, I was able to connected without an error.

1. Converted the SSL CA/issuer certificate from DER to PEM format:

```bash
openssl x509 -in ~/Downloads/GandiStandardSSLCA2.crt -inform DER -outform PEM -out GandiStandardSSLCA2.crt.pem
```

2. Ran ```NODE_EXTRA_CA_CERTS=GandiStandardSSLCA2.crt.pem node test.js```, which was successful.

[...]

PR #1192, opened 2020-01-28
Option to include system root certificates when a trustedCertificates API option is used

okanaganrusty:
Update for open issue 1191 to include system root certificates when a trustedCertificates API option is used, otherwise you have to specify the entire end-to-end trust.  Alternatively, the parameter includeSystemTrustedCertificates can be set to `false` to preserve existing functionality.


intelliot:
@okanaganrusty Is this still an issue?

#1191 was closed a while ago, so I don't think this is still a problem.

okanaganrusty:
No, this is no longer an issue.  Thanks for resolving it.

PR #1195, opened 2020-01-30
Release 1.6.3-beta.1

intelliot:


intelliot:
> As an aside, why do you want to tag a beta version of ripple-lib? Do you anticipate anybody using the beta?

Yes, I use it for testing. It helps to catch mistakes that become obvious when trying to actually use a release. Things like the `export =` syntax, and making sure that we have a prepublish script so that we don't accidentally publish obsolete files.

> Do you not want the WebSocket logic to be in a non-beta release for the moment in case there are more related WebSocket bugs?

I'm fine with releasing the WebSocket logic as-is. Hopefully any bugs will be found quickly.

> Do you not want to wait for us to tag 1.0.0 of ripple-keypairs?

Doesn't matter either way. I expect we will be doing ripple-keypairs 1.0.0 eventually, hopefully in not too long. We should test it as much as we can, and also see if we can get an audit rolling.

nickewansmith:
Hey @intelliot, I'm wondering if we will be getting a non-beta release for the Websocket bug. It would be greatly appreciated if so!

intelliot:
@nickewansmith I expect that will be sometime this week. For now, could you get the beta release with something like `npm install --save ripple-lib@beta` or `yarn add ripple-lib@beta`?

PR #1251, opened 2020-04-03
Minor improvements to docs and comments

intelliot:
Improve `prepareTransaction` docs


docs/src/prepareTransaction.md.ejs
```diff
@@ -1,6 +1,6 @@
 ## prepareTransaction
 
-`prepareTransaction(transaction: object, instructions: object): Promise<object>`
+`prepareTransaction(transaction: [Transaction](https://xrpl.org/transaction-formats.html), instructions?: [Instructions](#transaction-instructions)): Promise<object>`
```
mDuo13, line 3: Sadly, this doesn't work the way you want it to. The `code` syntax blocks further parsing inside it, so you end up with something that looks like this:

> 

Should probably change it back here (but if possible, leave it as-is in the table below)

natenichols (approved):
LGTM 👍

PR #1275, opened 2020-04-27
Export hashing functions

intelliot:
The now-deprecated ripple-hashes library provided a number of hash
functions for users to compute the hashes/IDs for various XRP Ledger
objects:

* Transactions (to generate transaction hashes, also known as IDs)
* Transaction signing hashes (to sign transactions)
* Ledger Object IDs (to look up specific ledger objects in a ledger's
  state tree; see https://xrpl.org/ledger-object-ids.html)

This commit exports these utility methods from ripple-lib as static
methods. Access them on the RippleAPI class. Example:

```
import {RippleAPI} from 'ripple-lib'
const hash = RippleAPI.computeBinaryTransactionHash(...)
```

### DIFF
--- a/src/common/hashes/README.md
+++ b/src/common/hashes/README.md
@@ -2,7 +2,7 @@
 
 Methods to hash XRP Ledger objects
 
-## Methods
+## Computing a transaction hash (ID)
 
 ### computeBinaryTransactionHash = (txBlobHex: string): string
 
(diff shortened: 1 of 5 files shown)


src/common/hashes/README.md
```diff
@@ -12,19 +12,35 @@ Compute the hash of a binary transaction blob.
...
+
+_Removed:_ `computeTransactionSigningHash`, which took txJSON as a parameter. It was part of the deprecated ripple-hashes library. If you have txJSON, `encode` it with [ripple-binary-codec](https://github.com/ripple/ripple-binary-codec) first. Example: `return computeBinaryTransactionSigningHash(encode(txJSON))`
 
-### computeAccountHash = (address: string): string
+### computeAccountLedgerObjectID = (address: string): string
 
 Compute the hash of an account, given the account's classic address (starting with `r`).
 
-### computeSignerListHash = (address: string): string
+### computeSignerListLedgerObjectID = (address: string): string
```
hbergren, line 39: Is it worth making a breaking change by changing the function names?

intelliot: These functions were not previously exported, so I don't think it's a breaking change to rename them now, at the same time we are exporting them for the first time.

I renamed a few methods for clarity (the old name didn't make sense to me) but I don't feel strongly about it: there may be better names; I'm open to suggestions.

PR #1294, opened 2020-05-22
Deprecate submit() in favor of request('submit', {...})

intelliot:
Using request('submit', ...) allows the ripple-lib API to be effectively
identical to the underlying rippled API. Whereas submit() returns only
certain fields from rippled, using request('submit') gives you rippled's
full response, including the new fields added in rippled v1.5.0.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -5624,67 +5624,87 @@ return api.combine(signedTransactions);
 
 ## submit
 
-`submit(signedTransaction: string): Promise<object>`
+`request('submit', {tx_blob: string, fail_hard: boolean}): Promise<object>`
 
-Submits a signed transaction. The transaction is not guaranteed to succeed; it must be verified with [getTransaction](#gettransaction).
+The `submit` method applies a transaction and sends it to the network to be confirmed and included in future ledgers.
+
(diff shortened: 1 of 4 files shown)


docs/index.md
```diff
@@ -5624,67 +5624,87 @@ return api.combine(signedTransactions);
 
 ## submit
 
-`submit(signedTransaction: string): Promise<object>`
+`request('submit', {tx_blob: string, fail_hard: boolean}): Promise<object>`
 
-Submits a signed transaction. The transaction is not guaranteed to succeed; it must be verified with [getTransaction](#gettransaction).
+The `submit` method applies a transaction and sends it to the network to be confirmed and included in future ledgers.
```
intelliot, line 5629: Docs are mostly copied from https://xrpl.org/submit.html except:
- There is no sign-and-submit mode.
- Some links were updated or removed.

src/api.ts
```diff
@@ -398,7 +398,8 @@ class RippleAPI extends EventEmitter {
   prepareSettings = prepareSettings
   sign = sign
   combine = combine
-  submit = submit
+
+  submit = submit // @deprecated Use api.request('submit', { tx_blob: signedTransaction }) instead
```
hbergren, line 402: 👍

intelliot:
It exists and already works (it passes thru the command to rippled). But it's currently untyped; we could add typings for it here: https://github.com/ripple/ripple-lib/blob/develop/src/api.ts#L228

PR #1302, opened 2020-06-27
clearing of awaiting response promises when response handled

### LINKED ISSUE #1301: memory build up when doing repeated calls to getTransactions
Hi there! 

We are doing repeated calls to getTransactions - with a from and to ledger version range, as we are watching the ledger for incoming transactions on a specific account. This is on a long-lived connection (days).

When using the old ripple-lib - v1.3.4 - our memory usage on the server remains stable.
When using v1.7.1 - we get a linear increase in memory consumption until node runs out of heap and our service restarts. A linear increase is in keeping with the service we have that repeatedly gets the transactions, as it runs every few seconds.

We wrap ripple-lib in a utility library, so I created a test in the library that performs getTransactions then does a GC and heapdump (rinse-repeat 9 times). With the old version of ripple-lib - our heapdumps stay the same size. WIth v1.7.1 they grow linearly - in the same fashion they do on our server.

I then compared the heap dumps - and found large amounts of transaction hashes in the deltas between the oldest and newest heapdumps of the test set, I then looked at what was retaining them, and it looks like the RequestManager is holding on to them, on further investigation I realized that better men or women than me (that would be the engineers who I am speaking to now :) ) would be able to work out how the request manager works in v1.7.1

Sorry - I am rambling a bit, my parting questions are:

- would you like the heap dumps? They are about 24MB compressed.
- could this be a usage thing on our part, and this is expected behaviour - are we doing something silly by having long-lived connections - should we destroy connections with every request or N requests?

southbite:
fixes #1301 
very much a WIP - as I am not sure how this effects streaming.

### DIFF
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -225,18 +225,21 @@ class RequestManager {
   cancel(id: number) {
     const {timer} = this.promisesAwaitingResponse[id]
     clearTimeout(timer)
+    delete this.promisesAwaitingResponse[id];
   }
 
   resolve(id: number, data: any) {
     const {timer, resolve} = this.promisesAwaitingResponse[id]
     clearTimeout(timer)
     resolve(data)
+    delete this.promisesAwaitingResponse[id];
   }
 
   reject(id: number, error: Error) {
     const {timer, reject} = this.promisesAwaitingResponse[id]
     clearTimeout(timer)
     reject(error)
+    delete this.promisesAwaitingResponse[id];
   }
 
   rejectAll(error: Error) {
@@ -278,7 +281,8 @@ class RequestManager {
       throw new ResponseFormatError('valid id not found in response', data)
     }
     if (!this.promisesAwaitingResponse[data.id]) {
-      throw new ResponseFormatError('response handler not found', data)
+      return;
+      //throw new ResponseFormatError('response handler not found', data)
     }
     if (data.status === 'error') {
       const error = new RippledError(data.error_message || data.error, data)
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -225,21 +225,21 @@ class RequestManager {
   cancel(id: number) {
     const {timer} = this.promisesAwaitingResponse[id]
     clearTimeout(timer)
-    delete this.promisesAwaitingResponse[id];
+    delete this.promisesAwaitingResponse[id]
   }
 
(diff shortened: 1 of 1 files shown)


src/common/connection.ts
```diff
@@ -225,18 +225,21 @@ class RequestManager {
   cancel(id: number) {
     const {timer} = this.promisesAwaitingResponse[id]
     clearTimeout(timer)
+    delete this.promisesAwaitingResponse[id];
```
intelliot, line 228:
```suggestion
    delete this.promisesAwaitingResponse[id]
```

src/common/connection.ts
```diff
@@ -225,18 +225,21 @@ class RequestManager {
...
   }
 
   resolve(id: number, data: any) {
     const {timer, resolve} = this.promisesAwaitingResponse[id]
     clearTimeout(timer)
     resolve(data)
+    delete this.promisesAwaitingResponse[id];
```
intelliot, line 235:
```suggestion
    delete this.promisesAwaitingResponse[id]
```

src/common/connection.ts
```diff
@@ -225,18 +225,21 @@ class RequestManager {
...
   }
 
   reject(id: number, error: Error) {
     const {timer, reject} = this.promisesAwaitingResponse[id]
     clearTimeout(timer)
     reject(error)
+    delete this.promisesAwaitingResponse[id];
```
intelliot, line 242:
```suggestion
    delete this.promisesAwaitingResponse[id]
```

src/common/connection.ts
```diff
@@ -278,7 +281,8 @@ class RequestManager {
       throw new ResponseFormatError('valid id not found in response', data)
     }
     if (!this.promisesAwaitingResponse[data.id]) {
-      throw new ResponseFormatError('response handler not found', data)
+      return;
```
intelliot, line 284:
```suggestion
      return
```

src/common/connection.ts
```diff
@@ -278,7 +281,8 @@ class RequestManager {
...
+      //throw new ResponseFormatError('response handler not found', data)
```
intelliot, line 285:
```suggestion
```

intelliot (approved):
As this has been waiting for 3 months, we'll merge this to try in a beta. Can easily revert if needed.

PR #1326, opened 2020-11-12
Improve descriptions of get-ledger response time fields.

mDuo13:
I noticed that the `closeTime` and `closeTimeResolution` descriptions weren't quite right, so I updated them.


intelliot (approved):
Thanks!

PR #1327, opened 2020-11-12
style: yarn format (run prettier)

intelliot:
Just as we did with https://github.com/ripple/ripple-lib/pull/1116, this PR only applies the changes made when running `yarn format`.

## Context

I noticed that the Tickets Support PR https://github.com/ripple/ripple-lib/pull/1324 includes a lot of unrelated formatting only changes. I believe it is best to get those changes out of the way separately (with this PR) so that when it comes to the Tickets Support PR, it's easier to see and review just the changes that are specific to Tickets.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -7,4 +7,4 @@
   "trailingComma": "none",
   "quoteProps": "consistent",
   "bracketSpacing": false
-  }
+}
--- a/src/api.ts
+++ b/src/api.ts
@@ -155,7 +155,7 @@ class RippleAPI extends EventEmitter {
     const serverURL = options.server
     if (serverURL !== undefined) {
       this.connection = new Connection(serverURL, options)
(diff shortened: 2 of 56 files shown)


intelliot:
The big diff here comes from the jump to Prettier v2 which [changed the default value of `arrowParens`](https://prettier.io/docs/en/options.html#arrow-function-parentheses) from `avoid` to `always`:

> At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option.

elmurci (approved):
LGTM, thanks @intelliot , yes, I realised the Tickets PR had a lot of formatting changes. This PR will make things easier.

PR #1339, opened 2020-12-08
Revive Browser Integration Tests

natenichols:
This PR revives a small number of the previous browser tests, specifically the integration tests.

### Context of Change
First, this PR changes the `integration-test.ts` to test by connecting to the altnet, as opposed to on a standalone server. As such, we have to use the altnet faucet to generate a pre-funded account. This also means that this test suite now takes longer to run.

Second, we changed `webpack.config.js` to configure `integration-test.ts` to run with the webpack generated file `ripple-latest.js`. this test is referenced in `localintegrationrunner.html`, an html file that will run mocha tests in the browser.

Third, we use puppeteer to test `localintegrationrunner.html`. We wait for the final test to complete, and then assert that no tests have failed in the browser. 

Fourth, we removed some of the more complicated tests, for one of two reasons. 
1. Added complexity due to using the altnet. An example of this is the multisigning and pathfinding tests, which required more than just a signer and a counterparty. We could revive these in the future. 

2. the `getSettings` test was removed because it was inconsistent after being moved over to the altnet. It would sometimes succeed and sometimes fail. Note that it would not fail due to ripple lib, but due to moving to the altnet

### Type of Change
- [ ] Documentation Updates
- [ ] Release

#### Before
ripple-lib was not tested in the browser

#### After
Basic browser integration tests run as a part of the test suite. Note that this would have caught this issues seen in https://github.com/ripple/ripple-lib/issues/1319.

## Future Tasks
The ripple-lib test directory has unused tests, like the browser unit tests or the broadcast-tests. It would be good to either revive more of these tests or remove them from the directory

Note: currently the tickets integration test doesn't run, as tickets are not activated on the altnet. We can wait to merge until tickets is live.


natenichols:
Closing in favor of #1468

PR #1352, opened 2021-01-23
Add missing transaction specifications

mDuo13:
While tracking down a broken link coming from the API reference, I found that not only was the "Ticket Create" specification missing, so were a couple of the other new transaction types. This PR rectifies that by doing the following:

- Add the following missing specifications to the list of transactionType specifications:

    - Ticket Create
    - Account Delete
    - Deposit Preauth

- Rearrange the transaction type specifications to be in alphabetical order.
- Add descriptions to the transaction type specifications instead of linking back to the type table.
- Link transaction types to their corresponding native transaction type(s).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -19,20 +19,23 @@
   - [Transaction ID](#transaction-id)
   - [Transaction Memos](#transaction-memos)
 - [Transaction Specifications](#transaction-specifications)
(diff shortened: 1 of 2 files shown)


intelliot (approved):
Nice, thanks!

PR #1355, opened 2021-02-09
Fix connectionTimeout option

intelliot:
If the rippled server is down or inaccessible, the connect() method could hang for a long time. Previously, a connection timeout was added. The default was 2000 ms. This was too short; slower connections could take longer than that to connect.

This PR changes the default connection timeout to 5000 ms.

The `RippleAPI` constructor was meant to be able to take a `connectionTimeout` option, but the option was missing from the JSON schema (`api-options.json`), so it was not actually allowed to be used.

This PR adds the `connectionTimeout` option to the JSON schema.

Thanks to:
- @southbite 
- @madshell
- @koresar
- @nickewansmith 
- @kincaidoneil

### DIFF
--- a/src/common/connection.ts
+++ b/src/common/connection.ts
@@ -119,7 +119,7 @@ function createWebSocket(url: string, config: ConnectionOptions): WebSocket {
  * ws.send(), but promisified.
  */
 function websocketSendAsync(ws: WebSocket, message: string) {
-  return new Promise((resolve, reject) => {
+  return new Promise<void>((resolve, reject) => {
     ws.send(message, undefined, (error) => {
       if (error) {
         reject(new DisconnectedError(error.message, error))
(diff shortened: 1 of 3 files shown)


src/common/connection.ts
```diff
@@ -326,7 +326,7 @@ export class Connection extends EventEmitter {
     this._url = url
     this._config = {
       timeout: 20 * 1000,
```
natenichols, line 328: Whats the advantage of having two different timeouts, one for `connect()` and one for requests? Are there scenarios where requests will take 20 seconds to complete?

natenichols (approved):
LGTM 👍  

I have one question, but that's just for my own understanding 😄

src/common/connection.ts
koresar: Oh yeah. There are. Talking from our daily operations.

No, it doesn't. But there were people who asked for two different timeouts.

intelliot: 👍 koresar -- thanks for the insight.

With my home internet, `connect()` has never taken more than 2 seconds for me. Requests can take a lot longer depending on the complexity of the request; some requests require rippled to do more work, I believe.

PR #1382, opened 2021-03-12
fix: deserialization and verification of payment paths

### LINKED ISSUE #1347: Serialized transaction does not match original txJSON due to wrong serialization / deserialization of paths
We have an issue with signing transaction that has paths field specified. It looks like that paths object needs to have all three fields specified (account, issuer and currency). In case one of them is not defined / not provided, then ripple-lib will throw following error:
`Serialized transaction does not match original txJSON. See error.data`.

Lets look at this example:
   ```
const pathfind = {
        source: {
            address: 'rNf1i8YPXbV3Nh4FDsJiyENqMmGN8PnaLR',
            maxAmount: {
                currency: 'drops',
                amount: '100'
            }
        },
        destination: {
          address: 'rKT4JX4cCof6LcDYRz8o3rGRu7qxzZ2Zwj',
          amount: {
            value: '1',
            currency: 'USD',
            counterparty: 'rVnYNK9yuxBz4uP8zC8LEFokM2nqH3poc'
          }
        }
    }

    xrp.getPaths(pathfind).then(data => {
        pathfind.paths = data[0].paths;
        xrp.preparePayment('rNf1i8YPXbV3Nh4FDsJiyENqMmGN8PnaLR', pathfind).then(ret => {
            const signed = xrp.sign(ret.txJSON, wallet.secret);
        }).catch(err => {
            console.log('ERR: ', err);
        })
    }).catch(err => {
        console.log('ERR: ', err);
    })
```

With this example we get error as described above. We also found out, that this happens because of `ripple-binary-codec`.

If we check following code:
```

const txJSON = {
    Fee: "45",
    Amount: "1000000",
    Account: "rMqzrqJgSVzfKEJNyTxJnkpybMBjTTfuef",
    Destination: "rVnYNK9yuxBz4uP8zC8LEFokM2nqH3poc",
    TransactionType: "Payment",
    Paths: [
        [
            { account: 'rVnYNK9yuxBz4uP8zC8LEFokM2nqH3poc'}
        ]
    ]
}


const encoded = binaryCodec.encode(txJSON)
const decoded = binaryCodec.decode(encoded);

console.log('DECODED: ', decoded['Paths']);
```

this is the output:

```
"DECODED: " [
  [
    {
      account: 'rVnYNK9yuxBz4uP8zC8LEFokM2nqH3poc',
      issuer: undefined,
      currency: undefined
    }
  ]
]
```

So issue is that everytime, paths will be defined and one of the PathFind fields will not be defined, we cannot sign transaction. I would also like to point out, that in paths, we cannot pass undefined values from our side, to ripple-lib, because txJSON must be passed as string, and we cannot stringify undefined values.

I would also like to ask following question:
Is it absolutely necessary to do encode, and then immediately decode here: https://github.com/ripple/ripple-lib/blob/develop/src/transaction/sign.ts#L59 (decode is performed inside checkTxSerialization function)

intelliot:
Fixes #1347

The fix is just to update to ripple-binary-codec 1.1.2. However I've added a new unit test and a couple of snippets in order to test/verify the fix.

Ref #1376


natenichols (approved):
LGTM 👍 I like to see the tests that catch the cases that were failing!

intelliot:
Going to try getting this out before it gets too late on a Friday 😀

PR #1386, opened 2021-03-16
Expose address codec methods

### LINKED ISSUE #1385: Expose ripple-address-codec functions
I'd love to access the functions of [ripple-address-codec](https://github.com/ripple/ripple-address-codec), especially the ones for encoding and decoding X-addresses, from this library so I don't have to separately pull in ripple-address-codec (especially when working working in web browsers).

Specifically, I'm looking to convert from (classicAddress, tag, network) into an X-address and back.

intelliot:
Fix #1385


mDuo13 (approved):
LGTM

PR #1401, opened 2021-04-19
Prevent getFee from returning NaN from Reporting Mode

### LINKED ISSUE #1398: Calculation in getFee sometimes returns NaN when using public (s1, s2) ripple servers
Hello, I have recently been experiencing an issue where the `getFee` returns NaN when calling `preparePayment` using Ripple s1 and s2 public servers as the server for the Ripple api. This is because sometimes they do not return the `load_factor` in the response to the request `server_info`. 

I queried the **rippled** library about this and this is their response as to why sometimes `load_factor` is missing. [Reason why load_factor is missing](https://github.com/ripple/rippled/issues/3812#issuecomment-816871100).

Is there any way to default `load_factor` to 1 if it is not defined in the `server_info` request?

intelliot:
When using a rippled server in Reporting Mode, the server_info response does not have a load_factor property.

https://xrpl.org/rippled-server-modes.html#reporting-mode

https://github.com/ripple/rippled/issues/3812

Now, when load_factor is missing, we default to a load_factor of 1.

Fix #1398

Thanks to @LynchEntrostat @anhcao142

### DIFF
--- a/src/common/serverinfo.ts
+++ b/src/common/serverinfo.ts
@@ -71,6 +71,10 @@ async function getFee(this: RippleAPI, cushion?: number): Promise<string> {
 
   const serverInfo = (await this.request('server_info')).info
   const baseFeeXrp = new BigNumber(serverInfo.validated_ledger.base_fee_xrp)
+  if (serverInfo.load_factor === undefined) {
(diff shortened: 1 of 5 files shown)


ximinez (approved):
Looks good

sappenin (approved):

PR #1420, opened 2021-05-28
Edit docs for style and update old URLs

mDuo13:
Updates the documentation to better follow our style guide and use [plain language](https://www.plainlanguage.gov/guidelines/words/use-simple-words-phrases/) where possible. Also replaces a lot of old ripple.com URLs with more updated equivalents.

This is not a comprehensive review based on reading and trying to better understand the docs. That would be useful, but for this PR I've mostly used find-and-replace to fix some of the more obvious issues.


test/fixtures/responses/get-transactions-one.json
```diff
@@ -5,7 +5,7 @@
     "sequence": 491,
     "id": "D868CFF0DF8C8AAF205404460EA764ACB3B8862527FA414BC8C1CA9A45B1F276",
     "specification": {
-      "domain": "ripple.com"
+      "domain": "example.com"
```
mDuo13, line 8: I'm not sure how this fixture is used; this line may need to change back.

intelliot (approved):

test/fixtures/responses/get-transactions-one.json
intelliot: `ripple.com` in hex is `726970706C652E636F6D` and `example.com` is `6578616D706C652E636F6D`. I will add a commit that replaces all of the former with the latter

mvadari (approved):
lgtm

test/fixtures/responses/get-transactions-one.json
mvadari: Looks like tests are failing because of this change, so I think it'll have to be added to this PR

natenichols (approved):
Once the unit tests pass, this LGTM.

PR #1429, opened 2021-06-01
fix #999: use 'current' ledger when preparing txs

### LINKED ISSUE #999: Auto-filled sequence should be based on 'current' ledger
The 'validated' ledger is obsolete if you are submitting more than 1 tx per ledger.

Since it is always necessary to asynchronously confirm whether a tx has actually succeeded, there should be no harm in using the account's sequence based on the 'current' ledger.

intelliot:
See #999 

This should make `tefPAST_SEQ` occur less often, and make the behavior consistent when connecting to a rippled server in Reporting Mode.

### DIFF
--- a/src/transaction/utils.ts
+++ b/src/transaction/utils.ts
@@ -346,9 +346,9 @@ function prepareTransaction(
     }
 
     try {
-      // Consider requesting from the 'current' ledger (instead of 'validated').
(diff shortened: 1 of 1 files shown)


theotherian (approved):

theotherian:
Thanks for making this fix @intelliot

PR #1459, opened 2021-07-13
Update boilerplate docs

mDuo13:
- Suggest XRPLF's public cluster; this is a better starter experience since it's full-history and blocks known scam accounts
- Update the Node.js compatibility language to match the README. (credit to nixerFFM for noticing the discrepancy)
- Remove dated language about ES5/ES6 Promises.
- Re-gen docs (this seems to have added a memo field that was previously modified in some other PR)


intelliot (approved):
Thank you!

Issue #1458, opened 2021-07-12
Fix TrustlineTransaction type


mrosendin:
The typings for `FormattedTrustlineTransaction` need to be updated. 

Gives TypeScript errors like `Property 'limit' does not exist on type 'FormattedTrustline'.` for each property.

<kbd>![Screen Shot 2021-07-12 at 4 24 53 PM](https://user-images.githubusercontent.com/3401412/125367819-b551c300-e32d-11eb-8056-52af1b9b1082.png)</kbd>

The error goes away if I instead do `transaction.specification.specification.limit!`, but this property doesn't exist on the object.

This leads me to believe that the typings are somehow incorrect for `FormattedTrustline`
(see [FormattedTrustlineTransaction](https://github.com/ripple/ripple-lib/blob/develop/src/transaction/types.ts#L142) and [FormattedTrustlineSpecification](https://github.com/ripple/ripple-lib/blob/develop/src/common/types/objects/trustlines.ts#L32))

Example of a trust line object returned from the API:

```javascript
{
  "type": "trustline",
  "address": "ranRvKcpressC7Ctkwi37qBacSwYEzFuqR",
  "sequence": 19183454,
  "id": "61F997882C45C7D9111E9E6F620BF6DA1A36CED755D46EDC9CF1EB199D8EF9FC",
  "specification": {
    "limit": "999999999",
    "currency": "AUR",
    "counterparty": "rJCygsdUaeUgsngvxSUsBjzMLfYrzpkAVj",
    "memos": [
      {
        "type": "test",
        "format": "text/plain",
        "data": "StablePay App"
      }
    ],
    "qualityIn": 1,
    "qualityOut": 1,
    "ripplingDisabled": false,
    "frozen": false
  },
  "outcome": {
    "result": "tesSUCCESS",
    "timestamp": "2021-07-12T21:15:00.000Z",
    "fee": "0.000012",
    "balanceChanges": {
      "ranRvKcpressC7Ctkwi37qBacSwYEzFuqR": [
        {
          "currency": "XRP",
          "value": "-0.000012"
        }
      ]
    },
    "orderbookChanges": {},
    "ledgerVersion": 19183460,
    "indexInLedger": 0
  }
```

I can submit a PR to fix this.

mrosendin:
I am unable to create a branch, but I pushed the fix to a fork: c3013dd39dee81e673f0e89f025fe99eb3a1b9fe

intelliot:
[...]

PR #1460, opened 2021-07-13
fix: TrustlineTransaction type (#1458)

mrosendin:
Tested with the local package. My IDE likes the typings now.

<kbd>Screen Shot 2021-07-13 at 4 30 50 PM</kbd>


intelliot (approved):
Thank you!

florent-uzio (approved):
Thanks, Matt!

PR #1463, opened 2021-07-19
Update multisig doc

peng-huang-cc:
According to the example, send tx failed. 

The result is: 
```
Tentative result code: telINSUF_FEE_P
Tentative result message: Fee insufficient.
```


intelliot (approved):
Thank you!

Issue #1222, opened 2020-02-27
Make ripple-lib smaller


intelliot:
It's currently about 751 KB minified for the browser, which is relatively large. We should investigate whether there are some simple things we can do to make it smaller and/or more efficient.

Stormtv:
https://bundlephobia.com/result?p=ripple-lib@1.6.4

Seems like Lodash and elliptic are pretty large.

Also https://bundlephobia.com/result?p=elliptic@6.5.2 is using bn.js and hash.js which are also dependencies of ripple-lib so have them both use the same version of bn.js and hash.js could save data.

intelliot:
- ripple-lib 1.9.1 really is much bigger than 0.11.0 - we're talking 933 KB (+ another 71 KB for lodash) vs. 352 KB.
- I would like to get rid of lodash.
- Probably get rid of the jsonschemas as well.
- I'd prefer the API to be more like the rippled API; the rippled API is the 1 thing most developers should learn. For most methods, we don't need client-side schemas + input validation. rippled already returns errors if anything is malformed.
- We probably don't need the full Lodash https://github.com/ripple/ripple-lib/blob/develop/package.json#L28
- ripple-lib is requiring the whole thing and also requiring isequal separately
- https://bundlephobia.com/result?p=ripple-lib@1.9.3
- Seems to be 3 different libraries for handling big numbers also. https://github.com/indutny/bn.js/ https://github.com/MikeMcl/bignumber.js https://github.com/peterolson/BigInteger.js and one for decimals also https://github.com/MikeMcl/decimal.js 
- The ultimate eventual goal is to get rid of lodash entirely. Requiring isequal was a step in the direction of not requiring the whole thing. But clearly, that hasn’t been achieved yet ;)
- I’m very much in favor of not requiring the full lodash if someone can figure it out. Last I looked, it seemed like it would be a decent amount of work.
- Doubt bundlephobia counts the jsonschemas since those aren’t a dependency: the actual schemas themselves, the json files under https://github.com/ripple/ripple-lib/tree/develop/src/common/schemas

camposfyi:
lodash is pretty useful... might be a good idea to use [lodash-webpack-plugin](https://www.npmjs.com/package/lodash-webpack-plugin). It will remove any lodash function not used in the code. This way there's no need to create a custom lodash build every time you want to use a new function.

Brietoe:
Coincidentally, I actually started working on some of these items today before I was even aware of this 'issue' post.
[...]

PR #1467, opened 2021-07-22
Reduce dependencies on lodash

mvadari:
This PR reduces the dependency on lodash, but doesn't eliminate it completely. It reduces the size of the browser build by 12kb (953kb to 941kb). It also uses a custom lodash build based on the methods that we actually use. Helps with #1222.

To rebuild the lodash package based on the methods used, you can run `yarn lojack`.

More work could be done to reduce the dependency on lodash, but it would require essentially rewriting the methods in native JS and storing them somewhere in the library. I think this is out of the scope of this task and can be done again later when the focus is on reducing the size of the browser build.

Useful site: https://youmightnotneed.com/lodash/


src/common/utils.ts
```diff
@@ -146,7 +145,7 @@ function convertKeysFromSnakeCaseToCamelCase(obj: any): any {
 }
 
 function removeUndefined<T extends object>(obj: T): T {
-  return _.omitBy(obj, _.isUndefined) as T
+  return _.omitBy(obj, value => value === undefined) as T
```
nutbreaker, line 148: I would have created the function isUndefined in the Utils to harmonize the undefined checks. The reason for this suggestion is that today, based on the branch develop, there are 10 calls to `isUndefined` but 95 occurrences of `== undefined` (without the test folder) However, I will question the creation of this function because most of the cases can be solved by a truthy/falsy check.

```javaScript
function isUndefined(value){
  return undefined === value
}

//...  
return _.omitBy(obj, isUndefined) as T
//...
```

src/common/validate.ts
```diff
@@ -8,9 +8,9 @@ function error(text) {
 
 function validateLedgerRange(options) {
   if (
-    !_.isUndefined(options) &&
-    !_.isUndefined(options.minLedgerVersion) &&
-    !_.isUndefined(options.maxLedgerVersion)
+    options !== undefined &&
+    options.minLedgerVersion !== undefined &&
+    options.maxLedgerVersion !== undefined
```
nutbreaker, line 13: How about something like this
```javaScript
if(!options?.minLedgerVersion && !options?.maxLedgerVersion)
```

src/ledger/balance-sheet.ts
```diff
@@ -18,29 +18,38 @@ export type GetBalanceSheet = {
...
+  obligations?: Record<string, string>,
+  ledger_current_index?: number,
+  validated?: boolean
+}
+
+function formatBalanceSheet(balanceSheet: BalanceSheet): GetBalanceSheet {
   const result: GetBalanceSheet = {}
 
-  if (!_.isUndefined(balanceSheet.balances)) {
+  if (balanceSheet.balances !== undefined) {
```
nutbreaker, line 33: How about ?
```javascript
if (balanceSheet.balances)
```

src/common/utils.ts
mvadari: A truthy/falsy check isn't always the right solution, because it'll return the same if it's `0` or `''` or `undefined`, which is often not desirable. It's safer to use `=== undefined`.

src/ledger/utils.ts
```diff
@@ -74,7 +74,7 @@ function renameCounterpartyToIssuerInOrder(order: RequestBookOffersArgs) {
   const taker_gets = renameCounterpartyToIssuer(order.taker_gets)
   const taker_pays = renameCounterpartyToIssuer(order.taker_pays)
   const changes = {taker_gets, taker_pays}
-  return _.assign({}, order, _.omitBy(changes, _.isUndefined))
+  return Object.assign({}, order, _.omitBy(changes, value => value === undefined))
```
natenichols, line 77: Looking at `_.forEach(...)`, replaced with `Object.entries(...).forEach([_, ...])`.

Is it possible to to `EmitBy` with something like:

```js
const omitted = Object.entries(...).filter([_, ...] => value !== undefined)`

Object.assign({}, order, Object.fromEntries(omitted))
```

It seems strange to be removing one lodash function from this line, but leaving the other.

src/transaction/sign.ts
```diff
@@ -95,7 +95,7 @@ function objectDiff(a: object, b: object): object {
       return
     }
     if (type1 === '[object Array]') {
-      if (!isEqual(i1, i2)) {
+      if (!_.isEqual(i1, i2)) {
```
natenichols, line 98: Why is it important to change the imports here?

test/localrunner.html
```diff
@@ -3,7 +3,7 @@
     <meta charset="utf-8">
     <!-- encoding must be set for mocha's special characters to render properly -->
     <link rel="stylesheet" href="../node_modules/mocha/mocha.css" />
-    <script src="./vendor/lodash.min.js"></script>
+    <script src="../lodash.custom.min.js"></script>
```
natenichols, line 6: Is there a way we can package lodash w/ the webpacked ripple-lib, so that clients don't have to import our custom lodash package?

src/ledger/utils.ts
mvadari: Yeah that's fair, I didn't want to get rid of all the methods and unless I remove all the `omitBy` calls we don't get any saved space.

src/transaction/sign.ts
mvadari: `lodash.isequal` is technically a separate package, which takes up separate space. I removed the package as a dependency.

test/localrunner.html
mvadari: Pretty sure that already happens with the custom file we have - when it's packaged it gets packaged based on our custom settings.

src/ledger/utils.ts
natenichols: Oh, yeah, that makes sense. Unless you wanna rip it out entirely, then I'm okay w/ leaving it.

mvadari: I want to see what lodash methods we still use in the new version, I don't think it's worth doing any additional refactoring rn

test/localrunner.html
natenichols: Iirc we expect clients to import lodash separately: https://xrpl.org/get-started-using-node-js.html#rippleapi-in-web-browsers.

If we wanna package lodash w/ ripple-lib, we can do that, but I think wed have to adjust the webpacking.

mvadari: I looked into that - the file is like 70 kb bigger if we package lodash with it :(

I think it's fine for now and we can revisit the issue later

natenichols: Sounds good. We should definitely revisit as a part of #1222

natenichols (approved):
LGTM 👍

intelliot (commented):
This is very interesting and it was great to experiment with this.

I'm not sure that using a custom build of lodash is a good approach. This PR seems to increase the size of the npm distribution by 30 KB (926 KB vs 896 KB). Not a big difference, but not an improvement, either.

I do love that this gets us a lot closer to getting rid of lodash entirely, though.

src/common/utils.ts
ledhed2222: i would in this case vote for:

```javascript
return _.omitBy(obj, value => value == null) as T
```

because that predicate will only be true for values of `null` or `undefined`. I like to save `=== undefined` for cases when I _REALLY_ want to check exclusively for `undefined`. what do you think?

src/common/validate.ts
ledhed2222: as above - `!= null`

ledhed2222:
Agree with @intelliot . I think we should keep your code changes, but not touch packages or mess with customized lodash build, at least for now. what do you think?

mvadari:
I removed all the custom lodash stuff and somehow the size of the build went up - it's now 945kb

natenichols:
This LGTM. We'll wanna revisit removing Lodash entirely as a part of #1222,

natenichols (approved):

ledhed2222 (approved):
excellent!

PR #1468, opened 2021-07-22
Run Integration Tests in Browser

natenichols:
Runs integration tests in the browser.

### Context of Change
This PR uses GitHub Actions, and a rippled standalone node as a GitHub Actions Service to run integration testing in the browser. This PR uses puppeteer to run these tests. The browser integration test is run with `yarn test:browser`, and it starts up a headless chromium browser, runs a webpacked integration test, waits until the last test shows up in the http, and then asserts that 0 test have been failed, and more than zero have succeeded. 

### Type of Change
- [ ] Documentation Updates
- [ ] Release


mvadari (commented):
Can you set this PR to be merging into the parent branch? It makes the diff much easier to read

natenichols:
Done, will rebase on develop once `github-actions-integration` is taken care of... These three PRs build on one another, so I couldn't cleanly make another branch off of master.

mvadari:
In Github you can set the "base" branch as a different branch, and Github will automatically correct the base branch when it gets merged in

mvadari (approved):

.github/workflows/nodejs.yml
```diff
@@ -10,29 +10,13 @@ on:
...
-    - run: yarn build
+  unit:
 
-  build:
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
-        node-version: [12.x, 13.x, 14.x]
+        node-version: [12.x, 14.x, 16.x]
```
ledhed2222: 🔥

test/localintegrationrunner.html
```diff
@@ -3,13 +3,13 @@
...
     <!-- encoding must be set for mocha's special characters to render properly -->
     <link rel="stylesheet" href="../node_modules/mocha/mocha.css" />
-    <script src="./vendor/lodash.min.js"></script>
   </head>
   <body>
     <div id="deb"></div>
     <div id="mocha"></div>
     <script src="../node_modules/mocha/mocha.js"></script>
     <script src="hacks/phantomhacks.js"></script>
+    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.20/lodash.min.js" integrity="sha512-90vH1Z83AJY9DmlWa8WkjkV79yfS2n2Oxhsi2dZbIv0nC4E6m5AbH8Nh156kkM7JePmqD6tcZsfad1ueoaovww==" crossorigin="anonymous"></script>
```
ledhed2222, line 12: why are we doing this, especially given that `lodash` is still a dependency of the package?

webpack.config.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 const path = require('path');
 const webpack = require('webpack');
+const assert = require("assert/");
```
ledhed2222, line 4: that trailing `/` and the use of double quotes instead of single is bothering me, lol

webpack.config.js
```diff
@@ -40,14 +41,77 @@ function getDefaultConfiguration() {
 };
 }
 
+function webpackForTest(testFileName) {
+  var match = testFileName.match(/\/?([^\/]*)-test.ts$/);
```
ledhed2222, line 45: ew `const` instead of `var` please

webpack.config.js
```diff
@@ -40,14 +41,77 @@ function getDefaultConfiguration() {
...
+  if (!match) {
+    assert(false, 'wrong filename:' + testFileName);
+  }
+
+  let test = {
```
ledhed2222, line 50: `const` instead of `let`

ledhed2222 (changes requested):
helllllll yeah man. excellent work. few comments.

webpack.config.js
natenichols: That's what I get for reusing old code 😄

test/localintegrationrunner.html
ledhed2222: nathan and i spoke about this. i feel strongly that the experience for using `ripple-lib` in browser vs node should not be different. for that reason, `lodash` should be included (or not) regardless of target. the size savings are not worth the confusion to users who just want to pick this lib up and use it, especially considering that `ripple-lib` is already over 5x smaller than `web3.js`.

mvadari: If we're actually packaging lodash now, then the custom lodash build might be more worth it

natenichols: Yeah, I still think we wanna rip out lodash in v2... 

I'm ok w/ leaving it like this for this PR, thoughts?

intelliot: Personally, I agree with Nate:
1. I'm ok w/ leaving it like this for this PR
2. I want to get rid of lodash in v2. Removing it entirely is a significant effort, but very worthwhile, IMHO

Just my opinion though: it's ok if y'all decide otherwise

ledhed2222: sorry - i think we're talking about two different things.

1) if we want to remove lodash for v2 that's great
2) but as long as it's in v1 i think we should include it in the dist so that web users don't have to know to do this.

the second bit is outside the scope of this PR, but I'd love to see a ticket for v1 to add it back in for browser dist, unless/until is completely removed as a dependency for all platforms

ledhed2222 (approved):

PR #1476, opened 2021-07-26
fix #1474: add type argument to Promise

intelliot:
Fix #1474

PR #1477, opened 2021-07-26
fix: add void return type to Promises

intelliot:
Follow-up to #1476


natenichols (approved):
LGTM 👍 Those were the only tests I saw w/o the void type on them.

PR #1486, opened 2021-07-28
refactor combine logic for clarity

### LINKED ISSUE #1448: Combine requires a string array but the code handles it as an array of object (DEVX-83)
It is not possible to combine multiple signatures using the `api.combine` function.

```js
function combine(signedTransactions: Array<string>): object {
  validate.combine({signedTransactions})

  // TODO: signedTransactions is an array of strings in the documentation, but
  // tests and this code handle it as an array of objects. Fix!
  const txs: any[] = _.map(signedTransactions, binary.decode)
  const tx = _.omit(txs[0], 'Signers')
  if (!_.every(txs, (_tx) => _.isEqual(tx, _.omit(_tx, 'Signers')))) {
    throw new utils.common.errors.ValidationError(
      'txJSON is not the same for all signedTransactions'
    )
  }
  const unsortedSigners = _.reduce(
    txs,
    (accumulator, _tx) => accumulator.concat(_tx.Signers || []),
    []
  )
  const signers = unsortedSigners.sort(compareSigners)
  const signedTx = _.assign({}, tx, {Signers: signers})
  const signedTransaction = binary.encode(signedTx)
  const id = computeBinaryTransactionHash(signedTransaction)
  return {signedTransaction, id}
}
```

JST5000:

Refactor api.combine to be more readable.

### Context of Change
Resolves #1448 

I refactored the combine method to clearly highlight the core steps that were required instead of making the reader infer the high level goals from a bunch of lambda functions.

### Type of Change

## Before / After

Before it was hard to understand the point of each lambda function, now it should be easy to quickly find the part of the function that's causing problems or needs improving.

## Test Plan

CI Passes


src/transaction/combine.ts
```diff
@@ -5,6 +5,28 @@ import BigNumber from 'bignumber.js'
...
+  const tx = transactions[0];
+  const _txSignersTemp = tx.Signers;
+  delete tx.Signers
+  
+  transactions.forEach( _tx => {
+      const _txSignersTemp = _tx.Signers
+      delete _tx.Signers;
+      
+      if ( JSON.stringify(tx) !== JSON.stringify(_tx)) {
+          throw new utils.common.errors.ValidationError('txJSON is not the same for all signedTransactions')
```
mvadari, line 21: Import this directly from the module instead of using `import *` syntax

src/transaction/combine.ts
```diff
@@ -17,26 +39,35 @@ function compareSigners(a, b) {
...
     (accumulator, _tx) => accumulator.concat(_tx.Signers || []),
     []
   )
-  const signers = unsortedSigners.sort(compareSigners)
-  const signedTx = _.assign({}, tx, {Signers: signers})
-  const signedTransaction = binary.encode(signedTx)
+  //The documentation requires signers to be sorted
+  const sortedSigners = unsortedSigners.sort(compareSigners)
+  const tx = transactions[0];
+  tx.Signers = sortedSigners
```
mvadari, line 51: This is bad practice - this will modify the actual object at `transactions[0]`. That's why the syntax in deleted line 38 exists - so that a new object is created for that (I know it's a little weird). It may not matter for how you're using it, but someone else might use the same method and not realize.

src/transaction/combine.ts
```diff
@@ -17,26 +39,35 @@ function compareSigners(a, b) {
...
+  return tx
+}
+
+/**
+ * 
+ * @param signedTransactions A collection of the same transaction signed by different signers. The only difference
+ * between the elements of signedTransactions should be the Signers field.
+ * @returns An object with the combined transaction (now having a sorted list of all signers) which is encoded, along
+ * with a transaction id based on the combined transaction.
+ */
```
mvadari, line 62: Nice docs!

src/transaction/combine.ts
```diff
@@ -5,6 +5,28 @@ import BigNumber from 'bignumber.js'
 import {decodeAccountID} from 'ripple-address-codec'
 import {validate} from '../common'
 import {computeBinaryTransactionHash} from '../common/hashes'
+import { JsonObject } from 'ripple-binary-codec/dist/types/serialized-type'
+
+//The transactions should all be equal EXCEPT for the Signers field
```
mvadari, line 10: Can you add these types of quick explaining comments to the other methods in this file as well?

mvadari (changes requested):
Nice job!

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
 import * as _ from 'lodash'
 import binary from 'ripple-binary-codec'
-import * as utils from './utils'
 import BigNumber from 'bignumber.js'
+import { ValidationError } from '../common/errors'
```
natenichols, line 4: Can we adjust the spacing on these imports to be `{ValidationError}` to match the rest of the code. This should get fixed when we address linting, though.

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
 import * as _ from 'lodash'
 import binary from 'ripple-binary-codec'
-import * as utils from './utils'
 import BigNumber from 'bignumber.js'
+import {ValidationError} from '../common/errors'
 import {decodeAccountID} from 'ripple-address-codec'
 import {validate} from '../common'
 import {computeBinaryTransactionHash} from '../common/hashes'
+import {JsonObject} from 'ripple-binary-codec/dist/types/serialized-type'
```
natenichols, line 8: I _think_ this is the wrong JsonObject.. This is the one in ripple-binary-codec. Think theres a different one in ripple-lib, `TransactionJson`

natenichols: Really, we should put TransactionJson in the binary codec, and then import it into ripple-lib.

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
     []
   )
-  const signers = unsortedSigners.sort(compareSigners)
-  const signedTx = _.assign({}, tx, {Signers: signers})
-  const signedTransaction = binary.encode(signedTx)
+  //The documentation requires signers to be sorted
+  const sortedSigners = unsortedSigners.sort(compareSigners)
+
+  //This is used to make a new copy of the transaction with ALL signers
+  const signedTx = _.assign({}, transactions[0], {Signers: sortedSigners})
```
natenichols, line 58: Can we use `Object.assign()` here

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
+ * 
+ * @param signedTransactions A collection of the same transaction signed by different signers. The only difference
+ * between the elements of signedTransactions should be the Signers field.
+ * @returns An object with the combined transaction (now having a sorted list of all signers) which is encoded, along
+ * with a transaction id based on the combined transaction.
+ */
+function combine(signedTransactions: Array<string>): object {
+  validate.combine({signedTransactions})
+
+  const transactions: JsonObject[] = signedTransactions.map(binary.decode);
```
natenichols, line 73: Ahhh, I see why you're using that here. Disregard my early comments.

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
-  // tests and this code handle it as an array of objects. Fix!
-  const txs: any[] = _.map(signedTransactions, binary.decode)
-  const tx = _.omit(txs[0], 'Signers')
-  if (!_.every(txs, (_tx) => _.isEqual(tx, _.omit(_tx, 'Signers')))) {
-    throw new utils.common.errors.ValidationError(
-      'txJSON is not the same for all signedTransactions'
-    )
-  }
+function getATransactionWithAllSigners(transactions : Array<JsonObject>) : JsonObject {
   const unsortedSigners = _.reduce(
```
natenichols, line 49: Can we use `transactions.reduce` here. I want to try to add as little lodash as possible, as we're planning to remove it.

src/transaction/combine.ts
natenichols: This is totally fine, then, especially as you're not exposing this type to the client.

natenichols: Frankly, I made bad design decisions w/ those types I would love to go back and improve how ripple binary codec does types

natenichols (approved):
Awesome work @JST5000. Left a couple requests for change, but nothing that is important enough to hold up approval.

LGTM 👍 (pending those changes)

src/transaction/combine.ts
mvadari: we're about to do that ;)

mvadari (approved):
lgtm! just make sure you do what Nathan's comments said, and you're good to go

intelliot (changes requested):
This looks good. For the single-line comments with `//`, it would be ideal to put a space between the `//` and the text, as that's the way all the other comments in the project are styled (I believe). In the future, this is hopefully auto-fixable by the linter.

And as Nate mentioned, we would prefer to move away from using lodash `_` if possible.

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
 import {computeBinaryTransactionHash} from '../common/hashes'
+import {JsonObject} from 'ripple-binary-codec/dist/types/serialized-type'
+
+//The transactions should all be equal EXCEPT for the Signers field
+function validateAllSignedTransactionsAreEqual( transactions : Array<JsonObject> ) {
+  const tx = transactions[0];
+  const _txSignersTemp = tx.Signers;
+  delete tx.Signers
+  
+  transactions.forEach( _tx => {
```
ledhed2222, line 16: why the underscore before `_tx`? as a convention, a leading underscore should mean that you aren't using that argument in the function body, but in this case you are. please remove

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
+      const _txSignersTemp = _tx.Signers
+      delete _tx.Signers;
```
ledhed2222, line 18: there's no reason to delete this since you are re-assigning it later. using `delete` is weird

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
+      
+      if ( JSON.stringify(tx) !== JSON.stringify(_tx)) {
```
ledhed2222, line 20: errant space inside paren

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
 import {decodeAccountID} from 'ripple-address-codec'
 import {validate} from '../common'
 import {computeBinaryTransactionHash} from '../common/hashes'
+import {JsonObject} from 'ripple-binary-codec/dist/types/serialized-type'
+
+//The transactions should all be equal EXCEPT for the Signers field
+function validateAllSignedTransactionsAreEqual( transactions : Array<JsonObject> ) {
+  const tx = transactions[0];
+  const _txSignersTemp = tx.Signers;
+  delete tx.Signers
```
ledhed2222, line 14: no reason to delete this since you reassign it later

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
+import {ValidationError} from '../common/errors'
 import {decodeAccountID} from 'ripple-address-codec'
 import {validate} from '../common'
 import {computeBinaryTransactionHash} from '../common/hashes'
+import {JsonObject} from 'ripple-binary-codec/dist/types/serialized-type'
+
+//The transactions should all be equal EXCEPT for the Signers field
+function validateAllSignedTransactionsAreEqual( transactions : Array<JsonObject> ) {
+  const tx = transactions[0];
+  const _txSignersTemp = tx.Signers;
```
ledhed2222, line 13: leading underscore is confusing and indicates that the variable is not going to be used, but it will

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
-import * as utils from './utils'
 import BigNumber from 'bignumber.js'
+import {ValidationError} from '../common/errors'
 import {decodeAccountID} from 'ripple-address-codec'
 import {validate} from '../common'
 import {computeBinaryTransactionHash} from '../common/hashes'
+import {JsonObject} from 'ripple-binary-codec/dist/types/serialized-type'
+
+//The transactions should all be equal EXCEPT for the Signers field
+function validateAllSignedTransactionsAreEqual( transactions : Array<JsonObject> ) {
```
ledhed2222, line 11: why these extra spaces inside parens?

src/transaction/combine.ts
ledhed2222: should have have a space after `//`

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
+      
+      if ( JSON.stringify(tx) !== JSON.stringify(_tx)) {
+          throw new ValidationError('txJSON is not the same for all signedTransactions')
+      }
+
+      _tx.Signers = _txSignersTemp
+  })
+
+  tx.Signers = _txSignersTemp
+  return transactions
```
ledhed2222, line 28: overall it seems like this can be re-written much simpler like so:

```javascript
function validateAllSignedTransactionsAreEqual(transactions : Array<JsonObject>) {
  if (transactions.every(tx => JSON.stringify(tx) === JSON.stringify(transactions[0])) {
    return transactions
  }
  throw new ValidationError('txJSON is not the same for all signedTransactions')
}
```

src/transaction/combine.ts
ledhed2222: ah nevermind i see why you're doing this. but there's a better way - see my suggestion below!

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
   const unsortedSigners = _.reduce(
-    txs,
+    transactions,
     (accumulator, _tx) => accumulator.concat(_tx.Signers || []),
     []
   )
-  const signers = unsortedSigners.sort(compareSigners)
-  const signedTx = _.assign({}, tx, {Signers: signers})
-  const signedTransaction = binary.encode(signedTx)
+  //The documentation requires signers to be sorted
```
ledhed2222, line 54: space after `//`

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
+  const sortedSigners = unsortedSigners.sort(compareSigners)
+
+  //This is used to make a new copy of the transaction with ALL signers
```
ledhed2222, line 57: space after `//`

src/transaction/combine.ts
ledhed2222: also, since you have a function that is validating this condition you can rewrite this comment to be more clear in terms of what the function is doing, as opposed to the abstract requirement. See my suggested function below.

src/transaction/combine.ts
ledhed2222: this is also better described as a `map`.

src/transaction/combine.ts
ledhed2222: there's no benefit to assigning `unsortedSigners` to a variable, since it will only be used once. I think that makes your code unnecessarily verbose. instead:

```javascript
function getATransactionWithAllSigners(transactions : Array<JsonObject>) : JsonObject {
  // Signers must be sorted
  const sortedSigners = transactions
    .map(tx => tx.Signers)
    .filter(signer => signer)
    .sort(compareSigners)

  return Object.assign({}, {...transactions[0], Signers: sortedSigners})
}
```

src/transaction/combine.ts
```diff
@@ -1,42 +1,80 @@
...
+ * @param signedTransactions A collection of the same transaction signed by different signers. The only difference
+ * between the elements of signedTransactions should be the Signers field.
+ * @returns An object with the combined transaction (now having a sorted list of all signers) which is encoded, along
+ * with a transaction id based on the combined transaction.
+ */
+function combine(signedTransactions: Array<string>): object {
+  validate.combine({signedTransactions})
+
+  const transactions: JsonObject[] = signedTransactions.map(binary.decode);
+  validateAllSignedTransactionsAreEqual(transactions)
```
ledhed2222, line 74: if you're not using the return value of `transactions` then don't return anything from this. updated proposed function above.

ledhed2222 (changes requested):
using functional styles will let you further simplify this :) see comments

src/transaction/combine.ts
JST5000: That's much better! 

The "...object" syntax is super useful for this use case :)

src/transaction/combine.ts
JST5000: That looks cleaner to me :) 
(Unfortunately I still had to use lodash to get flatMap as it isn't built in, but it's much more readable than assign!)

src/transaction/combine.ts
```diff
@@ -1,42 +1,63 @@
...
   return addressToBigNumber(a.Signer.Account).comparedTo(
     addressToBigNumber(b.Signer.Account)
   )
 }
 
+function getATransactionWithAllSigners(transactions: Array<JsonObject>): JsonObject {
+  // Signers must be sorted - see compareSigners for more details
+  const sortedSigners = _.flatMap(transactions, tx => tx.Signers)
+    .filter(signer => signer)
+    .sort(compareSigners)
```
ledhed2222, line 41: check whether you can do `map` followed by `flat` if there's no polyfill for `flatMap`. otherwise, this is a good reason to use lodash or something similar

JST5000: "flat" is also not appearing in the polyfill unfortunately, so I think lodash is probably the best bet

src/transaction/combine.ts
```diff
@@ -1,42 +1,63 @@
...
-  const signers = unsortedSigners.sort(compareSigners)
-  const signedTx = _.assign({}, tx, {Signers: signers})
-  const signedTransaction = binary.encode(signedTx)
+  const transactions: JsonObject[] = signedTransactions.map(binary.decode);
+  validateAllSignedTransactionsAreEqual(transactions)
+
+  const oneTxWithAllSigners = getATransactionWithAllSigners(transactions)
+  const signedTransaction = binary.encode(oneTxWithAllSigners)
   const id = computeBinaryTransactionHash(signedTransaction)
   return {signedTransaction, id}
```
ledhed2222, line 62: i don't think you gain readability from assigning these variables. i think this is actually clearer, especially because the keys of the return are descriptive

```javascript
validateAllSignedTransactionsAreEqual(transactions)

const signedTransaction = binary.encode(getATransactionWithAllSigners(transactions))
return {
  id: computeBinaryTransactionHash(signedTransaction),
  signedTransaction,
}
```

JST5000: Yep, I'd agree with that. The function name makes it clear what it's doing.

src/transaction/combine.ts
```diff
@@ -1,41 +1,63 @@
...
+import {ValidationError} from '../common/errors'
 import {decodeAccountID} from 'ripple-address-codec'
 import {validate} from '../common'
 import {computeBinaryTransactionHash} from '../common/hashes'
+import {JsonObject} from 'ripple-binary-codec/dist/types/serialized-type'
+
+/**
+ * The transactions should all be equal EXCEPT for the 'Signers' field. 
+ */
+ function validateAllSignedTransactionsAreEqual(transactions: Array<JsonObject>) {
```
ledhed2222, line 13: you've got some very long function names that could be terser without losing detail. 

this could be `validateSignedTransactionEquality` (probably shorter still but a start)

src/transaction/combine.ts
```diff
@@ -1,41 +1,63 @@
...
+ * the numeric value of the signer addresses, with the lowest value first. 
+ * (If submitted as JSON, the submit_multisigned method handles this automatically.)
+ */
 function compareSigners(a, b) {
   return addressToBigNumber(a.Signer.Account).comparedTo(
     addressToBigNumber(b.Signer.Account)
   )
 }
 
+function getATransactionWithAllSigners(transactions: Array<JsonObject>): JsonObject {
```
ledhed2222, line 37: how about `firstTransactionWithAllSigners`

src/transaction/combine.ts
ledhed2222: actually - the `Signed` part is confusing, since we are validating transactions _regardless_ of signer. So it could just be `validateTransactionEquality`.

JST5000: Good point!

src/transaction/combine.ts
JST5000: I think 'getA' conveys the intent better than 'first' for this one.

src/transaction/combine.ts
JST5000: Checked out the compiler errors caused by the flatMap configuration change, but it seemed like it would take me a long time to figure out the root cause and debug it all. Didn't seem worth it for removing one instance of lodash in a file which already is using lodash elsewhere. (If you want to see the config change, you can see it in the commit c7f2d28)

ledhed2222: I mean I'm team lodash lol so I'm ok with this 😹

src/transaction/combine.ts
ledhed2222: there's something about the `ATransaction` in camel case that makes this hard to read. Is it "a transaction"? Or an "a-class transaction"? it looks strange to me. what about just `getTransactionWithAllSigners`

src/transaction/combine.ts
```diff
@@ -1,43 +1,66 @@
...
-import * as utils from './utils'
 import BigNumber from 'bignumber.js'
+import {ValidationError} from '../common/errors'
 import {decodeAccountID} from 'ripple-address-codec'
 import {validate} from '../common'
 import {computeBinaryTransactionHash} from '../common/hashes'
+import {JsonObject} from 'ripple-binary-codec/dist/types/serialized-type'
+
+/**
+ * The transactions should all be equal EXCEPT for the 'Signers' field.
```
ledhed2222, line 11: i don't think "except" needs to be all-caps in this case. the all-caps implies that a caller could do something wrong if they don't do whatever the all-caps indicates, but in this case this is just what the function does, NBD to user.

src/transaction/combine.ts
```diff
@@ -1,43 +1,66 @@
...
 function addressToBigNumber(address) {
   const hex = Buffer.from(decodeAccountID(address)).toString('hex')
   return new BigNumber(hex, 16)
 }
 
+/**
+ * According to the documentation on multi-signatures:
+ * If presented in binary form, the Signers array must be sorted based on 
+ * the numeric value of the signer addresses, with the lowest value first. 
+ * (If submitted as JSON, the submit_multisigned method handles this automatically.)
```
ledhed2222, line 29: Link to the documentation and remove "According to the documentation"

src/transaction/combine.ts
ledhed2222: if you're in here already, "equivalence" would be even more descriptive than "equality".

src/transaction/combine.ts
mvadari: It's totally fine to use lodash's `flatMap`, it's used elsewhere too I think

src/transaction/combine.ts
mvadari: Yeah I like that, or `getTransactionAllSigners`

src/transaction/combine.ts
mvadari: +1

src/transaction/combine.ts
mvadari: You generally don't put articles (a, an, the) in method names, unless they're serving some sort of specific purpose

src/transaction/combine.ts
JST5000: Yep, that makes sense :)

intelliot (approved):
LGTM. Thanks for taking this on!

PR #1497, opened 2021-08-02
Generate faucet wallet Testnet and Devnet

florent-uzio:

Solves https://github.com/ripple/ripple-lib/issues/1484
New function to easily generate a faucet wallet from Testnet and Devnet (https://xrpl.org/xrp-testnet-faucet.html).

### Context of Change

New feature to give the user the ability to easily create a pre-funded wallet on the Testnet and Devnet with 1000 XRP.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

The tests check that the correct faucet URL is used based on the user input.


test/fixtures/responses/index.js
```diff
@@ -1,7 +1,7 @@
-'use strict'; // eslint-disable-line strict
+'use strict' // eslint-disable-line strict
 
 function buildList(options) {
-  return new Array(options.count).fill(options.item);
+  return new Array(options.count).fill(options.item)
```
mvadari, line 4: Why are the semicolons removed here?

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  return axios
+    .post<FaucetWallet>(faucetUrl)
+    .then((response) => {
+      return response.data
+    })
+    .catch((error) => {
+      throw new errors.XRPLFaucetError(error.data)
+    })
+}
```
mvadari, line 32: This method isn't going to be very robust, because it takes a few seconds for the transaction to be confirmed on the ledger, so you can't use it immediately. See https://github.com/XRPLF/xrpl-py/blob/master/xrpl/asyncio/wallet/wallet_generation.py#L26

yarn.lock
```diff
@@ -3,21 +3,21 @@
 
 
 "@babel/code-frame@^7.0.0", "@babel/code-frame@^7.14.5":
```
mvadari, line 5: Why is there such a big diff here?

mvadari (changes requested):

test/fixtures/responses/index.js
florent-uzio: I think this is due to the `prettier`  configuration which indicates that the semicolons are never placed at the end of the rows:

```
{
  "parser": "typescript",
  "printWidth": 80,
  "tabWidth": 2,
  "semi": false,   <== Here
  "singleQuote": true,
  "trailingComma": "none",
  "quoteProps": "consistent",
  "bracketSpacing": false
}
```

src/wallet/wallet-generation.ts
florent-uzio: Yes, the function returns a `Promise` so my idea was that the user would use `async/await` to get the Wallet information (or use `.then()`).
What do you think?

yarn.lock
florent-uzio: That's a good question... I tried with Node 12 and Node 14 and every time there was a huge difference.
What Node version do you use?

src/wallet/wallet-generation.ts
florent-uzio: So, do you mean that when you get the response from the POST request, the response includes a balance of 1000 XRP but it can happen that the XRPL account doesn't have 1000 XRP yet?

yarn.lock
mvadari: The diff looks fine, the dependabot PRs also have a massive diff so idk what changed but clearly something did.

src/wallet/wallet-generation.ts
mvadari: Yeah exactly.

test/fixtures/responses/index.js
mvadari: That seems like a bad configuration. I guess it doesn't matter now anyways, because @ledhed2222 is working on setting up better linting in the repo anyways.

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Returns the URL of the faucet that should be used.
+ *
+ * @param onTestnet - Boolean to indicate which URL to retrieve.
+ * @returns - The URL of the matching faucet.
+ */
+export const getFaucetUrl = (onTestnet: boolean): string => {
```
natenichols, line 40: Could we maybe define the network with an `enum`?

```ts
enum Faucet {
    Testnet
    Devnet
}
```

docs/index.md
```diff
@@ -47,6 +47,8 @@
   - [xAddressToClassicAddress](#xaddresstoclassicaddress)
   - [renameCounterpartyToIssuer](#renamecounterpartytoissuer)
   - [formatBidsAndAsks](#formatbidsandasks)
+- [XRPL Faucets](#xrpl-faucets)
```
intelliot, line 50: I don't think we can edit docs/index.md directly. It is a generated file (using the docgen script): https://github.com/ripple/ripple-lib#generating-documentation

intelliot (changes requested):

docs/index.md
florent-uzio: Thanks I removed the modification I made before.

src/wallet/wallet-generation.ts
florent-uzio: The function doesn't accept an `enum` anymore but I kept the `enum` to use it elsewhere.

src/wallet/wallet-generation.ts
florent-uzio: Cool thank you, I learnt a lot here.
I made a lot of changes to make sure the account is funded before the function returns the faucet account similar to xrpl-py. Let me know what you think.

yarn.lock
florent-uzio: yeah, that creates merge conflicts... I need to find out how to solve that...
Do you have this big difference as well?

src/api.ts
```diff
@@ -399,6 +416,9 @@ class RippleAPI extends EventEmitter {
   computeLedgerHash = computeLedgerHash // @deprecated Invoke from top-level package instead
   signPaymentChannelClaim = signPaymentChannelClaim // @deprecated Invoke from top-level package instead
   verifyPaymentChannelClaim = verifyPaymentChannelClaim // @deprecated Invoke from top-level package instead
+
+  generateFaucetWallet = generateFaucetWallet
```
natenichols, line 420: If we're using `axios` to make a HTTP request to the faucet, IMO it doesn't make sense to include this in the `RippleAPI` class. 

Could we add this function to the utils we are exporting?

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,198 @@
...
+  balance: number
+}
+
+export enum FaucetNetwork {
+  Testnet = 'https://faucet.altnet.rippletest.net/accounts',
+  Devnet = 'https://faucet.devnet.rippletest.net/accounts'
+}
+
+const INTERVAL_SECONDS = 1 // Interval to check an account balance
+const MAX_ATTEMPS = 20 // Maximum attemptsto retrieve a balance
```
mvadari, line 21:
```suggestion
const MAX_ATTEMPTS = 20 // Maximum attempts to retrieve a balance
```

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,198 @@
...
+
+/**
+ * Generates a random wallet with some amount of XRP (usually 1000 XRP).
+ *
+ * @param wallet - An existing XRPL wallet to fund, if undefined, a new wallet will be created.
+ * @returns - A Wallet on the Testnet or Devnet that contains some amount of XRP.
+ */
+async function generateFaucetWallet(
+  this: RippleAPI,
+  wallet?: string
```
natenichols, line 31: can we name this `address`, if the HTTP request takes an optional Classic Address?

src/api.ts
mvadari, line 420: It needs the rippled connection to make sure the wallet has been funded properly.

src/wallet/wallet-generation.ts
mvadari, line 31: If `wallet` is just supposed to be the address, I think it should be renamed to something like `address` - `wallet` implies it has something to do with the keys

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  // Based on the connection URL, retrieve the correct faucet URL
+  const faucetNetwork = getFaucetUrl(this)
+
+  return new Promise((resolve, reject) => {
+    axios
+      .post<FaucetWallet>(faucetNetwork, body)
+      .then(async (response) => {
+        address = response.data.account.classicAddress
```
mvadari, line 72: this shouldn't be using the same variable as `address` above - it's confusing

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,198 @@
...
+ * Get the faucet URL based on the RippleAPI connection
+ * @param api - RippleAPI
+ * @returns A {@link FaucetNetwork}
+ */
+export function getFaucetUrl(api: RippleAPI) {
+  const connectionUrl = api.connection.getUrl()
+
+  // 'altnet' for Ripple Tesnet server and 'testnet' for XRPL Labs Testnet server
+  if (connectionUrl.includes('altnet') || connectionUrl.includes('testnet')) {
+    return FaucetNetwork.Testnet
```
mvadari, line 188: IMO this can just be constants, I don't think you really gain anything from having this be an `Enum` (but if this is good practice in JS then I'm happy to leave it alone)

src/wallet/wallet-generation.ts
natenichols, line 72: I prefer to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators#optional_chaining_operator) when parsing RPC responses in TypeScript.

It makes it such that if the response is formed differently than expected, the value of `address` is just undefined, as opposed to throwing an error.

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,198 @@
...
+ * @param api - RippleAPI
+ * @param address - Wallet address.
+ * @param currency - Currency to search, default XRP.
+ * @returns
+ */
+async function getWalletBalanceByCurrency(
+  api: RippleAPI,
+  address: string,
+  currency: string = 'XRP'
+): Promise<string> {
```
mvadari, line 112: This method doesn't have to be general to any currency - it's only being used for XRP

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,198 @@
...
+            address,
+            startingBalance
+          )
+
+          if (isAcctFunded) {
+            resolve(response.data)
+          } else {
+            reject(
+              new errors.XRPLFaucetError(
+                `Unable to fund address with faucet after waiting ${INTERVAL_SECONDS} seconds`
```
mvadari, line 86: This seems a bit of a small interval, given that it takes longer than this to validate a ledger

src/wallet/wallet-generation.ts
natenichols: ☝️

src/wallet/wallet-generation.ts
natenichols: I think `INTERVAL_SECONDS` is the wrong thing to report here...

`await hasBalanceIncreased(...)` can retry up to 20 times, meaning you could wait up to 20 seconds, if I'm understanding your code correctly...

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,198 @@
...
+      }
+
+      try {
+        const newBalance = +(await getWalletBalanceByCurrency(api, address))
+        if (newBalance > originalBalance) {
+          clearInterval(interval)
+          resolve(true)
+        }
+      } catch (err) {
+        reject(
```
natenichols, line 168: Should we also `clearInterval` here before we reject this promise.

natenichols: Or do we want it to keep retrying, even if it fails to retrieve balance one time.

src/wallet/wallet-generation.ts
natenichols: Do we want to add an optional `url?: string` field that allows users to pass in their own Faucet URL? For use w/ other faucets (like a Hooks devnet faucet).

src/api.ts
florent-uzio: Hi @mvadari I'm not sure to get what you mean here. Is there something missing in the code?

src/wallet/wallet-generation.ts
florent-uzio: Thanks for spotting that!

src/wallet/wallet-generation.ts
florent-uzio: I'll rename `wallet` to `address`.

I don't mind adding a `url` field as well.

src/wallet/wallet-generation.ts
florent-uzio: Ah yes good catch, I will correct that

src/wallet/wallet-generation.ts
florent-uzio: Agree, I will rename it to `classicAddress`.

src/wallet/wallet-generation.ts
florent-uzio: Thanks, that's a good input.
I did several tests and if the address balance isn't retrieved, the function `getAddressXrpBalance` handles the error. The `catch` block you mentioned here will not be triggered. 
It will be triggered if what happens after `const newBalance = +(await getWalletBalanceByCurrency(api, address))` creates an error.
So I think it's good to add `clearInterval` here.
Let me know if that doesn't make sense.

src/wallet/wallet-generation.ts
florent-uzio: I agree. 
It's a bit easier to manipulate the faucet URLs with an `enum` because I use it in the tests as well.
And it's easier to have one `enum` import than two `const` import :)
it's a small thing but I think it's ok to leave the `enum` like that.

src/wallet/wallet-generation.ts
florent-uzio: Yeah I agree.
Here only `classicAddress` is optional. `account` will always exist.

So I don't think we can use `?` here as `classicAddress` is at the "end" of the chain but what I did is to check if `classicAddress` is `undefined` in an `if else` block.

src/wallet/wallet-generation.ts
florent-uzio: Yes, I removed the currency field

yarn.lock
florent-uzio: I fixed the conflicts. I must have used `npm i` in the past which created those issues. It should be good now, Axios is added to yarn.lock.

natenichols (approved):
LGTM 👍

intelliot (commented):
@florent-uzio How bad would it be to use Node.js's built-in https module, instead of axios?

Here is an example: https://github.com/ripple/ripple-lib/blob/develop/snippets/src/reliableTransactionSubmission.ts#L156-L200

We could probably mostly copy that, maybe with minor changes..

(The benefit would be reducing our external dependencies by 1)

florent-uzio:
Hi @intelliot , yes it's possible to use the native Node.js modules. I've implemented it, let me know what you think.

I had to add a `devDependency`: `https-browserify`

Thanks

webpack.config.js
```diff
@@ -1,124 +1,138 @@
...
+    },
+    resolve: {
+      extensions: ['.js', '.json'],
+      fallback: {
+        buffer: require.resolve('buffer/'),
+        assert: require.resolve('assert/'),
+        url: require.resolve('url/'),
+        stream: require.resolve('stream-browserify'),
+        crypto: require.resolve('crypto-browserify'),
+        https: require.resolve('https-browserify'),
```
intelliot, line 44: Is this the only change in this file? Could we revert/undo the formatting changes? I think it will be safer/cleaner to avoid formatting changes, and consider doing them in a later PR that *only* changes formatting.

intelliot (approved):
I would prefer not to change the formatting of `webpack.config.js` and `test/fixtures/responses/index.js` since it is not truly necessary. (Unfortunately you will need to undo/revert the changes caused by Prettier, at least for now) But everything else looks good.

intelliot:
To be clear: I am fine with using `axios` if it's worth the cost, and you prefer it. I just want to make sure we also considered the "direct" way (`axios` also uses `https` under the hood)

src/api.ts
mvadari: This was in response to someone else's comment which isn't visible anymore for some reason.

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,218 @@
...
+  balance: number
+}
+
+export enum FaucetNetwork {
+  Testnet = 'faucet.altnet.rippletest.net',
+  Devnet = 'faucet.devnet.rippletest.net'
+}
+
+const INTERVAL_SECONDS = 1 // Interval to check an account balance
+const MAX_ATTEMPS = 20 // Maximum attempts to retrieve a balance
```
mvadari, line 20:
```suggestion
const MAX_ATTEMPTS = 20 // Maximum attempts to retrieve a balance
```

mvadari (approved):
One minor typo, but lgtm otherwise. Thanks!

src/wallet/wallet-generation.ts
florent-uzio: Thanks I will update that

florent-uzio:
Hi @intelliot , I changed the formatting of the files so the changes should be more obvious now.

I think using `https` here is fine. We are only sending HTTP requests in this section of ripple-lib.
If we have more HTTP requests in the future I think we should move to `axios` as it's a much easier library to use (in my view).

Let me know if all seems ok for you and I can merge a bit later today or tomorrow morning.

src/wallet/wallet-generation.ts
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * Get the faucet URL based on the RippleAPI connection
+ * @param api - RippleAPI
+ * @returns A {@link FaucetNetwork}
+ */
+export function getFaucetUrl(api: RippleAPI) {
+  const connectionUrl = api.connection.getUrl()
+
+  // 'altnet' for Ripple Tesnet server and 'testnet' for XRPL Labs Testnet server
```
intelliot, line 206:
```suggestion
  // 'altnet' for Ripple Testnet server and 'testnet' for XRPL Labs Testnet server
```

intelliot (commented):
I think we will also need docs. Could be a separate PR, though.

I wonder if we should consider the Hooks Testnet as well: https://hooks-testnet.xrpl-labs.com/

florent-uzio:
@intelliot , how can we use the Hooks Testnet: https://hooks-testnet.xrpl-labs.com/ ?

intelliot:
Looking at the site, it seems like it makes a POST request.

Request URL: https://hooks-testnet.xrpl-labs.com/newcreds

It's up to you. I don't think it's really necessary for this PR

intelliot (approved):

florent-uzio:
Yes I just found out after I asked the question hehe.

I think we can skip it for now, as you said I don't think it's really necessary now. We can always support it in another PR.

I will merge soon then.

PR #1498, opened 2021-08-02
Defines TypeScript types for rippled account method requests & responses

mvadari:

This PR creates new TypeScript types for all rippled account methods (https://xrpl.org/account-methods.html), both the request shapes and response shapes.

### Context of Change

xrpl.js v2.0 models

### Type of Change

## Test Plan


src/models/common/index.ts
```diff
@@ -11,10 +11,15 @@ export interface IssuedCurrency {
   issuer: string
 }
 
+export interface Balance {
+  currency: string
+  value: string
+}
+
```
ledhed2222, line 18: Whta's a balance?

mvadari: I made up a name for the object, happy to change it/move it. This type is used by `gateway_balances` and probably nowhere else.

ledhed2222: if it's only used by `GatewayBalances` then shouldn't it be defined in some `gatewayBalances.ts` file and _not_ exported?

src/models/methods/account_channels.ts
```diff
@@ -0,0 +1,39 @@
...
+    result: {
+        account: string
+        channels: Channel[]
+        ledger_hash: string
+        ledger_index: number
+        validated?: boolean
+        limit?: number
+        marker?: any
+    }
+}
```
ledhed2222, line 39: we should be using camel case for file names in JS/TS

src/models/methods/account_currencies.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface AccountCurrenciesResponse extends BaseResponse {
+    result: {
+        ledger_hash?: string
+        ledger_index: number
+        receive_currencies: string[]
+        send_currencies: string[]
+        validated: boolean
+    }
+}
```
ledhed2222, line 21: camelcase file names

src/models/methods/account_info.ts
```diff
@@ -0,0 +1,41 @@
...
+export interface AccountCurrenciesResponse extends BaseResponse {
+  result: {
+    account_data: AccountRoot
+    signer_lists?: SignerList[]
+    ledger_current_index?: number
+    ledger_index?: number
+    queue_data?: QueueData
+    validated?: boolean
+  }
+}
```
ledhed2222, line 41: camelcase

src/models/methods/account_offers.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export interface AccountOffer {
+  flags: number
+  seq: number
+  taker_gets: Amount
+  taker_pays: Amount
+  quality: string
+  expiration?: number
+}
```
ledhed2222, line 21: why is this one exported?

mvadari: Unintentional, it's fixed now

src/models/common/transaction.ts
```diff
@@ -0,0 +1,38 @@
...
+        LedgerIndex: string
+        NewFields: {[field: string]: any}
+    }
+}
+
+interface ModifiedNode {
+    ModifiedNode: {
+        LedgerEntryType: string
+        LedgerIndex: string
+        FinalFields: {[field: string]: any}
```
natenichols, line 15: Is there any way we can narrow this from `{[field: string]: any`? Does this come in the form of a `LedgerEntry` that we can use to replace `any` here?

src/models/methods/baseMethod.ts
```diff
@@ -0,0 +1,24 @@
...
+
+interface Warning {
+    id: number
+    message: string
+    details?: {[key: string]: string}
+}
+
+export interface BaseResponse {
+    id: number | string
+    status: string
```
natenichols, line 15: Can status be anything other than `success` or `error`?

src/models/methods/baseMethod.ts
```diff
@@ -0,0 +1,24 @@
...
+    type: string
+    result: any
+    warning?: "load"
+    warnings?: Warning[]
+    forwarded?: boolean
+    error?: string
+    request?: any
```
natenichols, line 22: Could this be type `BaseResponse`? IIRC, request just echos back the request sent to the server...

src/models/methods/baseMethod.ts
mvadari: I'm not entirely sure, which is why I left it as `string` - maybe @intelliot knows

src/models/methods/baseMethod.ts
mvadari: Good point

mvadari: No, it would actually have to be `Response`

src/models/common/transaction.ts
mvadari: You could in theory narrow it, but as far as I understand, it's a subset of `LedgerEntry` fields, and isn't necessarily a full `LedgerEntry` object. I suppose this is another question for @intelliot

natenichols (approved):
LGTM 👍 , given the two open questions for @intelliot are answered before merging.

src/models/common/transaction.ts
intelliot: This is probably any (modifiable) fields in any ledger object, so I believe there are a lot of possibilities, and it'll expand in the future as more kinds of ledger objects are added in the future. I would fine with leaving this as-is, but of course, if we want to make this more specific in the future, that would be fine as well.

src/models/methods/baseMethod.ts
intelliot: I don't know. I think this is fine as-is, or `"success" | string` if you want to document the fact that 'success' is the happy path value

PR #1499, opened 2021-08-02
Defines TypeScript types for all ledger objects

mvadari:

This PR makes new updated models for all XRPL ledger objects (https://xrpl.org/ledger-object-types.html).

### Context of Change

xrpl.js v2.0

### Type of Change

## Test Plan

This PR doesn't touch existing code, so all tests should still be fine.


src/models/ledger/pay_channel.ts
```diff
@@ -0,0 +1,18 @@
...
+  SettleDelay: number
+  OwnerNode: string
+  PreviousTxnID: string
+  PreviousTxnLgrSeq: number
+  Flags: number
+  Expiration?: number
+  CancelAfter?: number
+  SourceTag?: number
+  DestinationTag?: number
+  index: string
```
JST5000, line 17: Seems like you might be missing `DestinationNode?: string` (Last entry here: https://xrpl.org/paychannel.html#paychannel)

src/models/ledger/ripple_state.ts
```diff
@@ -0,0 +1,17 @@
+import { Amount } from "../common";
+
+export interface RippleState {
+  LedgerEntryType: 'RippleState'
+  Flags: number
+  Balance: Amount
+  LowLimit: Amount
+  HighLimit: Amount
```
JST5000, line 8: The RippleState documentation only says that fields like Balance can be Amounts (not strings). The Amount type you have here though can be a string. (Not sure if that's just an inconsistency in the documentation though) Should these Amount types be limited to just the IssuedCurrencyAmount object?

JST5000: Link to documentation for reference: https://xrpl.org/ripplestate.html#example-ripplestate-json

src/models/ledger/signer_list.ts
```diff
@@ -0,0 +1,17 @@
+interface SignerEntry {
+  SignerEntry: {
+    Account: string
+    SignerWeight: number
+  }
+}
```
JST5000, line 6: Is this supposed to have a nested 'SignerEntry:'? (I didn't see that in the other interfaces you used for sub-interfaces, but I might be missing something)

Link to documentation for reference: https://xrpl.org/signerlist.html#signer-entry-object

JST5000 (changes requested):
Seems like there's just one or two spots that might need changing, but after that looks good!

(For reference, I went through each file with the documentation and checked each field just to be sure, so maybe future reviewers don't have to duplicate that work if they don't want to :P)

src/models/ledger/signer_list.ts
mvadari: It's kind of weird, but this one does (this setup is in a couple of other places as well). You can see it in the examples.

src/models/ledger/ripple_state.ts
mvadari: Good catch!

src/models/ledger/deposit_preauth.ts
```diff
@@ -0,0 +1,9 @@
+export interface DepositPreauth {
+    LedgerEntryType: 'DepositPreauth'
+    Account: string
+    Authorize: string
+    Flags: 0
+    OwnerNode: string
+    PreviousTxnID: string
+    PreviousTxnLgrSeq: number
+  }
```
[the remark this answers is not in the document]
mvadari: Turns out all the objects have an `index` field because they're ledger entry objects, so I created a superclass

JST5000 (approved):
Looks good!

natenichols (approved):

PR #1500, opened 2021-08-03
build: include lodash in webpack build

natenichols:
Stop excluding lodash from webpack build.

### Context of Change
Previously, we would webpack ripple-latest without lodash, as it is a frequently used library. This really only saves ~80KB, which is not worth the hassle of not packaging these together.

### Type of Change
- [x] Build 
- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/test/localintegrationrunner.html
+++ b/test/localintegrationrunner.html
@@ -9,7 +9,6 @@
     <div id="mocha"></div>
     <script src="../node_modules/mocha/mocha.js"></script>
     <script src="hacks/phantomhacks.js"></script>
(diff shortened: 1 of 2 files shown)


natenichols:
I am dumb and it is getting late. When squashing this PR, please rename the commit message to 
```
build: include lodash in webpack build
```

mvadari (approved):

intelliot:
Curious if @mDuo13 happens to have any opinion on this

intelliot (approved):
I'm fine with this. Not needing lodash is a "feature" so we could release this as 1.10 😃

PR #1503, opened 2021-08-04
Implements TypeScript types for path and order book methods

mvadari:

This PR creates new TypeScript types for all rippled path and order book methods (https://xrpl.org/path-and-order-book-methods.html), both the request shapes and response shapes.

### Context of Change

xrpl.js v2.0 models

### Type of Change

## Test Plan


src/models/methods/pathFind.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export interface PathFindResponse extends BaseResponse {
+  result: {
+      alternatives: PathOption[]
+      destination_account: string
+      destination_amount: Amount
+      source_account: string
+      full_reply: boolean
+      id: number | string
```
JST5000, line 40: Will we need to update the documentation when 2.0 comes out to limit id's to numbers/strings? (It seems websocket supports any unique object, but we have a convention that is partially applied)

src/models/methods/ripplePathFind.ts
```diff
@@ -0,0 +1,31 @@
...
+    issuer?: string
+  }
+
+export interface RipplePathFindRequest extends BaseRequest {
+    command: "ripple_path_find"
+    source_account: string
+    destination_account: string
+    destination_amount: Amount
+    send_max?: Amount
+    source_currencies: SourceCurrencyAmount
```
JST5000, line 15: It seems like source_currencies should be optional - https://xrpl.org/ripple_path_find.html#ripple_path_find

JST5000 (commented):
Just one line seems like it should be optional, other than that good to go!

src/models/methods/pathFind.ts
natenichols: `id` should be optional IIRC, and a member of `BaseRequest`/`BaseResponse`

```ts
id?: number | string
```

I think `id` can be anything (so we could type it as `any`), and rippled just echoes it back to you, but I dont really see the harm in restricting it to a `number | string`.

natenichols (approved):
Left one comment, but nothing I really felt should hold up my approval.

src/models/methods/ripplePathFind.ts
mvadari: Good catch, thanks

src/models/methods/pathFind.ts
JST5000: @natenichols in this specific case id is echoed in the response, in addition to the id? at the top level in BaseRequest/BaseResponse

JST5000 (approved):

PR #1504, opened 2021-08-04
Removed legacy exit code conversion

JST5000:

Removed an exit code we were supporting temporarily due to a ws library bug.

### Context of Change

Responding to a TODO left 2 years ago to get rid of the conversion on the next breaking change. 
We were converting exit code 1005 to 1000 to be backwards compatible with old versions of the ws library which were incorrectly exiting with 1000 instead of 1005. (Link to issue: https://github.com/websockets/ws/issues/1257)
 

### Type of Change

## Before / After

Before we were turning the exit code 1005 

## Test Plan

CI tests pass

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -163,11 +163,10 @@ class RippleAPI extends EventEmitter {
       })
       this.connection.on('disconnected', (code) => {
         let finalCode = code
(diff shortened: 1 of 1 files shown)


mvadari:
This PR can be opened if you merge it into `2.0` instead of `develop`, assuming you don't want this to go into 1.x.

mvadari (approved):

natenichols (approved):

PR #1505, opened 2021-08-04
Fixes ledger version in `getTrustlines`

### LINKED ISSUE #1501: Relations between transactions and balances
Hello, 

I have a question concerning some behaviors in the API that I find it hard to explain, 
more precisely between the relationship that the transactions and the balances of an account share over time.

From my point of view, I would always expect some transaction to alter the balances of an account depending on what is actually happening in the transaction, however this does not seem to be the logic taken by the API. Let's have a concrete (quite simple) example to explain what I mean.

Let's say we take this random account: rGBP6jP1a7pJF7AtnEUa5Q4T2dwZxcQaWG which contains a few transactions.

- First of all the first transaction involving this account is this one : 

AEFD4306CA9F3CBEB991D4898C94DEFB2EA88C348BE1311E5C6758E214E28FE7

we can see in this transaction that the given account receives 20 XRP from this address: raMNd49mErtMaJMA7QUXWE8eL6Ftt7LWQY and it is the only transaction that perform a balance change for our initial account within ledger 62965182 meaning that at this ledger only this transaction could have an effect on the balances of this account

Now if I am performing a getBalances for this this address at this ledger I get : 
```json
[
    { "currency": "XRP", "value": "20" },
    {
      "currency": "EUR",
      "counterparty": "rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq",
      "value": "0.468895474611"
    }
  ]
```

here we can see that the account owns some EUR that were never shown in the transaction that created it, how am I supposed to retrieve this information ? 

- Second, let's get a big further in time and jump to this transaction that the account made: 5FF49424A4314292DC534C3E5BDA15912B0542D8B7E492C15B40237A8E34D3CE

here we can see that the account ordered some BTC and the order has been partially-filled, however the getBalances for this account at ledger 62965412 does not show any BTC : 

```json
[
    { "currency": "XRP", "value": "919.999746" },
    {
      "currency": "EUR",
      "counterparty": "rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq",
      "value": "0.468895474611"
    }
  ]
```

Can you explain me what I am doing wrong with my analysis ?

mvadari:

`getTrustlines` wasn't using `ledgerVersion` in the options at all, and was instead getting the current ledger version. This PR fixes that.

### Context of Change

Fixes #1501 

### Type of Change

## Test Plan

Added a test to check for the correct behavior. CI passes.

### DIFF
--- a/src/ledger/trustlines.ts
+++ b/src/ledger/trustlines.ts
@@ -31,7 +31,7 @@ async function getTrustlines(
   // 2. Make Request
   const responses = await this._requestAll('account_lines', {
     account: address,
-    ledger_index: options.ledgerVersion,
+    ledger_index: options.ledgerVersion ?? await this.getLedgerVersion(),
     limit: options.limit,
     peer: options.counterparty
   })
--- a/src/ledger/trustlines.ts
+++ b/src/ledger/trustlines.ts
@@ -31,7 +31,7 @@ async function getTrustlines(
   // 2. Make Request
   const responses = await this._requestAll('account_lines', {
     account: address,
-    ledger_index: await this.getLedgerVersion(),
+    ledger_index: options.ledgerVersion, //await this.getLedgerVersion(),
     limit: options.limit,
     peer: options.counterparty
   })
--- a/src/ledger/trustlines.ts
+++ b/src/ledger/trustlines.ts
@@ -31,7 +31,7 @@ async function getTrustlines(
   // 2. Make Request
   const responses = await this._requestAll('account_lines', {
     account: address,
-    ledger_index: options.ledgerVersion, //await this.getLedgerVersion(),
+    ledger_index: options.ledgerVersion,
     limit: options.limit,
     peer: options.counterparty
   })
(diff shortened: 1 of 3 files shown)


src/ledger/trustlines.ts
```diff
@@ -31,7 +31,7 @@ async function getTrustlines(
   // 2. Make Request
   const responses = await this._requestAll('account_lines', {
     account: address,
-    ledger_index: await this.getLedgerVersion(),
+    ledger_index: options.ledgerVersion,
```
natenichols, line 34: I think we still want to default to `this.getLedgerVersion()` when no options is provided.

Can we do
```ts
const currentLedger = await this.getLedgerVersion()

const ledger_index = options.ledgerVersion ?? currentLedger

const responses = await this._requestAll('account_lines', {
    ........
    ledger_index,
    ........
}
```

natenichols (approved):
LGTM 👍

intelliot (approved):
LGTM

PR #1506, opened 2021-08-05
[For Discussion Only] prepare-snippet: improve type safety

intelliot:

We cannot avoid the "as Promise<T>" type assertion due to a limitation of generic narrowing. We can only narrow transaction; we cannot narrow T because it might be a union.

Note that if we pass in a BaseTransaction, we return a Promise<BaseTransaction> (i.e. not narrowed).

### Context of Change

Type-safe version of `prepare`

### Type of Change

## Before / After

Added `BaseTransaction` back and renamed the transaction union type to just `Transaction` (less verbose vs. `XrpLedgerTransaction`) so that we can return the same type of transaction that was passed in.


intelliot:
function overloading might be a good alternative to this

PR #1507, opened 2021-08-05
refactor: Add common transaction fields Typescript definition

natenichols:
Adds a base-transaction type for defining out transaction models.

Note that these changes are on top of @mvadari 's account-methods branch.

### Context of Change
This provides an interface for both the base fields that will be used by our `Typescript` transaction models, and optional functionality that allows for clients to validate the form of a transaction at runtime. By doing this, we will be able to remove `jsonschemas`

### Type of Change
- [ ] Documentation Updates
- [ ] Release
- [ ] 
## Test Plan
Adds tests for runtime verification of common fields in `verifyCommonFields`


intelliot:
Any ideas/proposals/experiments for how we will generate reference documentation?

natenichols:
Right now we use [doctoc](https://www.npmjs.com/package/doctoc), and I'll look into how well it supports typescript. 

If it don't like the support, it looks like something like [typedoc](https://github.com/TypeStrong/TypeDoc) is build specifically for typescript. I haven't looked into it much, but we can totally switch away from `doctoc` if we need to.

@intelliot Do you have opinions?

intelliot:
Nope, I'm not sure what options are available. The main concern is that the current reference documentation (particularly method parameters) is generated from the jsonschemas. Examples are usually sourced from the unit test fixtures. It's kind of cool, and it's potentially painful to change all that up.

natenichols:
Okay, so I’ve been playing around w/ typedoc + typedoc-markdown-plugin, and I think it does pretty much what we want.
It turns a function + JsDoc comment into MD

```ts
/**
 * Signs a transaction JSON
 * 
 * @param this - RippleAPI to use for networking
 * @param txJSON - Transaction encoded in hex string
 * @param secret - Seed to generate wallet to sign
 * @param options - Options describing how to signed
 * @param keypair - Keypari to sign wallet with
 * @returns Transaction hash and signature
 */
function sign(
  this: RippleAPI,
  txJSON: string,
  secret?: any,
  options?: SignOptions,
  keypair?: KeyPair
): {signedTransaction: string; id: string} {....}
```
Into:

### sign

**sign**: (`txJSON`: `string`, `secret?`: `any`, `options?`: [`SignOptions`](../modules.md#signoptions), `keypair?`: [`KeyPair`](../modules.md#keypair)) => { `id`: `string` ; `signedTransaction`: `string`  }

#### Type declaration

(`txJSON`, `secret?`, `options?`, `keypair?`): `Object`

Signs a transaction JSON

##### Parameters

| Name | Type | Description |
| :------ | :------ | :------ |
| `txJSON` | `string` | Transaction encoded in hex string |
| `secret?` | `any` | Seed to generate wallet to sign |
| `options?` | [`SignOptions`](../modules.md#signoptions) | Options describing how to signed |
| `keypair?` | [`KeyPair`](../modules.md#keypair) | Keypari to sign wallet with |

##### Returns

`Object`

Transaction hash and signature

| Name | Type |
| :------ | :------ |
| `id` | `string` |
| `signedTransaction` | `string` |

You can do a similar thing with types…

```ts
export interface AccountInfoRequest {
  /** Account To get info for */
  account: string
  /** Is this strict */
  strict?: boolean
  /** Should we look in the queue */
  queue?: boolean
  /** ledger_hash to look at */
  ledger_hash?: string
  /** ledger_index to look at */
  ledger_index?: number | ('validated' | 'closed' | 'current')
  /** Signer list to use */
  signer_lists?: boolean
}
```
Produces:

## Properties

### account

• **account**: `string`

Account To get info for

#### Defined in

[src/common/types/commands/account_info.ts:9](https://github.com/ripple/ripple-lib/blob/b81c1dab/src/common/types/commands/account_info.ts#L9)

___

### ledger\_hash

• `Optional` **ledger\_hash**: `string`

ledger_hash to of ledger to look in

#### Defined in

[src/common/types/commands/account_info.ts:15](https://github.com/ripple/ripple-lib/blob/b81c1dab/src/common/types/commands/account_info.ts#L15)

I am reasonably sure that with a solid linting config + typedoc, we can generate documentation directly from typescript… We’ll still have to manually write examples but thats okay.

natenichols:
You can do a similar thing with types…

```ts
export interface AccountInfoRequest {
  /** Account To get info for */
  account: string
  /** Is this strict */
  strict?: boolean
  /** Should we look in the queue */
  queue?: boolean
  /** ledger_hash to look at */
  ledger_hash?: string
  /** ledger_index to look at */
  ledger_index?: number | ('validated' | 'closed' | 'current')
  /** Signer list to use */
  signer_lists?: boolean
}
```
Produces:

## Properties

### account

• **account**: `string`

Account To get info for

#### Defined in

[src/common/types/commands/account_info.ts:9](https://github.com/ripple/ripple-lib/blob/b81c1dab/src/common/types/commands/account_info.ts#L9)

___

### ledger\_hash

• `Optional` **ledger\_hash**: `string`

ledger_hash to of ledger to look in

#### Defined in

[src/common/types/commands/account_info.ts:15](https://github.com/ripple/ripple-lib/blob/b81c1dab/src/common/types/commands/account_info.ts#L15)

I am reasonably sure that with a solid linting config + typedoc, we can generate documentation directly from typescript… We’ll still have to manually write examples but thats okay.

natenichols:
Or, we can do what we did w/ `xrpl-py`, and just link to that section on xrpl.org `readthedocs`

src/models/common/index.ts
```diff
@@ -11,10 +11,27 @@ export interface IssuedCurrency {
...
+  Account: string;
+  TxnSignature: string;
+  SigningPubKey: string;
+}
+
+export interface Memo {
+  MemoData?: string;
+  MemoType?: string;
+  MemoFormat?: string;
+}
```
mvadari, line 37: nit:
```suggestion
}

```

src/models/transactions/common.ts
```diff
@@ -0,0 +1,136 @@
...
+ * @param common - An interface w/ common transaction fields
+ * @returns - Void
+ * @throws - When the common param is malformed. 
+ */
+export function verifyCommonFields(common: CommonFields): void {
+    if (common.Account === undefined)
+        throw new ValidationError("CommonFields: missing field Account")
+    
+    if (typeof common.Account !== 'string')
+        throw new ValidationError("CommonFields: Account not string")
```
mvadari, line 86: Doesn't TypeScript handle this already? Or are we specifically doing runtime checks? This seems a bit excessive.

test/verify-test.ts
```diff
@@ -0,0 +1,240 @@
+import { ValidationError } from 'ripple-api/common/errors'
```
mvadari, line 1: This file should probably be renamed to something like `verify-transaction`.

src/models/transactions/common.ts
```diff
@@ -0,0 +1,136 @@
...
+ * verify the common fields of a transaction. The verify functionality will be
+ * optional, and will check transaction form at runtime. This should be called
+ * any time a transaction will be verified.
+ * 
+ * @param common - An interface w/ common transaction fields
+ * @returns - Void
+ * @throws - When the common param is malformed. 
+ */
+export function verifyCommonFields(common: CommonFields): void {
+    if (common.Account === undefined)
```
intelliot, line 82: If y'all are ok with it, I would prefer to avoid these kinds of runtime checks and just lean on rippled to catch such errors. In practice, developers don't need us to provide this kind of verify code. rippled is the source of truth, and ideally, it's the single source of truth.

test/verify-test.ts
mvadari: Actually, even better would be like `test/models/verify.ts` or something.

src/models/transactions/common.ts
intelliot: TypeScript will also catch some of these mistakes (if the developer is using TS; JavaScript developers can still use this library -- without type safety -- too)

mvadari: Yeah, I'm in favor of not checking anything that TS will check automatically

src/models/transactions/common.ts
natenichols: I thought we were specifically doing runtime checks, similar to jsonschemas. If we're not checking the types, then we also shouldn't worry about any runtime verification IMO.

src/models/transactions/common.ts
natenichols: That's fair, what should we verify in `verify`? Should it be preflight checks, like elliot suggested? 

IMO, we can check almost everything w/ typescript, I was leaving this functionality in for JS devs.

mvadari:
Yeah that's fair, it could be useful to JS devs. You can take a look at
what we did in xrpl-py for some examples.

src/models/transactions/common.ts
```diff
@@ -0,0 +1,136 @@
...
+    const validFormat = memo.MemoFormat === undefined 
+        || typeof memo.MemoData === 'string'
+    const validType = memo.MemoType === undefined
+        || typeof memo.MemoType === 'string'
+
+    const otherKeys = Object.keys(memo).every((key: string) => {
+        return key === "MemoFormat" 
+            || key === "MemoData"
+            || key === "MemoType"
+    })
```
mvadari, line 40: Is there a better way to do this? Maybe a helper function that makes sure there aren't extra keys in an object would be useful (idk if that's possible) - could be used in `validate` methods.

src/models/transactions/common.ts
mvadari: @natenichols and I chatted about this, we're going to leave the functionality in but make it optional (so the function will exist for those who want it, but you can also choose to not use it).

src/models/common/metadata.ts
```diff
@@ -0,0 +1,38 @@
...
+    DeletedNode: {
+        LedgerEntryType: string
+        LedgerIndex: string
+        FinalFields: {[field: string]: any}
+    }
+}
+
+type Node = CreatedNode | ModifiedNode | DeletedNode
+
+export interface Metadata {
```
mvadari, line 32: Is the name `Metadata` potentially ambiguous? Is there any other place where metadata is relevant in rippled?

test/verify-test.ts
```diff
@@ -0,0 +1,240 @@
...
+                },
+                {
+                    Memo: {
+                        MemoType: "72656e74"
+                    }
+                }
+            ],
+            Signers: [
+                {
+                    Account: "r....",
```
mvadari, line 41: Should we check the shape of `Account` fields? e.g. right length, starts with `r`? Or is that too much?

test/verify-test.ts
```diff
@@ -0,0 +1,240 @@
...
+                    TxnSignature: "DEADBEEF",
+                    SigningPubKey: "hex-string"
```
mvadari, line 43: Same question here

test/verify-test.ts
```diff
@@ -0,0 +1,240 @@
...
+        
+        assert.doesNotThrow(() => verifyCommonFields(txJson))
+    })
+
+    it (`Handles invalid Fee`, () => {
+        const invalidFee = {
+            Account: "r97KeayHuEsDwyU1yPBVtMLLoQr79QcRFe",
+            TransactionType: "Payment",
+            Fee: 1000
+        } as any
```
mvadari, line 69: Why is this cast `as any`?

src/models/common/metadata.ts
natenichols: I dont think so, I mostly didn't want to have types `TransactionMetaData` and `TransactionAndMetaData`

test/verify-test.ts
natenichols: If we wanted checks on the value of account fields, we should use `ripple-blinary-codec` to check if its a valid address.

test/verify-test.ts
natenichols: Its a pretty easy check to verify its a hex string, but also, these will get caught in serialization.

test/verify-test.ts
natenichols: But it will get caught in serialization.

test/verify-test.ts
natenichols: Because I wanted to test at runtime, and if its missing a field it won't be a type BaseTransaction

test/verify-test.ts
mvadari: Ah that's fair.

src/models/methods/baseMethod.ts
```diff
@@ -21,6 +19,6 @@ export interface BaseResponse {
     warnings?: Warning[]
     forwarded?: boolean
     error?: string
-    request?: Response
+    request?: Request
```
mvadari, line 22: Does `Request` not need to be imported?

mvadari (commented):
You should change the PR name so it's more descriptive (common fields of what?)

src/models/transactions/common.ts
natenichols: I don't think this is a bad use of `every`. I don't love it, and am open to other ideas.

Maybe a helper like 

```ts
onlyHasFields(obj: Record<string, unkown>, fields: Array<string>) bool {
    return Object.key(obj).every((key:string) => fields.includes(key)
}
```

thoughts?

mvadari (approved):
Approved but please respond to remaining comments

src/models/common/metadata.ts
natenichols: Default export

PR #1508, opened 2021-08-05
refactor: Define typescript definition for OfferCreate

natenichols:
Defines a typescript type for OfferCreate transaction.

### Context of Change
OfferCreate is an extension of CommonFields, and is exported in a type union with all other transaction types defined as `Transaction`.

We also will provide an optional `verifyOfferCreate` function for runtime form checking of `OfferCreate` transactions, should the users want to verify transaction form at runtime.

`SignedTransaction` and `TransactionAndMetadata` both use this  `Transaction` type.

### Type of Change
- [ ] Documentation Updates
- [ ] Release

## Test Plan
Adds tests for `verifyOfferCreate`, which provides runtime form verification for `OfferCreate` transactions


src/models/transactions/offerCreate.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ * Verify the form and type of an OfferCreate at runtime.
+ * 
+ * @param tx - An OfferCreate Transaction
+ * @returns - Void.
+ * @throws - When the OfferCreate is Malformed.
+ */
+ export function verifyOfferCreate(tx: OfferCreate): void {
```
intelliot, line 20: I imagine we may want to add some preflight checks here in the future. (or have we decided against that?) If so, we should make this `async` and return a Promise in case we need to check the state of the ledger.

intelliot (commented):
I think this is reasonable.

I wonder if, in the future, it would make sense to have a `verify(transaction)` method that takes any type of transaction. Or I suppose we could go a step further and have a method like `prepare(transaction, {verify: true})` so prepare and verify can be done in one call. Or perhaps verify is the default and someone can do `prepare(transaction, {verify: false})` if they don't want us to verify.

Apologies if we've already decided about this, I'm just thinking out loud 😀

src/models/transactions/offerCreate.ts
intelliot: Well, I guess this isn't part of a client, so I'm unsure what value this `verify` actually adds. The most valuable/tricky preflight checks are ones that involve checking a few things about the ledger to see if we can prevent an obvious `tec`. Maybe it would be like `client.verify(transaction)`

natenichols: Yeah, that's a really good idea.

src/models/transactions/offerCreate.ts
```diff
@@ -0,0 +1,47 @@
...
+    verifyCommonFields(tx)
+
+    if (tx.TakerGets === undefined)
+        throw new ValidationError("OfferCreate: missing field TakerGets")
+
+    if (tx.TakerPays === undefined)
+        throw new ValidationError("OfferCreate: missing field TakerPays")
+
+    const isIssuedCurrency = (obj: IssuedCurrencyAmount): boolean => {
+        return Object.keys(obj).length === 3
```
intelliot, line 30: These kinds of runtime checks are ok, but not a big deal. I believe these are the mistakes that rippled would immediately throw out as invalid (probably a `tem`) so we don't really need our library to do this, IMHO

src/models/transactions/offerCreate.ts
natenichols: right now, verify just checks transaction form at runtime, essentially replacing `jsonschemas`

natenichols:
Yeah, I agree... Something like:

```ts
verify<T extends Transaction>(tx: T): void
```

that verifies a transaction based on its `TransactionType`

I like the idea of passing `verify` as an instruction/option, so you could do 

```ts
interface Instructions {
    verify: boolean,
    preflight: boolean
}
```

intelliot:
Consider:
1. Maybe there isn't a need for the method to be generic:
    ```ts
    verify(tx: Transaction): Promise<void>
    ```
2. Are we keeping `Instructions` or could we get rid of it? ... There are pros/cons both ways, but I was kind of hoping we could maybe simplify by putting the instructions within the tx object itself 🤔

natenichols:
I'm in favor of getting rid of `Instructions` as they exist currently. They mostly seem like a way to explicitly fill transaction fields like `Fee`. This shouldn't be needed, as we can choose to only autofill if fields aren't already populated.

intelliot:
The particularly tricky Instructions are `maxLedgerVersionOffset` and `signersCount`

src/models/transactions/offerCreate.ts
intelliot: I was hoping we could get rid of jsonschemas and just not check transaction form anymore. I don't think checking for missing fields, etc. adds much value in practice. Transaction form is generally not problematic, types already provide some safety, and it is ultimately verified by rippled anyway. I think it's best to let rippled do it since that is guaranteed to be 100% correct and never needs to be updated (i.e. future versions of rippled could add new fields or otherwise change the definition of what is considered well-formed).

natenichols:
Do we still need a `maxLedgerVersionOffset`??? It seems like you could just set manually that with `maxLedgerVersion`.

As for signersCount, we can add that into an option specifically for multisigned transactions, and we can do something like:

```ts
function autofill(tx: Transaction, signersCount = 1)
```

because currently the signerCount instruction is only used for calculating the fee

src/models/transactions/offerCreate.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export enum AccountSetFlag {
+    asfAccountTxnID = 5
+    .
+}
+export interface AccountSet extends CommonFields {
+    TransactionType: "OfferCreate";
+    Flags: undefined
+    setFlag?: number | AccountSetFlag
+}
```
mvadari, line 20: Why is this here?

src/models/transactions/offerCreate.ts
mvadari: I think we should still be doing some runtime checks for transactions, because the potential cost of screwing something up is high. For example, we could add the `AccountDelete` checks in this type of method. Anything that requires checking `rippled` can be done in `autofill` or something, but IMO the verify methods shouldn't require a client.

src/models/transactions/offerCreate.ts
mvadari: These kinds of simple errors are super easy to catch and can save someone a few milliseconds.

src/models/transactions/offerCreate.ts
```diff
@@ -0,0 +1,80 @@
...
+        throw new ValidationError("OfferCreate: invalid Expiration")
+
+    if (tx.OfferSequence !== undefined && typeof tx.OfferSequence !== 'number')
+        throw new ValidationError("OfferCreate: invalid OfferSequence")
+}
+
+const order: OfferCreate = {
+    TransactionType: "OfferCreate",
+    Account: "r...",
+    TakerPays: "...",
```
mvadari, line 70: Is this a placeholder or does this actually do something?

src/models/transactions/offerCreate.ts
natenichols: Left over from paired programing w/ Mukul, removing it

src/models/transactions/offerCreate.ts
natenichols: Same here, removing this. "..." was just a placeholder

src/models/transactions/offerCreate.ts
intelliot: The important `AccountDelete` checks require a client to do correctly. Note that the owner reserve may change in the future, so I think we should avoid hardcoding it. One way to get it is `server_info` -> `reserve_inc_xrp`

src/models/transactions/offerCreate.ts
intelliot: I'm neutral on this. Personally I don't think it's worth the extra code, but if you like it, then it is fine to include. 👍

intelliot:
> Do we still need a `maxLedgerVersionOffset`??? It seems like you could just set manually that with `maxLedgerVersion`.

Correct, it's just for convenience. I think we talked about this last week, but for the record, I like the idea of using `LastLedgerSequence` and interpreting values under 32570 as being a max ledger version offset, [the way XUMM does it](https://xumm.readme.io/docs/payload-transaction-data#lastledgersequence). The idea is that it rarely makes sense to choose a max ledger version offset >= 32570, and it never makes sense to have a LastLedgerSequence < 32570.

test/models/offerCreate.ts
```diff
@@ -0,0 +1,177 @@
...
+            Flags: 0,
+            LastLedgerSequence: 65453019,
+            Sequence: 40949322,
+            SigningPubKey: "03C48299E57F5AE7C2BE1391B581D313F1967EA2301628C07AC412092FDC15BA22",
+            Expiration: 10,
+            OfferSequence: 12,
+            TakerGets: {
+            currency: "DSH",
+            issuer: "rcXY84C4g14iFp6taFXjjQGVeHqSCh9RX",
+            value: "43.11584856965009"
```
mvadari, line 25: nit: fix indenting

test/models/offerCreate.ts
```diff
@@ -0,0 +1,177 @@
...
+            Fee: "10",
+            Flags: 0,
+            LastLedgerSequence: 65453019,
+            Sequence: 40949322,
+            SigningPubKey: "03C48299E57F5AE7C2BE1391B581D313F1967EA2301628C07AC412092FDC15BA22",
+            TakerGets: "12928290425",
+            TakerPays: {
+            currency: "DSH",
+            issuer: "rcXY84C4g14iFp6taFXjjQGVeHqSCh9RX",
+            value: "43.11584856965009"
```
mvadari, line 45: ditto

test/models/offerCreate.ts
```diff
@@ -0,0 +1,177 @@
...
+            SigningPubKey: "03C48299E57F5AE7C2BE1391B581D313F1967EA2301628C07AC412092FDC15BA22",
+            TakerGets: {
+                currency: "DSH",
+                issuer: "rcXY84C4g14iFp6taFXjjQGVeHqSCh9RX",
+                value: "43.11584856965009"
+            },
+            TakerPays: {
+            currency: "DSH",
+            issuer: "rcXY84C4g14iFp6taFXjjQGVeHqSCh9RX",
+            value: "43.11584856965009"
```
mvadari, line 69: ditto (and across all of these)

test/models/offerCreate.ts
```diff
@@ -0,0 +1,177 @@
...
+            },
+            TakerPays: "12928290425",
+            TransactionType: "OfferCreate",
+            TxnSignature: "3045022100D874CDDD6BB24ED66E83B1D3574D3ECAC753A78F26DB7EBA89EAB8E7D72B95F802207C8CCD6CEA64E4AE2014E59EE9654E02CA8F03FE7FCE0539E958EAE182234D91",
+        } as any
+        
+        assert.doesNotThrow(() => verifyOfferCreate(offer))
+
+        const offer2 = {
```
mvadari, line 34: Each of these offers should be in their own test IMO

mvadari (approved):
a few minor comments, but looks good

src/models/transactions/common.ts
```diff
@@ -48,13 +48,17 @@ const isSigner = (signer: Signer): boolean => {
         && typeof signer.SigningPubKey === 'string'
 }
 
-export interface CommonFields {
+export interface GlobalFlags {
+    tfFullyCanonicalSig: boolean,
```
intelliot, line 52: With [`RequireFullyCanonicalSig`](https://xrpl.org/known-amendments.html#requirefullycanonicalsig), this flag has no real effect. It's ok to leave it, but I think we might be able to save a few lines of code by leaving it out. When we `prepare` or `autofill` a transaction, I think it's still a good idea to set the flag, just in case.

intelliot (approved):

src/models/transactions/common.ts
intelliot: https://xrpl.org/transaction-common-fields.html#global-flags

PR #1509, opened 2021-08-06
define Wallet class

### LINKED ISSUE #1485: Add Wallet abstraction
> Wallet provides a layer of abstraction over a public and private key. 
> - Wallet class (that stores keys)
> - Generate wallet

khancode:
Defines a `Wallet` class used to derive keypairs (publicKey/privateKey) and sign/verify offline transactions.

Fixes #1485 

### Context of Change
- `Wallet` class is a utility for deriving a wallet composed of a keypair (publicKey/privateKey).
- A wallet can be derived from either a seed, mnemonic, or entropy (array of random numbers).
- It provides functionality to sign/verify transactions offline.

### Type of Change

## Test Plan

Added tests to verify Wallet class functions (instance & static) work: `walletFromSeed`, `walletFromMnemonic`, `walletFromEntropy`, `sign`, `verify`.


src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+  private static deriveWallet(seed: string, options: WalletOptions): Wallet {
+    options.algorithm = options.algorithm ?? Wallet.defaultAlgorithm
+    const {publicKey, privateKey} = deriveKeypair(seed, options)
+    return new Wallet(publicKey, privateKey)
+  }
+
+  sign(
+    txJSON: any,
+    options: SignOptions = {signAs: ''}
+  ): {signedTransaction: string; id: string} {
```
natenichols, line 76: is `signedTransaction` a hex-string representing the entire transaction with `TxnSignature` and `SigningPubKey` populated?

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  private static deriveWallet(seed: string, options: WalletOptions): Wallet {
+    options.algorithm = options.algorithm ?? Wallet.defaultAlgorithm
+    const {publicKey, privateKey} = deriveKeypair(seed, options)
+    return new Wallet(publicKey, privateKey)
+  }
+
+  sign(
+    txJSON: any,
```
natenichols, line 74: Will the `txJson` eventually be typed with a `Transaction` type?

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+    const node = masterNode.derivePath(
+      options.derivationPath ?? Wallet.defaultDerivationPath
+    )
+    if (node.privateKey === undefined) {
+      return undefined
+    }
+
+    const publicKey = Wallet.hexFromBuffer(node.publicKey)
+    const privateKey = Wallet.hexFromBuffer(node.privateKey)
+    return new Wallet(publicKey, `00${privateKey}`)
```
natenichols, line 52: What does the `00` prefix on the `privateKey` do?

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  sign(
+    txJSON: any,
+    options: SignOptions = {signAs: ''}
+  ): {signedTransaction: string; id: string} {
+    return signOffline(this, JSON.stringify(txJSON), options)
+  }
+
+  verify(signedTransaction: {signedTransaction: string; id: string}): boolean {
```
natenichols, line 80: Would it be better to have verify take a `txJson` (with `SigningPubKey` + `TxnSignature` and verify that? Or is this better because it it symmetrical w/ `Wallet.sign()`

I don't know the answer here... Thoughts @intelliot?

natenichols (changes requested):
This is awesome! Super clean and great tests.

Would you add some documentation similar to [JsDoc](https://jsdoc.app/about-getting-started.html) before each of the public functions in the Wallet class?

I'm unsure what the linter will be enforcing here, so doesn't need to be anything too specific, just a general description of parameters and return types.

src/offline/wallet.ts
intelliot: Looks like `id` is not used. Maybe we only need the `signedTransaction: string`:

```suggestion
  verify(signedTransaction: string): boolean {
```

...And callers need to take the `signedTransaction` out of the result from `sign`:
```ts
wallet.verify(prepared.signedTransaction)
```
This seems fine if it is well-documented and we have examples/tutorials.

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+    const txJSON = decode(signedTransaction.signedTransaction)
```
intelliot, line 81:
```suggestion
    const txJSON = decode(signedTransaction)
```

src/offline/wallet.ts
intelliot: I would consider naming this `verifyTransaction` so that it has a different name from the `verify` method within ripple-keypairs:
```suggestion
  verifyTransaction(signedTransaction: string): boolean {
```

intelliot (changes requested):

src/offline/wallet.ts
khancode: Yup, that's correct.

src/offline/wallet.ts
khancode: I believe the `00` prefix is used to complete the derivation of `privateKey` from a seed (derived by a mnemonic). I followed the [same implementation xpring-common-js](https://github.com/xpring-eng/xpring-common-js/blob/1ff6c39e021fefeffa08d08ace14df74cd8b70da/src/XRP/wallet.ts#L140) does for this use case of wallet from mnemonic.

mvadari:
This should be merging into `2.0`, not `develop`.

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+import {deriveKeypair, generateSeed, verify} from 'ripple-keypairs'
+import {signOffline} from '../transaction/sign'
+import {SignOptions} from '../transaction/types'
+
+export interface WalletOptions {
+  // The entropy to use to generate the seed.
+  entropy?: Uint8Array | number[]
+
+  // The digital signature algorithm to generate an address for. Can be `ecdsa-secp256k1` (default) or `ed25519`.
+  algorithm?: 'ecdsa-secp256k1' | 'ed25519'
```
ledhed2222, line 13: i think this should be a type defined someone outside this class. why call it `ecdsa-secp256k1` instead of `secp256k1`?

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * A utility for deriving a wallet composed of a keypair (publicKey/privateKey).
+ * A wallet can be derived from either a seed, mnemnoic, or entropy (random number).
+ * It provides functionality to sign/verify transactions offline.
+ */
+class Wallet {
+  readonly publicKey: string
+  readonly privateKey: string
+  private static readonly defaultAlgorithm: 'ecdsa-secp256k1' | 'ed25519' =
```
ledhed2222, line 27: isn't this value always `ed25519`? i don't see any way to change the default, or any reason to want to allow people to change the default

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+    'ed25519'
+  private static readonly defaultDerivationPath: string = "m/44'/144'/0'/0/0"
+
+  constructor(publicKey: string, privateKey: string) {
+    this.publicKey = publicKey
+    this.privateKey = privateKey
+  }
+
+  static walletFromSeed(seed: string, options: WalletOptions): Wallet {
```
ledhed2222, line 36: why not call this `fromSeed`? here's what each looks like when called; the shorter version doesn't lose anything IMO

```javascript
import Wallet from 'ripple-lib/wallet'

const walletA = Wallet.walletFromSeed(....)
const walletB = Wallet.fromSeed(....)
```

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+    return Wallet.deriveWallet(seed, options)
+  }
+
+  static walletFromMnemonic(mnemonic: string, options: WalletOptions): Wallet {
```
ledhed2222, line 40: ditto

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (node.privateKey === undefined) {
+      return undefined
+    }
+
+    const publicKey = Wallet.hexFromBuffer(node.publicKey)
+    const privateKey = Wallet.hexFromBuffer(node.privateKey)
+    return new Wallet(publicKey, `00${privateKey}`)
+  }
+
+  static walletFromEntropy(options: WalletOptions): Wallet {
```
ledhed2222, line 55: ditto

src/offline/wallet.ts
ledhed2222: also - this is misleading since only some functions use `entropy` and only others use `derivationPath`. Sharing them as a single options type isn't helping, but actually making it less clear, to me

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
...
+    return Wallet.deriveWallet(seed, options)
+  }
+
+  static walletFromMnemonic(mnemonic: string, options: WalletOptions): Wallet {
+    const seed = mnemonicToSeedSync(mnemonic)
+    const masterNode = fromSeed(seed)
+    const node = masterNode.derivePath(
+      options.derivationPath ?? Wallet.defaultDerivationPath
+    )
+    if (node.privateKey === undefined) {
```
ledhed2222, line 46: `== null`

src/offline/wallet.ts
```diff
@@ -0,0 +1,88 @@
+import {fromSeed} from 'bip32'
```
mvadari, line 1: We're getting rid of the `offline` folder, so we actually want this to be under `src` instead of in `offline`.

src/offline/utils.ts
```diff
@@ -61,4 +62,5 @@ export {
     deriveXAddress,
     signPaymentChannelClaim,
     verifyPaymentChannelClaim,
+    Wallet,
```
mvadari, line 65: This should be exported at the top level instead of `utils`.

src/offline/wallet.ts
mvadari, line 13: Default should be `ed25519`, it's a newer algorithm and is more secure.

src/offline/wallet.ts
mvadari: I agree with the rename, `verify` is a little vague

mvadari (changes requested):
This PR should be merging into `2.0` instead of `develop`. `develop` is just for v1.x stuff for now.

src/offline/utils.ts
khancode: Wallet is indirectly exported at the top level since `offline/utils` is exported at the top level.

src/offline/wallet.ts
khancode: Sure, I'll move it to be under `src`.

src/offline/wallet.ts
khancode: > i think this should be a type defined someone outside this class. why call it `ecdsa-secp256k1` instead of `secp256k1`?

Yes, initially I was going to include the refactor of replacing string literals `ecdsa-secp256k1` and `ed25519` to be referenced as enum types throughout our codebase. However, these file changes seemed like it would make this PR hard to read. If you want I can include these changes here or we can do it in a separate PR to define and apply ECDSA enum type??

khancode: > also - this is misleading since only some functions use `entropy` and only others use `derivationPath`. Sharing them as a single options type isn't helping, but actually making it less clear, to me

Ah I see, one way to refactor this is to remove this single `WalletOptions` param and instead have `Wallet` functions take in appropriate params. I'll go with this approach if that sounds good to you. 🙂

src/offline/wallet.ts
khancode: Ah you're right, the wallet prefix is unnecessary in these function names. I'll refactor them.

src/offline/wallet.ts
khancode: I followed [xpring-common-js implementation](https://github.com/xpring-eng/xpring-common-js/blob/1ff6c39e021fefeffa08d08ace14df74cd8b70da/src/XRP/wallet.ts#L134) of using `node.privateKey === undefined`. I'd rather keep it to `=== undefined` just to be safe.

src/offline/wallet.ts
khancode: Yes, I was thinking we should track in a separate ticket just to keep this PR clean. If not, I don't mind including these changes as well. 🙂

src/offline/wallet.ts
khancode: That makes sense. To make the names consistent, I'll rename both to `signTransaction` and `verifyTransaction`.

khancode:
> This is awesome! Super clean and great tests.
> 
> Would you add some documentation similar to [JsDoc](https://jsdoc.app/about-getting-started.html) before each of the public functions in the Wallet class?
> 
> I'm unsure what the linter will be enforcing here, so doesn't need to be anything too specific, just a general description of parameters and return types.

Thank you! 😀  Sure, I can include docs before each Wallet public function.

src/offline/wallet.ts
khancode: Yes, I have the default set to `ed25519` on line 68: `options.algorithm = options.algorithm ?? Wallet.defaultAlgorithm`

src/offline/utils.ts
ledhed2222: for version 2, it should be explicitly exported at the toplevel. i don't think that `offline` should exist, and given that this isn't a `util`, I don't think it belongs here in the first place. IMO it should be either `src/Wallet.js` or `src/Wallet/index.js`

src/offline/wallet.ts
mvadari: If you're saving it for later, add a `TODO` here so we don't forget (there's already a bunch of `TODO`s in the models waiting for transaction types)

src/common/types/enums/ecdsa.ts
```diff
@@ -0,0 +1,6 @@
+enum ECDSA {
```
khancode, line 1: @intelliot @ledhed2222 I created an enum type for ECDSA methods for ecdsa-secp256k1 and ed25519. This replaces string literals used throughout our codebase to improve maintainability. Let me know if this enum name and values look good. 🙂

src/offline/wallet.ts
khancode: Sure, I added a `TODO`.

src/offline/wallet.ts
khancode: > i think this should be a type defined someone outside this class. why call it `ecdsa-secp256k1` instead of `secp256k1`?

I decided to create a new enum called ECDSA that contains these algorithms as its values. It's in this updated PR now.

src/offline/utils.ts
khancode: I moved it under `src/Wallet.js` in this updated PR.

src/common/types/enums/ecdsa.ts
mvadari, line 1: We're about to get rid of this folder, you should put this file in `src/common`.

src/common/types/enums/ecdsa.ts
```diff
@@ -0,0 +1,6 @@
...
+    ed25519 = 'ed25519',
+    secp256k1 = 'ecdsa-secp256k1',
```
mvadari, line 3:
```suggestion
    secp256k1 = 'secp256k1',
```

src/Wallet.ts
```diff
@@ -0,0 +1,111 @@
...
+  static fromMnemonic(
+    mnemonic: string,
+    derivationPath: string = Wallet.defaultDerivationPath
+  ): Wallet {
+    const seed = mnemonicToSeedSync(mnemonic)
+    const masterNode = fromSeed(seed)
+    const node = masterNode.derivePath(derivationPath)
+    if (node.privateKey === undefined) {
+      return undefined
+    }
```
mvadari, line 51: Don't return `undefined`. If this is an error, then throw an error, otherwise return `null`.

src/transaction/sign.ts
```diff
@@ -241,4 +246,18 @@ function sign(
...
+  options?: SignOptions,
+): SignedTransaction {
+  const {publicKey, privateKey} = wallet
+  return signWithKeypair(
+    null,
+    txJSON,
+    {publicKey, privateKey},
+    options,
+  )
+}
```
mvadari, line 261: This should really be implemented in the `Wallet` instead of here, but looks like the typing makes that hard. Maybe for now just add a `TODO` placeholder

src/common/types/enums/ecdsa.ts
khancode: Sure.

src/common/types/enums/ecdsa.ts
khancode: I think this suggested change will cause code to break. ripple-keypairs selects `secp256k1` method as `'ecdsa-secp256k1'` in `select(algorithm)`. So that's why I set `secp256k1 = 'ecdsa-secp256k1',` - https://github.com/ripple/ripple-keypairs/blob/07ce053f05e0b14155575de2b60e08a98c09ed1b/src/index.ts#L101

src/transaction/sign.ts
khancode: Yes, I agree. The reason why I put it here is `signOffline()` call flow is exactly the same as `sign()` except it doesn't call `checkFee()`. They both call `signWithKeypair()` so I thought it would be best to keep `signOffline()` here too. I'll add a `TODO` placeholder here.

test/api/generateXAddress/index.ts
```diff
@@ -1,6 +1,7 @@
 import assert from 'assert-diff'
 import responses from '../../fixtures/responses'
 import {TestSuite} from '../../utils'
+import ECDSA from '../../../src/common/types/enums/ecdsa'
 import {GenerateAddressOptions} from '../../../src/offline/generate-address'
```
natenichols, line 5: Do we want to change `GenerateAddressOptions` to `WalletOptions` or `GenerateWalletOptions`?

natenichols (approved):
This LGTM 👍

test/api/generateXAddress/index.ts
khancode: I'm not sure. I personally think it should stay as `GenerateAddressOptions` for now since it's being used in the context of `generateAddressAPI()`.

src/common/types/enums/ecdsa.ts
khancode: I created a ticket to track this change - https://ripplelabs.atlassian.net/browse/DEVX-227

test/api/walletFromEntropy/index.ts
```diff
@@ -0,0 +1,54 @@
+import assert from 'assert-diff'
```
mvadari, line 1: All the wallet creation methods should be in one file under `test/wallet` or something, they don't need to be under `api` (which is intended to just be the rippled-connected stuff). You could put them in one file.

mvadari (approved):
a couple of small comments that you should handle before merging, but looks good

intelliot (approved):

PR #1510, opened 2021-08-09
Implements TypeScript types for subscribe requests/responses/streams

mvadari:

This PR creates new TypeScript types for all rippled subscribe methods (https://xrpl.org/subscription-methods.html), the request shapes, response shapes, and stream shapes.

### Context of Change

xrpl.js v2.0 models

### Type of Change

## Test Plan


src/models/methods/subscribe.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export interface SubscribeResponse extends BaseResponse {
+    result: any
+}
+
+interface BaseStream {
+    type: string
+}
+
+export interface LedgerStream extends BaseStream {
```
mvadari, line 32: I put streams here, I couldn't really figure out a better place for them but I can move them if you think it would be better suited elsewhere

natenichols (approved):
This LGTM 👍 

We'll have to add [function overloading](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) to `Connection.on()`, but IDK if that is in the scope of this PR.

mvadari:
Yeah that sounds like it's part of the client

src/models/methods/subscribe.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+export interface ValidationStream extends BaseStream {
+    type: "validationReceived"
+    amendments?: string[]
+    base_fee?: number
+    flags: number
+    full: boolean
+    ledger_hash: string
+    ledger_index: string
```
JST5000, line 52: This one can be string or integer, so probably should be LedgerIndex type
(https://xrpl.org/subscribe.html#validations-stream)

src/models/methods/subscribe.ts
```diff
@@ -0,0 +1,107 @@
...
+    load_fee?: number
+    master_key?: string
+    reserve_base?: number
+    reserve_inc?: number
+    signature: string
+    signing_time: number
+    validation_public_key: string
+}
+
+export interface TransactionStream extends BaseStream {
```
JST5000, line 62: There seems to be a hidden additional TransactionStream called "transactions_proposed stream" which is very similar to TransactionStream, except it has a couple different values (Ex. validated is "optional in transactions_proposed stream", but mandatory in the "transactions stream") - The details for the proposed stream are described exclusively in paragraph form here: https://xrpl.org/subscribe.html#transaction-streams

That paragraph also describes an "accounts_proposed stream" which is similarly loosely documented in paragraph form.

src/models/methods/unsubscribe.ts
```diff
@@ -0,0 +1,23 @@
...
+export interface UnsubscribeRequest extends BaseRequest {
+    command: "unsubscribe"
+    streams?: StreamType[]
+    accounts?: string[]
+    accounts_proposed?: string[]
+    books?: Book[]
+    url?: string
+    url_username?: string
+    url_password?: string
+}
```
JST5000, line 19: It seems the url parameters ("url", "url_username", and "url_password") are deprecated and may be removed without notice: https://xrpl.org/unsubscribe.html#request-format

JST5000 (changes requested):
Everything other than the comments is spot on! Just some weird cases the documentation doesn't flag very well to change

src/models/methods/subscribe.ts
mvadari: No, `String - Integer` means the object is a string, but is essentially an integer cast as a string (see the example). It's to avoid floating point/overflow errors.

src/models/methods/subscribe.ts
JST5000: @mvadari Do you think it makes sense to add a TransactionProposedStream interface, or is that outside the scope?

JST5000: Decision: Not worth it to make custom interfaces, instead just made these interfaces accommodate the "proposed" formats as well since it's just small tweaks.

JST5000 (approved):

PR #1512, opened 2021-08-09
refactor: define typescript types for OfferCancel Transaction Model

mukulljangid:
Defines typescript types for OfferCancel transaction.

### Context of Change
OfferCancel is an extension of CommonFields, and is exported in a type union with all other transaction types defined as Transaction.

We also will provide an optional verifyOfferCancel function for runtime form checking of OfferCancel transactions, should the users want to verify transaction form at runtime.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
Adds tests for `verifyOfferCancel`, which provides runtime form verification for `OfferCancel` transactions


mvadari:
This PR title should have an active verb as the first word

intelliot:
While we're on the topic: for consistency, I recommend using the [imperative mood](https://chris.beams.io/posts/git-commit/). So "Define" instead of "Defines".

intelliot:
An "upgrade" 😀   to the page I linked above: I like [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) even more. I'm not pedantic about it, but it's nice to see.

natenichols (approved):
LGTM

natenichols:
You'll need to address the conflict, but that should be pretty easy.

test/models/offerCancel.ts
```diff
@@ -0,0 +1,64 @@
...
+            Account: "rnKiczmiQkZFiDES8THYyLA2pQohC5C6EF",
+            Fee: "10",
+            LastLedgerSequence: 65477334,
+            OfferSequence: 60797528,
+            Sequence: 60797535,
+            SigningPubKey: "0361BFD43D1EEA54B77CC152887312949EBF052997FBFFCDAF6F2653164B55B21...",
+            TransactionType: "OfferCancel",
+            TxnSignature: "30450221008C43BDCFC68B4793857CA47DF454C07E5B45D3F80E8E6785CAB9292...",
+            date: "2021-08-06T21:04:11Z"
+        } as any
```
JST5000, line 23: Why do we cast the offer as any? (Curious)

src/models/transactions/offerCancel.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Verify the form and type of an OfferCancel at runtime.
+ * 
+ * @param tx - An OfferCancel Transaction
+ * @returns - Void.
+ * @throws - When the OfferCancel is Malformed.
+ */
+ export function verifyOfferCancel(tx: OfferCancel): void {
+    verifyBaseTransaction(tx)
+
+    if (tx.OfferSequence !== undefined && typeof tx.OfferSequence !== 'number')
```
JST5000, line 19: Should having an undefined offerSequence be an error? (This doc says that it's not an error if the specified offer doesn't exist, but it seems like it might require at least a proposed offer to cancel: https://xrpl.org/offercancel.html)

JST5000 (commented):
Couple of curiosities before I approve (Not sure what the expected behavior should be)

src/models/transactions/offerCancel.ts
khancode: Isn't this type verification unnecessary when we have a defined TypeScript type for it? Wouldn't TS catch this error at compile time?

intelliot: The library can also be used by JavaScript apps, which won't get type safety. These verifications are for them.

Even so, I don't think they are necessary or even particularly helpful. I would be in favor of getting rid of these checks, but it's not a big deal. I'll leave it to @mukulljangid @mvadari @natenichols to decide :)

src/models/transactions/offerCancel.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface OfferCancel extends BaseTransaction {
+    TransactionType: "OfferCancel";
+    OfferSequence: number;
+}
+
+/**
+ * Verify the form and type of an OfferCancel at runtime.
+ * 
+ * @param tx - An OfferCancel Transaction
+ * @returns - Void.
```
khancode, line 13: I think it makes more sense to have return type be `boolean` instead of `void`. We can `return true` at the end of this function.

khancode, line 13: I think it makes more sense to have return type be `boolean` instead of `void`. We can `return true` at the end of this function.

test/models/offerCancel.ts
mvadari: Otherwise, TypeScript would complain that it's not the right format. But we want it to check at runtime instead of at compile time, so we cast it as `any`, which basically tells TS to ignore this.

src/models/transactions/offerCancel.ts
```diff
@@ -0,0 +1,21 @@
+import { ValidationError } from "../../common/errors";
+import { BaseTransaction, verifyBaseTransaction } from "./common";
+
+export interface OfferCancel extends BaseTransaction {
+    TransactionType: "OfferCancel";
+    OfferSequence: number;
+}
+
+/**
+ * Verify the form and type of an OfferCancel at runtime.
```
khancode, line 10: Can you also specify this is an optional function to call for users?

khancode (changes requested):

src/models/transactions/offerCancel.ts
mvadari: We've talked it over with enough folks that we think that the runtime checks are useful.

src/models/transactions/offerCancel.ts
mvadari: I don't think it makes sense to `return true` if there's no path where you could `return false`.

src/models/transactions/index.ts
```diff
@@ -1,3 +1,3 @@
 export * from './transaction'
 export * from './offerCreate'
-export * from './checkCreate'
+export * from './offerCancel'
```
mvadari, line 3: merge issue here

src/models/transactions/transaction.ts
```diff
@@ -1,20 +1,19 @@
...
-
+import { OfferCancel } from "./offerCancel";
 
 export type Transaction =
 //     AccountSet
 //   | AccountDelete
 //   | CheckCancel
 //   | CheckCash
-      CheckCreate
+//   | CheckCreate
```
mvadari, line 10: ditto

test/models/offerCancel.ts
```diff
@@ -0,0 +1,64 @@
...
+            LastLedgerSequence: 65477763,
+            OfferSequence: 219792,
+            Sequence: 219793,
+            SigningPubKey: "02775208415B30A20F98C36AAB9A905A7BAC5889CFE7E55D968E7B9D28B93F342...",
+            TransactionType: "OfferCancel",
+            TxnSignature: "30440220043C67CCDFC513E4D8E9E427AB6E41047B322097EB5E44DE1AE8F58F3...",
+            date: "2021-08-06T21:31:12Z",
+        } as any
+
+        assert.doesNotThrow(() => verifyOfferCancel(offer2))
```
mvadari, line 40: this should be in a separate test

mvadari (changes requested):

src/models/transactions/offerCancel.ts
khancode: Hmm that's true (no pun). I'm used to expecting verify functions returning a `boolean` as a success indication. I'm assuming users will wrap `verify` functions in a try catch and handle whatever error is thrown.

src/models/transactions/transaction.ts
mukulljangid: Done!

src/models/transactions/index.ts
mukulljangid: Fixed!

src/models/transactions/offerCancel.ts
JST5000, line 19: I think if OfferSequence is undefined this should actually throw an error (I don't think the field is optional - https://xrpl.org/offercancel.html)

src/models/transactions/offerCancel.ts
```diff
@@ -0,0 +1,24 @@
+import { ValidationError } from "../../common/errors";
+import { BaseTransaction, verifyBaseTransaction } from "./common";
```
khancode, line 2: nit: remove semicolons

src/models/transactions/offerCancel.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface OfferCancel extends BaseTransaction {
+    TransactionType: "OfferCancel";
+    OfferSequence: number;
```
khancode, line 6: ditto

src/models/transactions/transaction.ts
```diff
@@ -1,19 +1,24 @@
 import Metadata from "../common/metadata";
 import { OfferCreate } from "./offerCreate";
+import { OfferCancel } from "./offerCancel";
 import { CheckCash } from "./checkCash";
 import { CheckCancel } from "./checkCancel";
 import { CheckCreate } from "./checkCreate";
```
khancode, line 6: remove semicolons to keep it consistent

khancode (approved):
LGTM. Left some nit comments on removing semicolons.

src/models/transactions/offerCancel.ts
mukulljangid: Done!

JST5000 (approved):

test/models/offerCancel.ts
```diff
@@ -0,0 +1,88 @@
...
+            Sequence: 60797535,
+            SigningPubKey: "0361BFD43D1EEA54B77CC152887312949EBF052997FBFFCDAF6F2653164B55B21...",
+            TransactionType: "OfferCancel",
+            TxnSignature: "30450221008C43BDCFC68B4793857CA47DF454C07E5B45D3F80E8E6785CAB9292...",
+            date: "2021-08-06T21:04:11Z"
+        } as any
+        
+        assert.doesNotThrow(() => verifyOfferCancel(offer))
+        
+    })
```
khancode, line 27:
```suggestion
describe('OfferCancel Transaction Verification', function () {
   let offer
   
   beforeEach(() => {
       offer = {
            Account: "rnKiczmiQkZFiDES8THYyLA2pQohC5C6EF",
            Fee: "10",
            LastLedgerSequence: 65477334,
            OfferSequence: 60797528,
            Sequence: 60797535,
            SigningPubKey: "0361BFD43D1EEA54B77CC152887312949EBF052997FBFFCDAF6F2653164B55B21...",
            TransactionType: "OfferCancel",
            TxnSignature: "30450221008C43BDCFC68B4793857CA47DF454C07E5B45D3F80E8E6785CAB9292...",
            date: "2021-08-06T21:04:11Z"
        } as any
   })

    it (`verifies valid OfferCancel`, () => {        
        assert.doesNotThrow(() => verifyOfferCancel(offer))        
    })
```

test/models/offerCancel.ts
```diff
@@ -0,0 +1,88 @@
...
+            Account: "rnfQBGzgJb2x26U2Tfe1GaYw4fNB87Dc6J",
+            Fee: "12",
+            Flags: 2147483648,
+            LastLedgerSequence: 65477763,
+            Sequence: 219793,
+            SigningPubKey: "02775208415B30A20F98C36AAB9A905A7BAC5889CFE7E55D968E7B9D28B93F342...",
+            TransactionType: "OfferCancel",
+            TxnSignature: "30440220043C67CCDFC513E4D8E9E427AB6E41047B322097EB5E44DE1AE8F58F3...",
+            date: "2021-08-06T21:31:12Z",
+        } as any
```
khancode, line 80:
```suggestion
        delete offer.OfferSequence
```

test/models/offerCancel.ts
```diff
@@ -0,0 +1,88 @@
...
+            Fee: "12",
+            Flags: 2147483648,
+            LastLedgerSequence: 65477763,
+            OfferSequence: "99",
+            Sequence: 219793,
+            SigningPubKey: "02775208415B30A20F98C36AAB9A905A7BAC5889CFE7E55D968E7B9D28B93F342...",
+            TransactionType: "OfferCancel",
+            TxnSignature: "30440220043C67CCDFC513E4D8E9E427AB6E41047B322097EB5E44DE1AE8F58F3...",
+            date: "2021-08-06T21:31:12Z",
+        } as any
```
khancode, line 60:
```suggestion
        offer.OfferSequence = '99'
```

test/models/offerCancel.ts
```diff
@@ -0,0 +1,88 @@
...
+            LastLedgerSequence: 65477763,
+            OfferSequence: 219792,
+            Sequence: 219793,
+            SigningPubKey: "02775208415B30A20F98C36AAB9A905A7BAC5889CFE7E55D968E7B9D28B93F342...",
+            TransactionType: "OfferCancel",
+            TxnSignature: "30440220043C67CCDFC513E4D8E9E427AB6E41047B322097EB5E44DE1AE8F58F3...",
+            date: "2021-08-06T21:31:12Z",
+        } as any
+
+        assert.doesNotThrow(() => verifyOfferCancel(offer2))
```
khancode, line 44:
```suggestion
        offer.Flags = 2147483648
        assert.doesNotThrow(() => verifyOfferCancel(offer))
```

khancode (changes requested):
Left some suggested changes to clean up tests by using `beforeEach()`

test/models/offerCancel.ts
```diff
@@ -0,0 +1,57 @@
...
+            date: "2021-08-06T21:04:11Z"
+        } as any
+   })
+
+    it (`verifies valid OfferCancel`, () => {        
+        assert.doesNotThrow(() => verifyOfferCancel(offer))        
+    })
+
+    it (`verifies valid OfferCancel with flags`, () => {
+
```
khancode, line 32: nit: delete empty line

test/models/offerCancel.ts
```diff
@@ -0,0 +1,57 @@
...
+        offer.Flags = 2147483648
+        assert.doesNotThrow(() => verifyOfferCancel(offer))
+        
+    })
+
+    it (`throws w/ invalid OfferSequence`, () => {
+        offer.OfferSequence = '99'
+
```
khancode, line 40: ditto

test/models/offerCancel.ts
```diff
@@ -0,0 +1,57 @@
...
+        assert.throws(
+            () => verifyOfferCancel(offer),
+            ValidationError,
+            "OfferCancel: invalid OfferSequence"
+        )
+    })
+
+    it (`throws w/ missing OfferSequence`, () => {
+        delete offer.OfferSequence
+
```
khancode, line 50: ditto

khancode (approved):
LGTM 👍 Added small nit comments that you can do before merging.

src/models/transactions/offerCancel.ts
JST5000: For `tx.OfferSequence` since it's not optional I think the condition is actually:
`(typeof tx.OfferSequence !== 'number')`

Because OfferSequence isn't optional, we want this to throw an error when OfferSequence is undefined (And since typeof undefined !== number, that's the only condition we need)

JST5000 (approved):

src/models/transactions/transaction.ts
```diff
@@ -1,20 +1,25 @@
...
-import { CheckCash } from "./checkCash";
-import { CheckCancel } from "./checkCancel";
-import { CheckCreate } from "./checkCreate";
+import Metadata from "../common/metadata"
+import { OfferCreate } from "./offerCreate"
+import { OfferCancel } from "./offerCancel"
+import { CheckCash } from "./checkCash"
+import { CheckCancel } from "./checkCancel"
+import { CheckCreate } from "./checkCreate"
 import { SignerListSet } from "./signerListSet";
```
mvadari, line 7: be uniform about semicolon use

mvadari (approved):

src/models/transactions/transaction.ts
mukulljangid: Done

PR #1514, opened 2021-08-09
Implements TypeScript types for server info method requests/responses

mvadari:

This PR creates new TypeScript types for all rippled server info methods (https://xrpl.org/server-info-methods.html), and the request shapes and response shapes.

### Context of Change

xrpl.js v2.0 models

### Type of Change

## Test Plan

### DIFF
--- a/src/models/methods/manifest.ts
+++ b/src/models/methods/manifest.ts
@@ -14,6 +14,6 @@ export interface ManifestResponse extends BaseResponse {
             seq: number
         }
         manifest?: string
-        request: string
+        requested: string
     }
 }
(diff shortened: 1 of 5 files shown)


JST5000 (changes requested):

src/models/methods/manifest.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface ManifestResponse extends BaseResponse {
+    result: {
+        details?: {
+            domain: string
+            ephemeral_key: string
+            master_key: string
+            seq: number
+        }
+        manifest?: string
+        request: string
```
mukulljangid, line 17: Looks like this should be requested. https://xrpl.org/manifest.html

mukulljangid (approved):
Looks good to me!

natenichols (approved):
LGTM 👍

PR #1515, opened 2021-08-09
refactor: define typescript type for AccountSet transaction

mukulljangid:
Defines typescript types for AccountSet transaction.

### Context of Change
AccountSet is an extension of CommonFields, and is exported in a type union with all other transaction types defined as Transaction.

We also will provide an optional verifyAccountSet function for runtime form checking of AccountSet transactions, should the users want to verify transaction form at runtime.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
Adds tests for verifyAccountSet, which provides runtime form verification for AccountSet transactions


src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    if (tx.ClearFlag !== undefined){
+        if (typeof tx.ClearFlag !== 'number')
+            throw new ValidationError("AccountSet: invalid ClearFlag")
+        if (0 > tx.ClearFlag || tx.ClearFlag > 9)
+            throw new ValidationError("AccountSet: invalid ClearFlag")
+    }
+
+    if (tx.Domain !== undefined && typeof tx.Domain !== 'string')
+        throw new ValidationError("AccountSet: invalid Domain")
```
khancode, line 47: I'm not a TypeScript expert but isn't verifying a type unnecessary when TS will catch this error at compile time?

test/models/accountSet.ts
```diff
@@ -0,0 +1,161 @@
...
+        } as any
+
+        assert.throws(
+            () => verifyAccountSet(offer),
+            ValidationError,
+            "AccountSet: invalid SetFlag"
+        )
+    })
+
+    it (`throws w/ invalid SetFlag`, () => {
```
JST5000, line 44: Can you make the description a bit more specific since this is a different case than the above test with the same name?

test/models/accountSet.ts
```diff
@@ -0,0 +1,161 @@
...
+            TransferRate : "1000000001"
+        } as any
+
+        assert.throws(
+            () => verifyAccountSet(offer),
+            ValidationError,
+            "AccountSet: invalid TransferRate"
+        )
+    })
+
```
JST5000, line 144: Can you add a test case or two for transfer rate values? (Special case of TransferRate = 0, along with upper limits/lower limits)

JST5000 (changes requested):

src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @returns - Void.
+ * @throws - When the AccountSet is Malformed.
+ */
+ export function verifyAccountSet(tx: AccountSet): void {
+    verifyBaseTransaction(tx)
+
+    if (tx.ClearFlag !== undefined){
+        if (typeof tx.ClearFlag !== 'number')
+            throw new ValidationError("AccountSet: invalid ClearFlag")
+        if (0 > tx.ClearFlag || tx.ClearFlag > 9)
```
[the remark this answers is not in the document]
mukulljangid: Sounds good!

src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (tx.EmailHash !== undefined && typeof tx.EmailHash !== 'string')
+        throw new ValidationError("AccountSet: invalid EmailHash")
+        
+    if (tx.MessageKey !== undefined && typeof tx.MessageKey !== 'string')
+        throw new ValidationError("AccountSet: invalid MessageKey")
+    
+    if (tx.SetFlag !== undefined){
+        if (typeof tx.SetFlag !== 'number')
+            throw new ValidationError("AccountSet: invalid SetFlag")
+        if (0 > tx.SetFlag || tx.SetFlag > 9)
```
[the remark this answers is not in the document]
mukulljangid: Done!

src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,71 @@
...
+            throw new ValidationError("AccountSet: invalid SetFlag")
+    }
+
+    if (tx.TransferRate !== undefined && typeof tx.TransferRate !== 'number')
```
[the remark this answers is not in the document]
mukulljangid: Could be an additional check, but these are anyways going to be validated by rippled! We can skip it!

test/models/accountSet.ts
mukulljangid: Done!

test/models/accountSet.ts
mukulljangid: Since we are leaving out the values checking here, we can leave out these tests as well.

src/models/transactions/accountSet.ts
natenichols: Yes, you are correct, in TS this type of runtime checking is unnecessary. We've been back and forth on whether or not to include this functionality. IMO I'd like to keep it + make it optional for the following reasons:
1. JavaScript users can use these `verify` functions
2. This functionality is similar to the jsonschemas functionality already in use in ripple-lib. This also lets us define exactly what error message we want to throw.
3. This functionality is pretty extendable to other runtime checks (something like an AccountDelete having a certain fee)
4. Typescript doesn't offer any runtime guarantees. Someone using TS wrong can still pass whatever they want to w/ an `any`.  Letting TS Clients choose to use this at runtime could be a useful feature.

I know @intelliot disagrees here, maybe he could write out his thoughts.

src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,71 @@
...
+    TransactionType: "AccountSet";
+    ClearFlag?: number;
+    Domain?: string;
+    EmailHash?: string;
+    MessageKey?: string;
+    SetFlag?: AccountSetFlagEnum;
+    TransferRate?: number;
+    TickSize?: number;
+    WalletLocator?: string;
+    WalletSize?: number;
```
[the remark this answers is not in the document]
natenichols: I would say go ahead and leave them out if our docs say they're not used.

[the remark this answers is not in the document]
mukulljangid: I missed it! It was a mistake! They are out now!

JST5000 (approved):
LGTM!

src/models/transactions/accountSet.ts
khancode: That makes sense! I see no harm of keeping this optional for JS users.

I think it's worth documenting that we recommend users to call `verify` functions before submitting transactions; it would be helpful to describe this in jsdoc comments too. I also like the idea that it allows us to define specific error messages to throw.

@natenichols With that being said, I'm now supportive of this change and thanks for this detailed response! 😀

intelliot: > JavaScript users can use these verify functions

👍

> This functionality is similar to the jsonschemas functionality already in use in ripple-lib. This also lets us define exactly what error message we want to throw.

I don't find this functionality to be very useful. It's fine, but I don't really think it's worth the complexity + additional lines of code + need for maintenance.

> This functionality is pretty extendable to other runtime checks (something like an AccountDelete having a certain fee)

Note that the AccountDelete fee is equal to the owner reserve which can change based on a vote of the validators. So verifying that properly will require a Client (to retrieve the current owner reserve value).

> Typescript doesn't offer any runtime guarantees. Someone using TS wrong can still pass whatever they want to w/ an any. Letting TS Clients choose to use this at runtime could be a useful feature.

👍   But it happens rarely and early enough that I think it's also ok to get the error from rippled slightly later on.

intelliot (approved):

khancode (approved):

src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * Verify the form and type of an AccountSet at runtime.
+ * 
+ * @param tx - An AccountSet Transaction
+ * @returns - Void.
+ * @throws - When the AccountSet is Malformed.
+ */
+ export function verifyAccountSet(tx: AccountSet): void {
```
khancode, line 34: Can you specify in the jsdoc that this is an optional function to call for JS users?

khancode: Also I think it makes more sense to have return type be `boolean` instead of `void`. We can return true at the end of the function.

mvadari: The other `verify` methods all return `void` and just raise errors if there's a problem, we should be consistent

src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,69 @@
...
+    Domain?: string;
+    EmailHash?: string;
+    MessageKey?: string;
+    SetFlag?: AccountSetFlagEnum;
+    TransferRate?: number;
+    TickSize?: number;
+}
+
+/**
+ * Verify the form and type of an AccountSet at runtime.
```
khancode, line 28: Can you also specify this is an optional function to call for users?

src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,69 @@
...
+ * 
+ * @param tx - An AccountSet Transaction
+ * @returns - Void.
```
khancode, line 31: I think it makes more sense to have return type be `boolean` instead of `void`. We can `return true` at the end of this function.

khancode (changes requested):

src/models/transactions/accountSet.ts
mvadari: The other verify methods all return void and just raise errors if there's a problem, we should be consistent

khancode: Ok, I think that's fine then. I agree with keeping things consistent.

src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,69 @@
+import { ValidationError } from "../../common/errors";
+import { BaseTransaction, verifyBaseTransaction } from "./common";
```
khancode, line 2: nit: remove semicolons

src/models/transactions/accountSet.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export interface AccountSet extends BaseTransaction {
+    TransactionType: "AccountSet";
+    ClearFlag?: number;
+    Domain?: string;
+    EmailHash?: string;
+    MessageKey?: string;
+    SetFlag?: AccountSetFlagEnum;
+    TransferRate?: number;
+    TickSize?: number;
```
khancode, line 24: ditto

khancode (approved):
LGTM. Added small nit comments to resolve before merging.

PR #1517, opened 2021-08-10
Implements TypeScript types for transaction method requests/responses

mvadari:

This PR creates new TypeScript types for all rippled transaction methods (https://xrpl.org/transaction-methods.html), and the request shapes and response shapes. `sign` and `sign_for` are purposely excluded, because the library has offline signing functionality, and it is unadvised to use those methods unless you are running your own rippled server.

### Context of Change

xrpl.js v2.0 models

### Type of Change

## Test Plan

### DIFF
--- a/src/models/methods/baseMethod.ts
+++ b/src/models/methods/baseMethod.ts
@@ -23,5 +23,4 @@ export interface BaseResponse {
     error?: string
     request?: Response
     api_version?: number
-    searched_all?: boolean // used in `tx`
 }
--- a/src/models/methods/baseMethod.ts
+++ b/src/models/methods/baseMethod.ts
@@ -23,4 +23,5 @@ export interface BaseResponse {
     error?: string
     request?: Response
     api_version?: number
+    searched_all?: boolean // used in `tx`
 }
--- a/src/models/methods/submit.ts
+++ b/src/models/methods/submit.ts
@@ -1,0 +1,26 @@
+import { BaseRequest, BaseResponse } from "./baseMethod";
+
+export interface SubmitRequest extends BaseRequest {
+  command: "submit"
+  tx_blob: string
+  fail_hard?: boolean
+}
+
+export interface SubmitResponse extends BaseResponse {
+  result: {
+      engine_result: string
+      engine_result_code: number
+      engine_result_message: string
+      tx_blob: string
+      tx_json: any // TODO: type this properly when we have Transaction types
+      accepted: boolean
+      account_sequence_available: number
+      account_sequence_next: number
+      applied: boolean
+      broadcast: boolean
+      kept: boolean
+      queued: boolean
+      open_ledger_cost: string
+      validated_ledger_index: number
+  }
+}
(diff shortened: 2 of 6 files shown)


src/models/methods/submit.ts
```diff
@@ -0,0 +1,26 @@
+import { BaseRequest, BaseResponse } from "./baseMethod";
+
+export interface SubmitRequest extends BaseRequest {
+  command: "submit"
+  tx_blob: string
+  fail_hard?: boolean
+}
+
```
JST5000, line 8: Is the 'Sign-and-Submit Mode' deprecated? (Otherwise it might be another format we need an interface for) https://xrpl.org/submit.html#sign-and-submit-mode

src/models/methods/submit.ts
```diff
@@ -0,0 +1,26 @@
...
+      accepted: boolean
+      account_sequence_available: number
+      account_sequence_next: number
+      applied: boolean
+      broadcast: boolean
+      kept: boolean
+      queued: boolean
+      open_ledger_cost: string
+      validated_ledger_index: number
+  }
```
JST5000, line 25: If sign and submit exists, 'accepted' through 'validated_ledger_index' should be optional. Otherwise this is correct.

src/models/methods/submit.ts
mvadari: We're not supporting signing via rippled nodes, because it's discouraged (it can be insecure, the node can mess with your transaction before signing it).

JST5000 (approved):

khancode (approved):

src/models/methods/baseMethod.ts
```diff
@@ -23,4 +23,5 @@ export interface BaseResponse {
     error?: string
     request?: Response
     api_version?: number
+    searched_all?: boolean // used in `tx`
```
natenichols, line 26: Why does this need to be in `baseResponse`? Why can't we just put it in `TxResponse`?

mvadari: whoops, u rite

src/models/methods/tx.ts
```diff
@@ -0,0 +1,19 @@
...
+  min_ledger?: number
+  max_ledger?: number
+}
+
+export interface TxResponse extends BaseResponse {
+  result: {
+    hash: string
+    ledger_index: number
+    meta: TransactionMetadata | string
+    validated?: boolean
```
natenichols, line 17: Is there a field for `transaction` (or something similar) here? Is that what the comment on the next line is referring to?

natenichols (approved):
Couple comments, nothing worth holding up approval for.

src/models/methods/tx.ts
mvadari: Yeah, all the `transaction` fields are a part of `result`, so the final thing will basically be `{...} & Transaction`

src/models/methods/submit.ts
intelliot: Aside from the node messing with your transaction; you'd potentially be sending your private key over the internet, which is generally not very safe.

PR #1518, opened 2021-08-10
define TypeScript types for Payment Channel methods

khancode:

Created new TypeScript types for Payment Channel methods - https://xrpl.org/payment-channel-methods.html
Each method contains a request and response type:
- `channel_verify`

`channel_authorize` is excluded because it requires sending a secret over the network to a node that a user likely won't have control of - that's not safe, and we don't want to encourage that.

### Context of Change

xrpl.js v2.0 models

### Type of Change

## Test Plan

No tests required since it doesn't change existing code.


src/models/methods/index.ts
```diff
@@ -26,6 +28,11 @@ type Request = // account methods
              | AccountTxRequest
              | GatewayBalancesRequest
              | NoRippleCheckRequest
+               // payment channel methods
+             | ChannelAuthorizeRequest
```
natenichols, line 32: `ChannelAuthorizeRequest` and `ChannelVerifyRequest`  go here, but the responses should go under `type Response`

khancode: Oops, went over my head haha 😅

src/models/methods/channelAuthorize.ts
```diff
@@ -0,0 +1,18 @@
+import {BaseRequest, BaseResponse} from './baseMethod'
+
+export interface ChannelAuthorizeRequest extends BaseRequest {
```
mvadari, line 3: Actually, I think we don't want to implement this at all (for the same reason we're not implementing `Sign` and `SignFor`, because it requires sending your secret over the network to a node that you likely don't have control of - that's not safe, and we don't want to encourage that).

src/models/methods/channelVerify.ts
```diff
@@ -0,0 +1,15 @@
+import {BaseRequest, BaseResponse} from './baseMethod'
+
+export interface ChannelVerifyRequest extends BaseRequest {
+  command: 'channel_verify'
+  channel_id: string
+  amount: string
+  public_key: string
+  signature: string
```
mvadari, line 8: For ease of reading, this should be in the same order as on xrpl.org

src/models/methods/channelAuthorize.ts
khancode: Oh I see, I'll delete this method then.

src/models/methods/index.ts
```diff
@@ -34,6 +35,8 @@ type Request = // account methods
              | AccountTxRequest
              | GatewayBalancesRequest
              | NoRippleCheckRequest
+               // payment channel methods
+             | ChannelVerifyRequest
```
mvadari, line 39: nit: move this to be in order with xrpl.org (so between path and order book methods and subscription methods)

src/models/methods/index.ts
```diff
@@ -66,6 +69,8 @@ type Response = // account methods
               | AccountTxResponse
               | GatewayBalancesResponse
               | NoRippleCheckResponse
+                // payment channel methods
+              | ChannelVerifyResponse
```
mvadari, line 73: ditto

src/models/methods/index.ts
```diff
@@ -110,6 +115,9 @@ export {
     GatewayBalancesResponse,
     NoRippleCheckRequest,
     NoRippleCheckResponse,
+    // payment channel methods
+    ChannelVerifyRequest,
+    ChannelVerifyResponse,
```
mvadari, line 120: ditto

mvadari (approved):
some minor comments but looks good

natenichols:
I dont see `channel_authorize` in the file changes.

mvadari:
`channel_authorize` is a signing method, so we shouldn't be supporting that

natenichols (approved):
LGTM 👍

khancode:
> `channel_authorize` is a signing method, so we shouldn't be supporting that

Yes, it was removed.

PR #1520, opened 2021-08-10
Rename RippleAPI client to Client

mvadari:

Title says it all. Also renames all instances of `RippleAPI`/`XrplClient` that were called `api` to `client`

### Context of Change

xrpl.js v2.0

### Type of Change

## Test Plan

CI passes.


test/api/constructor/index.ts
```diff
@@ -1,29 +0,0 @@
-import assert from 'assert-diff'
-import {TestSuite} from '../../utils'
-import {RippleAPI} from 'ripple-api'
```
mvadari, line 3: These files were just moved to `test/client`, not deleted - I think git had a hard time with the diff

intelliot:
I would be tempted to rename it `XRPL` because it's shorter and `Xrpl` looks like strange capitalization to me

mvadari:
@intelliot are you suggesting renaming it to `XRPLClient` or just `XRPL`?

khancode:
How are we handling code breaking for users of xrpl.js when this change is released? Do we expect users to read v2.0 docs to use `XrplClient` instead of `RippleAPI`?

mvadari:
We're going to write up docs for how to make the upgrade and navigate the breaking changes. `RippleAPI` will no longer exist, and is being replaced by `XrplClient`.

intelliot:
> @intelliot are you suggesting renaming it to `XRPLClient` or just `XRPL`?

Just `XRPL`.

mvadari:
After discussing with @natenichols and @ledhed2222, I'm changing the name to `Client`.

intelliot (approved):

scripts/ci.sh
```diff
@@ -25,7 +25,7 @@ unittest() {
   $(npm bin)/babel -D --optional runtime --ignore "**/node_modules/**" -d test-compiled/ test/
   echo "--reporter spec --timeout 5000 --slow 500" > test-compiled/mocha.opts
   mkdir -p test-compiled/node_modules
-  ln -nfs ../../dist/npm test-compiled/node_modules/ripple-api
+  ln -nfs ../../dist/npm test-compiled/node_modules/xrpl-client
```
natenichols, line 28: Do we use `ci.sh` anymore?

test/client/getTransactions/index.ts
```diff
@@ -1,41 +1,41 @@
-import {RippleAPI} from 'ripple-api'
+import {Client} from 'xrpl-client'
```
natenichols, line 1: xrpl-client is already taken on `npm.js`, I think we'll want this to be either `xrpl` or `xrpl-js`

natenichols (changes requested):
I don't think we can call it `xrpl-client`, that's already taken by [XRPL Labs](https://www.npmjs.com/package/xrpl-client) websocket client library.

I like `xrpl-js` or `XRPL`

intelliot:
The npm package name will be `xrpl`, similar to how the package for web3.js is `web3`. I've reserved the name.

`xrpl-client` in this PR is a rename of `ripple-api` which is just an [internal symlink (symbolic link)](https://github.com/ripple/ripple-lib/blob/develop/test/node_modules/ripple-api) to the [src directory](https://github.com/ripple/ripple-lib/tree/develop/src). Due to how imports work, this effectively points to [`./src/index.ts`](https://github.com/ripple/ripple-lib/blob/develop/src/index.ts). It's not related to the npm package name.

So I think this PR is ok as-is. I think renaming `xrpl-client` to `xrpl-local-working-copy` (or similar) would be more accurate / less confusing, but it's not a big deal and shouldn't affect the external API.

mvadari:
I renamed it to `xrpl-local` to make it a little more clear

natenichols (approved):

PR #1522, opened 2021-08-11
refactor: define typescript types for CheckCash Transaction Model

mukulljangid:
Defines typescript types for CheckCash transaction.

### Context of Change
CheckCash is an extension of CommonFields, and is exported in a type union with all other transaction types defined as Transaction.

We also will provide an optional verifyCheckCash function for runtime form checking of CheckCash transactions, should the users want to verify transaction form at runtime.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
Adds tests for `verifyCheckCash`, which provides runtime form verification for `CheckCash` transactions


natenichols (approved):
LGTM

natenichols:
You'll also have to resolve conflicts here.

src/models/transactions/checkCash.ts
```diff
@@ -0,0 +1,37 @@
+import { ValidationError } from "../../common/errors";
+import { Amount, IssuedCurrencyAmount } from "../common";
+import { BaseTransaction, verifyBaseTransaction } from "./common";
+
+export interface CheckCash extends BaseTransaction {
+    TransactionType: "CheckCash";
+    CheckID: string;
+    Amount?:	Amount;
+    DeliverMin?: Amount;
```
JST5000, line 9: Amount and DeliverMin should probably be 'IssuedCurrencyAmount' instead of 'Amount' because it doesn't seem like they have the option to be a string, and the common 'Amount' type does have that option: https://xrpl.org/checkcash.html#checkcash

src/models/transactions/checkCash.ts
```diff
@@ -0,0 +1,37 @@
...
+}
```
JST5000, line 10: Since a transaction must include either an Amount or DeliverMin, but not both, could we encode that into the type? (Currently our typing allows for neither Amount nor DeliverMin as an option)

Maybe something like: 
```
interface CheckCashAmount extends BaseTransaction {
    TransactionType: "CheckCash";
    CheckID: string;
    Amount: IssuedCurrencyAmount;
 }
 
 interface CheckCashDeliverMin extends BaseTransaction {
    TransactionType: "CheckCash";
    CheckID: string;
    DeliveryMin: IssuedCurrencyAmount;
}

export type CheckCash = (CheckCashAmount | CheckCashDeliveryMin)
```

src/models/transactions/checkCash.ts
```diff
@@ -0,0 +1,37 @@
...
+            && typeof obj.issuer === 'string'
+            && typeof obj.currency === 'string'
+    }
+
+    if (tx.Amount !== undefined && typeof tx.Amount !== 'string' && !isIssuedCurrency(tx.Amount))
+        throw new ValidationError("CheckCash: invalid Amount")
+
+    if (tx.DeliverMin !== undefined && typeof tx.DeliverMin !== 'string' && !isIssuedCurrency(tx.DeliverMin))
+        throw new ValidationError("CheckCash: invalid DeliverMin")
+
```
JST5000, line 34: Related to above, I don't think 'string' is a valid type for Amount/DeliverMin (If it is and I'm wrong this is good)

JST5000: (If Amount is the correct type for this, could you add a helper function to common where the Amount type is to validate 'Amount' types so we don't need to know that Amounts are String | IssuedCurrencyAmount in our validation checks?)

JST5000 (changes requested):

src/models/transactions/checkCash.ts
mukulljangid: It's of the type Amount which means you can transfer value using `IssuedCurrencyAmount` for non-XRP tokens and string for XRP.

src/models/transactions/checkCash.ts
mukulljangid: Changed the code to this! Thank you!

src/models/transactions/common.ts
```diff
@@ -48,6 +48,17 @@ const isSigner = (signer: Signer): boolean => {
         && typeof signer.SigningPubKey === 'string'
 }
 
+export function isIssuedCurrency (obj: IssuedCurrencyAmount): boolean {
+    return Object.keys(obj).length === 3 
+        && typeof obj.value === 'string'
+        && typeof obj.issuer === 'string'
+        && typeof obj.currency === 'string'
+}
```
mvadari, line 56: You should use this function in `validateOfferCreate` as well, which creates its own

src/models/transactions/transaction.ts
```diff
@@ -1,12 +1,15 @@
...
 import { OfferCreate } from "./offerCreate";
+import { CheckCash } from "./checkCash";
 import { CheckCancel } from "./checkCancel";
 import { CheckCreate } from "./checkCreate";
 
 export type Transaction =
 //     AccountSet
 //   | AccountDelete
-       CheckCancel
+//   | CheckCancel
```
mvadari, line 10: merge issues

src/models/transactions/common.ts
mukulljangid: Done!

src/models/transactions/transaction.ts
```diff
@@ -1,12 +1,15 @@
...
+       CheckCash
+     | CheckCancel
```
mvadari, line 12: This isn't in alphabetical order

mukulljangid: Done

khancode (approved):
LGTM

src/models/transactions/offerCreate.ts
```diff
@@ -34,17 +34,10 @@ export interface OfferCreate extends BaseTransaction {
...
-            && typeof obj.issuer === 'string'
-            && typeof obj.currency === 'string'
-    }
-
-    if (typeof tx.TakerGets !== 'string' && !isIssuedCurrency(tx.TakerGets))
+    if (typeof tx.TakerGets !== 'string' && !isAmount(tx.TakerGets))
         throw new ValidationError("OfferCreate: invalid TakerGets")
 
-    if (typeof tx.TakerPays !== 'string' && !isIssuedCurrency(tx.TakerPays))
+    if (typeof tx.TakerPays !== 'string' && !isAmount(tx.TakerPays))
```
JST5000, line 40: These two lines can just be `!isAmount(tx.TakerGets)` instead of `typeof tx.TakerGets !== 'string' && !isAmount(tx.TakerGets)` because isAmount does the string check already

JST5000 (approved):
Just one simplification, but the logic's right either way - The rest looks good to me :)

PR #1523, opened 2021-08-11
Update sign docs with more info about keypairs

### LINKED ISSUE #1045: Document the secp256k1 and Ed25519 private key formats
Follow-up to #797 

There may be some existing documentation here, but I think it would help to elaborate and provide some specific examples. We should also document/explain the difference between the key formats used by rippled vs. ripple-lib. Docs should appear at https://xrpl.org/rippleapi-reference.html#sign and/or be linked from there.

/cc @mDuo13

JST5000:

* Added examples of what private and public keypairs look like for the sign method to give readers an intuition for the data type
* Added links to how keypairs are derived in rippled and ripple-lib so that they can dive deeper if they want to

Fixes #1045 

### Context of Change

This is in response to #1045 to improve the documentation's covering of the two encodings ed25519 and secp256k1. 
There are also several changes being made in the xrpl-dev-portal to add examples of the keys to their derivation to help developers see what the keys actually look like.

### Type of Change

- [X] Documentation Updates

## Before / After

Added links in the public key/private key fields for the sign api, and added an example of what keypairs look like in the two formats.

## Test Plan

No tests for docs change.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -5659,8 +5659,8 @@ Name | Type | Description
 ---- | ---- | -----------
 txJSON | string | Transaction represented as a JSON string in rippled format.
 keypair | object | *Optional* The private and public key of the account that is initiating the transaction. (This field cannot be used with secret).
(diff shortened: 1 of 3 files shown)


natenichols (approved):
LGTM 👍

khancode (approved):
LGTM 👌

mvadari (approved):

PR #1525, opened 2021-08-11
docs: export and document asf Flags

### LINKED ISSUE #1483: AccountSet flags by name
AccountSet _transaction_ flags (tf*) have aliases in `txFlags.AccountSet` but [AccountSet flags (asf*)](https://xrpl.org/accountset.html#accountset-flags) don't have aliases in the API. So for example, preparing an AccountSet using `prepareTransaction()` currently has to look kind of like t his:

```js
const accountset_tx_json = {
    "Account": cold_address,
    "TransferFee": 0,
    "TickSize": 5,
    "SetFlag": 8, // enable Default Ripple
    "Flags": (api.txFlags.AccountSet.DisallowXRP | 
              api.txFlags.AccountSet.RequireDestTag)
  }
```

It would be nice if there were some constants that could be passed to `SetFlag` here. For example, something like this:

```js
const accountset_tx_json = {
    "Account": cold_address,
    "TransferFee": 0,
    "TickSize": 5,
    "SetFlag": api.accountsetFlags.DefaultRipple,
    "Flags": (api.txFlags.AccountSet.DisallowXRP | 
              api.txFlags.AccountSet.RequireDestTag)
  }
```

(Note, this example uses a mix of transaction Flags and AccountSet flags because that makes it possible to enable 3 flags in one transaction.)

natenichols:
Adds documentation for AccountSet flags.

fixes #1483 

### Context of Change
Exports asf Flags w/ txFlags, and adds a documentation section in index.md

### Type of Change
- [x] Documentation Updates
- [ ] Release


docs/index.md
```diff
@@ -101,6 +101,7 @@
   - [iso8601ToRippleTime](#iso8601torippletime)
   - [rippleTimeToISO8601](#rippletimetoiso8601)
   - [txFlags](#txflags)
+  - [txFlagIndices](#txflagindices)
```
intelliot, line 104: I know this was the pre-existing name, but `txFlagIndices` sounds weird to me. Naming is hard, but ... could we put some thought into a better name for this?

docs/index.md
```diff
@@ -6397,6 +6398,19 @@ The remaining transaction types do not have any flags at this time.
 * PaymentChannelCreate
 * PaymentChannelFund
 
+## txFlagIndices
+
+The following are integer (as opposed to bit) flags that can be set for particular transactions in the AccountSet SetFlag or ClearFlag field
```
intelliot, line 6403:
```suggestion
To modify account flags, you can use an AccountSet transaction and its `SetFlag` or `ClearFlag` fields.

The flags are called [AccountSet flags (asf*)](https://xrpl.org/accountset.html#accountset-flags):
```

docs/index.md
```diff
@@ -6397,6 +6398,19 @@ The remaining transaction types do not have any flags at this time.
...
+
+`txFlagIndices.AccountSet.asfRequireDest`: Require a destination tag to send transactions to this account.
```
intelliot, line 6405: Perhaps something like this (will require renames/aliases elsewhere):
```suggestion
`RippleAPI.asfRequireDest`: Require a destination tag to send transactions to this account.
```

In xrpl.js, this could be `XRPL.asfRequireDest`.

intelliot: "asf" is already short for "AccountSet flag", I believe.

docs/index.md
natenichols: If were willing to change this, I think we should call them `AccountSetFlags` or `AccountFlags`

intelliot (approved):

src/api.ts
```diff
@@ -487,6 +488,7 @@ class RippleAPI extends EventEmitter {
   rippleTimeToISO8601 = rippleTimeToISO8601 // @deprecated Invoke from top-level package instead
   iso8601ToRippleTime = iso8601ToRippleTime // @deprecated Invoke from top-level package instead
   txFlags = txFlags
+  accountSetFlags = constants.AccountSetFlags
```
intelliot, line 491: Do we need to make this `static` in order for the flags to be accessed in the way that the documentation says?

docs/src/txFlagIndices.md.ejs
```diff
@@ -0,0 +1,15 @@
+## accountSetFlags
```
JST5000, line 1: Is it possible to change the name of this file to accountSetFlagIndices.md.ejs instead of txFlagIndices.md.ejs? (Or would that mess up the docgen?)

PR #1529, opened 2021-08-16
docs: Add documentation for generateFaucetWallet()

natenichols:
Adds documentation for `generateFaucetWallet()`

### Context of Change
Adds a short section to docs w/ an example of usage + response.

### Type of Change

- [x] Documentation Updates
- [ ] Release

### DIFF
--- a/docs/src/generateFaucetWallet.md.ejs
+++ b/docs/src/generateFaucetWallet.md.ejs
@@ -2,7 +2,7 @@
 
 `generateFaucetWallet(onTestnet = true)`
 
-Generates a random wallet with some amount of XRP (usually 1000 XRP).
+Calls the Testnet or Devnet faucet API in order to generate a new, random wallet with some amount of test XRP. This is for testing purposes only.
 
 ### Example
 
--- a/docs/src/generateFaucetWallet.md.ejs
+++ b/docs/src/generateFaucetWallet.md.ejs
@@ -1,0 +1,19 @@
+## generateFaucetWallet
+
+`generateFaucetWallet(onTestnet = true)`
+
+Generates a random wallet with some amount of XRP (usually 1000 XRP).
+
+### Example
+
+**Request**
+
+Create a new wallet on the Tesnet:
+
+```
+const wallet = await api.generateFaucetWallet()
+```
+
+**Response**
+
+<%- renderFixture('responses/generate-faucet-wallet.json') %>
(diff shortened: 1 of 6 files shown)


docs/src/generateFaucetWallet.md.ejs
```diff
@@ -0,0 +1,19 @@
+## generateFaucetWallet
+
+`generateFaucetWallet(onTestnet = true)`
+
+Generates a random wallet with some amount of XRP (usually 1000 XRP).
```
intelliot, line 5:
```suggestion
Calls the Testnet or Devnet faucet API in order to generate a new, random wallet with some amount of test XRP. This is for testing purposes only.
```

src/wallet/wallet-generation.ts
```diff
@@ -4,6 +4,7 @@ import {RippleAPI} from '..'
 import {errors} from '../common'
 import {GeneratedAddress} from '../offline/generate-address'
 import {isValidAddress} from '../common/schema-validator'
+import { RippledError } from '../common/errors'
```
intelliot, line 7:
```suggestion
import {RippledError} from '../common/errors'
```
Just making the spacing a little more consistent here

src/wallet/wallet-generation.ts
```diff
@@ -29,6 +30,9 @@ async function generateFaucetWallet(
   this: RippleAPI,
   address?: string
 ): Promise<FaucetWallet | void> {
+  if(!this.isConnected())
```
intelliot, line 33: I know this would be a "new" feature, but: generally, the ideal would be to just auto-connect if we need to connect and aren't connected yet.

Perhaps beyond the scope of this particular PR

natenichols: Can we `await this.connect()` here, then?

intelliot: I think so... although: what's the harm in not connecting? The faucet API is a service provided by RippleX; it is independent of RippleAPI's rippled connection.

natenichols: The faucet waits on the balance of the account to change before returning.

So if there is not a connection, it'll time out

mvadari (approved):

intelliot (approved):
I know @florent-uzio is on PTO, so after he gets back: if there are other fixes/improvements to make to these docs, we can do it in a later PR :)

PR #1530, opened 2021-08-16
refactor: Define EscrowCreate transaction

natenichols:
Add a Typescript definition for `EscrowCreate` Transaction.

### Context of Change
Adds a typescript type for `EscrowCreate`, `verifyEscrowCreate()`, and tests for form verification.

### Type of Change
- [ ] Documentation Updates
- [ ] Release


src/models/transactions/escrowCreate.ts
```diff
@@ -0,0 +1,54 @@
+import { ValidationError } from "../../common/errors";
+import { BaseTransaction, GlobalFlags, verifyBaseTransaction } from "./common";
+
+export interface EscrowCreateFlags extends GlobalFlags {
+    tfPassive?: boolean;
+    tfImmediateOrCancel?: boolean;
+    tfFillOrKill?: boolean;
+    tfSell?: boolean;
+}
```
mvadari, line 9: Are we doing booleans instead of ints?

natenichols: Yes, we're doing booleans so that you can describe Flags as
```js
{
tfPassive: true,
tfFillOrKill: true,
}
```
It makes it easier to specify flags. We'll also have to specify a constant w/ ints, but that will only be used for processing.

src/models/transactions/escrowCreate.ts
```diff
@@ -0,0 +1,54 @@
+import { ValidationError } from "../../common/errors";
+import { BaseTransaction, GlobalFlags, verifyBaseTransaction } from "./common";
```
khancode, line 2: nit: remove semicolons

khancode (changes requested):
Suggesting the same changes to be made just like with [EscrowCreate PR](https://github.com/ripple/ripple-lib/pull/1511#pullrequestreview-733379253):
- make tests easier to write/read (using beforeEach())
- make error messages more concise

src/models/transactions/escrowCreate.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    if (tx.FinishAfter !== undefined && typeof tx.FinishAfter !== 'number')
+        throw new ValidationError("EscrowCreate: invalid FinishAfter")
+
+    if (tx.Condition !== undefined && typeof tx.Condition !== 'string')
+        throw new ValidationError("EscrowCreate: invalid Condition")
+
+    if (tx.DestinationTag !== undefined && typeof tx.DestinationTag !== 'number')
+        throw new ValidationError("EscrowCreate: invalid DestinationTag")
+}
```
JST5000, line 54: It would be good to check that "Either CancelAfter or FinishAfter must be specified." - https://xrpl.org/escrowcreate.html#escrowcreate

JST5000: (It's ok if both are specified, the important case is CancelAfter and FinishAfter both being missing is disallowed)

JST5000: There is a second case with a similar rule: "With the fix1571 amendment enabled, you must supply FinishAfter, Condition, or both."

test/models/escrowCreate.ts
```diff
@@ -0,0 +1,177 @@
...
+            SourceTag: 11747
+        } as any
+        
+        assert.throws(
+            () => verifyEscrowCreate(invalidDestTag),
+            ValidationError,
+            "EscrowCreate: invalid DestinationTag"
+        )    
+    })
+
```
JST5000, line 176: It'd be good to add a test for each of the for the two cases above:
1. Both CancelAfter and FinishAfter being missing (Should fail validation)
2. Both FinishAfter and Condition being missing (Should fail validation)

JST5000 (changes requested):

mukulljangid:
> * make tests easier to write/read (using beforeEach())
> * make error messages more concise
> * nit: remove semicolons
I suggest similar changes as well!

khancode (approved):
LGTM!

mvadari (approved):
lgtm, but make sure you re-rebase on 2.0

JST5000 (approved):
LGTM!

PR #1534, opened 2021-08-16
Clean utils

JST5000:

Updated references to deprecated functions in api.ts to their new versions in all the tests. (Which removed dependencies on offline/utils.ts)

### Context of Change

This is in response to jira ticket DEVX-194 to clean up the utils folder.

### Type of Change

## Before / After

Before api.ts had a bunch of deprecated functions, now we are removing them for xrpl.js and just using the new architecture (which involves directly calling helper functions from /offline instead of calling aliases for those functions in the api object).

One specific change of note is that submit's new function had a slightly different return value so I updated the test to reflect deprecated return values not showing up, and a new field "success" being part of the output.

## Test Plan

CI Passes

## Future Tasks

[] We will have to update the documentation for Submit since it seems to have a new return structure


src/offline/utils.ts
```diff
@@ -25,12 +25,7 @@ import {
     computeEscrowHash,
```
mvadari, line 25: You can move the utils file to `src/utils` - we're going to get rid of the `offline` directory

src/common/schemas/output/submit.json
```diff
@@ -3,13 +3,9 @@
...
-    "resultCode": {
-      "type": "string",
-      "description": "Deprecated: Use `engine_result` instead."
-    },
-    "resultMessage": {
-      "type": "string",
-      "description": "Deprecated: Use `engine_result_message` instead."
+    "success": {
+      "type": "boolean",
+      "description": "Indicates whether the transaction successfully submitted. On immediate failure it will return false."
```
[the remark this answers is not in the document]
JST5000: Yeah it was causing tests to fail, part of a test was to run validate, which didn't like the new input unless I fixed the jsonschema

src/offline/utils.ts
JST5000: Should I also move the other files in the offline directory that utils borrows functions from? (Or would that be part of getting rid of the `offline` directory?

test/integration/integration-test.ts
```diff
@@ -5,8 +5,14 @@ import requests from '../fixtures/requests'
...
 import {isValidClassicAddress} from 'ripple-address-codec'
 import {payTo, ledgerAccept} from './utils'
+<<<<<<< HEAD
+import {errors} from 'ripple-api/common'
+import {isValidSecret} from 'ripple-api/common/utils'
+import { generateAddressAPI } from '../../src/offline/generate-address'
+=======
 import {errors} from 'xrpl-local/common'
 import {isValidSecret} from 'xrpl-local/common/utils'
+>>>>>>> 2.0
```
mvadari, line 15: You have some merge conflict stuff left here

test/utils.ts
```diff
@@ -29,7 +29,7 @@ export interface TestSuite {
  * If no test suite exists, we return this object with `isMissing: true`
  * so that we can report it.
  */
-interface LoadedTestSuite {
+export interface LoadedTestSuite {
```
mvadari, line 32: These tests can be in the normal format, instead of the special API format (you don't need the API to test them). See the model tests for examples on how to write them.

test/ripple-client-test.ts
```diff
@@ -27,11 +27,30 @@ describe('Client [Test Runner]', function () {
 
   // Collect all the tests:
   const allPublicMethods = getAllPublicMethods(new Client())
-  const allTestSuites = loadTestSuites()
+  const utilTestSuites = loadTestSuitesFromFolder("utils")
+  const apiTestSuites = loadTestSuitesFromFolder("client")
```
mvadari, line 31: `utils` tests don't need to be run in this format, they don't have to be run with two types of addresses and with the client.

test/utils.ts
mvadari: If any tests are written to need the client, then rewrite the tests to not need it.

test/ripple-client-test.ts
JST5000: One test seems to require client.getLedger and I couldn't find any alternatives. Do you happen to know any ways to get a ledger without the Client object? (The specific test is `test/client/computeLedgerHash/index.ts` - I moved it back into the api tests so that it could have access to client until I learn a viable alternative).

Moved the rest into the default format though!

src/client/index.ts
```diff
@@ -451,48 +417,6 @@ class Client extends EventEmitter {
...
-  // @deprecated Invoke from top-level package instead
-  static computeAccountLedgerObjectID = computeAccountLedgerObjectID // (address: string): string
-  // Compute the hash (ID) of an account's SignerList.
-  // @deprecated Invoke from top-level package instead
-  static computeSignerListLedgerObjectID = computeSignerListLedgerObjectID // (address: string): string
-  // Compute the hash of an order, given the owner's classic address (starting with `r`) and the account sequence number of the `OfferCreate` order transaction.
-  // @deprecated Invoke from top-level package instead
-  static computeOrderID = computeOrderID // (address: string, sequence: number): string
-  // Compute the hash of a trustline, given the two parties' classic addresses (starting with `r`) and the currency code.
-  // @deprecated Invoke from top-level package instead
```
mvadari, line 476: These should all be renamed to `compute...Index` as that's the term used by rippled

src/client/index.ts
```diff
@@ -451,48 +417,6 @@ class Client extends EventEmitter {
...
-  static computeEscrowHash = computeEscrowHash // (address, sequence): string
-  // Compute the hash of a payment channel, given the owner's classic address (starting with `r`), the classic address of the destination, and the account sequence number of the `PaymentChannelCreate` payment channel transaction.
-  // @deprecated Invoke from top-level package instead
-  static computePaymentChannelHash = computePaymentChannelHash // (address, dstAddress, sequence): string
-
-  xrpToDrops = xrpToDrops // @deprecated Invoke from top-level package instead
-  dropsToXrp = dropsToXrp // @deprecated Invoke from top-level package instead
-  rippleTimeToISO8601 = rippleTimeToISO8601 // @deprecated Invoke from top-level package instead
-  iso8601ToRippleTime = iso8601ToRippleTime // @deprecated Invoke from top-level package instead
```
mvadari, line 495: These should be named better (see xrpl-py for examples)

src/index.ts
```diff
@@ -4,7 +4,7 @@ export * from './transaction/types'
 
 export * from './common/types/objects/ledger'
 
-export * from './offline/utils'
+export * from './utils';
```
mvadari, line 7: be uniform about semicolon usage/lack thereof

src/utils.ts
```diff
@@ -1,7 +1,7 @@
-import { deriveKeypair, deriveAddress, deriveXAddress } from './derive'
-import computeLedgerHeaderHash from './ledgerhash'
-import signPaymentChannelClaim from './sign-payment-channel-claim'
-import verifyPaymentChannelClaim from './verify-payment-channel-claim'
+import { deriveKeypair, deriveAddress, deriveXAddress } from './offline/derive'
+import computeLedgerHeaderHash from './offline/ledgerhash'
+import signPaymentChannelClaim from './offline/sign-payment-channel-claim'
+import verifyPaymentChannelClaim from './offline/verify-payment-channel-claim'
```
mvadari, line 4: If these are all being treated as utils, you should put all of this in some sort of `utils` folder

src/utils.ts
```diff
@@ -10,7 +10,7 @@ import { dropsToXrp,
     rippleTimeToISO8601,
     iso8601ToRippleTime,
     isValidSecret,
-} from '../common/utils'
+} from './common/utils'
```
mvadari, line 13: ditto

src/utils.ts
```diff
@@ -24,13 +24,8 @@ import {
     computeLedgerHash,
     computeEscrowHash,
     computePaymentChannelHash,
-} from '../common/hashes'
-import { generateAddressAPI, GenerateAddressOptions, GeneratedAddress } from '../offline/generate-address'
```
mvadari, line 28: ditto

src/utils.ts
```diff
@@ -24,13 +24,8 @@ import {
...
-
-// @deprecated Use X-addresses instead
-const generateAddress = (options: GenerateAddressOptions = {}): GeneratedAddress => (
-    generateAddressAPI({...options, includeClassicAddress: true})
```
mvadari, line 32: `generateAddressAPI` is a terrible name, we should change it

test/client/computeLedgerHash/index.ts
```diff
@@ -29,7 +30,7 @@ export default <TestSuite>{
     ledger.parentCloseTime = ledger.closeTime
     let hash
     try {
-      hash = client.computeLedgerHash(ledger, {computeTreeHashes: true})
+      hash = computeLedgerHeaderHash(ledger, {computeTreeHashes: true})
```
mvadari, line 33: See #1550 for how to make these tests not use the client.

test/ripple-client-test.ts
```diff
@@ -27,11 +27,27 @@ describe('Client [Test Runner]', function () {
...
   // Collect all the tests:
   const allPublicMethods = getAllPublicMethods(new Client())
-  const allTestSuites = loadTestSuites()
+  const apiTestSuites = loadTestSuitesFromFolder("client")
 
-  // Run all the tests:
-  for (const {name: methodName, tests, config} of allTestSuites) {
-    describe(`client.${methodName}`, () => {
+  runTests(apiTestSuites, "client.")
+
```
mvadari, line 33: Why were the changes in this file necessary?

yarn.lock
```diff
@@ -338,6 +338,16 @@
...
+"@types/node@10.12.18":
+  version "10.12.18"
+  resolved "https://registry.yarnpkg.com/@types/node/-/node-10.12.18.tgz#1d3ca764718915584fcd9f6344621b7672665c67"
+  integrity sha512-fh+pAqt4xRzPfqA6eh3Z2y6fyZavRIumvjhaCL753+TVkGKGhpPeyrJG2JftD0T9q4GF00KjefsQ+PQNDdWQaQ==
+
+"@types/node@11.11.6":
+  version "11.11.6"
+  resolved "https://registry.yarnpkg.com/@types/node/-/node-11.11.6.tgz#df929d1bb2eee5afdda598a41930fe50b43eaa6a"
+  integrity sha512-Exw4yUWMBXM3X+8oqzJNRqZSwUAaS4+7NdvHqQuFi/d+synz++xmX3QIf+BFqneW8N31R8Ky+sikfZUXq07ggQ==
+
```
mvadari, line 350: Why did this file change? There weren't any changes to `package.json`

mvadari (changes requested):

test/ripple-client-test.ts
JST5000: I suppose they aren't anymore. I made it more extensible for when I had multiple folders that wanted to use the functionality and figured it didn't do harm to keep it in. (Although I also see that it's outside the scope of the task, which  makes it weird to include) Do you think I should revert it back to how it was running before?

src/utils.ts
JST5000: Changed the name to generateXAddress since that's the default behavior

yarn.lock
JST5000: Not sure! After running yarn install the extra lines disappeared. Reverted the changes.

test/ripple-client-test.ts
mvadari: Yes

mvadari:
> [] We will have to update the documentation for Submit since it seems to have a new return structure

It doesn't really matter - we're changing the structure for requests as a whole and removing a bunch of analogous stuff anyways.

test/ripple-client-test.ts
JST5000: Reverted

test/client/computeLedgerHash/index.ts
JST5000: Done

src/client/index.ts
JST5000: @mvadari so would `computeOrderId` become `computerOrderIDIndex`, `computeOrderIndex` or something else? (Are there docs with those Rippled terms somewhere?)

src/common/schema-validator.ts
```diff
@@ -3,7 +3,7 @@ import * as assert from 'assert'
 const {Validator} = require('jsonschema')
 import {ValidationError} from './errors'
 import {isValidClassicAddress, isValidXAddress} from 'ripple-address-codec'
-import {isValidSecret} from './utils'
+import {isValidSecret} from '../utils/index'
```
mvadari, line 6: This should just be `../utils` - the `index` part isn't necessary, all index files are just the "home" for the folder

src/utils/generate-address.ts
```diff
@@ -26,7 +26,7 @@ export interface GenerateAddressOptions {
   includeClassicAddress?: boolean
 }
 
-function generateAddressAPI(options: GenerateAddressOptions = {}): GeneratedAddress {
+function generateXAddress(options: GenerateAddressOptions = {}): GeneratedAddress {
```
mvadari, line 29: This should actually be moved to the `Wallet` class as a static method. But that's probably out of the scope of this PR, so you can just add a `TODO` for now.

src/utils/index.ts
```diff
@@ -189,12 +208,31 @@ function iso8601ToRippleTime(iso8601: string): number {
...
-  iso8601ToRippleTime,
-  isValidSecret
+    computeLedgerHeaderHash,
+    dropsToXrp,
+    xrpToDrops,
+    toRippledAmount,
+    convertKeysFromSnakeCaseToCamelCase,
+    removeUndefined,
+    rippleTimeToISO8601,
+    iso8601ToRippleTime,
```
mvadari, line 218: These names should be changed, they aren't very descriptive (see xrpl-py for naming examples)

src/utils/ledgerhash.ts
```diff
@@ -1,22 +1,24 @@
...
+import _ from 'lodash'
+import { iso8601ToRippleTime } from '.'
+import { ValidationError } from '../common/errors'
 import {
   computeLedgerHash,
   computeTransactionTreeHash,
   computeStateTreeHash
 } from '../common/hashes'
-import * as common from '../common'
+//import * as common from '../common'
```
mvadari, line 9: Remove this commented line

src/utils/sign-payment-channel-claim.ts
```diff
@@ -1,15 +1,16 @@
...
 
 function signPaymentChannelClaim(
   channel: string,
   amount: string,
   privateKey: string
 ): string {
+  console.log("Entered method")
   validate.signPaymentChannelClaim({channel, amount, privateKey})
-
+  console.log("Validated")
```
mvadari, line 13: Remove `console.log`

test/fixtures/get-ledger-full.json
```diff
@@ -0,0 +1,64 @@
+{
+    "stateHash": "2C23D15B6B549123FB351E4B5CDE81C564318EB845449CD43C3EA7953C4DB452",
```
mvadari, line 2: Wasn't this fixture here before? Why is it being re-added here?

test/integration/integration-test.ts
```diff
@@ -74,15 +75,15 @@ function testTransaction(
   const signedData = testcase.client.sign(txJSON, secret)
   console.log('PREPARED...')
   return testcase.client
-    .submit(signedData.signedTransaction)
+    .request({command: "submit", tx_blob: signedData.signedTransaction})
```
mvadari, line 78: Out of the scope of the PR (handled in #1550)

src/utils/index.ts
JST5000: Changed iso8601 to isoTime to make it more intuitive for folks who don't know the acronym (And still keep the unique identifier since iso time is uniquely iso8601)

src/utils/generate-address.ts
JST5000: Added todo

src/common/schema-validator.ts
JST5000: Removed extra /index from all utils imports

src/client/index.ts
JST5000: Replaced 'ID' with 'Index' for these functions (So `computeOrderId` became `computeOrderIndex`) - The docs referenced are the api calls on xrpl.org: e.g. https://xrpl.org/offer.html

test/fixtures/get-ledger-full.json
JST5000: I added this was added by mistake because I saw a ledger-full in a different PR's test, but I couldn't find the file it in the fixture folder (I guess I just missed it). Removed the duplicate!

test/integration/integration-test.ts
JST5000: Merged to include #1550's changes

src/utils/index.ts
```diff
@@ -1,8 +1,27 @@
...
+    computeTrustlineHash,
+    computeTransactionTreeHash,
+    computeStateTreeHash,
+    computeLedgerHash,
+    computeEscrowHash,
+    computePaymentChannelHash,
+} from '../common/hashes'
+import { generateXAddress } from './generate-address'
+
 import * as _ from 'lodash'
```
mvadari, line 21:
```suggestion
import _ from 'lodash'
```

src/utils/index.ts
```diff
@@ -176,25 +195,48 @@ function unixToRippleTimestamp(timestamp: number): number {
   return Math.round(timestamp / 1000) - 0x386d4380
 }
 
-function rippleTimeToISO8601(rippleTime: number): string {
+/**
+ * @param {number} rippleTime is the number of seconds since ripple epoch (1/1/2000 GMT)
```
mvadari, line 199:
```suggestion
 * @param {number} rippleTime is the number of seconds since Ripple Epoch (1/1/2000 GMT)
```

src/utils/index.ts
```diff
@@ -176,25 +195,48 @@ function unixToRippleTimestamp(timestamp: number): number {
...
+function rippleTimeToISOTime(rippleTime: number): string {
   return new Date(rippleToUnixTimestamp(rippleTime)).toISOString()
 }
 
 /**
  * @param {string} iso8601 international standard date format
  * @return {number} seconds since ripple epoch (1/1/2000 GMT)
  */
-function iso8601ToRippleTime(iso8601: string): number {
+function isoTimeToRippleTime(iso8601: string): number {
```
mvadari, line 210: Standardize between this function and the above function on how you capitalize `iso`/`ISO`

src/utils/ledgerhash.ts
```diff
@@ -1,22 +1,23 @@
-import * as _ from 'lodash'
```
mvadari, line 1: rename all these files to be camelCase

src/common/hashes/index.ts
```diff
@@ -84,36 +84,36 @@ export const computeBinaryTransactionSigningHash = (
...
+ * index is unique.
  *
  * See [Ledger Object IDs](https://xrpl.org/ledger-object-ids.html)
  *
  * @param address The classic account address
- * @returns {string} The Ledger Object ID for the account
+ * @returns {string} The Ledger Object Index for the account
  */
-export const computeAccountLedgerObjectID = (address: string): string => {
+export const computeAccountLedgerObjectIndex = (address: string): string => {
```
mvadari, line 99:
```suggestion
export const computeAccountRootIndex = (address: string): string => {
```
The ledger object part is implied, and just makes the function name super long/annoying to type

src/common/hashes/index.ts
```diff
@@ -84,36 +84,36 @@ export const computeBinaryTransactionSigningHash = (
...
  *   * The SignerListID (currently always 0)
  *
- * This method computes a SignerList Ledger Object ID.
+ * This method computes a SignerList Ledger Object index.
  *
  * @param address The classic account address of the SignerList owner (starting with r)
  * @return {string} The ID of the account's SignerList object
  */
-export const computeSignerListLedgerObjectID = (address: string): string => {
+export const computeSignerListLedgerObjectIndex = (address: string): string => {
```
mvadari, line 116:
```suggestion
export const computeSignerListIndex = (address: string): string => {
```

src/common/hashes/index.ts
```diff
@@ -122,17 +122,17 @@ export const computeSignerListLedgerObjectID = (address: string): string => {
...
  *
- * This method computes an Offer ID (aka Order ID).
+ * This method computes an Offer Index (aka Order Index).
  *
  * @param address The classic account address of the SignerList owner (starting with r)
- * @returns {string} The ID of the account's Offer object
+ * @returns {string} The index of the account's Offer object
  */
-export const computeOrderID = (address: string, sequence: number): string => {
+export const computeOrderIndex = (address: string, sequence: number): string => {
```
mvadari, line 135:
```suggestion
export const computeOfferIndex = (address: string, sequence: number): string => {
```
ripple-lib renamed offers to orders, but that's just confusing, so we should undo that

test/hashes-test.ts
```diff
@@ -47,7 +47,7 @@ describe('Ledger', function () {
       var account = 'rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh'
       var expectedEntryHash =
         '2B6AC232AA4C4BE41BF49D2459FA4A0347E1B543A4C92FCEE0821C0201E2E9A8'
-      var actualEntryHash = hashes.computeAccountLedgerObjectID(account)
+      var actualEntryHash = hashes.computeAccountLedgerObjectIndex(account)
```
mvadari, line 50: Don't use * imports

test/utils/computeLedgerHash/index.ts
```diff
@@ -1,35 +1,35 @@
...
 import requests from '../../fixtures/requests'
 import responses from '../../fixtures/responses'
-import {assertResultMatch, TestSuite} from '../../utils'
+import getLedgerFull from '../../fixtures/responses/get-ledger-full.json'
+import {assertResultMatch} from '../../utils'
 const {computeLedgerHash: REQUEST_FIXTURES} = requests
 
 function getNewLedger() {
-  return JSON.parse(JSON.stringify(responses.getLedger.full))
+  return JSON.parse(JSON.stringify(getLedgerFull))
```
mvadari, line 11: These two lines are identical, there's no need to change it - it's a little easier to understand with the old version, actually

test/utils/computeLedgerHash/index.ts
```diff
@@ -1,35 +1,35 @@
...
+    ) 
+    */
```
mvadari, line 23: You can remove this

test/utils/deriveAddress/index.ts
```diff
@@ -0,0 +1,11 @@
+import assert from 'assert-diff'
+import {deriveAddress} from 'ripple-keypairs/dist'
+
+describe('Derive Address', function () {
```
mvadari, line 4: We don't need to test ripple-keypairs code here, it's already tested there - this can be deleted

test/utils/deriveKeypair/index.ts
```diff
@@ -0,0 +1,34 @@
+import assert from 'assert-diff'
+import {deriveKeypair} from 'ripple-keypairs/dist'
```
mvadari, line 2: We don't need to test ripple-keypairs code here, it's already tested there - this can be deleted

src/utils/index.ts
```diff
@@ -1,8 +1,27 @@
...
+    computeAccountRootIndex,
+    computeSignerListIndex,
+    computeOfferIndex,
+    computeTrustlineHash,
+    computeTransactionTreeHash,
+    computeStateTreeHash,
+    computeLedgerHash,
+    computeEscrowHash,
+    computePaymentChannelHash,
+} from '../common/hashes'
```
mvadari, line 18: `hashes` should be moved into `utils` as well

test/utils/generateAddress/index.ts
```diff
@@ -1,78 +1,73 @@
 import assert from 'assert-diff'
 import responses from '../../fixtures/responses'
-import {TestSuite} from '../../utils'
-import {GenerateAddressOptions} from '../../../src/offline/generate-address'
+import {generateXAddress, GenerateAddressOptions} from '../../../src/utils/generateAddress'
+import xAddressResponse from '../../fixtures/responses/generate-x-address.json'
```
mvadari, line 4: I suggest you just use `responses.generateAddress` instead of directly importing the JSON - you're using `responses` anyways, so it helps keep everything symmetrical

test/utils/signPaymentChannelClaim/index.ts
```diff
@@ -0,0 +1,21 @@
+import {assertResultMatch} from '../../utils'
+import {channel, amount} from '../../fixtures/requests/sign-payment-channel-claim.json'
+import signedPaymentChannelClaim from '../../fixtures/responses/sign-payment-channel-claim.json'
```
mvadari, line 3: ditto here

test/utils/signPaymentChannelClaim/index.ts
```diff
@@ -0,0 +1,21 @@
+import {assertResultMatch} from '../../utils'
+import {channel, amount} from '../../fixtures/requests/sign-payment-channel-claim.json'
```
mvadari, line 2: you should just hardcode in the channel and amount, this is more confusing than anything else

test/hashes-test.ts
```diff
@@ -1,6 +1,15 @@
 import assert from 'assert'
```
mvadari, line 1: This file can then also be moved into the `utils` folder

test/utils.ts
```diff
@@ -146,4 +146,4 @@ export function ignoreWebSocketDisconnect(error: Error): void {
     return
   }
   throw error
-}
```
mvadari, line 149: To avoid confusion, I suggest renaming this `testUtils`, so it doesn't seem like this is testing the utils folder.

mvadari (commented):
a few small items left

test/utils/computeLedgerHash/index.ts
```diff
@@ -1,35 +1,27 @@
 import assert from 'assert-diff'
```
mvadari, line 1: I wouldn't put this in an `index` file - it's much neater if it's just `test/utils/computeLedgerHash.ts`. Same with all the other utils files.

mvadari (commented):
Almost there

natenichols:
Should we rename `rippleTimeToISO` to `xrplTimeToISO`?

@intelliot Thoughts?

mvadari (approved):

PR #1537, opened 2021-08-17
refactor: define typescript type for AccountDelete transaction

mukulljangid:
Defines typescript types for AccountDelete transaction.

### Context of Change
AccountDelete is an extension of CommonFields, and is exported in a type union with all other transaction types defined as Transaction.

We also will provide an optional verifyAccountDelete function for runtime form checking of AccountDelete transactions, should the users want to verify transaction form at runtime.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
Adds tests for verifyAccountDelete, which provides runtime form verification for AccountDelete transactions


src/models/transactions/accountDelete.ts
```diff
@@ -0,0 +1,28 @@
...
+    verifyBaseTransaction(tx)
+
+    if (tx.Destination === undefined)
+        throw new ValidationError("AccountDelete: missing field Destination")
+    
+    if (typeof tx.Destination !== 'string')
+        throw new ValidationError("AccountDelete: invalid Destination")
+
+    if (tx.DestinationTag !== undefined && typeof tx.DestinationTag !== 'number')
+        throw new ValidationError("AccountDelete: invalid DestinationTag")
```
mvadari, line 27: Check the fee too - it should be 5 XRP

mukulljangid: Done!

src/models/transactions/accountDelete.ts
```diff
@@ -0,0 +1,28 @@
+import { ValidationError } from "../../common/errors";
+import { BaseTransaction, verifyBaseTransaction } from "./common";
+
+export interface AccountDelete extends BaseTransaction {
+    TransactionType: "AccountDelete";
+    Destination: string;
```
natenichols, line 6: I think we can add `Fee: 5` to this typescript definition. Fee should always be 5 in `AccountDelete` Transactions IIRC.

natenichols (approved):
Left one comment, nothing to hold up approval, though. LGTM 👍 

Just make sure you add that `Fee: 5`

intelliot (changes requested):
The fee can change based on a vote of the validators. We should at least explore/consider the possibility of retrieving the current owner reserve from rippled and using that as the fee.

intelliot (approved):
Actually, I'm fine with removing the check from the model. We should consider a preflight check in the client, as well as ensuring that autofill does the right thing.

PR #1543, opened 2021-08-18
Remove ledger methods from Connection

mvadari:

The server was previously subscribed to ledger updates. We can instead just ask for information about the ledger when we need it.

### Context of Change

xrpl v2.0 client revamp

### Type of Change

## Test Plan

Existing tests were edited/commented out so that tests pass.


mvadari:
Broadcast tests are still not passing, I could use some help with that.

intelliot:
Why no CI?

src/client/broadcast.ts
```diff
@@ -40,23 +40,12 @@ class BroadcastClient extends Client {
...
 
     clients.forEach((client) => {
-      client.on('ledger', this.onLedgerEvent.bind(this))
       client.on('error', (errorCode, errorMessage, data) =>
         this.emit('error', errorCode, errorMessage, data)
       )
     })
   }
 
-  onLedgerEvent(ledger) {
```
natenichols, line 50: Why are we getting rid of this function?

mvadari: This method is only being used because the client was initially automatically subscribed to ledger events, but that's not really necessary. Now that it's not, this method doesn't do anything, so it should be removed.

src/ledger/utils.ts
```diff
@@ -101,13 +102,28 @@ function compareTransactions(
   return first.outcome.ledgerVersion < second.outcome.ledgerVersion ? -1 : 1
 }
 
+function hasLedgerVersions(connection: Connection, min: number, max: number | undefined): Promise<boolean> {
```
natenichols, line 105: This function should be `async` if it returns a promise. Also, I generally prefer `async`/`await` syntax to `.then()` syntax, but its up to you.

src/client/broadcast.ts
natenichols: Gotcha, so using `server_info` to fetch the range of ledgers, instead of listening to the subscriptions.

natenichols (approved):
LGTM 👍

src/client/broadcast.ts
mvadari: Yep, or just making a `ledger` call to get information about a particular ledger

src/client/connection.ts
```diff
@@ -380,41 +324,21 @@ export class Connection extends EventEmitter {
...
+  //   return new Promise((resolve, reject) => {
+  //     if (!this._shouldBeConnected) {
+  //       reject(new NotConnectedError())
+  //     } else if (this._state === WebSocket.OPEN) {
+  //       resolve()
+  //     } else {
+  //       this.once('connected', () => resolve())
+  //     }
+  //   })
+  // }
```
JST5000, line 341: Should this comment be removed now?

src/ledger/utils.ts
```diff
@@ -29,26 +30,24 @@ function getXRPBalance(
...
-  return connection
+  const data = await connection
     .request(request)
-    .then((data) => common.dropsToXrp(data.result.account_data.Balance))
+  return common.dropsToXrp(data.result.account_data.Balance)
 }
 
 // If the marker is omitted from a response, you have reached the end
-function getRecursiveRecur(
+async function getRecursiveRecur(
```
JST5000, line 39: What is this function supposed to do? (I'm not quite clear on when I'd use it from the name)

src/ledger/utils.ts
```diff
@@ -101,38 +100,57 @@ function compareTransactions(
...
-function hasCompleteLedgerRange(
+async function hasLedgerVersions(connection: Connection, min: number, max: number | undefined): Promise<boolean> {
+  const response = await connection.request({
+    command: 'server_info'
+  })
+  const ledgerRange = response.result.info.complete_ledgers
+  const rangeset = new RangeSet()
+  rangeset.parseAndAddRanges(ledgerRange)
+  if (!max) {
+    return rangeset.containsValue(min)
```
JST5000, line 111: This condition will trigger when max=0, and I'm not sure if that's intentional. If it is could we break this into a more explicit condition such as `max == undefined || max == 0`

test/client/preparePayment/index.ts
```diff
@@ -266,19 +266,20 @@ export default <TestSuite>{
...
+  //     maxLedgerVersion: version + 100,
+  //     fee: '0.000012'
+  //   }
+  //   const response = await client.preparePayment(
+  //     address,
+  //     REQUEST_FIXTURES.allOptions,
+  //     localInstructions
+  //   )
+  //   assertResultMatch(response, RESPONSE_FIXTURES.allOptions, 'prepare')
+  // },
```
JST5000, line 282: Should this be removed?

src/ledger/utils.ts
mvadari: I'm not really sure because I don't understand the getter/setter paradigm, but messing with this is out of the scope of this PR.

src/client/connection.ts
JST5000: (Same question for the other large commented out sections below)

src/ledger/utils.ts
mvadari: I'm not really sure because I don't understand the getter/setter paradigm, but messing with this is out of the scope of this PR.

test/mock-rippled.ts
```diff
@@ -46,6 +46,8 @@ function createLedgerResponse(request, response) {
       newResponse.result.ledger.parent_close_time =
         newResponse.result.ledger.close_time - 10
     }
+    newResponse.result.ledger_index = newResponse.result.ledger.ledger_index
+    newResponse.result.ledger_index = newResponse.result.ledger.ledger_index
```
JST5000, line 50: Why is this line there twice?

JST5000 (commented):
A couple questions before I'm comfortable approving

src/ledger/utils.ts
mvadari: Turns out the method isn't used in the repo, so I just deleted it.

JST5000 (approved):
LGTM!

PR #1544, opened 2021-08-18
Edit Client constructor to take a server URI

mvadari:

Title says it all. This PR also renames `ClientBroadcast` to `BroadcastClient`.

### Context of Change

xrpl.js v2.0 client revamp

We're moving all the methods that don't require a client outside of the client, so you should no longer ever need a client without a URI.

### Type of Change

## Test Plan

Edited tests to pass.

### DIFF
--- a/snippets/src/getTransaction.ts
+++ b/snippets/src/getTransaction.ts
@@ -1,8 +1,6 @@
 import {Client} from '../../dist/npm'
 
-const client = new Client({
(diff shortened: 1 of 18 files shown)


JST5000 (approved):
LGTM!

mukulljangid (approved):

natenichols (approved):
LGTM 👍

PR #1548, opened 2021-08-18
refactor: define typescript type for SetRegularKey transaction

mukulljangid:
Defines typescript types for SetRegularKey transaction.

### Context of Change
`SetRegularKey` is an extension of CommonFields, and is exported in a type union with all other transaction types defined as Transaction.

We also will provide an optional `verifySetRegularKey` function for runtime form checking of `SetRegularKey` transactions, should the users want to verify transaction form at runtime.

### Type of Change

## Test Plan
Adds tests for `verifySetRegularKey`, which provides runtime form verification for `SetRegularKey` transactions


mvadari (commented):
Re-rebase on 2.0 - I can't review this until then

natenichols (approved):
LGTM 👍

mvadari:
I recommend rebasing instead of merging with 2.0

mvadari (approved):
looks good, I recommend rebasing instead of merging so the commits aren't as weird

JST5000 (approved):

PR #1550, opened 2021-08-19
Removes methods that were just rippled wrappers

mvadari:

This PR removes all methods from the client that were just wrappers of rippled methods, and didn't really add anything else.

### Context of Change

xrpl.js 2.0

### Type of Change

## Test Plan

CI passes. Tests that tested these methods were deleted.


intelliot:
Nice - it will be interesting to see how much smaller the library gets, with this cruft taken out.

mukulljangid (approved):

snippets/src/getTransaction.ts
```diff
@@ -1,15 +1,19 @@
...
+  const ledger = await client.request({command: 'ledger', transactions: true})
   console.log(ledger)
-  const tx = await client.getTransaction(ledger.transactionHashes[0])
+  const tx = await client.request({
+    command: 'tx',
+    transaction: ledger.result.ledger.transactions[0] as string
+  })
   console.log(tx)
-  console.log('deliveredAmount:', tx.outcome.deliveredAmount)
+  console.log('deliveredAmount:', (tx.result.meta as TransactionMetadata).DeliveredAmount)
```
natenichols, line 17: Shouldn't `ledger` be of type `LedgerResponse`, and thus have metadata w/o needing the typecast?

mvadari: It's possible for `meta` to be a string, if you ask for it to be returned in binary format

natenichols (approved):
👍

snippets/src/getTransaction.ts
natenichols: Ah gotcha, then can we check and not have to type cast.

```ts
if (typeof tx.result.meta !== 'string') {
    console.log(...)
}
```

mvadari: We know that from the request that we sent, it won't be in binary format

PR #1551, opened 2021-08-20
refactor: moves fixtures closer to tests

mvadari:

This PR makes it more intuitive to write tests, by beginning the process of moving fixtures closer to the tests they work with. Now, you can add a fixture in the test itself instead of needing to dig around in `mock-rippled.ts`.

This is just a sample of what the result will be like (just did `account_info`, `account_objects`, and `server_info` for now), because I wanted to get feedback before doing everything.

The most relevant file is `mock-rippled.ts`; the rest of the files are mostly scattering the changes across the tests.

### Context of Change

Making it easier to write/modify tests

### Type of Change

## Test Plan

Tests still pass.

## Future

* Move the rest of the mocks

### DIFF
--- a/test/client/prepareTransaction/index.ts
+++ b/test/client/prepareTransaction/index.ts
@@ -2,7 +2,6 @@ import {ValidationError} from 'xrpl-local/common/errors'
 // import requests from '../../fixtures/requests'
 import responses from '../../fixtures/responses'
 import rippled from '../../fixtures/rippled'
-import { addRippledResponse } from '../../mock-rippled'
 import {assertRejects, assertResultMatch, TestSuite} from '../../utils'
 const instructionsWithMaxLedgerVersionOffset = {maxLedgerVersionOffset: 100}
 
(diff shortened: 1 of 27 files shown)


test/client/prepareTransaction/index.ts
```diff
@@ -569,24 +569,24 @@ export default <TestSuite>{
...
-      Authorize: 'rpZc4mVfWUif9CRoHRKKcmhu1nx2xktxBo'
-    }
-    await assertRejects(
-      client.prepareTransaction(txJSON, localInstructions),
-      RippledError,
-      'Account not found.'
-    )
-  },
+  // 'rejects Promise when Account is valid but non-existent on the ledger': async (
+  //   client
```
mvadari, line 573: These tests will be removed when we get rid of the `prepare` methods anyways.

intelliot (commented):
This looks good, but for your consideration, I came up with an idea to make it a bit more concise: https://github.com/ripple/ripple-lib/pull/1555

Let me know what you think :)

mvadari:
I like it! Mind if I just merge it into this branch? Or you can :)

intelliot:
Done!

intelliot (approved):
LGTM

natenichols:
So, if we wanna do this w/ an `addResponse` function, can we just rip out most of the `createMockRippled` function?

natenichols (approved):

PR #1552, opened 2021-08-20
refactor: add verify() function

mukulljangid:
Adds `verify()` which can take the `transaction` and check specifications for a particular `type`.

### Context of Change
Having one generic verification function to verify all types of transactions.

### Type of Change

## Test Plan
Added tests for all types of transactions and the `verify()` function validates it and gives out the right `error`.


src/models/transactions/verify.ts
```diff
@@ -0,0 +1,86 @@
...
+         verifyPaymentChannelClaim, 
+         verifyPaymentChannelCreate, 
+         verifyPaymentChannelFund, 
+         verifyPaymentTransaction, 
+         verifySetRegularKey, 
+         verifySignerListSet, 
+         verifyTicketCreate, 
+         verifyTrustSet } from ".";
+
+export function verify(tx: Transaction) {
```
mvadari, line 25: This method should be in `Transaction` and should also be exported at the `src/models/transactions` level

test/models/verify.ts
```diff
@@ -0,0 +1,393 @@
...
+import { assert } from 'chai'
+
+/**
+ * Transaction Verification Testing
+ *
+ * Providing runtime verification testing for each specific transaction type
+ */
+describe('Global Transaction Verification', () => {
+
+    it (`verifies valid AccountDelete`, () => {
```
mvadari, line 12: Nest tests that are for the same transaction type in another `describe`

src/models/transactions/verify.ts
```diff
@@ -0,0 +1,86 @@
...
+        verifySetRegularKey(tx)
+
+    if (tx.TransactionType === 'SignerListSet')
+        verifySignerListSet(tx)
+
+    if (tx.TransactionType === 'TicketCreate')
+        verifyTicketCreate(tx)
+
+    if (tx.TransactionType === 'TrustSet')
+        verifyTrustSet(tx)
```
natenichols, line 85: I would personally use a [switch](https://www.tutorialsteacher.com/typescript/typescript-switch) here, but

test/models/verify.ts
natenichols: What you could do for the tests is to just add `verify` to each of the existing tests, and expect the same result as the specific `verify` functions.

src/models/transactions/verify.ts
mukulljangid: Used switch!

test/models/verify.ts
mukulljangid: Added verify to all!

src/models/transactions/verify.ts
mukulljangid: Done!

src/models/transactions/transaction.ts
```diff
@@ -44,3 +47,98 @@ export interface TransactionAndMetadata {
...
+/**
+ * 
+ * @param {Transaction} tx A Transaction.
+ * @returns {void}
+ * @throws {ValidationError} When the Transaction is malformed.
+ */
+ export function verify(tx: Transaction) {
+    
+  if(!isEqual(decode(encode(tx)),tx))
+      throw new ValidationError(`Invalid Transaction: ${tx.TransactionType}`)
```
mvadari, line 60: I'd put this after the switch, because those errors are more specific.

src/models/transactions/transaction.ts
```diff
@@ -44,3 +47,98 @@ export interface TransactionAndMetadata {
...
+    case 'TicketCreate':
+      verifyTicketCreate(tx)
+      break
+  
+    case 'TrustSet':
+      verifyTrustSet(tx)
+      break
+
+    default:
+      throw new ValidationError(`Invalid TransactionType`)
```
mvadari, line 141: Add the tr
```suggestion
      throw new ValidationError(`Invalid TransactionType: ${tx.TransactionType}`)
```

src/models/transactions/transaction.ts
```diff
@@ -44,3 +47,98 @@ export interface TransactionAndMetadata {
...
+
+/**
+ * 
+ * @param {Transaction} tx A Transaction.
+ * @returns {void}
+ * @throws {ValidationError} When the Transaction is malformed.
+ */
+ export function verify(tx: Transaction) {
+    
+  if(!isEqual(decode(encode(tx)),tx))
```
mvadari, line 59: Maybe also add a check for `onlyHasFields` (a helper function in `models`) before this equality is checked, as it will also help to provide better error messages

test/models/paymentTransaction.ts
```diff
@@ -14,23 +15,33 @@ describe('Payment Transaction Verification', () => {
...
+            Amount: {
+                currency: "USD",
+                value: "1",
+                issuer: "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn"
+             },
             Destination: 'rfkE1aSy9G8Upk4JssnwBxhEv5p4mn2KTy',
             DestinationTag: 1,
-            InvoiceID: '6F1DFD1D0FE8A32E40E1F2C05CF1C15545BAB56B617F9C6C2D63A6B704BEF59B',
-            Paths: [[{ account: 'aw0efji', currency: 'XRP', issuer: 'apsoeijf90wp34fh'}]],
-            SendMax: '100000000',
```
mvadari, line 22: Why did this change?

test/models/accountDelete.ts
```diff
@@ -34,7 +38,10 @@ describe('AccountDelete Transaction Verification', function () {
...
 
         assert.throws(
-            () => verifyAccountDelete(invalidDestination),
+            () => {
+                verifyAccountDelete(invalidDestination)
+                verify(invalidDestination)
+            },
             ValidationError,
             "AccountDelete: missing field Destination"
         )
```
mvadari, line 47: Each of the `assert.throws` needs a separate `assert.throws` for `verify[SpecificTransaciton]` vs `verify` - `assert.throws` only checks if an error is thrown somewhere in the block, not that each throws an error

mvadari (changes requested):

src/models/transactions/transaction.ts
```diff
@@ -1,24 +1,39 @@
...
+import {
+  PaymentChannelFund,
+  verifyPaymentChannelFund,
+} from "./paymentChannelFund";
+import { SetRegularKey, verifySetRegularKey } from "./setRegularKey";
+import { SignerListSet, verifySignerListSet } from "./signerListSet";
+import { TicketCreate, verifyTicketCreate } from "./ticketCreate";
+import { TrustSet, verifyTrustSet } from "./trustSet";
+import { ValidationError } from "../../common/errors";
+import { isEqual, omitBy, isUndefined } from "lodash";
```
mvadari, line 35: You should instead do `import _ from "lodash"` and then use `_.function` elsewhere. Also, don't use `isUndefined` - it's much easier just to use `== null`.

src/models/transactions/transaction.ts
```diff
@@ -1,24 +1,39 @@
...
+import { encode, decode } from "ripple-binary-codec";
```
mvadari, line 36: Run the linter on your new code - this doesn't look like it's even autofixing right now.

src/models/transactions/transaction.ts
```diff
@@ -45,3 +60,96 @@ export interface TransactionAndMetadata {
...
 }
+
+/**
+ *
+ * @param {Transaction} tx A Transaction.
+ * @returns {void}
+ * @throws {ValidationError} When the Transaction is malformed.
+ */
+export function verify(tx: Transaction): void {
+  tx = <Transaction>omitBy(tx, isUndefined);
```
mvadari, line 71: What is the objective of this?

test-compiled-for-web/integration-test.js
```diff
@@ -0,0 +1,2 @@
+/*! For license information please see integration-test.js.LICENSE.txt */
```
mvadari, line 1: This folder should be deleted - it was renamed to `testCompiledForWeb`

test/models/depositPreauth.ts
```diff
@@ -22,12 +21,15 @@ describe("DepositPreauth Transaction Verification", function () {
...
-  it("verifies valid DepositPreauth when only Unauthorize is provided", function () {
-    depositPreauth.Unauthorize = "raKEEVSGnKSD9Zyvxu4z6Pqpm4ABH8FS6n";
-    assert.doesNotThrow(() => verifyDepositPreauth(depositPreauth));
-  });
+  // it("verifies valid DepositPreauth when only Unauthorize is provided", function () {
+  //   depositPreauth.Unauthorize = "raKEEVSGnKSDv9Zyvxu4z6Pqpm4ABH8FS6n";
+  //   assert.doesNotThrow(() => verifyDepositPreauth(depositPreauth));
+  //   // assert.doesNotThrow(() => verify(depositPreauth));
+  //   verify(depositPreauth);
+  // });
```
mvadari, line 32: Why was this test removed?

mvadari (changes requested):
A few more changes + questions

src/models/transactions/transaction.ts
```diff
@@ -1,24 +1,42 @@
+// eslint-disable max-lines-per-function
+// eslint-disable import/max-dependencies
```
mvadari, line 2: This should disable it for the whole file, instead of re-enabling at the bottom

src/models/transactions/transaction.ts
```diff
@@ -45,3 +63,104 @@ export interface TransactionAndMetadata {
   transaction: Transaction;
   metadata: Metadata;
 }
+
+/**
+ *
+ * @param tx - A Transaction.
+ * @returns
+ * @throws {ValidationError} When the Transaction is malformed.
+ */
```
mvadari, line 72: This is mildly incorrect TSdocs formatting.

natenichols: Linter will fix this

natenichols (approved):
My only question is about the test that was removed, and that shouldn't hold up approval imo.

src/models/transactions/transaction.ts
mvadari: I know, I just want to resolve linter errors in new code now, so we don't have to deal with it later

test/models/depositPreauth.ts
mukulljangid: Added it back!

src/models/transactions/transaction.ts
mukulljangid: This has now been changed and done correctly

natenichols (approved):

natenichols:
This looks good in its current form, just resolve the conficts.

test/models/accountDelete.ts
```diff
@@ -36,9 +36,15 @@ describe('AccountDelete', function () {
     assert.throws(
       () => verifyAccountDelete(invalidDestination),
       ValidationError,
-      'AccountDelete: missing field Destination',
-    )
-  })
+      "AccountDelete: missing field Destination"
```
mvadari, line 39: these files need to be linted properly

test/models/depositPreauth.ts
mvadari: It's still commented out?

test/models/escrowCreate.ts
```diff
@@ -109,14 +139,19 @@ describe('EscrowCreate', function () {
...
+    )
   })
 
   it(`Missing both CancelAfter and FinishAfter`, function () {
     delete escrow.CancelAfter
     delete escrow.FinishAfter
 
     assert.throws(
-      () => verifyEscrowCreate(escrow),
+      () => verify(escrow),
```
mvadari, line 154: This should also throw instead of throwing instead

test/models/offerCancel.ts
```diff
@@ -14,27 +13,30 @@ describe('OfferCancel', function () {
 
   beforeEach(function () {
     offer = {
-      Account: 'rnKiczmiQkZFiDES8THYyLA2pQohC5C6EF',
```
mvadari, line 17: Why was this replaced?

test/models/payment.ts
```diff
@@ -17,20 +18,30 @@ describe('Payment', function () {
 
   beforeEach(function () {
     paymentTransaction = {
+      Account: 'rEDkMjt2bUtbwJDrp2Z2G54p8XauGjDkfq',
```
mvadari, line 21: Why was this replaced?

test/models/payment.ts
```diff
@@ -103,18 +149,26 @@ describe('Payment', function () {
...
     paymentTransaction.DeliverMin = '10000'
     paymentTransaction.Flags = PaymentTransactionFlagsEnum.tfPartialPayment
     assert.doesNotThrow(() => verifyPayment(paymentTransaction))
+    assert.doesNotThrow(() => verify(paymentTransaction))
   })
 
   it(`verifies valid DeliverMin with tfPartialPayment flag set as a boolean`, function () {
     paymentTransaction.DeliverMin = '10000'
-    paymentTransaction.Flags = { tfPartialPayment: true }
+    paymentTransaction.Flags = PaymentTransactionFlagsEnum.tfPartialPayment
```
mvadari, line 168: This should work correctly as previously, right?

test/models/paymentTransaction.ts
```diff
@@ -0,0 +1,239 @@
+import { ValidationError } from 'xrpl-local/common/errors'
```
mvadari, line 1: This is an old file and shouldn't exist.

test/models/setRegularKey.ts
```diff
@@ -24,11 +25,13 @@ describe('SetRegularKey', function () {
 
   it(`verifies valid SetRegularKey`, function () {
     assert.doesNotThrow(() => verifySetRegularKey(account))
+    assert.doesNotThrow(() => verify(account))
   })
 
   it(`verifies w/o SetRegularKey`, function () {
     account.RegularKey = undefined
-    assert.doesNotThrow(() => verifySetRegularKey(account))
+    // assert.doesNotThrow(() => verifySetRegularKey(account));
```
mvadari, line 33: This line shouldn't be commented out

test/models/signerListSet.ts
```diff
@@ -74,5 +86,19 @@ describe('SignerListSet', function () {
...
     )
+    assert.throws(
+      () => verify(SignerListSetTx),
+      ValidationError,
+      'SignerListSet: invalid SignerEntries',
+    )
+  })
+
+  it(`throws w/ extra field`, function () {
+    SignerListSetTx.ExtraField = 'garbage'
```
mvadari, line 97: Run this through `verifySignerListSet` as well

test/models/signerListSet.ts
```diff
@@ -44,6 +45,7 @@ describe('SignerListSet', function () {
 
   it(`verifies valid SignerListSet`, function () {
     assert.doesNotThrow(() => verifySignerListSet(SignerListSetTx))
+    assert.doesNotThrow(() => verify(SignerListSetTx))
```
mvadari, line 48: this should be camelCase: 
```suggestion
    assert.doesNotThrow(() => verify(signerListSetTx))
```

mvadari (commented):
looks pretty good, a few minor comments

test/models/offerCancel.ts
```diff
@@ -19,22 +18,24 @@ describe('OfferCancel', function () {
...
       Sequence: 60797535,
+      Flags: 2147483648,
       SigningPubKey:
-        '0361BFD43D1EEA54B77CC152887312949EBF052997FBFFCDAF6F2653164B55B21...',
+        '0369C9BC4D18FAE741898828A1F48E53E53F6F3DB3191441CC85A14D4FC140E031',
       TransactionType: 'OfferCancel',
       TxnSignature:
-        '30450221008C43BDCFC68B4793857CA47DF454C07E5B45D3F80E8E6785CAB9292...',
-      date: '2021-08-06T21:04:11Z',
+        '304402203EC848BD6AB42DC8509285245804B15E1652092CC0B189D369E12E563771D049022046DF40C16EA05DC99D01E553EA2E218FCA1C5B38927889A2BDF064D1F44D60F0',
```
mvadari, line 26: This was left as before because it's probably a linter error for a too-long line - these models don't actually care whether the signature is correct or not

test/models/payment.ts
```diff
@@ -17,20 +18,27 @@ describe('Payment', function () {
...
       DestinationTag: 1,
+      Fee: '12',
+      Flags: 2147483648,
+      LastLedgerSequence: 65953073,
+      Sequence: 65923914,
+      SigningPubKey:
+        '02F9E33F16DF9507705EC954E3F94EB5F10D1FC4A354606DBE6297DBB1096FE654',
+      TransactionType: 'Payment',
+      TxnSignature:
+        '3045022100E3FAE0EDEC3D6A8FF6D81BC9CF8288A61B7EEDE8071E90FF9314CB4621058D10022043545CF631706D700CEE65A1DB83EFDD185413808292D9D90F14D87D3DC2D8CB',
```
mvadari, line 33: ditto

test/models/payment.ts
```diff
@@ -17,20 +18,27 @@ describe('Payment', function () {
 
   beforeEach(function () {
     paymentTransaction = {
-      TransactionType: 'Payment',
```
mvadari, line 20: Why were things deleted from this object?

mvadari (approved):
please address all the comments before merging

test/models/offerCancel.ts
mukulljangid: enode/decode won't work. it decodes it back to the sign without the dots and it does not match the original tx!

PR #1554, opened 2021-08-23
fix: websocket wouldn't close properly on erroring test

mvadari:

When a test failed because of an error in its code, the `afterEach` script would fail because the websocket wouldn't properly disconnect. This fixes that problem. A different error is thrown, but it's also pretty descriptive, so it's fine.

### Context of Change

test cleanup

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

CI passes.


mvadari:
I'm gonna write a test for this first

natenichols (approved):
LGTM 👍

mukulljangid (approved):

PR #1556, opened 2021-08-23
refactor: clean up Client and associated files

mvadari:

This PR:
* removes `_PRIVATE` from the client and has tests directly import things
* makes `client.requestAll` a public function
* types `connection.request` with `any`, so that if someone wants to send something that isn't a request type we support (such as a hooks-related request), they are able to with a simple workaround
* Fixes lodash imports so they don't use *
* Makes the default fee cushion 1

### Context of Change

xrpl.js refactor

### Type of Change

## Test Plan

CI passes.


natenichols:
Why are we changing the default fee to 1?

natenichols (approved):
LGTM 👍

mvadari:
I don't really see a reason it should be 1.2 - seems like it could be a source of confusion.

intelliot:
I don't think I've ever seen someone actually confused by 1.2 -- my preference would be to keep it at 1.2 so that people know that fees aren't fixed -- but it's fine either way. In practice, there's no difference (yet) between 1 and 1.2 in terms of how long it takes for a tx to get processed. Curious if @mDuo13 has an opinion?

PR #1557, opened 2021-08-24
Refactoring existing hash utils and adding tests

JST5000:

Updating the naming and validation used by hash utils in ripple-lib and adding test cases for robustness.

### Context of Change

As part of refactoring for ripple-lib 2.0

### Type of Change

- [X] Breaking change / Refactor

## Before / After

Added new toHash function and tests for it.

## Test Plan

Added a CI test for correct hashing and a test for throwing if there is an unsigned transaction.


src/models/utils/index.ts
```diff
@@ -1,3 +1,8 @@
+import { encode } from "ripple-binary-codec/dist"
```
mvadari, line 1: You don't need the `/dist` in `src` - that's a testing-specific thing. You treat it like any other package here.

test/toHash.ts
```diff
@@ -0,0 +1,22 @@
+import {ValidationError} from "../src/common/errors"
+import {OfferCreate} from "../src/models/transactions"
+import {toHash} from "../src/models/utils"
+import {assertResultMatch} from "./utils"
+import {assert} from 'chai'
+import fixtures from './fixtures/rippled'
+
+describe('AccountDelete Transaction Verification', function () {
```
mvadari, line 8: this should be renamed

src/models/utils/index.ts
```diff
@@ -19,3 +24,23 @@ export function onlyHasFields(obj: object, fields: Array<string>): boolean {
...
+/**
+ * Hashes the Transaction object as the ledger does. Only valid for signed Transactions.
+ * 
+ * @param {Transaction} tx A transaction to hash. Tx must be signed.
+ * @returns {string} A hash of tx
+ * @throws {ValidationError} if the Transaction is unsigned
+ */
+
+// TODO: Investigate 0x53545800 if single-signing, or 0x534D5400 if multisigning - https://xrpl.org/serialization.html#blob-fields 
+// Could be a bug in the xrpl-py method!
```
mvadari, line 37: This is not a bug, that's a signing-specific prefix and not relevant for already-signed transactions.

src/models/utils/index.ts
```diff
@@ -19,3 +24,23 @@ export function onlyHasFields(obj: object, fields: Array<string>): boolean {
...
+
+export function toHash(tx: Transaction): string {
+    if(tx.TxnSignature === undefined) {
+        throw new ValidationError("The transaction must be signed to hash it.")
+    }
+    const _TRANSACTION_HASH_PREFIX = "54584E00"
+    const encodedStr = _TRANSACTION_HASH_PREFIX.concat(encode(tx))
+    return sha512Half(encodedStr)
+}
```
mvadari, line 46: You should delete/move `computeBinaryTransactionHash`

mvadari (changes requested):

src/common/hashes/index.ts
```diff
@@ -59,13 +61,31 @@ const addLengthPrefix = (hex: string): string => {
...
-  return computeBinaryTransactionHash(encode(txJSON))
+/**
+ * Hashes the Transaction object as the ledger does. Throws if the transaction is unsigned.
+ * 
+ * @param {Transaction} tx A transaction to hash. Tx must be signed.
+ * @returns {string} A hash of tx
+ * @throws {ValidationError} if the Transaction is unsigned
+ */
+
+export const computeSignedTransactionHash = (tx: Transaction): string => {
```
mvadari, line 84: I would combine `computeSignedTransactionHash` and `computeBinaryTransactionHash`, and have `tx` be either a string or a `Transaction`. Then you can just check the type and encode if needed.

src/models/utils/index.ts
JST5000: Handled in #1534 which moves the hashes functions to the utils folder

src/client/index.ts
```diff
@@ -472,10 +471,7 @@ class Client extends EventEmitter {
    */
   // Compute the hash of a binary transaction blob.
   // @deprecated Invoke from top-level package instead
-  static computeBinaryTransactionHash = computeBinaryTransactionHash // (txBlobHex: string): string
-  // Compute the hash of a transaction in txJSON format.
-  // @deprecated Invoke from top-level package instead
-  static computeTransactionHash = computeTransactionHash // (txJSON: any): string
+  static computeSignedTransactionHash = computeSignedTransactionHash // (txBlobHex: string): string
```
mvadari, line 474: Eh no point in adding it back if you don't need to

src/offline/ledgerhash.ts
```diff
@@ -113,7 +113,7 @@ function computeLedgerHeaderHash(
   options: ComputeLedgerHeaderHashOptions = {}
 ): string {
   const subhashes = {
-    transactionHash: computeTransactionHash(ledger, options),
+    transactionHash: computeSignedTransactionHash(ledger, options),
```
mvadari, line 116: Why are options being passed in here?

test/hashesTest.ts
```diff
@@ -149,4 +164,44 @@ describe('Ledger', function () {
       assert.equal(actualEntryHash, expectedEntryHash)
     })
   })
+
+  describe('Transaction Hash Verification', function () {
```
mvadari, line 168: why is this not part of the above `describe`?

src/offline/ledgerhash.ts
JST5000: This function happens to have the exact same name as the one that was being exported, but it's serving a slightly different function. (It gets the Transaction Hash from a Ledger object instead of from a Transaction object and also performs some additional verifications to ensure the Ledger object is properly formed)

I renamed it to be more descriptive (To `computeTransactionHashFromLedger`)

test/hashesTest.ts
JST5000: The rest of the tests in this file have a structure of `Describe( functionName )` followed by `it( 'will do this thing and see what happens' ) - @mvadari, Should I rewrite the structure to match that of the other tests we have?

src/transaction/combine.ts
```diff
@@ -1,12 +1,12 @@
 import BigNumber from "bignumber.js";
-import _ from "lodash";
+import * as _ from "lodash";
```
mvadari, line 2: This is incorrect - `_` basically is already `* as _`

src/utils/hashes/index.ts
```diff
@@ -62,13 +65,37 @@ const addLengthPrefix = (hex: string): string => {
...
-export const computeBinaryTransactionHash = (txBlobHex: string): string => {
-  const prefix = HashPrefix.TRANSACTION_ID.toString(16).toUpperCase();
-  return sha512Half(prefix + txBlobHex);
-};
+/**
+ * Hashes the Transaction object as the ledger does. Throws if the transaction is unsigned.
+ *
+ * @param {Transaction | string} tx - A transaction to hash. Tx may be in binary blob form. Tx must be signed.
+ * @returns {string} A hash of tx.
+ * @throws {ValidationError} If the Transaction is unsigned.
```
mvadari, line 73: This is the incorrect JSdocs format. You can run the linter on just the files you're messing with using `eslint path/to/file`

mvadari (approved):

src/utils/hashes/index.ts
```diff
@@ -44,32 +50,56 @@ const currencyToHex = (currency: string): string => {
...
     return bytesToHex([length]) + hex;
   }
   if (length <= 12480) {
     const x = length - 193;
+    // eslint-disable-next-line no-bitwise -- adding a prefix to hex requires bitwise operations
     return bytesToHex([193 + (x >>> 8), x & 0xff]) + hex;
   }
   if (length <= 918744) {
     const x = length - 12481;
+    // eslint-disable-next-line no-bitwise -- adding a prefix to hex requires bitwise operations
```
natenichols, line 67: you can probably just disable this `no-bitwise` at the top of the file, since it is widely used throughout the file

src/transaction/combine.ts
```diff
@@ -1,12 +1,12 @@
...
 import { decodeAccountID } from "ripple-address-codec";
 import binary from "ripple-binary-codec";
 import { JsonObject } from "ripple-binary-codec/dist/types/serialized-type";
 
 import { validate } from "../common";
 import { ValidationError } from "../common/errors";
-import { computeBinaryTransactionHash } from "../utils/hashes";
+import { computeSignedTransactionHash } from "../utils/hashes";
```
natenichols, line 9: Why'd we rename this to ComputeSignedTransactionHash?

natenichols (approved):
Awesome work. Looks good to me 👍

PR #1564, opened 2021-08-25
Refactor generateFaucetWallet to return a Wallet

khancode:

Refactor `generateFaucetWallet()` to return a `Wallet`

### Context of Change
- Adds `Wallet.create()` that generates a new seed and creates a new `Wallet`
- `Wallet.create()` is used in `generateFaucetWallet()` to create and return a new Wallet to be funded
- Refactored tests for `Wallet` class.

### Type of Change

## Test Plan

adds/updates tests for `Wallet`


src/Wallet.ts
```diff
@@ -16,12 +16,26 @@ import {ValidationError} from './common/errors'
 class Wallet {
   readonly publicKey: string
   readonly privateKey: string
+  readonly classicAddress: string
+  readonly seed?: string
```
mvadari, line 20: What is the purpose in adding a `seed` instance variable to the wallet?

khancode: In the case of creating a new Wallet (using `Wallet.create()`), I think it's helpful for users to know what `seed` was used that created a `Wallet` instance so they can reuse this seed to derive the same wallet in the future.

mvadari: Can you not create a wallet from the private key?

src/Wallet.ts
```diff
@@ -16,12 +16,26 @@ import {ValidationError} from './common/errors'
...
+    this.classicAddress = deriveAddress(publicKey)
+    this.seed = seed
+  }
+
+  /**
+   * Creates a new Wallet using a generated seed
+   * @param {ECDSA} algorithm The digital signature algorithm to generate an address for.
+   * @returns {Wallet} A new Wallet derived from a generated seed.
+   */
+  static create(algorithm: ECDSA = Wallet.defaultAlgorithm): Wallet {
```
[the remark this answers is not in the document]
khancode: Sure, I renamed it to `Wallet.generate`.

test/wallet/index.ts
```diff
@@ -0,0 +1,210 @@
+import {assert} from 'chai'
+import ECDSA from '../../src/common/ecdsa'
+import Wallet from '../../src/Wallet'
+
+/**
+ * Wallet testing
+ *
+ * Provides tests for Wallet class
+ */
+describe('Wallet', () => {
```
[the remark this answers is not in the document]
khancode: These were deleted in [add getXAddress to Wallet commit](https://github.com/ripple/ripple-lib/commit/64132defe7c948c8c3e122419b7759065f20faa3)

src/wallet/generateFaucetWallet.ts
```diff
@@ -5,6 +5,7 @@ import { errors } from "../common";
 import { RippledError } from "../common/errors";
 import { isValidAddress } from "../common/schema-validator";
 import { GeneratedAddress } from "../utils/generateAddress";
+import Wallet from ".";
```
mvadari, line 8: This likely causes a dependency cycle

mvadari (commented):
can you make sure your code doesn't have any linter issues, so we don't have to re-lint new code?

src/wallet/generateFaucetWallet.ts
khancode: From my understanding, a dependency cycle would occur if `wallet/index.ts` and `generateFaucetWallet.ts `referenced each other which isn't the case. Only `generateFaucetWallet.ts` references `wallet/index.ts`.

khancode:
> can you make sure your code doesn't have any linter issues, so we don't have to re-lint new code?

Done.

src/wallet/index.ts
```diff
@@ -23,12 +23,27 @@ import { SignOptions } from "./transaction/types";
 class Wallet {
   readonly publicKey: string;
   readonly privateKey: string;
+  readonly classicAddress: string;
+  readonly seed?: string;
   private static readonly defaultAlgorithm: ECDSA = ECDSA.ed25519;
   private static readonly defaultDerivationPath: string = "m/44'/144'/0'/0/0";
 
-  constructor(publicKey: string, privateKey: string) {
+  constructor(publicKey: string, privateKey: string, seed?: string) {
```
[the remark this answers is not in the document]
khancode: Yeah it does look weird but `Wallet.generate()` calls this constructor to save the new seed created with its public/private keypair.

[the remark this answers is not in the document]
mvadari: Like here: https://github.com/XRPLF/xrpl-py/blob/master/xrpl/wallet/main.py#L19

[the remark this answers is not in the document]
khancode: This would be possible except the issue is with `Wallet.fromMnemonic()` where its seed value is in a different format than the other derive functions. It uses `bip32` instead of `ripple-keypairs` to derive public/private keypair from its seed value.

src/wallet/generateFaucetWallet.ts
```diff
@@ -23,40 +23,45 @@ const MAX_ATTEMPTS = 20; // Maximum attempts to retrieve a balance
...
-    if (addressToFundBalance && !isNaN(Number(addressToFundBalance))) {
-      startingBalance = Number(addressToFundBalance);
-    } else {
-      startingBalance = 0;
-    }
+  const faucetUrl = getFaucetUrl(client);
+
+  // If no existing Wallet is provided or its address is invalid to fund
+  if (!wallet || !isValidAddress(wallet.classicAddress)) {
+    wallet = Wallet.generate();
```
ledhed2222, line 45: you're re-assigning the parameter here. no bueno

src/wallet/generateFaucetWallet.ts
```diff
@@ -23,40 +23,45 @@ const MAX_ATTEMPTS = 20; // Maximum attempts to retrieve a balance
...
+  }
+
+  // Create the POST request body
+  body = new TextEncoder().encode(
```
ledhed2222, line 49: AFAICT the `body` could be a `const` defined here. no need to `let`. IMO `let` should be used only when really really necessary

src/wallet/generateFaucetWallet.ts
```diff
@@ -23,40 +23,45 @@ const MAX_ATTEMPTS = 20; // Maximum attempts to retrieve a balance
...
+  const addressToFundBalance = await getAddressXrpBalance(
+    client,
+    wallet.classicAddress
+  );
+
+  // Check the address balance is not undefined and is a number
+  if (addressToFundBalance && !isNaN(Number(addressToFundBalance))) {
+    startingBalance = Number(addressToFundBalance);
+  } else {
+    startingBalance = 0;
```
ledhed2222, line 64: you could clean this up like so

```typescript
// remove the let starting balance from above
....

let startingBalance = 0
if (addressToFundBalance && !isNaN(Number(addressToFundBalance)) {
  startingBalance = Number(addressToFundBalance)
}
```

this removes the `else` entirely

src/wallet/generateFaucetWallet.ts
khancode: Ah you're right, I just noticed this `body` can be refactored to be a `const` now.

src/wallet/generateFaucetWallet.ts
khancode: Using a ternary operator would be better in this case:
```TypeScript
let startingBalance =
    addressToFundBalance && !isNaN(Number(addressToFundBalance))
      ? Number(addressToFundBalance)
      : 0;
```

ledhed2222 (approved):

mvadari (approved):

PR #1565, opened 2021-08-26
refactor: migrate fixtures to use typescript

natenichols:
Uses typescript to import fixtures. 

### Context of Change
Fixtures were previously imported using Javascript, so the form of the fixtures was not taken into account when writing tests. With typescript, you get an error when using a property on a fixture that is the wrong type or does not exist.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


khancode (approved):
This is awesome, LGTM!

PR #1566, opened 2021-08-26
test: removes the use of `TestSuite`

mvadari:

This PR makes it easier for other people to add tests by making the testing framework more approachable for an outsider. It removes the use of `TestSuite` and returns to `describe`/`it` syntax. It kind of undoes #1086, but still keeps the spirit.

There are two main types of tests:
* those that are tested on both classic addresses and X-addresses (e.g. `getBalances`)
* those that don't use addresses at all (e.g. `combine`)

The changes were reviewed carefully for each test type and then programmatically applied across the rest of the files with a Python script that basically did a bunch of Ctrl-F replaces. 

A few exceptions to the two main types:
* `sign` has some of each type.
* Some of the `prepare` methods (such as `prepareTransaction`) had X-Addresses turned off, so they directly use only a classic address. The tests _could_ be edited to use both, but since we're getting rid of all the `prepare` methods soon, I didn't think it was worth the effort.

Note for reviewers: The large number of lines is mostly due to indenting changes.

### Context of Change

The `TestSuite` framework was super confusing for anyone who doesn't already understand the system well, which means it was hard/intimidating for people new to ripple-lib to come into the ecosystem and write tests.

### Type of Change

## Test Plan

CI passes.


test/client/preparePayment.ts
```diff
@@ -11,17 +11,12 @@ const { preparePayment: REQUEST_FIXTURES } = requests;
...
- * - Check out "test/client/index.ts" for more information about the test runner.
- */
-describe("preparePayment", () => {
+describe("client.preparePayment", function () {
   beforeEach(setupClient.setup);
   afterEach(setupClient.teardown);
 
   addressTests.forEach(function (test) {
-    describe(test.type, () => {
+    describe(test.type, function () {
```
khancode: What's the reason behind this change or ESLint rule?

khancode (approved):
LGTM

test/client/preparePayment.ts
mvadari: I don't have a good enough understanding of JS/TS to really understand why, but for some reason, when I used arrow functions, the compiler complained about instance variables (like `this.client`). It went away when I switched to regular functions. I'm guessing it's something to do with scope.

PR #1560, opened 2021-08-24
Nn/linting

natenichols:
## High Level Overview of Change
Initial Linter setup. 

### Context of Change
This is the base linter config. I have some files refactored, but after merging 2.0 I realized I will likely need to get other folks working on this to get this done. 

### Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

PR #1572, opened 2021-08-27
Lints `src/models`

mvadari:

Title says it all

### Context of Change

Lints the whole folder

### Type of Change

## Test Plan

CI passes.


src/models/transactions/escrowFinish.ts
```diff
@@ -14,10 +14,9 @@ export interface EscrowFinish extends BaseTransaction {
  * Verify the form and type of an EscrowFinish at runtime.
  *
  * @param tx - An EscrowFinish Transaction.
- * @returns Void.
  * @throws When the EscrowFinish is Malformed.
  */
-export function verifyEscrowFinish(tx: EscrowFinish): void {
+export function verifyEscrowFinish(tx: Record<string, unknown>): void {
```
JST5000, line 19: Why are we getting rid of EscrowFinish here? (Curious)

JST5000 (changes requested):
Just one change to the ModifiedOfferCreateTransaction

src/models/transactions/escrowFinish.ts
mvadari: Because this method is only going to be used by JS folks. So we don't have any of the benefit of typing anyways, and we shouldn't be typing it as such.

natenichols (approved):
LGTM 👍

ledhed2222 (commented):
should we be doing this yet? nathan isn't done with linter config so we'll have to come back around to this anyway

mvadari:
The linter config PR has already been merged: #1560. This PR uses that setup.

src/models/methods/ledger.ts
```diff
@@ -16,12 +18,24 @@ export interface LedgerRequest extends BaseRequest {
...
+interface ModifiedOfferCreateTransaction {
+  transaction: Transaction;
+  metadata: ModifiedMetadata;
+}
+
 interface LedgerQueueData {
   account: string;
   // TODO: Retype tx once we have transaction types
   // Also include tx_blob as possible type: https://xrpl.org/ledger.html
   // Also handle the special case where 'owner_funds: string' is a field of OfferCreate sometimes - https://xrpl.org/ledger.html#response-format
```
JST5000, line 34: You can get rid of these comments now

JST5000 (approved):
Other than one extra comment, looks good!

PR #1573, opened 2021-08-27
Signer

JST5000:

Adding the Signer class implementation along with tests

### Context of Change

Part of the xrpl.js refactor. 

### Type of Change

## Before / After

* Added new Signer class
* Added new test cases for each method in it

## Test Plan

Adding unit tests to the CI pipeline


test/signerTest.ts
```diff
@@ -0,0 +1,130 @@
...
+                }              
+            ],
+            SigningPubKey: "",
+            TransactionType: "TrustSet",
+            //hash: "BD636194C48FD7A100DE4C972336534C8E710FD008C0F3CF7BC5BF34DAF3C3E6"
+          }
+
+        const transactions: Transaction[] = new Array(tx1, tx2)
+
+        multisign(transactions)
```
JST5000, line 94: This is where the type issue is coming up: If I type multisign as `Transaction[] | string[]` it gives me an error when trying to pass in `Transaction[]` - Specifically: `Argument of type Transaction[] is not assignable to parameter of type Transaction[] | string[]`

src/ledger/signer.ts
```diff
@@ -0,0 +1,130 @@
+import { BigNumber } from "bignumber.js";
```
mvadari, line 1: This should be at the `src` level, not in `ledger`

natenichols: Do you think it should go in the `wallet` repository?

mvadari: That seems reasonable as well.

src/models/transactions/common.ts
```diff
@@ -79,7 +79,7 @@ export interface BaseTransaction {
   Flags?: number | GlobalFlags;
   LastLedgerSequence?: number;
   Memos?: Array<{ Memo: Memo }>;
-  Signers?: Signer[];
+  Signers?: Array<{ Signer: Signer }>;
```
mvadari, line 82: Instead of this, I would change the signer object to be `Signer: {...whatever-was-there-before}`

src/wallet/signer.ts
```diff
@@ -0,0 +1,221 @@
...
+    id: computeBinaryTransactionHash(signedTransaction),
+  };
+}
+
+function getDecodedTransaction(txOrBlob: Transaction | string): Transaction {
+  if (typeof txOrBlob === "object") {
+    return txOrBlob;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Type assertions let us use well-typed transactions
```
mvadari, line 131: This reason doesn't make sense - the reason should be about why you need to disable the line, not why the rule exists

src/wallet/signer.ts
```diff
@@ -0,0 +1,221 @@
...
+
+/**
+ * Sign uses your wallet to cryptographically sign your transaction which proves
+ * that you are the one issuing this transaction.
+ *
+ * @param wallet - A Wallet that holds your cryptographic keys.
+ * @param tx - The Transaction that is being signed.
+ * @returns A signed Transaction and a Transaction id that corresponds to the pre-signed Transaction.
+ */
+function sign(wallet: Wallet, tx: Transaction): SignedTransaction {
```
mvadari, line 32: Rearrange the methods so that all the private helpers are at the bottom and the exported methods are at the top (so it's easier to follow what's going on).

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+
+  const encodedTransactions: string[] = transactions.map((txOrBlob) =>
+    getEncodedTransaction(txOrBlob)
+  );
+
+  return combine(encodedTransactions);
+}
+
+/**
+ * AuthorizeChannel creates a signature that can be used to redeem a specific amount of XRP from a payment channel.
```
mvadari, line 81: nit:
```suggestion
 * Creates a signature that can be used to redeem a specific amount of XRP from a payment channel.
```

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+ * @param wallet - A Wallet that holds your cryptographic keys.
+ * @param tx - The Transaction that is being signed.
+ * @returns A signed Transaction and a Transaction id that corresponds to the pre-signed Transaction.
+ */
+function sign(wallet: Wallet, tx: Transaction): SignedTransaction {
+  return wallet.signTransaction(tx, { signAs: "" });
+}
+
+/**
+ * Multisign takes several transactions (in object or blob form) and creates a single transaction with all Signers
```
mvadari, line 37: nit:
```suggestion
 * Takes several transactions (in object or blob form) and creates a single transaction with all Signers.
```

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+import { ValidationError } from "../common/errors";
+import { SignedTransaction } from "../common/types/objects";
+import { Signer } from "../models/common";
+import { Transaction } from "../models/transactions";
+import { verifyBaseTransaction } from "../models/transactions/common";
+import { computeBinaryTransactionHash } from "../utils";
+import Wallet from "../Wallet";
+
+/**
+ * Sign uses your wallet to cryptographically sign your transaction which proves
```
mvadari, line 25:
```suggestion
 * Uses a wallet to cryptographically sign a transaction which proves
```

src/wallet/signer.ts
mvadari: ditto for the other TSdocs

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+ * Combine takes a collection of signedTransactions with the same underlying transaction and produces a
+ * Transaction with all Signers. It then signs that Transaction and gives an id based on the combined Transaction.
+ *
+ * @param signedTransactions - A collection of the same transaction signed by different signers. The only difference
+ * between the elements of signedTransactions should be the Signers field.
+ * @returns An object with the combined transaction (now having a sorted list of all signers) which is encoded, along
+ * with a transaction id based on the combined transaction.
+ */
+function combine(signedTransactions: string[]): SignedTransaction {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- decode gives a JsonObject which we know is a Transaction, so we cast it to get strong-typing in this code and our exported interface
```
mvadari, line 126: This comment is too long, did the linter not complain?

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+  const transactions: Transaction[] = signedTransactions.map((tx: string) => {
+    return decode(tx);
+  }) as unknown as Transaction[];
+
+  // This will throw a more clear error for JS users if there's a problem with any of the transaction's formatting
+  transactions.forEach((tx) =>
+    verifyBaseTransaction(tx as unknown as Record<string, unknown>)
```
mvadari, line 133: This should be replaced with `verify` later

test/models/baseTransaction.ts
```diff
@@ -42,9 +42,11 @@ describe("Transaction Verification", function () {
       ],
       Signers: [
         {
-          Account: "r....",
```
mvadari, line 45: Why were the changes in this file necessary?

test/wallet/signerTest.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  // it('authorizeChannel succeeds with ed25519 seed'), () => {
+  //     const wallet = Wallet.fromSeed("sEdnGHNrPbHrdUcszeuDEigMdC1Lyyd")
+  //     const channelId = '5DB01B7FFED6B67E6B0414DED11E051D2EE2B7619CE0EAA6286D67A3A4D5BDB3'
+  //     const amount = '1000000'
+  //
+  //     assert.equal(authorizeChannel(wallet, channelId, amount),
+  // eslint-disable-next-line max-len -- It's cleaner on one line
+  //     '304402207936D71315E0F2AC349A33F988111A6C6D92CD96347AF42BDB5C34EE61D4EC1F0220684CB495EA93DBF0B5A7EC003A1501BC87B2635AA592E5362D240A43DC1BC801')
+  // }
```
mvadari, line 136: Why is this commented out?

test/models/baseTransaction.ts
JST5000: The test case was incorrectly formatted - It didn't include the "Signers: { ... }" wrapper which rippled transactions have to have.

test/wallet/signerTest.ts
JST5000: I didn't know how to get a valid ed25519 secret, added a correct version of this test

src/wallet/signer.ts
JST5000: It didn't comment, odd. I shortened them to just a couple words.

src/models/common/index.ts
```diff
@@ -26,9 +26,11 @@ export interface IssuedCurrencyAmount extends IssuedCurrency {
...
 
 export interface Signer {
-  Account: string;
-  TxnSignature: string;
-  SigningPubKey: string;
+  Signer: {
+    Account: string;
+    TxnSignature: string;
+    SigningPubKey: string;
+  };
```
ledhed2222, line 33: this is weird...why are you nesting it like this?

src/models/transactions/common.ts
```diff
@@ -55,13 +55,18 @@ const SIGNER_SIZE = 3;
...
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Only used by JS and Signer is previously unknown
+    const signer = signerWrapper.Signer as Record<string, unknown>;
+    return (
+      Object.keys(signer).length === SIGNER_SIZE &&
+      typeof signer.Account === "string" &&
+      typeof signer.TxnSignature === "string" &&
+      typeof signer.SigningPubKey === "string"
+    );
+  }
+  return false;
```
ledhed2222, line 69: i think this will be easier to read if you invert the conditional, since all your heavy lifting is within the if. like so:

```typescript
function isSigner(obj: unknown): boolean {
  const signerWrapper = obj as Record<string, unknown>
  if (signerWrapper.Signer === undefined) {
    return false
  }
  const signer = signerWrapper.Signer as Record<string, unknown>
  return (
    Object.keys(signer).length === SIGNER_SIZE &&
    typeof signer.Account === "string" &&
    typeof signer.TxnSignature === "string" &&
    typeof signer.SigningPubKey === "string"
  )
}
```

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+    // This will throw a more clear error for JS users if any of the supplied transactions has incorrect formatting
+    verifyBaseTransaction(tx as unknown as Record<string, unknown>);
+
+    if (tx.SigningPubKey !== "") {
+      throw new ValidationError(
+        "For multisigning the transaction must include the SigningPubKey field as an empty string."
+      );
+    }
+
+    if (tx.Signers === undefined) {
```
ledhed2222, line 66: i greatly prefer `tx.Signers == null` just in case

ledhed2222 (changes requested):
see comments

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+): SignedTransaction {
+  if (transactions.length === 0) {
+    throw new ValidationError("There were 0 transactions given to multisign");
+  }
+
+  transactions.forEach((txOrBlob) => {
+    const tx = getDecodedTransaction(txOrBlob);
+
+    // This will throw a more clear error for JS users if any of the supplied transactions has incorrect formatting
+    verifyBaseTransaction(tx as unknown as Record<string, unknown>);
```
natenichols, line 58: should be replaced w/ `verify()`

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+ */
+function multisign(
+  transactions: Array<Transaction | string>
+): SignedTransaction {
+  if (transactions.length === 0) {
+    throw new ValidationError("There were 0 transactions given to multisign");
+  }
+
+  transactions.forEach((txOrBlob) => {
+    const tx = getDecodedTransaction(txOrBlob);
```
natenichols, line 55: would prefer if `getDecodedTransaction` returned a `Transaction`

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * @param signedTransactions - A collection of the same transaction signed by different signers. The only difference
+ * between the elements of signedTransactions should be the Signers field.
+ * @returns An object with the combined transaction (now having a sorted list of all signers) which is encoded, along
+ * with a transaction id based on the combined transaction.
+ */
+function combine(signedTransactions: string[]): SignedTransaction {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- We are casting here to get strong typing
+  const transactions: Transaction[] = signedTransactions.map((tx: string) => {
+    return decode(tx);
```
natenichols, line 128: call `getDecodedTransaction` here, which should return a `Transaction`

src/wallet/signer.ts
natenichols: You shouldn't have to use type casting here.

src/models/common/index.ts
mvadari: That's how the XRPL does it

JST5000: This mirrors the request format (See the example here: https://xrpl.org/submit_multisigned.html#submit_multisigned) - An alternative is pulling the `Signer:` property out (Ex. `Signers: Array<{ Signers: Signer }>` instead of `Array<Signer>`)

Having it in the SIgner interface makes it clear how to construct a Signer (And makes forming a Transaction's typing for it look cleaner), but it makes it less intuitive to access Signer's fields.

I'm leaning towards this encapsulation (With `Signer:` within the type) because it ensures you don't have something like `const signer: { Signer: Signer } = tx.Signers[0]` which feels awkward.

What are your thoughts? (Am I missing a better way to match the Transaction interface?)

src/wallet/signer.ts
```diff
@@ -0,0 +1,219 @@
...
+    return txOrBlob;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- We are casting here to get strong typing
+  return decode(txOrBlob) as unknown as Transaction;
+}
+
+function getEncodedTransaction(txOrBlob: Transaction | string): string {
+  if (typeof txOrBlob === "object") {
+    return encode(JSON.parse(JSON.stringify(txOrBlob)));
```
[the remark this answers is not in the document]
JST5000: This turns a Transaction into a JsonObject for encode to take. Is there a cleaner way to do that?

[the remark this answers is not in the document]
mvadari: A Transaction is already a JsonObject. A JsonObject is basically just a TS object. If you have to cast, then cast. (we probably need to redo typing in ripple-binary-codec now anyways, given all the typing we just did).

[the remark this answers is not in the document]
natenichols: I would just do a typecast here. 

`JSON.stringify` and `JSON.parse` have runtime complexity, whereas `as Transaction` does not.

src/wallet/signer.ts
JST5000: I'm not sure how to avoid it, I suppose an alternative would be to turn off the Typescript warning somehow? (Or maybe just have the inputs for Verify be Transactions, since those type descriptions only apply to TypeScript anyway?) - @mvadari What do you think would be best? (Or does Verify make this error irrelevant?)

mvadari: There's a lot of annoying linter errors to wade through if we type them as `Transaction`s, and I also don't think that's actually correct. 
I'm surprised that TypeScript gets mad at this issue, because a `Transaction` is technically a `Record<string, unknown`. That's one error that we can't just "turn off", though.

src/wallet/signer.ts
```diff
@@ -0,0 +1,222 @@
...
+    throw new ValidationError("There were 0 transactions given to multisign");
+  }
+
+  transactions.forEach((txOrBlob) => {
+    const tx: Transaction = getDecodedTransaction(txOrBlob);
+
+    // This will throw a more clear error for JS users if any of the supplied transactions has incorrect formatting
+    // TODO: Replace this with verify() (The general validation function for all Transactions)
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- verify does not accept Transaction type
+    verifyBaseTransaction(tx as unknown as Record<string, unknown>);
```
mvadari, line 61: This should directly cast to `Record<string, unknown>` (and that linting error can be disabled)

src/wallet/signer.ts
```diff
@@ -0,0 +1,222 @@
...
+      throw new ValidationError(
+        "For multisigning the transaction must include the SigningPubKey field as an empty string."
+      );
+    }
+
+    if (tx.Signers == null) {
+      throw new ValidationError(
+        "For multisigning the transaction must include a Signers field containing an array of signatures."
+      );
+    }
```
mvadari, line 73: Don't the transactions need to be signed individually for them to be combined?

src/wallet/signer.ts
```diff
@@ -0,0 +1,222 @@
...
+/**
+ * Takes a collection of signedTransactions with the same underlying transaction and produces a
+ * Transaction with all Signers. It then signs that Transaction and gives an id based on the combined Transaction.
+ *
+ * @param signedTransactions - A collection of the same transaction signed by different signers. The only difference
+ * between the elements of signedTransactions should be the Signers field.
+ * @returns An object with the combined transaction (now having a sorted list of all signers) which is encoded, along
+ * with a transaction id based on the combined transaction.
+ */
+function combine(signedTransactions: string[]): SignedTransaction {
```
mvadari, line 128: I would add the logic here to the `multisign` method. If you want to keep it as a separate function, then name it something a little more descriptive (maybe something like `combineForMultisign`).

src/wallet/signer.ts
```diff
@@ -0,0 +1,222 @@
...
+  const transactions: Transaction[] = signedTransactions.map((tx: string) => {
+    return getDecodedTransaction(tx);
+  });
+
+  // This will throw a more clear error for JS users if there's a problem with any of the transaction's formatting
+  transactions.forEach((tx) =>
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Casting here let's us do type validation for JS
+    verifyBaseTransaction(tx as unknown as Record<string, unknown>)
```
mvadari, line 136: Why do we verify all of the transactions in 2 separate places? We should only need to do it once. (see comment above about combining the two methods into one)

src/wallet/signer.ts
```diff
@@ -0,0 +1,222 @@
...
+ * @param transactions - An array of Transactions (in object or blob form) to combine and sign.
+ * @returns A single signed Transaction which has all Signers from transactions within it.
+ * @throws ValidationError if:
+ * - There were no transactions given to sign
+ * - The SigningPubKey field is not the empty string in any given transaction
+ * - Any transaction is missing a Signers field.
+ */
+function multisign(
+  transactions: Array<Transaction | string>
+): SignedTransaction {
```
mvadari, line 50: This should also just return a `Transaction`

test/wallet/signerTest.ts
```diff
@@ -0,0 +1,173 @@
...
+  Account: address,
+  Destination: "rQ3PTWGLCbPz8ZCicV5tCX3xuymojTng5r",
+  Amount: "20000000",
+  Sequence: 1,
+  Fee: "12",
+  SigningPubKey: publicKey,
+};
+
+describe("Signer tests", function () {
+  it("sign transaction offline", function () {
```
mvadari, line 32:
```suggestion
  it("sign transaction", function () {
```

test/wallet/signerTest.ts
```diff
@@ -0,0 +1,173 @@
...
+    const transactions: Transaction[] = [tx1, tx2];
+
+    multisign(transactions);
+  });
+
+  it("throws a validation error if multisigning with no transactions", function () {
+    const transactions: Transaction[] = [];
+    assert.throws(() => multisign(transactions), ValidationError);
+  });
+
```
mvadari, line 115: Add additional tests for multisign - 
* When the transactions aren't identical
* With tx blobs instead of Transaction objects

test/wallet/signerTest.ts
```diff
@@ -0,0 +1,173 @@
...
+          },
+        },
+      ],
+      SigningPubKey: "",
+      TransactionType: "TrustSet",
+    };
+
+    const transactions: Transaction[] = [tx1, tx2];
+
+    multisign(transactions);
```
mvadari, line 108: This test should also check the output and make sure it looks right

mvadari (changes requested):

src/models/common/index.ts
intelliot: I think `Memo`s are/should also be nested like this (?)

Whatever we decide, I suppose the ideal is for `Signer` and `Memo` to sort of be consistent with each other

mvadari: Yeah, memos are also like this. If we add a testcase with memos then it should fail

ledhed2222 (approved):

src/wallet/signer.ts
JST5000: Converting to unknown is how to tell Typescript that you meant to do a conversion which it doesn't see any overlap with. (It unfortunately won't compile if I directly convert, or if I try to implicitly let it convert by not casting at all)

src/wallet/signer.ts
JST5000: This actually sparked a lot of confusion for me and after talking to Nathan just now we realized we were missing a piece of the multisign workflow:

- Combining multisigned transactions requires them to be signed individually in a special format. In Rippled this is done by the "sign_for" request. Specifically, sign_for creates the "Signers" field instead of changing "TxnSignature" or "SigningPubKey" fields. (We didn't want to implement "sign_for" because it signs over the internet, but we do still need a way to sign using "Signers" instead of "TxnSignature" on the transaction)
- So, we need an offline 'multisign' function takes a transaction and signs it in the multsign format
- And we need a combineMultisigned function (which is what we were previously calling 'multisign' in the Signers ticket or 'combine' in 1.x)

JST5000: The current names I'm using are 'multisign' and 'combineMultisigned' for these two functions, @mvadari what are your thoughts on those names?

mvadari: Those names feel confusing to me. But could we instead have one multisign function that handles both parts? The 3 parts of a `Signer` object are pretty easy to derive in the same method as where we're combining stuff.

JST5000: Potentially, the issue would be that only one of the many signers would be able to use the combined version. The default workflow would be A multisigns, B multisigns, C multisigns, then those three are combined. So only by the time C signs do we have all the information needed for combine. 

If all signatures needed for multisigning were owned by the same node, we could simplify it to sign all in the combine method, but I think there are important use cases where you need signatures from independent networks (which sign offline on their own networks and share the end result). 

Maybe signForMultisign and combineMultisigned would be better names?

mvadari: I think you could still take in the existing structure - the `Signer` object needs `Account`, `SignerPublicKey`, and `TxnSignature`, which are all in a signed transaction anyways. 
Hm, I guess you still need to sign it differently with ripple-binary-codec, so you'd need to have that helper method anyways.
I don't like the `for` part of `signForMultisigned` - maybe just `signMultisign`?

src/models/common/index.ts
mvadari: If you don't want to fix this in this PR, add a `TODO` somewhere

src/wallet/signer.ts
```diff
@@ -0,0 +1,196 @@
...
+ * that then gets signed and returned.
+ *
+ * @param transactions - An array of Transactions (in object or blob form) to combine and sign.
+ * @returns A single signed Transaction which has all Signers from transactions within it.
+ * @throws ValidationError if:
+ * - There were no transactions given to sign
+ * - The SigningPubKey field is not the empty string in any given transaction
+ * - Any transaction is missing a Signers field.
+ */
+function combineMultisigned(
```
mvadari, line 52: Do you think `combineMultisigned` is a better name than just `multisign`?

test/wallet/index.ts
```diff
@@ -117,12 +117,9 @@ describe("Wallet", function () {
...
       };
       const wallet = new Wallet(publicKey, privateKey);
-      const signedTx: { signedTransaction: string; id: string } =
-        wallet.signTransaction(txJSON);
+      const signedTx: string = wallet.signTransaction(txJSON);
 
-      assert.hasAllKeys(signedTx, ["id", "signedTransaction"]);
-      assert.isString(signedTx.id);
-      assert.isString(signedTx.signedTransaction);
+      assert.isString(signedTx);
```
mvadari, line 122: Signing is deterministic. So we might want to add an `expected` that was calculated directly via ripple-binary-codec.

test/wallet/signerTest.ts
```diff
@@ -0,0 +1,276 @@
+import { assert } from "chai";
```
mvadari, line 1: This file should just be `test/wallet/signer.ts` - it's already in the test folder, so you know it's a test file

test/wallet/signerTest.ts
mvadari: This wasn't resolved

src/wallet/signer.ts
JST5000: Now that we've moved the signing prep into `sign` I think `multisign` is a better name than `combineMultisigned` (Because it's no longer ambiguous which part of the process it's referring to)

src/wallet/signer.ts
```diff
@@ -0,0 +1,192 @@
...
+  if (transactions.length === 0) {
+    throw new ValidationError("There were 0 transactions to multisign");
+  }
+
+  transactions.forEach((txOrBlob) => {
+    const tx: Transaction = getDecodedTransaction(txOrBlob);
+
+    // This will throw a more clear error for JS users if any of the supplied transactions has incorrect formatting
+    // TODO: Replace this with verify() (The general validation function for all Transactions)
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- verify does not accept Transaction type
```
mvadari, line 62: Another way to fix this would be to type the verify method `Record<string, unknown> | Transaction`

src/wallet/signer.ts
```diff
@@ -0,0 +1,192 @@
...
+    verifyBaseTransaction(tx as unknown as Record<string, unknown>);
+
+    if (tx.SigningPubKey !== "") {
+      throw new ValidationError(
+        "SigningPubKey must be an empty string for all transactions when multisigning.s"
```
mvadari, line 67:
```suggestion
        "SigningPubKey must be an empty string for all transactions when multisigning."
```

src/wallet/signer.ts
```diff
@@ -0,0 +1,192 @@
...
+  return signWithKeypair(signingData, wallet.privateKey);
+}
+
+/**
+ * Verifies that the given transaction has a valid signature based on public-key encryption.
+ *
+ * @param tx - A transaction to verify the signature of. (Can be in object or encoded string format).
+ * @returns Returns true if tx has a valid signature, and returns false otherwise.
+ */
+function verify(tx: Transaction | string): boolean {
```
mvadari, line 116: This should be `verifySignature` or `verifySignedTransaction` or something slightly more descriptive.

test/wallet/signer.ts
```diff
@@ -0,0 +1,300 @@
...
+        TxnSignature:
+          "304502210098009CEFA61EE9843BB7FC29B78CFFAACF28352A4A7CF3AAE79EF12D79BA50910220684F116266E5E4519A7A33F7421631EB8494082BE51A8B03FECCB3E59F77154A",
+      },
+    },
+  ],
+  SigningPubKey: "",
+  TransactionType: "TrustSet",
+};
+
+const expectedmultisign: Transaction = {
```
mvadari, line 123: nit:
```suggestion
const expectedMultisign: Transaction = {
```

test/wallet/signer.ts
```diff
@@ -0,0 +1,300 @@
...
+        TxnSignature:
+          "304502210098009CEFA61EE9843BB7FC29B78CFFAACF28352A4A7CF3AAE79EF12D79BA50910220684F116266E5E4519A7A33F7421631EB8494082BE51A8B03FECCB3E59F77154A",
+      },
+    },
+  ],
+  SigningPubKey: "",
+  TransactionType: "TrustSet",
+};
+
+describe("Signer tests", function () {
```
mvadari, line 157:
```suggestion
describe("Signer", function () {
```
You're running tests, you already know that these are tests :P

test/wallet/signer.ts
```diff
@@ -0,0 +1,300 @@
...
+  it("sign transaction", function () {
```
mvadari, line 158: I would label all these tests with the function name (e.g. this one could be `"sign"`, the next one is `"sign with multisign"`, etc.

src/wallet/signer.ts
JST5000: I think `verifySignature` makes the most sense, because that's the only check it does and I don't want to confuse it with the more general verify(Tx)

src/wallet/signer.ts
JST5000: That makes sense to me, I added that to the TODO here

test/wallet/index.ts
JST5000: Do you have a specific function in mind?

src/wallet/signer.ts
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable import/max-dependencies -- These imports are required for Signer */
+import { BigNumber } from "bignumber.js";
```
mvadari, line 2: Needs to be properly prettier-ed

test/wallet/index.ts
mvadari: Basically check your work by doing what happens under the hood (by hand), just like you did with the multisign tests.

test/wallet/signer.ts
```diff
@@ -0,0 +1,300 @@
...
+          Signer: {
+            SigningPubKey:
+              "02A8A44DB3D4C73EEEE11DFE54D2029103B776AA8A8D293A91D645977C9DF5F544",
+            TxnSignature:
+              "3044022077BCE143B9A0B51A7716BB93CBC0C99FB41BA339D91A87CB9E47DA80A7EF660802205C81AA49D408771F65A131200CCBFC536ACFE212C1414E05E43B56BE1F9380F2",
+            Account: "rHLEki8gPUMnF72JnuALvnAMRhRemzhRke",
+          },
+        },
+      ],
+    };
```
mvadari, line 235: Given that we just discussed this, add a test to make sure that transactions signed in non-multisign mode are errored correctly (with the correct error message)

test/wallet/signer.ts
```diff
@@ -0,0 +1,300 @@
...
+
+  it("multisign runs successfully with tx_blobs", function () {
+    const transactions: Transaction[] = [
+      multisignTxToCombine1,
+      multisignTxToCombine2,
+    ];
+
+    const encodedTransactions: string[] = transactions.map(encode);
+
+    assert.deepEqual(multisign(encodedTransactions), expectedMultisign);
```
mvadari, line 207: Extra type definitions unneeded here, TS is smart enough to figure that out

src/wallet/signer.ts
JST5000: Added a TODO

src/models/common/index.ts
JST5000: Added a TODO for Memos

src/common/types/objects/transactions.ts
```diff
@@ -1,3 +1,5 @@
+import { Signer } from '../../../models/common'
+
```
mvadari, line 2: The types here shouldn't be used at all (we're switching to the ones in `src/model`), so you don't need to worry about the right type being here.

test/wallet/index.ts
```diff
@@ -165,12 +165,9 @@ describe('Wallet', function () {
...
       }
       const wallet = new Wallet(publicKey, privateKey)
-      const signedTx: { signedTransaction: string; id: string } =
-        wallet.signTransaction(txJSON)
+      const signedTx: string = wallet.signTransaction(txJSON)
 
-      assert.hasAllKeys(signedTx, ['id', 'signedTransaction'])
-      assert.isString(signedTx.id)
-      assert.isString(signedTx.signedTransaction)
+      assert.isString(signedTx)
```
mvadari, line 170: Add a TODO at the least to make this a more rigorous check that actually checks the output of the signed transaction instead of just the type

test/fixtures/rippled/accountTx.js
```diff
@@ -1,30 +1,34 @@
 /* eslint-disable max-len */
-'use strict';
```
mvadari, line 2: Looks like a bunch of fixtures got caught up in this too - can you undo the changes for the JS files?

test/wallet/signer.ts
```diff
@@ -0,0 +1,303 @@
...
+      Destination: 'rPT1Sjq2YGrBMTttX4GZHjKu9dyfzbpAYe',
+      Flags: 2147483648,
+      LastLedgerSequence: 20582339,
+      Fee: '12',
+      Sequence: 20582260,
+    }
+    const signedTxBlob =
+      '120000228000000024013A0F74201B013A0FC36140000000014FB18068400000000000000C732102A8A44DB3D4C73EEEE11DFE54D2029103B776AA8A8D293A91D645977C9DF5F544744730450221009ECB5324717E14DD6970126271F05BC2626D2A8FA9F3797555D417F8257C1E6002206BDD74A0F30425F2BA9DB69C90F21B3E27735C190FB4F3A640F066ACBBF06AD98114B3263BD0A9BF9DFDBBBBD07F536355FF477BF0E98314F667B0CA50CC7709A220B0561B85E53A48461FA8'
+
+    const wallet = Wallet.fromSeed(seed)
```
mvadari, line 172: You can have this be a global variable, instead of running this in every test - the wallet setup won't affect anything.

test/wallet/signer.ts
```diff
@@ -0,0 +1,303 @@
...
+    ]
+
+    const encodedTransactions: string[] = transactions.map(encode)
+
+    assert.deepEqual(multisign(encodedTransactions), expectedMultisign)
+  })
+
+  it('multisign throws a validation error when there are no transactions', function () {
+    const transactions: Transaction[] = []
+    assert.throws(() => multisign(transactions), ValidationError)
```
mvadari, line 210: You don't need the types here - TS is smart enough to figure it out on its own

test/wallet/signer.ts
```diff
@@ -0,0 +1,303 @@
...
+    assert.equal(
+      authorizeChannel(wallet, channelId, amount),
+      '7E1C217A3E4B3C107B7A356E665088B4FBA6464C48C58267BEF64975E3375EA338AE22E6714E3F5E734AE33E6B97AAD59058E1E196C1F92346FC1498D0674404',
+    )
+  })
+
+  it('verifySignature succeeds for valid signed transaction blob', function () {
+    const wallet = new Wallet(publicKey, privateKey)
+
+    const signedTx: string = sign(wallet, tx)
```
mvadari, line 278: ditto

test/wallet/signer.ts
```diff
@@ -0,0 +1,303 @@
...
+
+    assert.isTrue(verifySignature(decode(signedTx) as unknown as Transaction))
+  })
+
+  it('verify throws for invalid signing key', function () {
+    const wallet = new Wallet(publicKey, privateKey)
+    const signedTx: string = sign(wallet, tx)
+
+    const decodedTx: Transaction = decode(signedTx) as unknown as Transaction
+
```
mvadari, line 296: ditto

test/wallet/signer.ts
```diff
@@ -0,0 +1,303 @@
...
+              '02A8A44DB3D4C73EEEE11DFE54D2029103B776AA8A8D293A91D645977C9DF5F544',
+            TxnSignature:
+              '3044022077BCE143B9A0B51A7716BB93CBC0C99FB41BA339D91A87CB9E47DA80A7EF660802205C81AA49D408771F65A131200CCBFC536ACFE212C1414E05E43B56BE1F9380F2',
+            Account: 'rHLEki8gPUMnF72JnuALvnAMRhRemzhRke',
+          },
+        },
+      ],
+    }
+
+    const transactions: Transaction[] = [
```
mvadari, line 235: ditto

mvadari (approved):
A few comments, but nothing to hold up approval - please address them before merging. 

Nice job!

src/wallet/signer.ts
```diff
@@ -0,0 +1,185 @@
...
+ * Takes several transactions (in object or blob form) and creates a single transaction with all Signers
+ * that then gets signed and returned.
+ *
+ * @param transactions - An array of Transactions (in object or blob form) to combine and sign.
+ * @returns A single signed Transaction which has all Signers from transactions within it.
+ * @throws ValidationError if:
+ * - There were no transactions given to sign
+ * - The SigningPubKey field is not the empty string in any given transaction
+ * - Any transaction is missing a Signers field.
+ */
```
ledhed2222, line 44: i think we could make the docstrings more clear to show that `transactions` are all already signed

ledhed2222:
Big picture - we should think about code organization here more. We now have signing functionality scattered among a few files: `wallet/Wallet`, `wallet/signer`, `transaction/sign`. I think they should all be in one place and, as necessary, re-exported elsewhere

natenichols (approved):

mvadari:
> Big picture - we should think about code organization here more. We now have signing functionality scattered among a few files: `wallet/Wallet`, `wallet/signer`, `transaction/sign`. I think they should all be in one place and, as necessary, re-exported elsewhere

`transaction/sign` will be deleted soon, so everything will just be in the `wallet` folder.

PR #1574, opened 2021-08-28
implement Autofill Transaction

khancode:

Implements `autofill()` and `setTransactionFlagsToNumber()` to allow a `Client` to autofill fields in a `Transaction`.

### Context of Change
`autofill()` does the following operations:
- Converts X-address to classic address for each account field.
- Calls `setTransactionFlagsToNumber()` that sets a transaction's flags (object of boolean flags) to its numeric representation.
- Autofills missing `Transaction` fields: `Fee`, `Sequence`, `LastLedgerSequence`.

### Type of Change

## Test Plan

Adds tests for `autofill()` and `setTransactionFlagsToNumber()`


src/client/index.ts
```diff
@@ -338,6 +339,16 @@ class Client extends EventEmitter {
     return transactionUtils.convertStringToHex(string);
   }
 
+  /**
+   * Autofills missing fields in a transaction
+   *
+   * @param {Transaction} tx A transaction to autofill missing fields
+   * @returns {Transaction} An autofilled transaction
```
mvadari, line 346: nit: we're using a different doc format that the linter is enforcing

src/models/transactions/common.ts
```diff
@@ -66,6 +66,10 @@ export function isAmount(amount: Amount): boolean {
   return typeof amount === "string" || isIssuedCurrency(amount);
 }
 
+export enum GlobalFlagsEnum {
+  tfFullyCanonicalSig = 0x80000000,
+}
+
```
mvadari, line 72: This flag isn't actually used anymore: https://xrpl.org/transaction-common-fields.html#global-flags

src/models/utils/index.ts
```diff
@@ -19,3 +22,129 @@ export function onlyHasFields(obj: object, fields: string[]): boolean {
...
+      tx.Flags = convertPaymentChannelClaimFlagsToNumber(tx.Flags as PaymentChannelClaimFlags);
+      break;
+    case 'Payment':
+      tx.Flags = convertPaymentTransactionFlagsToNumber(tx.Flags as PaymentTransactionFlags);
+      break;
+    case 'TrustSet':
+      tx.Flags = convertTrustSetFlagsToNumber(tx.Flags as TrustSetFlags);
+      break;
+    default:
+      tx.Flags = convertGlobalFlagsToNumber(tx.Flags);
```
mvadari, line 47: `convertGlobalFlagsToNumber` should be applied on all transactions, regardless of their type

src/models/utils/index.ts
```diff
@@ -1,3 +1,6 @@
+import { OfferCreateFlags, OfferCreateFlagsEnum, PaymentChannelClaimFlags, PaymentChannelClaimFlagsEnum, PaymentTransactionFlags, PaymentTransactionFlagsEnum, Transaction, TrustSetFlags, TrustSetFlagsEnum } from "../transactions";
```
mvadari, line 1: Run `eslint --fix` on the new code you're adding

src/transaction/utils.ts
```diff
@@ -412,7 +417,136 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
+ * @param {Transaction} tx A transaction to autofill fields.
+ * @param {Client} client A client.
+ * @returns {Transaction} An autofilled transaction.
+ */
+async function autofillTransaction(
+  tx: Transaction,
+  client: Client
+): Promise<Transaction> {
+  validateAccountAddress(tx, 'Account', 'SourceTag');
+  if (tx['Destination']) {
```
mvadari, line 432:
```suggestion
  if (tx['Destination'] == null) {
```

src/transaction/utils.ts
```diff
@@ -412,7 +417,136 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
+  convertToClassicAddress(tx, 'Authorize');
+  convertToClassicAddress(tx, 'Unauthorize');
+  // EscrowCancel, EscrowFinish:
+  convertToClassicAddress(tx, 'Owner');
+  // SetRegularKey:
+  convertToClassicAddress(tx, 'RegularKey');
+
+  setTransactionFlagsToNumber(tx);
+
+  if (tx.Sequence === undefined) {
```
mvadari, line 446:
```suggestion
  if (tx.Sequence == null) {
```
We want to avoid using `undefined` to be slightly more general

src/transaction/utils.ts
```diff
@@ -412,7 +417,136 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
+    const sequence = await getNextValidSequenceNumber(tx.Account, client);
+    tx.Sequence = sequence;
+  }
+
+  if (tx.Fee === undefined) {
```
mvadari, line 451: ditto

src/transaction/utils.ts
```diff
@@ -412,7 +417,136 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
+    tx.Fee = await calculateFeePerTransactionType(tx, client);
+  }
+
+  if (tx.LastLedgerSequence === undefined) {
```
mvadari, line 455: ditto

src/transaction/utils.ts
```diff
@@ -412,7 +417,136 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
+    account,
+  };
+  const data = await client.request(request);
+  return data.result.account_data.Sequence;
+}
+
+async function calculateFeePerTransactionType(
+  transaction: Transaction,
+  client?: Client
+): Promise<string> {
```
mvadari, line 501: Something that's different about xrpl-py vs xrpl.js (which needs to be changed in Python) is that you'll definitely have a client for this. There should not be a default fee.

src/transaction/utils.ts
```diff
@@ -412,7 +417,136 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
+  let netFee;
+  if (client === undefined) {
+    netFee = 10; // 10 drops
+  } else {
+    netFee = await client.getFee(); // Usually 0.00001 XRP (10 drops)
+  }
+
+  let baseFee = netFee;
```
mvadari, line 509: You might need to do some more processing on this - https://github.com/ripple/ripple-lib/blob/develop/src/transaction/utils.ts#L278

mvadari (changes requested):

src/transaction/utils.ts
khancode: This should be `!= null` instead since it's checking if Destination is defined.

src/models/transactions/common.ts
khancode: Since this is deprecated and the only flag in `GlobalFlags`, we should remove `GlobalFlags` since it'll become an empty interface. @natenichols defined the `GlobalFlags` interface in another [commit](https://github.com/ripple/ripple-lib/commit/3231c9e08c6ff469876d5ff99093e32a72f12852). How do you feel about my suggestion?

khancode: Since this is deprecated and the only flag in GlobalFlags, we should remove GlobalFlags since it'll become an empty interface. @natenichols defined the GlobalFlags interface in another commit. What do you think about my suggestion?

src/models/utils/index.ts
khancode: It technically is in my PR but looks like it's not apparent reading it. I call `convertGlobalFlagsToNumber` in each `convert{TransactionType}FlagsToNumber` so that these functions return the full numeric value including GlobalFlags.

If we still want to keep `GlobalFlags` interface, I can move `convertGlobalFlagsToNumber` to the top of `setTransactionFlagsToNumber` and specify `convert{TransactionType}FlagsToNumber` exclude `GlobalFlags`.

khancode: It technically is in my PR but looks like it's not apparent reading it. I call `convertGlobalFlagsToNumber` in each `convert{TransactionType}FlagsToNumber` so that these functions return the full numeric value including `GlobalFlags`.

If we still want to keep `GlobalFlags` interface, I can move `convertGlobalFlagsToNumber` to the top of `setTransactionFlagsToNumber` and specify `convert{TransactionType}FlagsToNumber` exclude `GlobalFlags`.

khancode: If we still want to keep `GlobalFlags` interface, I can move `convertGlobalFlagsToNumber` to the top of `setTransactionFlagsToNumber` and specify `convert{TransactionType}FlagsToNumber` exclude `GlobalFlags`.

src/client/index.ts
natenichols: This should be resolvable with a `yarn lint --fix`

src/models/transactions/common.ts
natenichols: We need to have a `GlobalFlags`, because otherwise we'd have to specify flags in the baseTransaction as `Flags: number | Record<string,unkown>` We can do that, but i think its a lot cleaner to have `GlobalFlags` serve as a base class.

src/transaction/utils.ts
```diff
@@ -412,7 +417,136 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
 
+/**
+ * Autofills fields in a transaction.
+ *
+ * @param {Transaction} tx A transaction to autofill fields.
+ * @param {Client} client A client.
+ * @returns {Transaction} An autofilled transaction.
+ */
+async function autofillTransaction(
+  tx: Transaction,
```
natenichols, line 428: We'll want this on the client, so I think we'll want to define this as 

```ts
async function autofillTransaction(
    this: Client,
    tx: Transaction
```
and then 
```ts
class Client {
    autofillTransaction = autofillTransaction
}

**Review by natenichols — APPROVED** [review]
LGTM 👍

File: src/client/index.ts [hunk 12]
```diff
@@ -338,6 +339,16 @@ class Client extends EventEmitter {
...
+   */
+  async autofillTransaction(tx: Transaction): Promise<Transaction> {
+    return transactionUtils.autofillTransaction(tx, this)
+  }
```
ledhed2222, line 350: why not call this `autofill`? the signature will make it clear that we expect a transaction, and there aren't other things to autofill. Also, shouldn't you `await` the function from `transactionUtils`? Otherwise you could instead write this as:

```typescript
import { partial } from 'lodash'

autofillTransaction = partial(transactionUtils.autofillTransaction, _, this)
```

src/models/utils/index.ts
ledhed2222: we'll get to this later - since the linter rules aren't there if i were you omar I would wait actually

src/models/utils/index.ts
```diff
@@ -19,3 +22,129 @@ export function onlyHasFields(obj: object, fields: string[]): boolean {
...
+ * Sets a transaction's flags to its numeric representation
+ *
+ * @param {Transaction} tx A transaction to set its flags to its numeric representation
+ * @returns {void}
+ */
+export function setTransactionFlagsToNumber(tx: Transaction): void {
+  switch (tx.TransactionType) {
+    case 'OfferCreate':
+      tx.Flags = convertOfferCreateFlagsToNumber(tx.Flags as OfferCreateFlags);
+      break;
```
ledhed2222, line 36: maybe `return` instead of `break`. no real reason other than that the matched expression should be the last expression for the function anyway, so i think that that makes it more clear to the reader, who knows there is nothing below the `switch` statement

src/models/utils/index.ts
```diff
@@ -19,3 +22,129 @@ export function onlyHasFields(obj: object, fields: string[]): boolean {
...
+  if (typeof flags === 'number') {
+    return flags;
+  }
+  return flags?.tfFullyCanonicalSig ? GlobalFlagsEnum.tfFullyCanonicalSig : 0;
+}
+
+function convertOfferCreateFlagsToNumber(flags: OfferCreateFlags): number {
+  if (flags == null) {
+    return 0;
+  } else if (typeof flags === 'number') {
```
ledhed2222, line 61: this is a personal pet peeve of mine for some reason, but the `else` is unnecessary here since you returned in the first conditional

```typescript
if (flags == null) {
  return 0
}
if (typeof flags === 'number') {
  return flags
}
....
```

src/models/utils/index.ts
```diff
@@ -19,3 +22,129 @@ export function onlyHasFields(obj: object, fields: string[]): boolean {
...
+  }
+  if (flags.tfImmediateOrCancel) {
+    resultFlags |= OfferCreateFlagsEnum.tfImmediateOrCancel;
+  }
+  if (flags.tfFillOrKill) {
+    resultFlags |= OfferCreateFlagsEnum.tfFillOrKill;
+  }
+  if (flags.tfSell) {
+    resultFlags |= OfferCreateFlagsEnum.tfSell;
+  }
```
ledhed2222, line 78: seems like there is a way to express this as a `reduce` instead of a big series of conditions. not a blocker but i think that would be more elegant

src/models/utils/index.ts
```diff
@@ -19,3 +22,129 @@ export function onlyHasFields(obj: object, fields: string[]): boolean {
...
+
+  return resultFlags;
+}
+
+function convertPaymentChannelClaimFlagsToNumber(flags: PaymentChannelClaimFlags): number {
+  if (flags == null) {
+    return 0;
+  } else if (typeof flags === 'number') {
```
ledhed2222, line 86: as above

src/models/utils/index.ts
```diff
@@ -19,3 +22,129 @@ export function onlyHasFields(obj: object, fields: string[]): boolean {
...
+
+  return resultFlags;
+}
+
+function convertPaymentTransactionFlagsToNumber(flags: PaymentTransactionFlags): number {
+  if (flags == null) {
+    return 0;
+  } else if (typeof flags === 'number') {
+    return flags;
+  }
```
ledhed2222, line 107: this is now repeated a few times. i think you should factor this out into its own function

src/transaction/utils.ts
ledhed2222: also agree that the `client` should be the first argument, as that makes usage with partial application (see above) or currying easier

src/transaction/utils.ts
ledhed2222: i would also again make the client the first arg

src/transaction/utils.ts
```diff
@@ -412,7 +417,136 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
+  const account = tx[fieldName];
+  if (typeof account === "string") {
+    const { classicAccount } = getClassicAccountAndTag(account);
+    tx[fieldName] = classicAccount;
+  }
+}
+
+async function getNextValidSequenceNumber(
+  account: string,
+  client: Client
```
ledhed2222, line 488: client shoudl be the first arg

src/transaction/utils.ts
```diff
@@ -412,7 +417,136 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
+
+  // EscrowFinish Transaction with Fulfillment
+  if (transaction.TransactionType === 'EscrowFinish') {
+    if (transaction.Fulfillment != null) {
+      const fulfillmentBytesSize = Math.ceil(
+        transaction.Fulfillment.length / 2
+      );
+      // 10 drops × (33 + (Fulfillment size in bytes / 16))
+      baseFee = Math.ceil(netFee * (33 + fulfillmentBytesSize / 16));
+    }
```
ledhed2222, line 519: this conditional nesting is unnecessary since there is nothing that happens if the outer condition is true but the inner condition is false. as a result you should unnest this. here's one suggestion

```typescript
const isFulfillment = transaction.TransactionType === 'EscrowFinish' && transaction.Fulfillment != null
if (isFullfillment) {
  const fulfillmentBytesSize = Math.ceil(.....)
  ....
}
```

src/transaction/utils.ts
ledhed2222: i also think this should just be called `autofill` since it's already in the `transaction/utils.ts` file - that makes it clear what is being autofilled

ledhed2222 (changes requested):
nice work - see comments

src/client/index.ts
khancode: I agree with your suggestions. However, I'm not a fan of importing a module when this can be done natively with JavaScript's `bind()`:
```typescript
autofill = transactionUtils.autofill.bind(null, this);
```
What do you think?

src/models/utils/index.ts
khancode: Sure, makes sense.

src/models/utils/index.ts
khancode: Haha, I personally like using `else if` so it's easy to tell that only one of the possible conditions will execute.

src/client/index.ts
ledhed2222: i'm no javascript expert, but in my opinion the weirdness of `this` in javascript is such that it should be avoided. but if you'd like to chat about this more i'm down! maybe this is the right move.

src/transaction/utils.ts
khancode: Yup you're right, this totally went over my head copying our Python implementation. 😄

src/models/transactions/common.ts
mvadari: I'm fine with leaving it as an empty interface - it makes it easier to add other global flags if more get added in the future.

src/models/utils/index.ts
mvadari: The linter rules are already in `2.0` courtesy of #1560

src/models/utils/index.ts
```diff
@@ -23,3 +26,129 @@ export function isFlagEnabled(Flags: number, checkFlag: number): boolean {
   // eslint-disable-next-line no-bitwise -- Flags require bitwise operations
   return (checkFlag & Flags) === checkFlag;
 }
+
+/**
+ * Sets a transaction's flags to its numeric representation
+ *
+ * @param {Transaction} tx A transaction to set its flags to its numeric representation
+ * @returns {void}
+ */
```
mvadari, line 35: This isn't the right TSdocs format

src/models/utils/index.ts
mvadari: I don't quite understand what you mean

khancode: I just refactored it, it should be clear now. 🙂

src/models/utils/index.ts
khancode: I'll be fixing the eslint errors in another commit.

src/transaction/utils.ts
```diff
@@ -412,7 +417,130 @@ function convertMemo(memo: Memo): { Memo: ApiMemo } {
...
 }
 
+/**
+ * Autofills fields in a transaction.
+ *
+ * @param client - A client.
+ * @param tx - A transaction to autofill fields.
+ * @returns An autofilled transaction.
+ */
+async function autofill(client: Client, tx: Transaction): Promise<Transaction> {
```
mvadari, line 427: This file should go in the `ledger` folder - something like `src/ledger/autofill.ts`
(yes, this doesn't really have to do anything with the ledger, but that's the folder all the sugar is going in, and it'll be renamed soon)

src/client/index.ts
khancode: @ledhed2222 and I further discussed this in a meeting and we agreed to create a sugar function for JavaScript `bind()`. These changes have been included.

src/ledger/autofill.ts
```diff
@@ -0,0 +1,209 @@
...
+    tag: expectedTag,
+  };
+}
+
+/**
+ * Autofills fields in a transaction.
+ *
+ * @param client - A client.
+ * @param tx - A transaction to autofill fields.
+ * @param signersCount - A signers count used for multisign.
```
mvadari, line 68:
```suggestion
 * @param signersCount - The expected number of signers for this transaction. Used for multisign.
```

src/ledger/autofill.ts
```diff
@@ -0,0 +1,209 @@
...
+ * @returns An autofilled transaction.
+ */
+async function autofill(
```
mvadari, line 71: The helper functions should all be above or below `autofill` - it's confusing when some are above and some are below

src/ledger/autofill.ts
```diff
@@ -0,0 +1,209 @@
...
+    promises.push(setNextValidSequenceNumber(client, tx));
+  }
+  if (tx.Fee == null) {
+    promises.push(calculateFeePerTransactionType(client, tx, signersCount));
+  }
+  if (tx.LastLedgerSequence == null) {
+    promises.push(setLatestValidatedLedgerSequence(client, tx));
+  }
+
+  return Promise.all(promises).then(() => tx);
```
mvadari, line 91: This is really nice

src/ledger/autofill.ts
```diff
@@ -0,0 +1,209 @@
...
+ *
+ * @param client - A client.
+ * @param tx - A transaction to autofill fields.
+ * @param signersCount - A signers count used for multisign.
+ * @returns An autofilled transaction.
+ */
+async function autofill(
+  client: Client,
+  tx: Transaction,
+  signersCount?: number
```
mvadari, line 74: @natenichols thoughts on an optional param for a `LastLedgerSequence` offset?

natenichols: I would say leave it out IMO. You can accomplish the same thing w/ `LastLedgerSequence`. 

The big advantage to adding be that we could have a default`LastLedgerSequenceOffset`, such that clients would ensure that their transactions would succeed or fail within 10 ( or so ) ledgers.

mvadari: There's already a default here, so I think we're good on that front. 
I guess we can always add that feature later.

src/client/index.ts
```diff
@@ -20,9 +20,11 @@ import {
   decodeXAddress,
 } from "ripple-address-codec";
 
+import { bind } from "..";
```
ledhed2222, line 23: why call this bind? the whole point i thought was to call it `partial`

khancode: Ah you're right, slipped my mind!

src/ledger/autofill.ts
khancode: Thanks! 😊

src/client/index.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable import/max-dependencies -- Client needs a lot of dependencies by definition */
 /* eslint-disable @typescript-eslint/member-ordering -- TODO: remove when instance methods aren't members */
 /* eslint-disable max-lines -- This might not be necessary later, but this file needs to be big right now */
-import { EventEmitter } from 'events'
```
mvadari, line 4: Please incorporate prettier changes

src/client/index.ts
```diff
@@ -198,6 +200,9 @@ class Client extends EventEmitter {
   // number. Defaults to '2'.
   public readonly maxFeeXRP: string
 
+  // eslint-disable-next-line @typescript-eslint/no-invalid-this -- this is expected
+  public autofill = partial(autofill, this)
```
mvadari, line 204: can you add a TODO to use `partial` for the other methods as well? (you don't need to do it now, especially because we're going to delete a bunch of these methods anyways)

khancode: Done.

src/utils/index.ts
```diff
@@ -214,7 +214,21 @@ function ISOTimeToRippleTime(iso8601: string): number {
...
 
+/**
+ * It returns a function that prepends params to the given func.
+ * A sugar function for JavaScript .bind().
+ *
+ * @param func - A function to prepend params.
+ * @param params - Parameters to prepend to a function.
+ * @returns A function bound with params.
+ */
+function partial(func: Function, ...params: any[]): Function {
```
mvadari, line 225: Given that this is a helper method for the client, I'd recomment putting this in `src/client` so it doesn't need to be exported.

test/client/autofill.ts
```diff
@@ -0,0 +1,202 @@
...
+    this.mockRippled.addResponse('ledger', {
+      status: 'success',
+      type: 'response',
+      result: {
+        ledger_index: 9038214,
+      },
+    })
+    const txResult = await this.client.autofill(tx)
+    assert.strictEqual(txResult.LastLedgerSequence, 9038234)
+  })
```
mvadari, line 201: I would add one more test that checks if everything works all together (if all of the autofillable fields are empty, will it autofill all of them correctly), given that you're using a lot of promises.

test/models/utils.ts
```diff
@@ -18,13 +32,124 @@ describe('Models Utils', function () {
...
+        Flags: {
+          tfPassive: true,
+          tfImmediateOrCancel: false,
+          tfFillOrKill: true,
+          tfSell: false,
+        },
+      }
+
+      const { tfPassive, tfFillOrKill } = OfferCreateFlagsEnum
+      // eslint-disable-next-line no-bitwise -- expected use of bitwise
```
mvadari, line 69: Just disable this for the whole file - it'll be easier/cleaner

mvadari (approved):
A few notes, nothing to hold up approval. Please address them before merging.

src/utils/index.ts
khancode: I put it here because I see we use Lodash's version of `partial` in other files too. I'm planning on creating another PR to replace all Lodash partial with this one.

mvadari: Everywhere we're using it will be scrapped, so don't worry about that. You can even make the method tailored to the client.

src/client/index.ts
```diff
@@ -169,6 +170,20 @@ function getCollectKeyFromCommand(command: string): string | null {
...
+/**
+ * It returns a function that prepends params to the given func.
+ * A sugar function for JavaScript .bind() without the "this" (keyword) binding.
+ *
+ * @param func - A function to prepend params.
+ * @param params - Parameters to prepend to a function.
+ * @returns A function bound with params.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types -- expected param types
+function prepend(func: Function, ...params: unknown[]): Function {
```
khancode, line 182: @ledhed2222 Since `partial` will only be used in `Client` (@mvadari), I moved it here and renamed it to `prepend` to tailor it more to `Client` class.

ledhed2222 (approved):

PR #1575, opened 2021-08-30
Lints `test/utils`

mvadari:

Title says it all

### Context of Change

Lints the whole folder

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/test/utils/dropsToXrp.ts
+++ b/test/utils/dropsToXrp.ts
@@ -87,38 +87,38 @@ describe("dropsToXrp", function () {
   it("throws with an amount with too many decimal places", function () {
     assert.throws(() => {
       dropsToXrp("1.2");
-    }, /has too many decimal places/);
+    }, /has too many decimal places/u);
 
     assert.throws(() => {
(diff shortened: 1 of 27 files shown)


test/utils/dropsToXrp.ts
```diff
@@ -87,38 +87,38 @@ describe("Drops To XRP", function () {
   it("throws with an amount with too many decimal places", function () {
     assert.throws(() => {
       dropsToXrp("1.2");
-    }, /has too many decimal places/);
+    }, /has too many decimal places/u);
```
JST5000, line 90: Why are these regex strings? (Or are they some other kind of string? - I'm not quite sure what /.../u does)

JST5000 (approved):
LGTM!

test/utils/dropsToXrp.ts
mvadari: It's regex - it checks that the regex is in the error message somewhere

natenichols (approved):

PR #1576, opened 2021-08-31
feat: Add typescript types to subscribe

### LINKED ISSUE #1306: Stream event typings missing
As the project supports TypeScript, it would be nice to have typed stream events.

For example I could not find the type (interface) of LedgerEvent in the code (neither exported nor internal). However, there are JSON schemas for the events:
https://github.com/ripple/ripple-lib/blob/d82703f41b41153f647084a50013bad35f301f9c/src/common/schemas/output/ledger-event.json

PR1: A good first step would be providing the types of such events exported standalone, for example
```ts
// definition
export interface LedgerEvent {
  ledgerVersion: number;
  // ...
}

// usage
import { RippleAPI, LedgerEvent } from "ripple-lib";
const api = new RippleAPI({ /* ... */ });
api.on('ledger', (ledger) => console.log((ledger as LedgerEvent).ledgerVersion));
```

PR2: After the types are defined [TypeScript function overloads](https://www.typescriptlang.org/docs/handbook/functions.html#overloads) could be used to type the EventEmitter api in a way that would only allow specified event names and automatically get the type of the event callback similar to how the `request` method is typed already.

natenichols:
Adds typescript types for subscription streams.

Fixes #1306 

### Context of Change
This uses function overloading to add an `on` function to the `Client` class. This allows the user to attach specific handlers for the different transaction streams.

### Type of Change
- [ ] Documentation Updates
- [ ] Release


src/client/index.ts
```diff
@@ -261,6 +273,27 @@ class Client extends EventEmitter {
...
+  public on(event: "transaction", listener: (tx: TransactionStream) => void);
+  public on(
+    event: "peerStatusChange",
+    listener: (status: PeerStatusStream) => void
+  );
+  public on(
+    event: "consensusPhase",
+    listener: (phase: ConsensusStream) => void
+  );
+  public on(event: "path_find", listener: (path: PathFindStream) => void);
```
mvadari, line 290: Thoughts on whether we need additional sugar on top of this?

test/client/combine.ts
```diff
@@ -3,29 +3,28 @@ import binary from "ripple-binary-codec";
 
 import requests from "../fixtures/requests";
```
mvadari, line 4: Looks like you need to re-merge/rebase

src/models/methods/subscribe.ts
```diff
@@ -115,10 +115,25 @@ export interface ConsensusStream extends BaseStream {
...
+  type: "path_find";
+  source_account: string;
+  destination_account: string;
+  destination_amount: Amount;
+  full_reply: boolean;
+  id: any;
+  send_max?: Amount;
+  alternatives: {
+    paths_computed: Path[];
+    source_amount: string;
```
JST5000, line 128: I think the source_amount should be an "Amount" based on this documentation for PathFindResponse (Which might not be the same thing you're going off of here, so I could be wrong!) - https://xrpl.org/path_find.html#path_find

JST5000 (commented):
Just one thing to check before I feel comfortable approving.

src/client/index.ts
mvadari: We decided the answer was "no"

.github/workflows/nodejs.yml
```diff
@@ -7,6 +7,7 @@ on:
   push:
     branches: [ develop, master, 2.0 ]
   pull_request:
+    branches: [ develop, 2.0 ]
```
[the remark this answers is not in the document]
natenichols: I did this because ci wasn't running on my PR...

src/client/index.ts
```diff
@@ -250,6 +260,8 @@ class Client extends EventEmitter {
   public request(
     r: SubmitMultisignedRequest
   ): Promise<SubmitMultisignedResponse>;
+  public request(r: SubscribeRequest): Promise<SubscribeResponse>;
+  public request(r: UnsubscribeRequest): Promise<UnsubscribeResponse>;
```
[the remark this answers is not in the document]
natenichols: Yeah, IMO that seems reasonable. Thats also how it's [expected to work in `1.x`](https://xrpl.org/rippleapi-reference.html#listening-to-streams)

mvadari (approved):

JST5000 (approved):
LGTM

src/models/methods/subscribe.ts
```diff
@@ -115,10 +115,25 @@ export interface ConsensusStream extends BaseStream {
   consensus: "open" | "establish" | "accepted";
 }
 
+export interface PathFindStream {
```
mvadari, line 118: This type should extend `BaseStream` like the other stream types

test/client/subscribe.ts
```diff
@@ -0,0 +1,109 @@
...
+  it("Emits transaction", async function (done) {
+    this.client.on("transaction", (tx) => {
+      assert(tx.type === "transaction");
+      done();
+    });
+
+    this.client.connection.onMessage(
+      JSON.stringify(rippled.streams.transaction)
+    );
+  });
```
mvadari, line 39: Given that you're doing this for all the types, you could make this a helper function

natenichols: Yeah, but I'd still have to call `onMessage` w/ the appropriate `rippled.streams` fixture, which would be a lil trickier. What specifically would you be looking for here?

IMO I like having them all separate, that way you can see exactly which one broke.

src/models/methods/subscribe.ts
```diff
@@ -111,10 +111,25 @@ export interface ConsensusStream extends BaseStream {
   consensus: 'open' | 'establish' | 'accepted'
 }
 
+export interface PathFindStream extends BaseStream {
+  type: "path_find";
+  source_account: string;
+  destination_account: string;
+  destination_amount: Amount;
+  full_reply: boolean;
+  id: any;
```
mvadari, line 120: Streams don't have an ID - that's just the first response

mvadari: Unless that's special about `PathFind` - in which case it should be typed as `number | string`

test/client/subscribe.ts
mvadari: I was thinking something similar to what we're doing with addresses for the client tests - you can have an `it` in the loop, so you can still tell which one broke (because they're individual tests).

mvadari (approved):
nothing to hold up approval

src/models/methods/subscribe.ts
natenichols: PathFind does have an ID

PR #1577, opened 2021-08-31
Lints `src/client`

mvadari:

Title says it all.

### Context of Change

Adding linting to xrpl.js

### Type of Change

## Test Plan

CI passes. Linter doesn't have any errors for `src/client`.


src/models/methods/baseMethod.ts
```diff
@@ -1,4 +1,5 @@
 export interface BaseRequest {
+  [x: string]: unknown;
```
mvadari, line 2: Not sure if this is the right way to solve the problem, but this fixes TypeScript's complaint in `client.request` that `account` doesn't exist on all `Request` objects (even though we don't need it to)

src/client/backoff.ts
```diff
@@ -1,42 +1,69 @@
...
+   * @returns Number of attempts.
+   */
+  public get attempts(): number {
+    return this.numAttempts;
   }
 
   /**
    * Return the backoff duration.
+   *
+   * @returns The backoff duration.
```
JST5000, line 55: nit:
```suggestion
   * @returns The backoff duration in milliseconds.
```

src/client/broadcastClient.ts
```diff
@@ -1,44 +1,42 @@
 import { Client, ClientOptions } from ".";
 
-class BroadcastClient extends Client {
-  ledgerVersion: number | undefined = undefined;
-  private readonly _clients: Client[];
+/**
+ *
+ */
```
JST5000, line 5: Is this supposed to be blank?

src/client/broadcastClient.ts
```diff
@@ -1,44 +1,42 @@
...
+export default class BroadcastClient extends Client {
```
JST5000, line 6: Is this class description comment supposed to be blank?

src/client/broadcastClient.ts
```diff
@@ -1,44 +1,42 @@
...
-    this.isConnected = function () {
-      return clients.map((client) => client.isConnected()).every(Boolean);
-    };
-
-    // synchronous methods are all passed directly to the first client instance
-    const defaultClient = clients[0];
-    const syncMethods = ["sign"];
-    syncMethods.forEach((name) => {
-      this[name] = defaultClient[name].bind(defaultClient);
-    });
```
JST5000, line 41: Why do we no longer need to pass in these methods to the first client?

mvadari: We're removing `sign` from the client anyways, with the `Signer`.

natenichols (approved):

mukulljangid (approved):

PR #1578, opened 2021-08-31
Lints `test/models`

mvadari:

Title says it all.

### Context of Change

Adding linting to xrpl.js

### Type of Change

## Test Plan

CI passes. Linter doesn't have any errors for `test/models`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -90,6 +90,21 @@ module.exports = {
 
         // We need to mess with internal things to generate certain testing situations
         "@typescript-eslint/no-unsafe-member-access": "off",
+
+        // We need to be able to import xrpl-local
+        "node/no-extraneous-import": [
+          "error",
+          {
+            allowModules: ["xrpl-local"],
+          },
+        ],
+      },
+    },
+    {
+      files: ["test/models/*.ts"],
+      rules: {
+        "@typescript-eslint/consistent-type-assertions": "off",
+        "@typescript-eslint/no-explicit-any": "off",
       },
     },
     {
--- a/test/models/accountSet.ts
+++ b/test/models/accountSet.ts
@@ -70,7 +70,7 @@ describe("AccountSet", function () {
   });
 
   it(`throws w/ invalid EmailHash`, function () {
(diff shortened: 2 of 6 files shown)


.eslintrc.js
```diff
@@ -2,98 +2,120 @@ module.exports = {
   root: true,
 
   // Make ESLint compatible with TypeScript
-  parser: '@typescript-eslint/parser',
+  parser: "@typescript-eslint/parser",
```
ledhed2222, line 5: why are double-quotes required? should be single quotes.... @natenichols

test/models/accountSet.ts
```diff
@@ -70,7 +70,7 @@ describe("AccountSet Transaction Verification", function () {
   });
 
   it(`throws w/ invalid EmailHash`, function () {
-    account.EmailHash = 657861645678909876543456789876543;
+    account.EmailHash = 6578656789876543;
```
ledhed2222, line 73: why is this a linter change? and why are semi colons not being removed as per our `prettier` config @natenichols

mvadari: This is a linter change because the number was too big, and the linter said the precision would change on runtime. Since it doesn't matter how big the number is for this check, I just made the numbers smaller.

.eslintrc.js
```diff
@@ -2,98 +2,120 @@ module.exports = {
...
+        "node/no-extraneous-import": [
+          "error",
+          {
+            allowModules: ["xrpl-local"],
+          },
+        ],
+      },
+    },
+    {
+      files: ["test/models/*.ts"],
```
JST5000, line 104: When we're doing other linting tickets will we need to mess with the config like this? (Ex. when linting test/client)

JST5000 (approved):
Changes LGTM

.eslintrc.js
mvadari: Perhaps - it depends on the specific errors you're getting. You should probably talk to @natenichols or me if you want to disable a rule for a whole folder.

natenichols (approved):

PR #1594, opened 2021-09-02
Lints top-level test files

mvadari:

This PR lints all of `test/*.ts`.

### Context of Change

Linter

### Type of Change

## Test Plan

CI passes.


.eslintrc.js
```diff
@@ -88,6 +88,8 @@ module.exports = {
 
         // We need to test things without type guards sometimes
         "@typescript-eslint/no-unsafe-assignment": "off",
+        "@typescript-eslint/no-unsafe-call": "off",
+        "@typescript-eslint/consistent-type-assertions": "off",
```
ledhed2222, line 92: turning both of these off seems rather aggressive. i'd rather see individual `eslint-disable` comments for the areas that we can't fix which violate these.

.eslintrc.js
```diff
@@ -99,6 +101,14 @@ module.exports = {
...
           },
         ],
+
+        // Tests are already in 2 callbacks, so max 3 is pretty restrictive
+        "max-nested-callbacks": [
+          "error",
+          {
+            max: 4,
+          },
+        ],
```
ledhed2222, line 111: i would turn this off completely for test files.

.eslintrc.js
ledhed2222: ah this is tests, nevermind

test/broadcastClient.ts
```diff
@@ -1,35 +1,45 @@
...
     });
     assert(this.client.isConnected());
-    return this.client.request({ command: "server_info" }).then((response) => {
-      return checkResult(responses.getServerInfo, response.result.info);
-    });
+    this.client
+      .request({ command: "server_info" })
+      .then(async (response: ServerInfoResponse) => {
+        checkResult(responses.getServerInfo, response.result.info);
+      });
```
ledhed2222, line 42: if you're using the `then` syntax there is no reason to mark this function as `async` (the other one that is. The inner one also doesn't need to because you're not `await`ing it or doing anything with the promise

test/client/getPaths.ts
```diff
@@ -1,7 +1,7 @@
 import addresses from "../fixtures/addresses.json";
 import requests from "../fixtures/requests";
 import rippled from "../fixtures/rippled";
-import setupClient from "../setupClient";
+import { setupClient, teardownClient } from "../setupClient";
 import { assertRejects } from "../testUtils";
 // import responses from '../fixtures/responses'
```
ledhed2222, line 6: let's get rid of this while you're here

mvadari: I left it in because the tests here will be rewritten after the sugar is.

ledhed2222 (approved):

test/mockRippledTest.ts
```diff
@@ -17,10 +17,14 @@ describe('mock rippled tests', function () {
   })
 
   it('provide bad response shape', async function () {
-    assert.throws(
-      () => this.mockRippled.addResponse('account_info', { data: {} }),
-      Error,
-    )
+    try {
```
natenichols, line 20: Why is this preferable to `assert.throws()`

natenichols (approved):
One question, but otherwise LGTM 👍

test/mockRippledTest.ts
mvadari: The linter complained - `no-unsafe-return`. If you'd rather I disable the error instead of doing this, I can do that.

natenichols: Dont worry about disabling the rule, this is okay.

PR #1595, opened 2021-09-02
Lints `src/common`

mvadari:

Title says it all. I didn't lint `src/types` because that's going to be deleted soon.

### Context of Change

Linting the whole codebase

### Type of Change

## Test Plan

CI passes. Linter doesn't have any errors in `src/common/*.ts`.

### DIFF
--- a/src/transaction/sign.ts
+++ b/src/transaction/sign.ts
@@ -10,7 +10,6 @@ import { xrpToDrops } from '../utils'
 import { computeBinaryTransactionHash } from '../utils/hashes'
 
 import { SignOptions, KeyPair, TransactionJSON } from './types'
-import * as utils from './utils'
 
 function computeSignature(tx: object, privateKey: string, signAs?: string) {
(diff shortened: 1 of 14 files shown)


mukulljangid (approved):

src/transaction/sign.ts
```diff
@@ -182,14 +182,15 @@ function checkTxSerialization(serialized: string, tx: TransactionJSON): void {
...
-    const error = new utils.common.errors.ValidationError(
-      "Serialized transaction does not match original txJSON. See `error.data`"
-    );
-    error.data = {
+    const data = {
       decoded,
       tx,
       diff: objectDiff(tx, decoded),
     };
+    const error = new utils.common.errors.ValidationError(
```
natenichols, line 190: I would prefer if we just imported `ValidationError` and used `new ValidationError` here.

natenichols (approved):
This LGTM 👍

PR #1596, opened 2021-09-03
Lint integration and browser tests

JST5000:

Title says it all

### Type of Change

## Before / After

* Rewrote several of the tests in 'await' form instead of '.then()' form to improve readability and make types more explicit.
* Removed retry logic for integration test because it's no longer needed with our setup, and it was very confusing to use

## Test Plan

CI continues to pass


src/client/index.ts
```diff
@@ -252,7 +252,9 @@ class Client extends EventEmitter {
   ): Promise<SubmitMultisignedResponse>;
   public request(r: TransactionEntryRequest): Promise<TransactionEntryResponse>;
   public request(r: TxRequest): Promise<TxResponse>;
-  public request<R extends Request, T extends Response>(r: R): Promise<T> {
+  public async request<R extends Request, T extends Response>(
+    r: R
+  ): Promise<T> {
     // TODO: should this be typed with `extends BaseRequest/BaseResponse`?
```
JST5000, line 258: These are auto-changes because I clicked "ctrl-s" while in here.

test/browser/browserIntegration.ts
```diff
@@ -1,3 +1,5 @@
+import * as path from "path";
```
mvadari, line 1: * imports is bad practice

test/browser/browserIntegration.ts
```diff
@@ -6,7 +8,12 @@ describe("Browser Tests", function () {
     const browser = await puppeteer.launch({ headless: true });
     try {
       const page = await browser.newPage().catch();
-      await page.goto(`file:///${__dirname}/../localIntegrationRunner.html`);
+      const fileLocation = path.join(
+        "file:///",
+        __dirname,
+        "../localIntegrationRunner.html"
+      );
+      await page.goto(fileLocation);
```
mvadari, line 16: This should be `await path.goto(path.join(...))`

test/integration/integration.ts
```diff
@@ -1,124 +1,187 @@
...
-                account
-              ).then(resolve, reject),
-            INTERVAL
-          );
-        });
-      }
-      console.log(error.stack);
-      assert(false, `Transaction not successful: ${error.message}`);
-    });
+  const client: Client = testcase.client;
```
mvadari, line 55: The added linter rules I sent you should make this not necessary.
The goal with linting is to make code nicer, not to make it a pain to satisfy the linter. This type of solution is excessive and annoying, and honestly makes the code a little less readable, so we want to avoid doing things like this.

test/integration/integration.ts
```diff
@@ -1,124 +1,187 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call -- Is better than setting type by extracting to variable
+    testcase.transactions.push(hash);
+  }
+
+  return {
+    txJSON: JSON.stringify(data.result),
+  };
 }
 
-function testTransaction(
```
mvadari, line 72: Why is this code commented out?

test/integration/integration.ts
```diff
@@ -1,124 +1,187 @@
...
+  assert.strictEqual(submittedResponse.result.engine_result, "tesSUCCESS");
+  const options = {
+    minLedgerVersion: lastClosedLedgerVersion,
+    maxLedgerVersion: txData.LastLedgerSequence,
+  };
+  await ledgerAccept(testcase.client);
+  return verifyTransaction(testcase, signedData.id, type, options, address);
 }
 
-function setup(this: any, server = serverUrl) {
```
mvadari, line 118: Why is this commented out?

test/integration/utils.ts
```diff
@@ -0,0 +1,99 @@
+import { Client } from "../../src";
+import { SignedTransaction } from "../../src/common/types/objects";
+
+const masterAccount = "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh";
+const masterSecret = "snoPBrXtMeMyMHUVTgbuqAfg1SUTb";
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- This function is actually returning any
+export async function ledgerAccept(client: Client): Promise<any> {
```
mvadari, line 8: This method shouldn't be both here and in `integration.ts` - pick one place

test/integration/integration.ts
```diff
@@ -1,124 +1,187 @@
...
 const HOST = process.env.HOST ?? "0.0.0.0";
+// eslint-disable-next-line node/no-process-env -- Allows the user to pass in different ports for local rippled
 const PORT = process.env.PORT ?? "6006";
 const serverUrl = `ws://${HOST}:${PORT}`;
 
 console.log(serverUrl);
 
-function acceptLedger(client) {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- The return type is legitimately any
+async function acceptLedger(client: Client): Promise<any> {
```
mvadari, line 42: https://xrpl.org/ledger_accept.html

test/integration/integration.ts
```diff
@@ -1,124 +1,187 @@
...
 
-function verifyTransaction(testcase, hash, type, options, txData, account) {
+// eslint-disable-next-line max-params -- Required for running checks while we move through the test
+async function verifyTransaction(
+  testcase: Mocha.Context,
+  hash: string,
+  type: string,
+  options: { minLedgerVersion: number; maxLedgerVersion?: number },
+  account: string
+): Promise<{ txJSON: string }> {
```
mvadari, line 53: If this is returning an object with only one element, you should just return that element - less effort on the other side.

test/integration/integration.ts
```diff
@@ -1,124 +1,187 @@
...
+
+// eslint-disable-next-line max-params -- The extra parameters let us call verifyTransaction
+async function testTransaction(
+  testcase: Mocha.Context,
+  type: string,
+  lastClosedLedgerVersion: number,
+  prepared: Prepare,
+  address = walletAddress,
+  secret = walletSecret
+): Promise<{ txJSON: string }> {
```
mvadari, line 118: ditto

test/integration/integration.ts
```diff
@@ -1,124 +1,187 @@
...
 
-function acceptLedger(client) {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- The return type is legitimately any
+async function acceptLedger(client: Client): Promise<any> {
   return client.connection.request({ command: "ledger_accept" });
 }
 
-function verifyTransaction(testcase, hash, type, options, txData, account) {
+// eslint-disable-next-line max-params -- Required for running checks while we move through the test
+async function verifyTransaction(
```
mvadari, line 47: All of these helper methods should move to the `utils` file. These methods (plus the methods in the `utils` file) might need to be split between a couple of files in order to be easier to parse.

test/integration/integration.ts
```diff
@@ -142,63 +210,72 @@ function makeTrustLine(testcase, address, secret) {
...
       client.request({ command: "submit", tx_blob: signed.signedTransaction })
     )
-    .then(() => ledgerAccept(client));
+    .then(async () => ledgerAccept(client));
 }
 
-function setupAccounts(testcase) {
-  const client = testcase.client;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- The return type is legitimately any
+async function setupAccounts(testcase: Mocha.Context): Promise<any> {
```
mvadari, line 246: Can you rework the logic of this like you did with some of the other methods, so it's easier to read?

src/client/connection.ts
```diff
@@ -156,10 +156,11 @@ export class Connection extends EventEmitter {
   private ws: WebSocket | null = null;
   private reconnectTimeoutID: null | NodeJS.Timeout = null;
   private heartbeatIntervalID: null | NodeJS.Timeout = null;
-  private readonly retryConnectionBackoff = new ExponentialBackoff({
-    min: 100,
-    max: SECONDS_PER_MINUTE * 1000,
-  });
+  private readonly retryConnectionBackoff: ExponentialBackoff =
```
mvadari, line 159: This doesn't need to be typed - it's already got a type with the ` = new ExponentialBackoff`

src/client/connection.ts
```diff
@@ -1,91 +1,91 @@
 /* eslint-disable max-lines -- Connection is a big class */
-import { EventEmitter } from 'events'
```
mvadari, line 2: You need to update your prettier config + run that

mvadari:
Closing in favor of #1612

PR #1600, opened 2021-09-07
Lints `src/wallet` and `test/wallet`

mvadari:

Title says it all.

### Context of Change

Linting the whole repo

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/src/wallet/generateFaucetWallet.ts
+++ b/src/wallet/generateFaucetWallet.ts
@@ -118,7 +118,7 @@ async function onEnd(
   startingBalance: number,
   fundWallet: Wallet,
   resolve: (wallet?: Wallet) => void,
-  reject: (err: ErrorConstructor | Error) => void,
+  reject: (err: ErrorConstructor | Error | unknown) => void,
 ): Promise<void> {
   const body = Buffer.concat(chunks).toString()
 
@@ -152,7 +152,7 @@ async function processSuccessfulResponse(
   startingBalance: number,
   fundWallet: Wallet,
   resolve: (wallet?: Wallet) => void,
-  reject: (err: ErrorConstructor | Error) => void,
+  reject: (err: ErrorConstructor | Error | unknown) => void,
 ): Promise<void> {
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- We know this is safe and correct
   const faucetWallet: FaucetWallet = JSON.parse(body)
@@ -182,7 +182,10 @@ async function processSuccessfulResponse(
       )
     }
   } catch (err) {
-    reject(new XRPLFaucetError(err))
+    if (err instanceof Error) {
+      reject(new XRPLFaucetError(err.message))
+    }
+    reject(err)
   }
 }
 
--- a/src/wallet/generateFaucetWallet.ts
+++ b/src/wallet/generateFaucetWallet.ts
@@ -120,7 +120,7 @@ async function onEnd(
   client: Client,
   startingBalance: number,
   fundWallet: Wallet,
-  resolve: (wallet: Wallet | undefined) => void,
+  resolve: (wallet?: Wallet) => void,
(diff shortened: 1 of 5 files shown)


src/wallet/generateFaucetWallet.ts
```diff
@@ -162,6 +210,7 @@ async function getAddressXrpBalance(
     )
     return xrpBalance[0].value
   } catch (err) {
+    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- Is going to be an error
```
natenichols, line 213: Since you can `throw` anything in JS, I usually do something like
```ts
} catch (err) {
    if (err instanceOf Error) {
        return `Unable to retrieve ${address} balance. Error: ${err.message}`
    }
    
    throw err
}
```

natenichols: And that should get rid of the `restrict-template-expressions` issue

src/wallet/generateFaucetWallet.ts
```diff
@@ -198,7 +248,8 @@ async function hasAddressBalanceIncreased(
       } catch (err) {
         clearInterval(interval)
         reject(
-          new errors.XRPLFaucetError(
+          new XRPLFaucetError(
+            // eslint-disable-next-line @typescript-eslint/restrict-template-expressions -- Is going to be an error
             `Unable to check if the address ${address} balance has increased. Error: ${err}`,
```
natenichols, line 253: Ditto

src/wallet/generateFaucetWallet.ts
```diff
@@ -141,12 +97,104 @@ async function generateFaucetWallet(
...
+}
+
+// eslint-disable-next-line max-params -- Helper function created for organizational purposes
+async function onEnd(
+  response: IncomingMessage,
+  chunks: Uint8Array[],
+  client: Client,
+  startingBalance: number,
+  fundWallet: Wallet,
+  resolve: (wallet: Wallet | undefined) => void,
```
natenichols, line 123:
```ts
(wallet?: Wallet) => void
```

src/wallet/generateFaucetWallet.ts
```diff
@@ -141,12 +97,104 @@ async function generateFaucetWallet(
...
+  }
+}
+
+// eslint-disable-next-line max-params -- Only used as a helper function
+async function processSuccessfulResponse(
+  client: Client,
+  body: string,
+  startingBalance: number,
+  fundWallet: Wallet,
+  resolve: (wallet: Wallet | undefined) => void,
```
natenichols, line 157: ditto

natenichols (approved):
All my comments are about small stuff and nitpicks.

LGTM 👍

src/wallet/generateFaucetWallet.ts
```diff
@@ -1,40 +1,43 @@
-import https = require('https')
+import { IncomingMessage } from 'http'
+import * as https from 'https'
```
JST5000, line 2: Should we avoid the * import here?

src/wallet/generateFaucetWallet.ts
```diff
@@ -229,3 +286,9 @@ export function getFaucetUrl(client: Client) {
 }
 
 export default generateFaucetWallet
+
+const _private = {
+  FaucetNetwork,
+  getFaucetUrl,
+}
+export { _private }
```
JST5000, line 294: Why does `_private` have an underscore in it's name? (I'm not quite sure what that naming convention implies)

JST5000 (approved):
LGTM - The refactored onEnd feels a lot easier to read!

src/wallet/generateFaucetWallet.ts
mvadari: It means it's a private variable. These methods are used for testing, so they need to be exported for that, but we don't want people to have access to it otherwise. Technically they still have access, but it's bad practice.

PR #1605, opened 2021-09-09
Removes all `prepare` methods

mvadari:

This PR removes all methods that prepare different types of transactions from the client and the codebase.

### Context of Change

The `prepare` methods have been replaced by the model types and the `autofill` method. This allows developers to reference rippled documentation _as well as_ xrpl.js documentation to figure out how to build their transactions, instead of just needing to rely on one set of documentation.

### Type of Change

## Test Plan

Removed tests that were basically just checking `prepare`. Commented out some integration tests for sugar that will be greatly changed from its current state. Fixed other tests so that they didn't use the now-deleted methods. CI passes.


src/transaction/sign.ts
```diff
@@ -5,10 +5,11 @@ import keypairs from 'ripple-keypairs'
 
 import type { Client, Wallet } from '..'
 import { ValidationError } from '../common/errors'
+import { Transaction } from '../models/transactions'
 import { xrpToDrops } from '../utils'
 import { computeBinaryTransactionHash } from '../utils/hashes'
 
-import { SignOptions, KeyPair, TransactionJSON } from './types'
+import { SignOptions, KeyPair } from './types'
```
natenichols, line 12: Are we going to move types? or will that wait for the "Rip out `src/transaction` and `src/ledger`" PRs

mvadari: We have "rip out `src/transaction/sign`" and "rip out all the other old `src` stuff" tickets lined up - these types won't be needed once we do the former

natenichols (approved):

khancode (approved):
LGTM

PR #1609, opened 2021-09-09
feat: cleanup flag naming

natenichols:
Cleans up how flags are named.

### Context of Change
Replaces `<TxType>FlagEnum` with `<TxType>TransactionFlags`, and exports flag enums. This makes it such that a client trying to access a flag value would use `OfferCreateTransactionFlags.tfPassive` instead of `OfferCreateFlagsEnum.tfPassive`.

It also adds `AccountSetTransactionFlags`, which is import *AND* different from `AccountSetFlags`

### Type of Change
- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ledger/autofill.ts
+++ b/src/ledger/autofill.ts
@@ -25,11 +25,13 @@ interface ClassicAccountAndTag {
  * @param signersCount - The expected number of signers for this transaction. Used for multisign.
  * @returns An autofilled transaction.
  */
-async function autofill(
+async function autofill<T extends Transaction>(
   client: Client,
-  tx: Transaction,
+  transaction: T,
   signersCount?: number,
-): Promise<Transaction> {
+): Promise<T> {
+  const tx = {...transaction}
+
   setValidAddresses(tx)
 
   setTransactionFlagsToNumber(tx)
--- a/src/ledger/autofill.ts
+++ b/src/ledger/autofill.ts
@@ -21,7 +21,7 @@ interface ClassicAccountAndTag {
  * Autofills fields in a transaction.
  *
  * @param client - A client.
- * @param tx - A transaction to autofill fields.
+ * @param transaction - A transaction to autofill fields.
  * @param signersCount - The expected number of signers for this transaction. Used for multisign.
  * @returns An autofilled transaction.
  */
@@ -30,7 +30,7 @@ async function autofill<T extends Transaction>(
   transaction: T,
   signersCount?: number,
 ): Promise<T> {
-  const tx = {...transaction}
+  const tx = { ...transaction }
 
   setValidAddresses(tx)
 
--- a/src/models/transactions/accountSet.ts
+++ b/src/models/transactions/accountSet.ts
@@ -35,7 +35,7 @@ export interface AccountSetFlagsInterface {
 
 export interface AccountSet extends BaseTransaction {
   TransactionType: 'AccountSet'
-  Flags?: number | AccountSetTransactionFlags
+  Flags?: number | AccountSetFlagsInterface
   ClearFlag?: number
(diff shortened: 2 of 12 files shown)


src/ledger/autofill.ts
```diff
@@ -25,11 +25,13 @@ interface ClassicAccountAndTag {
  * @param signersCount - The expected number of signers for this transaction. Used for multisign.
  * @returns An autofilled transaction.
  */
-async function autofill(
+async function autofill<T extends Transaction>(
```
mvadari, line 28: If you want it to extend, it might need to be `<T extends BaseTransaction>` so that it can handle other transaction types

src/models/transactions/accountSet.ts
```diff
@@ -15,13 +15,32 @@ enum AccountSetFlagEnum {
...
+  tfOptionalDestTag?: boolean
+  tfRequireAuth?: boolean
+  tfOptionalAuth?: boolean
+  tfDisallowXRP?: boolean
+  tfAllowXRP?: boolean
+}
+
 export interface AccountSet extends BaseTransaction {
   TransactionType: 'AccountSet'
+  Flags?: number | AccountSetTransactionFlags
```
mvadari, line 38: Isn't this supposed to be
```suggestion
  Flags?: number | AccountSetFlagsInterface
```

src/models/transactions/transaction.ts
```diff
@@ -11,15 +11,17 @@ import { EscrowCreate } from './escrowCreate'
...
+import { PaymentChannelClaim, PaymentChannelClaimTransactionFlags } from './paymentChannelClaim'
 import { PaymentChannelCreate } from './paymentChannelCreate'
 import { PaymentChannelFund } from './paymentChannelFund'
 import { SetRegularKey } from './setRegularKey'
 import { SignerListSet } from './signerListSet'
 import { TicketCreate } from './ticketCreate'
-import { TrustSet } from './trustSet'
+import { TrustSet, TrustSetTransactionFlags } from './trustSet'
+
+
```
mvadari, line 24: nit: we don't need these extra lines

src/models/transactions/transaction.ts
```diff
@@ -46,3 +48,12 @@ export interface TransactionAndMetadata {
...
 }
+
+export {
+  AccountSetFlags,
+  AccountSetTransactionFlags,
+  OfferCreateTransactionFlags,
+  PaymentTransactionFlags,
+  PaymentChannelClaimTransactionFlags,
+  TrustSetTransactionFlags
+}
```
mvadari, line 59: Should these be exported here, or in the separate transaction type files?

natenichols: We can export wherever we want to.

I put them in here because it was easiest and it made sense, but I'm open to putting them somewhere else.

src/ledger/autofill.ts
natenichols: There shouldn't be any "other transaction types", though, right? Or are you thinking about someone using xrpl.js on a sidechain or something?

mvadari: Yeah something like that - by `T extends Transaction` did you mean if someone added extra fields or something?

khancode (approved):
LGTM!

src/ledger/autofill.ts
natenichols: I wrote `T extends Transaction` so that `autofill(Payment)` returns a `Payment` type and not a `Transaction` type.

natenichols: Other than that, theres no difference between this and the original type signature.

mvadari (approved):

PR #1611, opened 2021-09-10
add submit transaction methods

khancode:

Adds submit transaction methods: `submitTransaction` and `submitSignedTransaction`.

### Context of Change

Adds submit transactions methods:
- `submitTransaction` - Autofills, signs & encodes, and submits a transaction.
- `submitSignedTransaction` - encodes and submits a signed transaction

### Type of Change

## Test Plan

Adds tests for `submitTransaction` and `submitSignedTransaction`.


src/ledger/submit.ts
```diff
@@ -0,0 +1,64 @@
...
+  txSerialized: string,
+): Promise<SubmitResponse> {
+  const request: SubmitRequest = {
+    command: 'submit',
+    tx_blob: txSerialized,
+  }
+  const response: SubmitResponse = await client.request(request)
+  if (response.result.engine_result !== 'tesSUCCESS') {
+    throw new RippledError(response.result.engine_result_message)
+  }
```
mvadari, line 60: This error handling isn't necessary - the client automatically handles that. So this helper function isn't really needed.

test/client/submit.ts
```diff
@@ -0,0 +1,108 @@
...
+import { RippledError } from 'xrpl-local/common/errors'
+
+import { Wallet } from '../../src'
+import { Transaction } from '../../src/models/transactions'
+import rippled from '../fixtures/rippled'
+import { setupClient, teardownClient } from '../setupClient'
+import { assertRejects } from '../testUtils'
+
+describe('submit', function () {
+  describe('client.submitTransaction', function () {
```
mvadari, line 12: since they don't need any overlapping infrastructure, can you put the two methods in their own test files?

test/client/submit.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      this.mockRippled.addResponse('submit', rippled.submit.success)
+
+      try {
+        const response = await this.client.submitSignedTransaction(signedTx)
+        assert(response.result.engine_result, 'tesSUCCESS')
+      } catch (_error) {
+        assert(false, 'Did not expect an error to be thrown')
+      }
+    })
```
mvadari, line 95: You can use `assert.doesNotThrow` for a slightly nicer interface

test/client/submit.ts
```diff
@@ -0,0 +1,108 @@
...
+      this.mockRippled.addResponse('account_info', rippled.account_info.normal)
+      this.mockRippled.addResponse('ledger', rippled.ledger.normal)
+      this.mockRippled.addResponse('server_info', rippled.server_info.normal)
+      this.mockRippled.addResponse('submit', rippled.submit.success)
+
+      try {
+        const response = await this.client.submitTransaction(wallet, tx)
+        assert(response.result.engine_result, 'tesSUCCESS')
+      } catch (_error) {
+        assert(false, 'Did not expect an error to be thrown')
```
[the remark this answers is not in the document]
mvadari: Actually, you should probably use `assert.doesNotThrow` instead

test/client/submit.ts
```diff
@@ -0,0 +1,108 @@
+import { assert } from 'chai'
+
+import { RippledError } from 'xrpl-local/common/errors'
+
+import { Wallet } from '../../src'
+import { Transaction } from '../../src/models/transactions'
```
mvadari, line 6: use `xrpl-local` here

src/ledger/submit.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @param client - A Client.
+ * @param signedTransaction - A signed transaction to encode and submit.
+ * @returns A promise that contains SubmitResponse.
+ * @throws RippledError if submit request fails.
+ */
+async function submitSignedTransaction(
+  client: Client,
+  signedTransaction: Transaction,
+): Promise<SubmitResponse> {
+  const serialized = encode(signedTransaction)
```
natenichols, line 45: Should we check if this transaction is signed before submitting?

natenichols (approved):
One question from me. Nothing to hold up approval.

src/ledger/submit.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Encodes and submits a signed transaction.
+ *
+ * @param client - A Client.
+ * @param signedTransaction - A signed transaction to encode and submit.
+ * @returns A promise that contains SubmitResponse.
+ * @throws RippledError if submit request fails.
+ */
+async function submitSignedTransaction(
+  client: Client,
+  signedTransaction: Transaction,
```
mvadari, line 43: We could have this take either a `string` or a `Transaction`, so the encoding is optional and only done if it's a `Transaction`

test/client/submit.ts
khancode: I tried using `assert.doesNotThrow` but it complains that `this.client.submitSignedTransaction(signedTx)` is an async function and it expects a void return type. So that's why I resorted to using this Try...Catch.

I even tried doing this but the assertion doesn't work because it's expecting a void return type:
```TypeScript
assert.doesNotThrow(async () => await this.client.submitSignedTransaction(signedTx))
```

src/ledger/submit.ts
khancode: Ah sure, that's a good check.

src/ledger/submit.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @param transaction - A transaction to autofill, sign & encode, and submit.
+ * @returns A promise that contains SubmitResponse.
+ * @throws RippledError if submit request fails.
+ */
+async function submitTransaction(
+  client: Client,
+  wallet: Wallet,
+  transaction: Transaction,
+): Promise<SubmitResponse> {
+  const tx: Transaction = await autofill(client, transaction)
```
mvadari, line 27:
```suggestion
  const tx: Transaction = await client.autofill(transaction)
```

test/client/submitSignedTransaction.ts
```diff
@@ -0,0 +1,54 @@
+import { assert } from 'chai'
+
+import { Transaction } from 'xrpl-local/models/transactions'
+
+import rippled from '../fixtures/rippled'
+import { setupClient, teardownClient } from '../setupClient'
+
+describe('client.submitSignedTransaction', function () {
```
mvadari, line 8: this should use the `addressTests` setup so it checks our X-Address infrastructure as well

mvadari (approved):
small comment, but looks good

src/ledger/submit.ts
```diff
@@ -0,0 +1,55 @@
...
+  const signedTxEncoded: string = sign(wallet, tx)
```
mvadari, line 28:
```suggestion
  const signedTxEncoded= sign(wallet, tx)
```

src/ledger/submit.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+async function submitSignedTransaction(
+  client: Client,
+  signedTransaction: Transaction | string,
+): Promise<SubmitResponse> {
+  const signedTxEncoded =
+    typeof signedTransaction === 'string'
+      ? signedTransaction
+      : encode(signedTransaction)
+  const request: SubmitRequest = {
```
mvadari, line 48:
```suggestion
  const request = {
```

intelliot (approved):
I also like @mvadari 's suggestions

src/ledger/submit.ts
khancode: When I remove `: SubmitRequest`, TS complains because `client.request()` is overloaded with multiple request types. We need to keep this change as a way to tell TS we're calling the `SubmitRequest` version of `client.request()`

mvadari: It should handle the overload automatically - let me take a look

mvadari: Hm that's strange - ok, leave it in for now

src/ledger/submit.ts
khancode: Added TODO to fix this in another PR.

test/client/submitSignedTransaction.ts
```diff
@@ -51,4 +53,18 @@ describe('client.submitSignedTransaction', function () {
       assert(false, 'Did not expect an error to be thrown')
     }
   })
+
+  it('should throw a ValidationError when submiting an unsigned transaction', async function () {
```
mvadari, line 57: extreme nit: 
```suggestion
  it('should throw a ValidationError when submitting an unsigned transaction', async function () {
```

khancode: Haha done 😄

PR #1614, opened 2021-09-13
Remove deprecated deriveAddress from Client

JST5000:

* Remove deriveAddress from Client
* Remove alias in client for deriveXAddress (So the workflow is just importing derive from the utils folder now)
* Move the deriveXAddress test from the Client to Utils folder

### Context of Change

Follow up to #1534 for deriveAddress and deriveXAddress.

### Type of Change

## Before / After

Instead of using Client to access these functions, now you'll have to import them directly from the derive.ts util.

## Test Plan

CI Passes


natenichols (approved):
LGTM 👍

khancode (approved):

src/client/index.ts
```diff
@@ -109,7 +109,6 @@ import {
 import { BaseRequest, BaseResponse } from '../models/methods/baseMethod'
 import combine from '../transaction/combine'
 import { sign } from '../transaction/sign'
-import { deriveAddress, deriveXAddress } from '../utils/derive'
```
mvadari, line 112: If `deriveAddress` is being removed, can we remove the code from `'../utils/derive'`?

mvadari (changes requested):
see comment

src/client/index.ts
```diff
@@ -203,6 +202,7 @@ class Client extends EventEmitter {
    * @param server - URL of the server to connect to.
    * @param options - Options for client settings.
    */
+  // eslint-disable-next-line max-lines-per-function -- Client needs to space for initializing subscriptions
```
mvadari, line 205: don't bother adding these eslint comments - the rule will be edited a bit later and that'll handle this error

mvadari (approved):

PR #1615, opened 2021-09-13
Change multisign to return an encoded transaction instead of a Transaction object

JST5000:

Encode multisign's return instead of just returning the Transaction object 

### Context of Change

This change was inspired by using multisign and realizing the Transaction shouldn't be modified after combining signatures. (All Transaction modifications should happen before then)

### Type of Change

## Before / After

Before multisign returned a Transaction object, now it encodes that object before returning.

## Test Plan

CI Passes


natenichols (approved):
👍

mvadari (approved):

PR #1617, opened 2021-09-13
Resolves TODOs in ./src/client

natenichols:
Resolves various `TODO`s in `./src/client`

### Context of Change
I will go through each change in the "files changed" and leave a note about how I resolved.

### Type of Change
- [ ] Documentation Updates
- [ ] Release


src/client/broadcastClient.ts
```diff
@@ -23,7 +23,7 @@ export default class BroadcastClient extends Client {
     this.clients = clients
     this.getMethodNames().forEach((name: string) => {
       this[name] = async (...args): Promise<unknown> =>
-        // eslint-disable-next-line max-len -- Need a long comment, TODO: figure out how to avoid this weirdness
+        // eslint-disable-next-line max-len -- Need a long comment
```
natenichols, line 26: I don't know how to resolve this one.

natenichols: Disabling `max-len` is fine by me.

src/client/connection.ts
```diff
@@ -57,10 +55,9 @@ export const INTENTIONAL_DISCONNECT_CODE = 4000
 type WebsocketState = 0 | 1 | 2 | 3
 
 function getAgent(url: string, config: ConnectionOptions): Agent | undefined {
-  // TODO: replace deprecated method
   if (config.proxy != null) {
-    const parsedURL = parseURL(url)
-    const parsedProxyURL = parseURL(config.proxy)
+    const parsedURL = new URL(url)
```
natenichols, line 59: Uses the built in `URL`

src/client/connection.ts
```diff
@@ -74,16 +71,16 @@ function getAgent(url: string, config: ConnectionOptions): Agent | undefined {
       (value) => value == null,
     )
     const proxyOptions = { ...parsedProxyURL, ...proxyOverrides }
-    let HttpsProxyAgent
+    let HttpsProxyAgent: new (opt: typeof proxyOptions) => Agent
     try {
-      // eslint-disable-next-line max-len -- Long eslint-disable-next-line TODO: figure out how to make this nicer
-      // eslint-disable-next-line import/max-dependencies, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-require-imports, node/global-require, global-require, -- Necessary for the `require`
+      // eslint-disable-next-line max-len -- Long eslint-disable-next-line required.
```
natenichols, line 76: Similarly didn't know how to resolve, as theres not really a good way to do this in typescript.

src/client/connection.ts
```diff
@@ -244,8 +241,9 @@ export class Connection extends EventEmitter {
     this.ws.on('error', () => clearTimeout(connectionTimeoutID))
     this.ws.on('close', (reason) => this.onConnectionFailed(reason))
     this.ws.on('close', () => clearTimeout(connectionTimeoutID))
-    // eslint-disable-next-line @typescript-eslint/no-misused-promises -- TODO: resolve this
-    this.ws.once('open', async () => this.onceOpen(connectionTimeoutID))
+    this.ws.once('open', () => {
```
natenichols, line 244: Spawn these promises as `void`

src/client/index.ts
```diff
@@ -153,21 +152,6 @@ function getCollectKeyFromCommand(command: string): string | null {
...
 
-/**
- * It returns a function that prepends params to the given func.
- * A sugar function for JavaScript .bind() without the "this" (keyword) binding.
- *
- * @param func - A function to prepend params.
- * @param params - Parameters to prepend to a function.
- * @returns A function bound with params.
- */
-// TODO Need to refactor prepend so TS can infer the correct function signature type
```
natenichols, line 164: We get rid of this TODO by getting rid of `prepend`

src/client/connection.ts
natenichols: There's nothing awaiting them, so this should be fine.

src/client/connection.ts
natenichols: You have to specify each `key` in `proxyOptions`, because w/ the `URL` class `{...URL}` doesn't give you the values of getters.

src/client/connection.ts
```diff
@@ -58,6 +58,22 @@ function getAgent(url: string, config: ConnectionOptions): Agent | undefined {
...
+        username: parsedProxyURL.username,
+        password: parsedProxyURL.password,
+        host: parsedProxyURL.host,
+        hostname: parsedProxyURL.hostname,
+        port: parsedProxyURL.port,
+        pathname: parsedProxyURL.pathname,
+        search: parsedProxyURL.search,
+        hash: parsedProxyURL.hash,
+      },
+      (value) => value == null,
```
mvadari, line 75: Do we even need to do this `omitBy`? Isn't JS going to handle that on its own?

src/client/connection.ts
mvadari: @ledhed2222 I believe did this by doing a normal disable and then re-enabling it below. idk which one is preferable

src/client/connection.ts
natenichols: Yeah, you're prolly right that we can just get rid of it.

src/client/connection.ts
JST5000: Should this be a comment in the code for context?

src/client/index.ts
JST5000: Why are we removing prepend? (Does it mess up typescript?)

If I understand correctly the prepend's purpose here is to turn:
`client.autofill(client, transaction)` into `client.autofill(transaction)` which seems like a nice simplification from the interface perspective.

JST5000: Oh, are these functions now like instance methods, just defined outside the class?

JST5000 (commented):
I had some confusions reading through which I'd like to figure out before I feel comfortable approving :) (See comments for details)

src/client/connection.ts
```diff
@@ -57,10 +55,23 @@ export const INTENTIONAL_DISCONNECT_CODE = 4000
...
+      protocol: parsedProxyURL.protocol,
+      username: parsedProxyURL.username,
+      password: parsedProxyURL.password,
+      host: parsedProxyURL.host,
+      hostname: parsedProxyURL.hostname,
+      port: parsedProxyURL.port,
+      pathname: parsedProxyURL.pathname,
+      search: parsedProxyURL.search,
+      hash: parsedProxyURL.hash,
+    }
```
mvadari, line 73: Instead of this, can we just pass `parsedProxyURL` straight into `proxyOptions` instead of copying it like this?

src/client/index.ts
mvadari: pretty much yeah

src/ledger/submit.ts
```diff
@@ -55,7 +52,7 @@ async function submitSignedTransaction(
     command: 'submit',
```
mvadari, line 52: The request here probably doesn't need to be typed

src/client/broadcastClient.ts
ledhed2222: does this work?

```javascript
/* eslint-disable @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call ---
* types are outlined in Client class */
......
/* eslint-enable @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call */
```

src/client/connection.ts
ledhed2222: what's the purpose of using `void` here?

natenichols: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md#ignorevoid

ledhed2222:
interesting

src/ledger/autofill.ts
```diff
@@ -20,13 +20,13 @@ interface ClassicAccountAndTag {
...
  *
- * @param client - A client.
+ * @param this - A client.
  * @param transaction - A transaction to autofill fields.
  * @param signersCount - The expected number of signers for this transaction. Used for multisign.
  * @returns An autofilled transaction.
  */
 async function autofill<T extends Transaction>(
-  client: Client,
+  this: Client,
```
khancode, line 29: Is it safe to name a param using `this` keyword?

src/client/index.ts
khancode: Before deleting this, I'd like to see if there's a possible fix to see if we can have TS infer the correct result function signature type. I think there's value keeping `prepend`.

src/client/connection.ts
natenichols: No we cannot. I tried that and because `URL` is a class instance, the `...URL` operator seems to not deconstruct into an object in a clean way.

src/ledger/autofill.ts
natenichols: The `Client` parameter is required to be named `this` if you want to bind it to a class member.

src/client/connection.ts
mvadari: Can we just put those straight into `proxyOptions` instead of having this separate `proxyInfo` object then?

src/client/broadcastClient.ts
natenichols: Yeah, that works, I'll do that.

src/client/connection.ts
natenichols: Nah, I think its okay without it.

src/client/index.ts
natenichols: So, there is probably a way to get this to work, but I don't personally see the benefit of explicitly binding `this` to a member of `Client`. Especially when typescript will do it implicitly for us by naming the `Client` parameter `this`.

src/client/broadcastClient.ts
```diff
@@ -19,13 +19,13 @@ export default class BroadcastClient extends Client {
...
-    // exposed for testing
     this.clients = clients
     this.getMethodNames().forEach((name: string) => {
       this[name] = async (...args): Promise<unknown> =>
-        // eslint-disable-next-line max-len -- Need a long comment, TODO: figure out how to avoid this weirdness
-        /* eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call -- Types are outlined in Client class */
+        /* eslint-disable @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call -- Generates types
+          from the Client */
         Promise.race(clients.map(async (client) => client[name](...args)))
+      /* eslint-enable @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call */
```
JST5000, line 28: Interesting! This is a good technique to know for the more verbose lints :)

JST5000 (approved):
LGTM!

src/client/connection.ts
natenichols: Yeah duh, thats a better way to do this.

src/client/index.ts
```diff
@@ -529,24 +513,24 @@ class Client extends EventEmitter {
     return this.connection.isConnected()
   }
 
-  // TODO: Use prepend for other instance methods as well.
-  public autofill = prepend(autofill, this)
+  public autofill = autofill
```
mvadari, line 516: Does this work okay for docs? Or does docgen still think that these are properties? (out of the scope of this ticket, just curious)

src/sugar/orderbook.ts
```diff
@@ -35,35 +37,33 @@ interface OrderbookOptions {
...
  * @returns An object containing buy and sell objects.
  */
 // eslint-disable-next-line max-params -- Function needs 4 params.
 async function getOrderbook(
-  client: Client,
+  this: Client,
   takerPays: TakerAmount,
   takerGets: TakerAmount,
-  options: OrderbookOptions,
+  options?: OrderbookOptions,
```
mvadari, line 51:
```suggestion
  options: OrderbookOptions = {},
```

src/client/connection.ts
```diff
@@ -57,35 +55,49 @@ export const INTENTIONAL_DISCONNECT_CODE = 4000
...
-    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions, @typescript-eslint/no-unsafe-call -- Necessary
-    return new HttpsProxyAgent(proxyOptions) as unknown as Agent
+  if (config.proxy == null) {
+    return undefined
+  }
+
+  const parsedURL = new URL(url)
+  const parsedProxyURL = new URL(config.proxy)
+
+  const proxyOptions = _.omitBy(
```
mvadari, line 65: Why does this have more options than the old version?

natenichols: Moved the options from `parsedProxyURL` into the same object as the proxyOptions object.

src/client/index.ts
natenichols: I think docgen still sees these as properties, but thats something to tackle w/ docgen.

mvadari (approved):
some questions about `getAgent` but nothing to hold up approval

PR #1618, opened 2021-09-13
Add Wallet.fromSecret(...) as an alias for Wallet.fromSeed(...)

JST5000:

Title says it all

### Context of Change

It seemed odd to have to use .fromSeed(...) on variables named 'secret', even if it works. This makes more intuitive sense.

### Type of Change

## Before / After

Added a Wallet.fromSecret(...) method to generate a wallet which just uses fromSeed underneath.

## Test Plan

Added a CI tests to ensure the wallet works as expected.

### DIFF
--- a/src/wallet/index.ts
+++ b/src/wallet/index.ts
@@ -79,12 +79,7 @@ class Wallet {
    * @param algorithm - The digital signature algorithm to generate an address fro.
    * @returns A Wallet derived from a secret (AKA a seed).
    */
-  public static fromSecret(
-    secret: string,
-    algorithm: ECDSA = DEFAULT_ALGORITHM,
-  ): Wallet {
-    return Wallet.fromSeed(secret, algorithm)
-  }
+  public static fromSecret = Wallet.fromSeed
 
   /**
    * Derives a wallet from a mnemonic.
--- a/src/wallet/index.ts
+++ b/src/wallet/index.ts
@@ -73,6 +73,20 @@ class Wallet {
   }
 
   /**
+   * Derives a wallet from a secret (AKA a seed)
+   *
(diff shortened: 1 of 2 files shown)


mvadari (approved):

src/wallet/index.ts
```diff
@@ -72,6 +72,20 @@ class Wallet {
...
   }
 
+  /**
+   * Derives a wallet from a secret (AKA a seed)
+   *
+   * @param secret - A string used to generate a keypair (publicKey/privateKey) to derive a wallet.
+   * @param algorithm - The digital signature algorithm to generate an address fro.
+   * @returns A Wallet derived from a secret (AKA a seed).
+   */
+  public static fromSecret(
```
mvadari, line 82: You might be able to do this more simply by just doing `public static fromSecret = fromSeed`

JST5000: Good to know. I think for this case it's good to have it like this so that the documentation gets picked up by docgen. My thought process is that `fromSecret` is not inferior to `fromSeed` from a naming perspective, and I don't think an alias using `public static fromSecret = fromSeed` would get picked up in docgen. (Although I could be wrong)

ledhed2222: there's got to be a way to document object attributes, IE `fromSecret = fromSeed`. if you just add the docstrings for this above that assignment I would expect it to work

khancode (approved):
Besides addressing @mvadari comment, LGTM.

src/wallet/index.ts
JST5000: You're right, that works - Switched to @mvadari's simpler syntax :)

PR #1619, opened 2021-09-14
refactor: adds error response object

mvadari:

Adds an error response object to account for error responses. Also ran the auto-linter.

### Context of Change

We didn't have an error response object. It's not super necessary since the client throws an error if an error response is received, but it's still good to have one.

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/src/models/transactions/transaction.ts
+++ b/src/models/transactions/transaction.ts
@@ -6,7 +6,7 @@ import _ from 'lodash'
 import { encode, decode } from 'ripple-binary-codec'
 
 import { ValidationError } from '../../common/errors'
-import TransactionMetadata from './metadata'
+import { setTransactionFlagsToNumber } from '../utils'
 
 import { AccountDelete, verifyAccountDelete } from './accountDelete'
 import {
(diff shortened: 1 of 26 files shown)


src/wallet/index.ts
```diff
@@ -72,13 +72,13 @@ class Wallet {
...
   }
 
-  /**
-   * Derives a wallet from a secret (AKA a seed)
-   *
-   * @param secret - A string used to generate a keypair (publicKey/privateKey) to derive a wallet.
-   * @param algorithm - The digital signature algorithm to generate an address fro.
-   * @returns A Wallet derived from a secret (AKA a seed).
-   */
+  // .
```
natenichols, line 75: What happened to this comment?

test/models/checkCash.ts
```diff
@@ -1,7 +1,9 @@
-import { ValidationError } from 'xrpl-local/common/errors'
-import { verifyCheckCash } from './../../src/models/transactions/checkCash'
 import { assert } from 'chai'
+
+import { ValidationError } from 'xrpl-local/common/errors'
+
 import { verify } from '../../src/models/transactions'
+import { verifyCheckCash } from '../../src/models/transactions/checkCash'
```
natenichols, line 6: I'm assuming the linter changed the ordering of all these files? Is that true?

src/models/transactions/transaction.ts
```diff
@@ -79,6 +79,7 @@ export interface TransactionAndMetadata {
  * Encode/decode and individual type validation.
  *
  * @param tx - A Transaction.
+ * @param transaction
```
natenichols, line 82: Parameter name changed.

natenichols: Also change in the JSDoc

src/wallet/index.ts
mvadari: Not sure, but I fixed it.

test/models/checkCash.ts
mvadari: Yes - I didn't make that many changes really, I just ran the autolinter and there were a bunch of changes from there.

natenichols (approved):

JST5000 (approved):
LGTM!

PR #1620, opened 2021-09-14
Remove sign.ts and move functionality into Wallet

JST5000:

Moved over the core functionality from sign.ts into Wallet and refactored sign's tests to work with Wallet.

### Context of Change

Wallet/Signer are replacing sign.ts, so we want to remove sign.ts while ensuring Wallet and Signer still work.

### Type of Change

## Before / After

* Before Sign.ts handled the actual signing logic, now that's all in Wallet
* Before Sign.ts had individual tests, now the tests are in Wallet's tests

## Test Plan

CI Passes (Moved over the old Sign.ts tests to test the same logic in Wallet)


src/wallet/index.ts
```diff
@@ -1,18 +1,24 @@
...
+  encodeForSigning,
+  encodeForMultisigning,
+  encode,
+} from 'ripple-binary-codec'
 import {
   deriveAddress,
   deriveKeypair,
   generateSeed,
   verify,
+  sign,
```
natenichols, line 16: is this the right sign?

src/wallet/index.ts
```diff
@@ -190,4 +200,109 @@ class Wallet {
...
+
+/**
+ *  Decode a serialized transaction, remove the fields that are added during the signing process,
+ *  and verify that it matches the transaction prior to signing. This gives the user a sanity check
+ *  to ensure that what they try to encode matches the message that will be recieved by rippled.
+ *
+ * @param serialized - A signed and serialized transaction.
+ * @param tx - The transaction prior to signing.
+ *
+ * @returns This method does not return a value, but throws an error if the check fails.
```
natenichols, line 252: There is a `@throws` jsdoc option. Use that here.

natenichols (approved):

src/wallet/index.ts
```diff
@@ -152,8 +158,12 @@ class Wallet {
     transaction: Transaction,
     options: SignOptions = { signAs: '' },
```
mvadari, line 159: We should be able to get rid of `SignOptions` now

mvadari: Since we only need an optional `signAs`, you could make just that an optional param, or rename it to something like `multisignSigner` (so that it's more clear that it's just for multisign)

src/wallet/index.ts
```diff
@@ -190,4 +200,111 @@ class Wallet {
   }
 }
 
+function signWithKeypair(
+  txJSON: string,
+  publicKey: string,
+  privateKey: string,
+  options: SignOptions = {
```
mvadari, line 207: ditto

src/wallet/index.ts
```diff
@@ -190,4 +200,111 @@ class Wallet {
   }
 }
 
+function signWithKeypair(
```
mvadari, line 203: This can be a private method on the `Wallet` class - then you don't need to pass in the public and private keys

src/wallet/index.ts
```diff
@@ -190,4 +200,111 @@ class Wallet {
...
+    }
+
+    if (memo.Memo.MemoFormat) {
+      memo.Memo.MemoFormat = memo.Memo.MemoFormat.toUpperCase()
+    }
+
+    return memo
+  })
+
+  assert.deepEqual(
```
mvadari, line 296: Can we put this in a try/catch and throw a different error than an `AssertionError`?

test/wallet/index.ts
```diff
@@ -186,6 +189,15 @@ describe('Wallet', function () {
       '00141BA006D3363D2FB2785E8DF4E44D3A49908780CB4FB51F6D217C08C021429F'
     const address = 'rhvh5SrgBL5V8oeV9EpDuVszeJSSCEkbPc'
 
+    const { sign: REQUEST_FIXTURES } = requests
+    const { sign: RESPONSE_FIXTURES } = responses
```
mvadari, line 193: This should be at the top, with the imports

test/wallet/index.ts
```diff
@@ -186,6 +189,15 @@ describe('Wallet', function () {
...
+    const wallet = Wallet.fromSeed('ss1x3KLrSvfg7irFc1D929WXZ7z9H')
+    const wallet2 = Wallet.fromSeed('shsWGZcmZz6YsWWmcnpfr6fLTdtFV')
+    const wallet3 = new Wallet(
+      '02F89EAEC7667B30F33D0687BBA86C3FE2A08CCA40A9186C5BDE2DAA6FA97A37D8',
+      '00ACCD3309DB14D1A4FC9B1DAE608031F4408C85C73EE05E035B7DC8B25840107A',
+    )
```
mvadari, line 199: This should be in a `beforeEach`, so that error handling is a little better if something breaks in one of these `Wallet` methods

test/wallet/index.ts
```diff
@@ -202,6 +214,252 @@ describe('Wallet', function () {
...
+      }
+
+      const result = Wallet.fromSeed(secret).signTransaction(lowercaseMemoTx)
+      assert.equal(
+        result,
+        '120000228000000023000022B8240000000C2E0000270F201B00D5A36761400000000098968068400000000000000C73210305E09ED602D40AB1AF65646A4007C2DAC17CB6CDACDE301E74FB2D728EA057CF744730450221009C00E8439E017CA622A5A1EE7643E26B4DE9C808DE2ABE45D33479D49A4CEC66022062175BE8733442FA2A4D9A35F85A57D58252AE7B19A66401FE238B36FA28E5A081146C1856D0E36019EA75C56D7E8CBA6E35F9B3F71583147FB49CD110A1C46838788CD12764E3B0F837E0DDF9EA7C1F687474703A2F2F6578616D706C652E636F6D2F6D656D6F2F67656E657269637D0472656E74E1F1',
+      )
+    })
+
+    it('EscrowExecution', async function () {
```
mvadari, line 256:
```suggestion
    it('EscrowFinish', async function () {
```

test/wallet/index.ts
```diff
@@ -202,6 +214,252 @@ describe('Wallet', function () {
...
+      const secret = 'snoPBrXtMeMyMHUVTgbuqAfg1SUTb'
+      const result = Wallet.fromSeed(secret).signTransaction(
+        JSON.parse(REQUEST_FIXTURES.escrow.txJSON) as unknown as Transaction,
```
mvadari, line 259: Instead of parsing it, you should change the fixture so that it's an object instead of a JSON string

test/wallet/index.ts
```diff
@@ -202,6 +214,252 @@ describe('Wallet', function () {
...
+      )
+      assert.deepEqual(result, RESPONSE_FIXTURES.escrow.signedTransaction)
+    })
+
+    it('signAs', async function () {
```
mvadari, line 264: If all of these tests are copied from `test/client/sign`, the testing of X-Addresses should be included

mvadari (changes requested):
See comments

mvadari:
Please also resolve all linter issues in the code you've moved

test/wallet/index.ts
```diff
@@ -185,6 +191,18 @@ describe('Wallet', function () {
     const privateKey =
       '00141BA006D3363D2FB2785E8DF4E44D3A49908780CB4FB51F6D217C08C021429F'
     const address = 'rhvh5SrgBL5V8oeV9EpDuVszeJSSCEkbPc'
+    let wallet
+    let wallet2
+    let wallet3
+
+    this.beforeAll(function () {
```
mvadari, line 198:
```suggestion
    beforeAll(function () {
```

test/wallet/index.ts
```diff
@@ -199,9 +217,278 @@ describe('Wallet', function () {
       const wallet = new Wallet(publicKey, privateKey)
       const signedTx: string = wallet.signTransaction(txJSON)
 
-      // TODO: Check the output of the signature against a known result
```
mvadari, line 202: Why was this TODO removed without being resolved?

mvadari: This test is probably not needed because it's the same thing as the one right under it

test/wallet/index.ts
```diff
@@ -199,9 +217,278 @@ describe('Wallet', function () {
...
+      }
+
+      const result = Wallet.fromSeed(secret).signTransaction(lowercaseMemoTx)
+      assert.equal(
+        result,
+        '120000228000000023000022B8240000000C2E0000270F201B00D5A36761400000000098968068400000000000000C73210305E09ED602D40AB1AF65646A4007C2DAC17CB6CDACDE301E74FB2D728EA057CF744730450221009C00E8439E017CA622A5A1EE7643E26B4DE9C808DE2ABE45D33479D49A4CEC66022062175BE8733442FA2A4D9A35F85A57D58252AE7B19A66401FE238B36FA28E5A081146C1856D0E36019EA75C56D7E8CBA6E35F9B3F71583147FB49CD110A1C46838788CD12764E3B0F837E0DDF9EA7C1F687474703A2F2F6578616D706C652E636F6D2F6D656D6F2F67656E657269637D0472656E74E1F1',
+      )
+    })
+
+    it('EscrowFinish', async function () {
```
mvadari, line 259:
```suggestion
    it('signTransaction with EscrowFinish', async function () {
```

test/wallet/index.ts
```diff
@@ -199,9 +217,278 @@ describe('Wallet', function () {
       const wallet = new Wallet(publicKey, privateKey)
       const signedTx: string = wallet.signTransaction(txJSON)
 
-      // TODO: Check the output of the signature against a known result
       assert.isString(signedTx)
     })
+
+    it('sign', async function () {
```
mvadari, line 223:
```suggestion
    it('signTransaction', async function () {
```
All of these tests should be renamed accordingly

test/fixtures/requests/sign.json
```diff
@@ -1,5 +1,14 @@
...
+    "Flags":2147483648,
+    "TransactionType":"AccountSet",
+    "Account":"r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59",
+    "Domain":"6578616D706C652E636F6D",
+    "LastLedgerSequence":8820051,
+    "Fee":"12",
+    "Sequence":23,
+    "SigningPubKey":"02F89EAEC7667B30F33D0687BBA86C3FE2A08CCA40A9186C5BDE2DAA6FA97A37D8"
+    },
   "instructions": {
```
mvadari, line 12: You can remove `instructions` from all of these - we're not using them anymore

src/wallet/index.ts
```diff
@@ -145,15 +150,45 @@ class Wallet {
...
-   * @param options - Options to include for signing.
+   * @param multisignAddress - Multisign only. An account address corresponding to the multi-signature being added. If this
+   * wallet represents your [master keypair](https://xrpl.org/cryptographic-keys.html#master-key-pair) you can get your account address
+   * with the Wallet.getClassicAddress() function.
    * @returns A signed transaction.
    */
   public signTransaction(
     transaction: Transaction,
-    options: SignOptions = { signAs: '' },
+    multisignAddress: string = '',
```
mvadari, line 160:
```suggestion
    multisignAddress?: string,
```
This shouldn't be an empty string if it's not needed, it should be undefined

src/wallet/index.ts
```diff
@@ -181,13 +216,89 @@ class Wallet {
...
+
+    // - If SigningPubKey was not in the original tx, then we should delete it.
+    //   But if it was in the original tx, then we should ensure that it has not been changed.
+    if (!tx.SigningPubKey) {
+      delete decoded.SigningPubKey
+    }
+
+    // - Memos have exclusively hex data which should ignore case.
+    //   Since decode goes to upper case, we set all tx memos to be uppercase for the comparison.
+    tx.Memos?.map((memo) => {
```
mvadari, line 265: This mutates `tx` - make a copy and modify the copy

src/wallet/index.ts
```diff
@@ -181,13 +216,89 @@ class Wallet {
   }
 
   /**
-   * Gets the classic address of the account this wallet represents.
+   * Gets the classic address of the account this wallet represents. This only is correct if this wallet corresponds
+   * to your [master keypair](https://xrpl.org/cryptographic-keys.html#master-key-pair). If this wallet represents a
+   * [regular keypair](https://xrpl.org/cryptographic-keys.html#regular-key-pair) this will provide an incorrect address.
```
mvadari, line 221: We need better support for regular keys in wallets anyways - the `getClassicAddress` is used in other places, such as in the Signer. There's a ticket for this.

JST5000: So should I remove the comment? (Or is that fine for now?)

src/wallet/index.ts
```diff
@@ -72,13 +86,14 @@ class Wallet {
...
   }
 
-  /**
-   * Derives a wallet from a secret (AKA a seed)
-   *
-   * @param secret - A string used to generate a keypair (publicKey/privateKey) to derive a wallet.
-   * @param algorithm - The digital signature algorithm to generate an address fro.
-   * @returns A Wallet derived from a secret (AKA a seed).
-   */
+  // .
```
mvadari, line 89: This should be changed to the `/**` syntax used above (you just need to add a period to the end of the line.

src/wallet/index.ts
```diff
@@ -145,15 +160,46 @@ class Wallet {
    * Signs a transaction offline.
    *
    * @param transaction - A transaction to be signed offline.
-   * @param options - Options to include for signing.
+   * @param multisignAddress - Multisign only. An account address corresponding to the multi-signature being added. If this
+   * wallet represents your [master keypair](https://xrpl.org/cryptographic-keys.html#master-key-pair) you can get your account address
+   * with the Wallet.getClassicAddress() function.
+   * @throws ValidationError if the transaction is already signed or does not encode/decode to same result.
```
mvadari, line 166: nit: @throws should be under @returns

src/wallet/index.ts
mvadari: Keep the comment and add a TODO

src/wallet/index.ts
```diff
@@ -181,12 +227,104 @@ class Wallet {
...
+   * @param tx - A transaction to sign.
+   * @param privateKey - = A key to sign the transaction with.
+   * @param signAs - Multisign only. An account address to include in the Signer field.
+   * Can be either a classic address or an XAddress.
+   * @returns A signed transaction in the proper format.
+   */
+  // eslint-disable-next-line class-methods-use-this -- Helper to add organizational clarity
+  private computeSignature(
+    tx: Transaction,
+    privateKey: string,
```
mvadari, line 317: This doesn't need to be passed in, you already have it from the `Wallet` object

test/fixtures/requests/sign.json
```diff
@@ -1,8 +1,12 @@
 {
-  "txJSON": "{\"Flags\":2147483648,\"TransactionType\":\"AccountSet\",\"Account\":\"r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59\",\"Domain\":\"6578616D706C652E636F6D\",\"LastLedgerSequence\":8820051,\"Fee\":\"12\",\"Sequence\":23,\"SigningPubKey\":\"02F89EAEC7667B30F33D0687BBA86C3FE2A08CCA40A9186C5BDE2DAA6FA97A37D8\"}",
-  "instructions": {
-    "fee": "0.000012",
-    "sequence": 23,
-    "maxLedgerVersion": 8820051
-  }
+  "txJSON": {
```
mvadari, line 2: You can take out the `txJSON` too, and bring everything up a level

test/fixtures/requests/signTicket.json
```diff
@@ -1,9 +1,14 @@
...
-  "instructions": {
+  "txJSON": {
+    "TransactionType":"TicketCreate",
+    "TicketCount":1,
+    "Account":"r4SDqUD1ZcfoZrhnsZ94XNFKxYL4oHYJyA",
+    "Sequence":0,
+    "TicketSequence":23,
+    "Fee":"10000"
+    },
+    "instructions": {
```
mvadari, line 10: instructions need to be removed here too

src/wallet/index.ts
```diff
@@ -181,12 +224,100 @@ class Wallet {
...
+   *  Decode a serialized transaction, remove the fields that are added during the signing process,
+   *  and verify that it matches the transaction prior to signing. This gives the user a sanity check
+   *  to ensure that what they try to encode matches the message that will be recieved by rippled.
+   *
+   * @param serialized - A signed and serialized transaction.
+   * @param tx - The transaction prior to signing.
+   * @throws A ValidationError if the transaction does not have a TxnSignature/Signers property, or if
+   * the serialized Transaction desn't match the original transaction.
+   */
+  // eslint-disable-next-line class-methods-use-this -- Helper for organization purposes
```
mvadari, line 247: If you don't need `this` in the method, the function should be a private method outside of the class

src/wallet/index.ts
```diff
@@ -181,12 +224,100 @@ class Wallet {
...
+  /**
+   * Signs a transaction with the proper signing encoding.
+   *
+   * @param tx - A transaction to sign.
+   * @param privateKey - = A key to sign the transaction with.
+   * @param signAs - Multisign only. An account address to include in the Signer field.
+   * Can be either a classic address or an XAddress.
+   * @returns A signed transaction in the proper format.
+   */
+  // eslint-disable-next-line class-methods-use-this -- Helper to add organizational clarity
```
mvadari, line 311: This does actually need to take `this` as a param, because it uses `this.privateKey`

test/wallet/index.ts
```diff
@@ -180,27 +186,300 @@ describe('Wallet', function () {
...
   describe('signTransaction', function () {
-    const publicKey =
-      '030E58CDD076E798C84755590AAF6237CA8FAE821070A59F648B517A30DC6F589D'
-    const privateKey =
-      '00141BA006D3363D2FB2785E8DF4E44D3A49908780CB4FB51F6D217C08C021429F'
-    const address = 'rhvh5SrgBL5V8oeV9EpDuVszeJSSCEkbPc'
+    let wallet
+    let wallet2
+    let wallet3
+    let wallet4
```
mvadari, line 192: Why do you have 4 wallets in this testing framework? And you're still using other wallets in some of the tests.

JST5000: The main reason there are 4 wallets is because the existing tests used different keys/secrets, and it's challenging to regenerate the expected result independent of our implementation, and I figured more wallets would only add to the testing coverage rather than hurting it.

src/wallet/index.ts
JST5000: Is it cleaner to accept 'this' explicitly as a parameter instead of just using it implicitly?

test/wallet/index.ts
mvadari: I think it makes these tests a little messy. 

It should be easy enough to generate expected results via local rippled/a different library. Since these tests are currently passing, you know that the code isn't broken atm, so you could also just run the test locally with a different wallet and use that result in the expected result (not optimal, but fine in this case).

src/wallet/index.ts
```diff
@@ -181,13 +230,104 @@ class Wallet {
...
+      throw error
+    }
+  }
+}
+
+/**
+ * Signs a transaction with the proper signing encoding.
+ *
+ * @param tx - A transaction to sign.
+ * @param privateKey - = A key to sign the transaction with.
```
mvadari, line 313: nit:
```suggestion
 * @param privateKey - A key to sign the transaction with.
```

test/wallet/index.ts
```diff
@@ -181,27 +186,263 @@ describe('Wallet', function () {
   })
 
   describe('signTransaction', function () {
-    const publicKey =
-      '030E58CDD076E798C84755590AAF6237CA8FAE821070A59F648B517A30DC6F589D'
-    const privateKey =
-      '00141BA006D3363D2FB2785E8DF4E44D3A49908780CB4FB51F6D217C08C021429F'
-    const address = 'rhvh5SrgBL5V8oeV9EpDuVszeJSSCEkbPc'
+    let wallet
```
mvadari, line 189:
```suggestion
    let wallet: Wallet
```
helps with typing

mvadari (approved):

PR #1622, opened 2021-09-15
fix: remove the sugar/parse directory

natenichols:
Removes old code from the `sugar/parse` directory.


mvadari (commented):
will review when tests are passing

src/sugar/orderbook.ts
```diff
@@ -81,7 +84,7 @@ async function getOrderbook(
   const buy: BookOffer[] = []
   const sell: BookOffer[] = []
   orders.forEach((order) => {
-    if (order.Flags === OfferCreateTransactionFlags.tfSell) {
+    if (order.Flags === SELL_OFFER_FLAGS) {
```
mvadari, line 87: This shouldn't be `===` - that doesn't account for the possibility that multiple flags are turned on

mvadari (approved):

src/sugar/orderbook.ts
```diff
@@ -9,7 +9,7 @@ import {
   TakerAmount,
 } from '../models/methods/bookOffers'
 
-import { orderFlags } from './parse/flags'
+const SELL_OFFER_FLAGS = 0x00020000
```
mvadari, line 12: Is it not possible to get this number from one of the `Flags` things that we already have? Seems bad to have it as a separate constant.

natenichols: Right, so these flags are `lsf` flags. The flags we have defined are `tf` flags. 

So,  we can define these flags in the `./src/models/ledger` directory with the associated ledger entry.

mvadari: I'd just add a TODO for now then - not worth doing in this PR

src/models/ledger/offer.ts
```diff
@@ -16,3 +16,8 @@ export default interface Offer extends BaseLedgerEntry {
   PreviousTxnLgrSeq: number
   Expiration?: number
 }
+
+export enum OfferLedgerFlags {
+  lsfPassive = 0x00010000,
+  lsfSell = 0x00020000, // True, offer was placed as a sell.
```
mvadari, line 22: need the comments above the line for the linter to be happy

JST5000 (approved):
LGTM!

src/sugar/orderbook.ts
natenichols: Just added it.

PR #1623, opened 2021-09-15
Resolves TODOs in `./src/models`

natenichols:
Resolves TODO's in `./src/models`.

### Type of Change
- [ ] Documentation Updates
- [ ] Release


src/models/methods/ping.ts
```diff
@@ -1,11 +1,9 @@
...
 
 export interface PingRequest extends BaseRequest {
   command: 'ping'
 }
 
 export interface PingResponse extends BaseResponse {
-  // TODO: figure out if there's a better way to type this
-  // eslint-disable-next-line @typescript-eslint/ban-types -- actually should be an empty object
-  result: {}
+  result: Record<string, never>
```
mvadari, line 8: the `PingResponse` does sometimes have things in its results - see the integration tests

src/models/transactions/common.ts
```diff
@@ -111,8 +111,7 @@ export interface BaseTransaction {
   AccountTxnID?: string
   Flags?: number | GlobalFlags
   LastLedgerSequence?: number
-  // TODO: Make Memo match the format of Signer (By including the Memo: wrapper inside the Interface)
-  Memos?: Array<{ Memo: Memo }>
+  Memos?: Memo[]
```
mvadari, line 114: the `verifyMemo` needs to be fixed as well

src/sugar/parse/ledger.ts
```diff
@@ -6,7 +6,7 @@ import { removeUndefined, rippleTimeToISOTime } from '../../utils'
 import parseTransaction from './transaction'
 
 export interface FormattedLedger {
-  // TODO: properties in type don't match response object. Fix!
```
mvadari, line 9: Why are we worrying about TODOs in files that are about to be deleted?

src/utils/generateAddress.ts
```diff
@@ -27,8 +27,7 @@ export interface GenerateAddressOptions {
 }
 
 /**
- * TODO: Move this function to be a static function of the Wallet Class.
```
mvadari, line 30: Is this going to be moved to the `Wallet` class?

src/wallet/signer.ts
```diff
@@ -12,7 +12,7 @@ import { sign as signWithKeypair, verify } from 'ripple-keypairs'
 import { ValidationError } from '../common/errors'
 import { Signer } from '../models/common'
 import { Transaction } from '../models/transactions'
-import { verifyBaseTransaction } from '../models/transactions/common'
+import { verify as verifyTransaction } from '../models/transactions'
```
mvadari, line 15: Why not just leave it as `verify`?

test/client/getOrderbook.ts
```diff
@@ -66,19 +67,9 @@ function normalRippledResponse(request: BookOffersRequest): object {
   ) {
     return rippled.book_offers.fabric.requestBookOffersAsksResponse(request)
   }
-  throw new Error('unexpected end')
+  throw new XrplError('unexpected end')
 }
 
-// function xrpRippledResponse(request: BookOffersRequest): object {
```
mvadari, line 72: This shouldn't be removed - it'll likely be useful when the `getOrderbook` tests are rewritten

test/integration/requests/utility.ts
```diff
@@ -19,7 +20,7 @@ describe('Utility method integration tests', function () {
     const response = await (this.client as Client).request({
       command: 'ping',
     })
-    const expected = {
+    const expected: unknown = {
```
mvadari, line 23: why does this need to be typed?

src/models/methods/baseMethod.ts
```diff
@@ -1,3 +1,5 @@
+import { Request } from '.'
```
mvadari, line 1:
```suggestion
import type { Request } from '.'
```
Otherwise there's a dependency cycle

src/models/transactions/common.ts
natenichols: It should work as-is. I think we were originally just passing in `tx.Memos` into `isMemo`

src/sugar/parse/ledger.ts
natenichols: Wasn't, just removed the `TODO` part and made a note that this will be removed.

test/integration/requests/utility.ts
natenichols: The `assert.deepEqual` is expecting the types to be the same, unless `unknown` or `any`

src/models/transactions/common.ts
mvadari: We had to modify the `Signer` one, so I suspect we'll have to modify the `Memo` one too. Can you add a test just to double check?

test/client.ts
```diff
@@ -34,8 +34,12 @@ describe('Client', function () {
     assert.throws(() => new Client('wss//s:1'))
   })
 
-  it('Client connect() times out after 2 seconds', function () {
-    // TODO: Use a timer mock like https://jestjs.io/docs/en/timer-mocks
-    //       to test that connect() times out after 2 seconds.
+  it('Client connect() times out after 2 seconds', async function () {
+    const client = new Client('wss://ObviouslyNotARealDomain.com')
```
mvadari, line 38: I think this is supposed to be checking that it actually times out in 2 seconds and not some other interval.

src/wallet/signer.ts
natenichols: conflict w/ `ripple-keypairs.verify(....)`.

src/models/methods/ping.ts
```diff
@@ -1,11 +1,9 @@
...
 
 export interface PingRequest extends BaseRequest {
   command: 'ping'
 }
 
 export interface PingResponse extends BaseResponse {
-  // TODO: figure out if there's a better way to type this
-  // eslint-disable-next-line @typescript-eslint/ban-types -- actually should be an empty object
-  result: {}
+  result: { role: string; unlimited: boolean }
```
mvadari, line 8:
```suggestion
  result: { role?: string; unlimited?: boolean }
```
I think it only has those fields when it's rippled in standalone mode, because normal rippled doesn't have anything: https://xrpl.org/websocket-api-tool.html#ping

src/wallet/signer.ts
mvadari: Is it worth renaming the function itself?

src/models/transactions/common.ts
```diff
@@ -29,7 +29,13 @@ const transactionTypes = [
...
 const MEMO_SIZE = 3
 
-function isMemo(obj: { Memo?: unknown }): boolean {
+/**
+ * Exported for testing only.
+ *
+ * @param obj - An unknown object.
+ * @returns Boolean if the object is a memo.
+ */
+export function isMemo(obj: { Memo?: unknown }): boolean {
```
mvadari, line 38: this can be tested via `verifyBaseTransaction` (just add some memos there), so you don't need to export it.

src/models/methods/subscribe.ts
```diff
@@ -25,9 +25,7 @@ export interface SubscribeRequest extends BaseRequest {
 }
 
 export interface SubscribeResponse extends BaseResponse {
-  // TODO: figure out if there's a better way to type this
-  // eslint-disable-next-line @typescript-eslint/ban-types -- actually should be an empty object
-  result: {} | Stream
+  result: Record<string, never> | Stream
```
JST5000, line 28: Just to make sure this doesn't get lost - This field should also be updated with whatever you decide in response to Mayukha's comment above

src/sugar/index.ts
```diff
@@ -18,7 +18,7 @@ export function ensureClassicAddress(account: string): string {
     // must not have an embedded tag. In other words,
     // `tag` should be `false`.
     if (tag !== false) {
-      throw new Error(
+      throw new errors.XrplError(
```
JST5000, line 21: Why is this `error.XrplError` instead of just `XrplError`?

test/models/payment.ts
```diff
@@ -45,6 +47,16 @@ describe('Payment', function () {
...
   })
 
+  it(`Verifies memos correctly`, function () {
+    const memos: Memo = {
+      Memo: {
+        MemoData: '32324324',
+      },
+    }
+
+    assert.doesNotThrow(() => isMemo(memos))
```
JST5000, line 57: Can you also check the opposite (I assume it throws if isMemo is false?)

JST5000 (changes requested):
Just one small test case I think should be added

test/mockRippled.ts
```diff
@@ -60,10 +61,10 @@ export default function createMockRippled(port: number): MockedWebSocketServer {
       try {
         request = JSON.parse(requestJSON)
         if (request.id == null) {
-          throw new Error(`Request has no id: ${requestJSON}`)
+          throw new XrplError(`Request has no id: ${requestJSON}`)
         }
         if (request.command == null) {
-          throw new Error(`Request has no id: ${requestJSON}`)
+          throw new XrplError(`Request has no id: ${requestJSON}`)
```
mukulljangid, line 67: It should be `Request has no command: `

src/sugar/index.ts
natenichols: Because we are we exporting errors from this file as well.

mvadari: We shouldn't be exporting errors here too

test/mockRippled.ts
```diff
@@ -3,6 +3,7 @@ import _ from 'lodash'
 import { Server as WebSocketServer } from 'ws'
 
 import type { Request } from '../src'
+import { XrplError } from '../src/errors'
```
JST5000, line 6: This import should probably from 'xrpl-local/errors' I believe

test/models/accountDelete.ts
```diff
@@ -1,6 +1,8 @@
 import { assert } from 'chai'
 
-import { validateAccountDelete, validate, ValidationError } from 'xrpl-local'
+import { validate, ValidationError } from 'xrpl-local'
+
+import { validateAccountDelete } from '../../src/models/transactions/accountDelete'
```
JST5000, line 5: ditto on xrpl-local import

test/models/accountSet.ts
```diff
@@ -1,6 +1,8 @@
 import { assert } from 'chai'
 
-import { validateAccountSet, validate, ValidationError } from 'xrpl-local'
+import { validate, ValidationError } from 'xrpl-local'
+
+import { validateAccountSet } from '../../src/models/transactions/accountSet'
```
JST5000, line 5: ditto

JST5000: (Apply the pattern to all files in test with a relative path import to `src`)

JST5000 (changes requested):
I think there's one TODO left (which was hiding in the line underneath a TODO in signer)

src/utils/generateAddress.ts
mvadari: bump on this - there was a TODO for it to be moved but it wasn't

src/utils/hashes/ledgerHash.ts
```diff
@@ -76,7 +76,6 @@ export function computeSignedTransactionHash(tx: Transaction | string): string {
   let txObject: Transaction
   if (typeof tx === 'string') {
     txBlob = tx
-    // TODO: type ripple-binary-codec with Transaction
```
mvadari, line 79: is there a reason we aren't leaving this TODO in here? Or at least creating an issue in r-b-c

PR #1624, opened 2021-09-15
refactor: renames `verify...` methods to `validate...`

mvadari:

This PR renames all the transaction model `verify...` methods to `validate...`. It also exports the transaction models at the top level, and fixes all corresponding imports. 

### Context of Change

There was a namespace collision at `verifyPaymentChannelClaim` - one method actually verifies the cryptographic claim, and the other is a JS object-validation-checker.

### Type of Change

## Test Plan

CI passes.


src/index.ts
```diff
@@ -9,7 +9,7 @@ export * from './common/types/objects/ledger'
 
 export * from './models/methods'
 
-export * from './models/methods'
+export * from './models/transactions'
```
mvadari, line 12: This line was originally a copy of line 10

JST5000 (approved):
LGTM!

natenichols (approved):

PR #1625, opened 2021-09-15
refactor: remove unused methods/types in `src/sugar`

mvadari:

Title says it all.

### Context of Change

xrpl.js cleanup

### Type of Change

## Test Plan

CI passes.


mvadari:
waiting on #1622

src/sugar/utils.ts
```diff
@@ -1,125 +1,27 @@
...
+ * @returns The account's classic address.
+ * @throws Error if the X-Address has an associated tag.
+ */
+export function ensureClassicAddress(account: string): string {
+  if (isValidXAddress(account)) {
+    const { classicAddress, tag } = xAddressToClassicAddress(account)
+
+    // Except for special cases, X-addresses used for requests
+    // must not have an embedded tag. In other words,
+    // `tag` should be `false`.
```
JST5000, line 16: Do those special cases use this function, or do they handle it separately?

JST5000 (approved):
LGTM (One question, but it's for my curiousity)

src/sugar/utils.ts
mvadari: Pretty sure that's talking about transactions, which do handle X-Addresses separately (see `autofill`).

natenichols: Should this function go to `./src/utils` or something. isn't it only uses in `Client.request`?

src/sugar/index.ts
```diff
@@ -1,33 +0,0 @@
-import { xAddressToClassicAddress, isValidXAddress } from 'ripple-address-codec'
```
natenichols, line 1: Why did we delete `sugar/index.ts` How are we exporting from sugar?

natenichols (approved):
One question, but its not directly relevant to the PR, so I'm okay w/ giving approval

PR #1626, opened 2021-09-15
feat: extra protection for AccountDelete transactions

### LINKED ISSUE #1117: Decide how to handle the AccountDelete transaction fee
The `AccountDelete` Transaction has a transaction cost of 5,000,000 drops (before load scaling), per https://xrpl.org/transaction-cost.html#special-transaction-costs

This bumps into the 2,000,000 drop fee limit imposed on all transactions by default, so `AccountDelete` can't be used unless `maxFeeXRP` is set to '5' or higher. https://xrpl.org/rippleapi-reference.html

Is this acceptable, or do we want to make a special case for `AccountDelete`?

### LINKED ISSUE #1406: Check conditions prior AccountDelete call
One of the conditions for AccountDelete to succeed, is that the account must not be linked to any of the following types of ledger objects (as a sender or receiver): Escrow, PayChannel, RippleState, Check.

Is there an easy call to check this prior to executing a submit action? When the submit action fails because of non-eligibility, there is still 5 XRP deducted..

Thanks!

mvadari:

This PR adds extra checks for AccountDelete transactions, since there is a high transaction cost associated with them. It adds a deletion blockers check to autofill (there's already a fee exception for `maxFeeXRP`).

### Context of Change

Fixes #1406. Fixes #1117.

AccountDelete transactions have a high transaction cost associated with them (5 XRP), compared to normal transactions (10 drops). We want to add extra protections to these transactions, so people don't accidentally submit one of these transactions, have it fail, and lose the fee.

### Type of Change

## Test Plan

CI passes.


src/sugar/autofill.ts
```diff
@@ -193,4 +200,19 @@ async function setLatestValidatedLedgerSequence(
...
+  const request: AccountObjectsRequest = {
+    command: 'account_objects',
+    account: tx.Account,
+    ledger_index: 'validated',
+    deletion_blockers_only: true,
+  }
+  const response = await client.request(request)
+  if (response.result.account_objects.length > 0) {
+    throw new XrplError(
+      `Account ${tx.Account} cannot be deleted; there are Escrows, PayChannels, RippleStates, or Checks associated with the account.`,
```
intelliot, line 213: Could we also include the verbatim `account_objects` response so clients can see what the deletion blockers are?

src/sugar/submit.ts
```diff
@@ -54,6 +54,7 @@ async function submitSignedTransaction(
   const request: SubmitRequest = {
     command: 'submit',
     tx_blob: signedTxEncoded,
+    fail_hard: true,
```
intelliot, line 57: I think this is good default behavior. There are some cases where users may not want to use fail_hard. In that case, would we tell them to do a `submit` request directly? Or, could we offer this as an option?

mvadari: We could add it as an optional param that defaults to false. But I'm inclined to say that if users know enough about rippled to know that they don't want to `fail_hard`, then it's not hard to call autofill/sign yourself and then directly `submit`. Thoughts?

ximinez: I'm not familiar with this code, or the flow of logic, but one significant downside of using `fail_hard` is that it's a blunt tool, and prevents the transaction from processing in the case of "benign" failures - most notably if the transaction would have been queued. Getting queued usually means that the transaction will succeed, just not right away. In most cases, AFAIK, that's going to meet the user's needs.

All that to say, my vote is to only pass `fail_hard: true` only for `AccountDelete` transactions, or if the user specifies it explicitly.

src/sugar/submit.ts
```diff
@@ -54,6 +54,7 @@ async function submitSignedTransaction(
   const request: SubmitRequest = {
     command: 'submit',
     tx_blob: signedTxEncoded,
+    fail_hard: isAccountDelete(signedTransaction),
```
JST5000, line 57: With this setup, what's the workflow for users who want to `fail_hard`, but aren't submitting an AccountDelete request?

src/sugar/autofill.ts
```diff
@@ -46,6 +50,9 @@ async function autofill<T extends Transaction>(
   if (tx.LastLedgerSequence == null) {
     promises.push(setLatestValidatedLedgerSequence(client, tx))
   }
+  if (tx.TransactionType === 'AccountDelete') {
+    promises.push(checkAccountDeleteBlockers(client, tx as AccountDelete))
```
natenichols, line 54: Do we need this `as AccountDelete` there?

src/sugar/autofill.ts
```diff
@@ -193,4 +200,20 @@ async function setLatestValidatedLedgerSequence(
...
+async function checkAccountDeleteBlockers(client: Client, tx: AccountDelete) {
+  const request: AccountObjectsRequest = {
+    command: 'account_objects',
+    account: tx.Account,
+    ledger_index: 'validated',
+    deletion_blockers_only: true,
+  }
+  const response = await client.request(request)
+  if (response.result.account_objects.length > 0) {
+    throw new XrplError(
```
natenichols, line 212: I dont think throwing here will give us the right error handling flow. 

I think this will just throw and kill the event loop.

What I think we'll wanna do is to make this a promise and return a `Promise.reject()` if this fails

natenichols:
I would wager that this won't work. Throwing from a pending promise will just kill the event loop. 
Sample code 1: This is what I think is happening in this PR.
```js
function MakeProm() {
	return new Promise((res, rej) => {
		setTimeout(() => res(), 5000)
	})
}

function throwProm() {
	return new Promise((res, rej) => {
		setTimeout(() => {
			throw new Error("cannot resolve")
		}, 2000)
	})
}

async function doStuff() {
	const promises = []

	promises.push(MakeProm())
	promises.push(MakeProm())
	promises.push(throwProm())

	await Promise.all(promises).catch(() => {
		console.log("I CAUGHT IT")
	})
	console.log("DONE")
}

doStuff()
// logs
// Error: cannot resolve
//  at Timeout._onTimeout (/home/runner/s49zvzizjxc/index.js:10:10)
//  at listOnTimeout (internal/timers.js:554:17)
//  at processTimers (internal/timers.js:497:7)
```

versus sample code 2:
```js
function MakeProm() {
	return new Promise((res, rej) => {
		setTimeout(() => res(), 5000)
	})
}

function throwProm() {
	return new Promise((res, rej) => {
		setTimeout(() => {
			rej(new Error("cannot resolve")) /// Rejects the promise w/ an error
		}, 2000)
	})
}

async function doStuff() {
	const promises = []

	promises.push(MakeProm())
	promises.push(MakeProm())
	promises.push(throwProm())

	await Promise.all(promises).catch((err) => {
		console.log(`I CAUGHT IT ${err.message}`)
	})
	console.log("DONE")
}

doStuff()
// console.logs
// I CAUGHT IT cannot resolve
// DONE
```

natenichols (changes requested):
It made me leave a comment. See the code snippets right above this...

src/sugar/submit.ts
mvadari: They can write a submit `request` themselves

src/sugar/autofill.ts
mvadari: I typed the method as requiring an `AccountDelete`, but I could type it as just `Transaction` and then the casting wouldn't be needed

natenichols (approved):

JST5000 (approved):
LGTM! Moving the response out makes it a lot easier to read :)

PR #1627, opened 2021-09-16
refactor: move address-codec methods from client to utils

mvadari:

The ripple-address-codec methods don't need to be exported as static methods on the client - they don't require a network connection, and they really are utils.

Happy to also delete them if you'd rather that they export them from ripple-address-codec.

### Context of Change

xrpl.js cleanup

### Type of Change

## Test Plan

CI passes.


JST5000 (approved):
LGTM!

natenichols (approved):

PR #1628, opened 2021-09-16
fix: disallow two pending requests with the same id

mvadari:

There was a bug in `RequestManager` where if someone hand-inputted an ID into their request, they could have 2 requests with the same IDs. This PR fixes that (and improves handling of string request IDs).

### Context of Change

@ledhed2222 pointed out a bug

### Type of Change

## Test Plan

CI passes.

### DIFF
--- a/src/client/requestManager.ts
+++ b/src/client/requestManager.ts
@@ -138,7 +138,10 @@ export default class RequestManager {
    * @throws ResponseFormatError if the response format is invalid, RippledError if rippled returns an error.
    */
   public handleResponse(response: Partial<Response | ErrorResponse>): void {
-    if (response.id == null) {
+    if (
+      response.id == null ||
+      !(typeof response.id === 'string' || typeof response.id === 'number')
+    ) {
       throw new ResponseFormatError('valid id not found in response', response)
     }
     if (!this.promisesAwaitingResponse.has(response.id)) {
--- a/src/client/requestManager.ts
+++ b/src/client/requestManager.ts
@@ -1,4 +1,9 @@
-import { ResponseFormatError, RippledError, TimeoutError } from '../errors'
+import {
+  ResponseFormatError,
+  RippledError,
+  TimeoutError,
+  XrplError,
+} from '../errors'
 import { Response } from '../models/methods'
 import { BaseRequest, ErrorResponse } from '../models/methods/baseMethod'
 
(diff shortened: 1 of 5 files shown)


src/client/requestManager.ts
```diff
@@ -110,6 +118,9 @@ export default class RequestManager {
     if (timer.unref) {
       timer.unref()
     }
+    if (this.promisesAwaitingResponse.has(newId)) {
```
JST5000, line 121: Do we care about the case where someone passes a number as a string? Ex. Id = '1' and a pending id = 1 (number), should we count those as the same id? (From a quick test online, map.has() seems to not treat '1' = 1)

JST5000 (commented):
Not sure what the expected result is in my comment above

test/connection.ts
```diff
@@ -619,4 +619,13 @@ describe('Connection', function () {
       .then(() => new Error('Should not have succeeded'))
       .catch(done())
   })
+
+  it('should throw error if pending response with same ID', async function () {
+    await this.client.connection.request({ id: 'test', command: 'ping' })
```
natenichols, line 624: `await`-ing this should remove that request from the `pending` responses, right?

You should be able to make as many promises w/ the same `id` as you want, as long as you await them one at a time. 

IMO this should only throw if both promises are `pending` AND have the same `id`.

natenichols (changes requested):
Question before I re-review. Are we removing promises from the `requestManager` as they resolve? Can we consecutively have the same `id` in promises as long as they are `awaited`?

mvadari:
> Question before I re-review. Are we removing promises from the `requestManager` as they resolve? Can we consecutively have the same `id` in promises as long as they are `awaited`?

1. Yes - that's what `this.deletePromise` does.
2. Looks like no, because that's exactly the scenario that the testcase I wrote does

src/client/requestManager.ts
mvadari: We'd want those to be treated differently, because they're different objects - and in the current implementation they would be, assuming rippled treats them differently (@natenichols do you know if this is true?)

mvadari:
Turns out there was a bug in `this.deletePromise` _and_ in the test - both have been resolved now

natenichols (approved):

JST5000 (approved):
LGTM!

PR #1629, opened 2021-09-16
refactor: move everything out of the common folder

mvadari:

This PR:
* deletes `src/common/constants.ts` (all in `src/models` now)
* deletes `src/common/txFlags.ts` (all in `src/models` now)
* moves `src/common/ecdsa.ts` up to the `src` level
* moves `src/common/errors.ts` up to the `src` level, and exports them at the top

### Context of Change

We don't really need a `common` folder if there's only going to be 2 files that should be in there - they can also fit at the `src` level. 

### Type of Change

## Test Plan

CI passes.


JST5000 (commented):
Should ecdsa.ts be moved into src/wallet since it's only used by functions which are (or will be moved into) in the Wallet folder?

mvadari:
It's at the top level in xrpl-py, so I'm inclined to leave it where it is. If anyone else has opinions on this lmk

JST5000 (approved):
LGTM!

src/index.ts
```diff
@@ -13,4 +13,6 @@ export * from './models/transactions'
 
 export * from './utils'
 
+export * from './errors'
```
natenichols, line 16: Do we want to be exporting errors?

natenichols (approved):
I have one question. I don't really see the need to export `errors`, but also this is not a hill I'm willing to die on.

src/index.ts
mvadari: `catch` blocks or to check the error type

mvadari: I'm happy to take it out, but we can discuss this more in my PR for exporting things at the top level (which I'll put up by tomorrow)

PR #1630, opened 2021-09-16
Remove old types

JST5000:

Deleted all the old types and resolved any dependencies.

### Context of Change

We no longer need them since we've got the 2.0 models to replace them.

### Type of Change

## Before / After

* Removed the old types and commands
* ~~For files that are removed in other PR's I just added the types they needed locally (That way deleting the file in the merge will resolve everything, and the build won't break if we merge just these changes) - Specifically applies to sign.ts and sugar/utils~~ 
   * Updated to just to replace the types with any/inferred types in those files instead based on mvadari's comment
* Fixed a model which accidentally was using the old OfferCreateTransaction instead of the current OfferCreate model

## Test Plan

CI Passes


mvadari (commented):
Your work depends on two different PRs (#1625 and #1620), which is kind of complicated to deal with. Honestly, I'd just ignore the types/not worry about the errors, and wait until those PRs are merged, at which point the errors will automatically go away. That's easier than needing to remember to remove types later.

mvadari (approved):
well that's a better solution than just ignoring the errors - nice! 

Actually yeah you should just remove `toRippledAmount`, it's not used anywhere

mvadari:
I'd also still wait for the other PRs to be merged before merging this one.

natenichols (approved):

PR #1631, opened 2021-09-16
test: write tests for sugar functions

mukulljangid:
Write Tests for Sugar Functions

### Context of Change
Tests according to the new structure of the sugar functions.

### Type of Change

## Test Plan
Passes all CI tests.

### DIFF
--- a/src/sugar/balances.ts
+++ b/src/sugar/balances.ts
@@ -41,16 +41,19 @@ async function getBalances(
   options: GetBalancesOptions = {},
 ): Promise<Balance[]> {
   // 1. Get XRP Balance
-  const xrpRequest: AccountInfoRequest = {
-    command: 'account_info',
-    account,
-    ledger_index: options.ledger_index ?? 'validated',
-    ledger_hash: options.ledger_hash,
+  const xrpBalance: Balance[] = []
+  if (!options.peer) {
+    const xrpRequest: AccountInfoRequest = {
+      command: 'account_info',
+      account,
+      ledger_index: options.ledger_index ?? 'validated',
+      ledger_hash: options.ledger_hash,
+    }
+    const balance = await this.request(xrpRequest).then(
+      (response) => response.result.account_data.Balance,
+    )
+    xrpBalance.push({ currency: 'XRP', value: dropsToXrp(balance) })
   }
-  const balance = await this.request(xrpRequest).then(
-    (response) => response.result.account_data.Balance,
-  )
-  const xrpBalance = { currency: 'XRP', value: dropsToXrp(balance) }
   // 2. Get Non-XRP Balance
   const linesRequest: AccountLinesRequest = {
     command: 'account_lines',
@@ -64,7 +67,7 @@ async function getBalances(
   const accountLinesBalance = _.flatMap(responses, (response) =>
     formatBalances(response.result.lines),
   )
-  return [xrpBalance, ...accountLinesBalance]
+  return [...xrpBalance, ...accountLinesBalance].slice(0, options.limit)
 }
 
 export default getBalances
--- a/src/sugar/orderbook.ts
+++ b/src/sugar/orderbook.ts
@@ -54,7 +54,7 @@ async function getOrderbook(
     command: 'book_offers',
     taker_pays: takerPays,
     taker_gets: takerGets,
-    ledger_index: options.ledger_index ?? 'validated',
+    ledger_index: options.ledger_index,
     ledger_hash: options.ledger_hash,
     limit: options.limit ?? DEFAULT_LIMIT,
     taker: options.taker,
--- a/src/sugar/orderbook.ts
+++ b/src/sugar/orderbook.ts
@@ -58,7 +58,6 @@ async function getOrderbook(
     limit: options.limit ?? DEFAULT_LIMIT,
     taker: options.taker,
   }
-  console.log(request)
   // 2. Make Request
   const directOfferResults = await this.requestAll(request)
   request.taker_gets = takerPays
@@ -74,12 +73,8 @@ async function getOrderbook(
     (reverseOfferResult) => reverseOfferResult.result.offers,
   )
 
-  // Sort the orders
-  // for both buys and sells, lowest quality is closest to mid-market
-  // we sort the orders so that earlier orders are closer to mid-market
-
   const orders = [...directOffers, ...reverseOffers]
-  // separate out the orders amongst buy and sell
+  // separate out the buy and sell orders
   const buy: BookOffer[] = []
   const sell: BookOffer[] = []
   orders.forEach((order) => {
@@ -90,6 +85,9 @@ async function getOrderbook(
       sell.push(order)
     }
   })
+  // Sort the orders
+  // for both buys and sells, lowest quality is closest to mid-market
+  // we sort the orders so that earlier orders are closer to mid-market
   return { buy: sortOffers(buy), sell: sortOffers(sell) }
 }
 
--- a/src/sugar/orderbook.ts
+++ b/src/sugar/orderbook.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines-per-function -- Needs to process orderbooks. */
 import BigNumber from 'bignumber.js'
 import _ from 'lodash'
 
@@ -88,7 +89,10 @@ async function getOrderbook(
   // Sort the orders
   // for both buys and sells, lowest quality is closest to mid-market
   // we sort the orders so that earlier orders are closer to mid-market
-  return { buy: sortOffers(buy), sell: sortOffers(sell) }
+  return {
+    buy: sortOffers(buy).slice(0, options.limit),
+    sell: sortOffers(sell).slice(0, options.limit),
+  }
 }
 
 export default getOrderbook
--- a/src/sugar/orderbook.ts
+++ b/src/sugar/orderbook.ts
@@ -58,6 +58,7 @@ async function getOrderbook(
     limit: options.limit ?? DEFAULT_LIMIT,
     taker: options.taker,
   }
+  console.log(request)
   // 2. Make Request
   const directOfferResults = await this.requestAll(request)
   request.taker_gets = takerPays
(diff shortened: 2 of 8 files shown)


src/sugar/orderbook.ts
```diff
@@ -74,21 +72,21 @@ async function getOrderbook(
...
-
   const orders = [...directOffers, ...reverseOffers]
-  // separate out the orders amongst buy and sell
+  // separate out the buy and sell orders
   const buy: BookOffer[] = []
   const sell: BookOffer[] = []
   orders.forEach((order) => {
-    if ((order.Flags & OfferLedgerFlags.lsfSell) !== 0) {
+    // eslint-disable-next-line no-bitwise -- Bitwise AND
+    if (order.Flags & OfferLedgerFlags.lsfSell) {
```
mvadari, line 81: It's good practice to explicitly state the condition
```suggestion
    if (order.Flags & OfferLedgerFlags.lsfSell !== 0) {
```

test/client/getOrderbook.ts
```diff
@@ -72,142 +55,118 @@ function normalRippledResponse(
...
-//   }
-//   if (request.taker_gets.issuer === 'rp8rJYTpodf8qbSCHVTNacf8nSW8mRakFw') {
-//     return rippled.book_offers.usd_xrp
-//   }
-//   throw new Error('unexpected end')
-// }
+function xrpRippledResponse(
+  request: BookOffersRequest,
+): Record<string, unknown> {
+  if (request.taker_pays.issuer === 'rp8rJYTpodf8qbSCHVTNacf8nSW8mRakFw') {
```
mvadari, line 61: Shouldn't this be `takerPays`?

test/fixtures/requests/getOrderbookWithXrp.json
```diff
@@ -1,9 +1,9 @@
 {
-  "base": {
+  "taker_pays": {
```
mvadari, line 2:
```suggestion
  "takerPays": {
```
We should be consistent everywhere

test/client/getOrderbook.ts
mukulljangid: No. The input is the well formed request and that has parameters as the rippled ones!

src/sugar/orderbook.ts
mukulljangid: Linter finds it unnecessary!

test/client/getOrderbook.ts
```diff
@@ -72,142 +55,118 @@ function normalRippledResponse(
...
-        const response = await this.client.getOrderbook(
-          requests.getOrderbook.normal.taker_pays,
-          requests.getOrderbook.normal.taker_gets,
-          { limit: 1 },
-        )
-        assert.deepEqual(response, responses.getOrderbook.normal)
-      })
+  it('normal', async function () {
+    this.mockRippled.addResponse('book_offers', normalRippledResponse)
+    const response = await this.client.getOrderbook(
```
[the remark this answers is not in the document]
mukulljangid: Added!

src/sugar/orderbook.ts
mvadari: As in the linter has a warning when you add that?

test/client/getOrderbook.ts
mvadari: It should be corrected then, so that both aligned

src/sugar/orderbook.ts
mukulljangid: Yes

test/client/getBalances.ts
```diff
@@ -29,86 +29,72 @@ describe('getBalances', function () {
...
+        this.mockRippled.addResponse(
+          'account_info',
+          rippled.account_info.normal,
+        )
+        this.mockRippled.addResponse(
+          'account_lines',
+          rippledAccountLines.normal,
+        )
+        this.mockRippled.addResponse('ledger', rippled.ledger.normal)
+        const expectedResponse = responses.getBalances.slice(0, 10)
```
JST5000, line 48: Could you extract the number 10 in this slice into a variable so that you're showing the limit in `options` above should match the length of this expected result?

JST5000: Actually, you can use the options object itself here :) - 

`const expectedResponse = responses.getBalances.slice(0, request.options.limit)`

test/client/getBalances.ts
```diff
@@ -29,86 +29,72 @@ describe('getBalances', function () {
...
-      //     this.mockRippled.addResponse("ledger", rippled.ledger.normal);
-      //     const expectedResponse = responses.getBalances
-      //       .filter((item) => item.currency === "USD")
-      //       .slice(0, 3);
-      //     const result = await this.client.getBalances(test.address, options);
-      //     console.log(expectedResponse);
-      //     assertResultMatch(result, expectedResponse, "getBalances");
-      //   });
+        const expectedResponse = responses.getBalances.filter(
+          (item) => item.issuer === 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B',
```
JST5000, line 71: Same idea here, you should probably use the option variable instead of re-adding this constant.

`(item) => item.issuer === options.peer,`

test/client/getBalances.ts
```diff
@@ -29,86 +29,72 @@ describe('getBalances', function () {
...
-      //           item.currency === "USD" &&
-      //           item.issuer === "rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B"
-      //       )
-      //       .slice(0, 3);
-      //     const result = await this.client.getBalances(test.address, options);
-      //     console.log(expectedResponse);
-      //     assertResultMatch(result, expectedResponse, "getBalances");
-      //   });
+        const expectedResponse = responses.getBalances
+          .filter((item) => item.issuer === 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B')
```
JST5000, line 93: (Ditto for the rest like this)

test/client/getBalances.ts
```diff
@@ -29,86 +29,72 @@ describe('getBalances', function () {
...
+          .slice(0, 10)
```
JST5000, line 94: Ditto for both of these, and the rest that follow the pattern

test/client/getOrderbook.ts
```diff
@@ -1,44 +1,27 @@
...
+  for (const order of orders) {
+    assert(
+      new BigNumber(order.quality).isGreaterThanOrEqualTo(previousRate),
+      `Rates must be sorted from least to greatest: ${
+        order.quality as number
+      } should be >= ${previousRate}`,
+    )
+    previousRate = order.quality
+  }
+  return true
```
JST5000, line 23: Is this return used? (Since this only returns if it doesn't throw an error, it seems like you could get the same effect without returning anything)

test/client/getOrderbook.ts
```diff
@@ -72,142 +55,136 @@ function normalRippledResponse(
...
+    const response = await this.client.getOrderbook(
+      requests.getOrderbook.normal.takerPays,
+      requests.getOrderbook.normal.takerGets,
+      {
+        limit: 1,
+      },
+    )
+    const expectedResponse = {
+      buy: responses.getOrderbook.normal.buy.slice(0, 1),
+      sell: responses.getOrderbook.normal.sell.slice(0, 1),
```
JST5000, line 85: Flagging this as potentially sneaky spot where it would be good to use the limit defined above instead of new constants.

test/client/getOrderbook.ts
```diff
@@ -72,142 +55,136 @@ function normalRippledResponse(
...
+    this.mockRippled.addResponse('book_offers', xrpRippledResponse)
+    const response = await this.client.getOrderbook(
+      requests.getOrderbook.withXRP.takerPays,
+      requests.getOrderbook.withXRP.takerGets,
+    )
+    checkSortingOfOrders(response.buy)
+    checkSortingOfOrders(response.sell)
+  })
+
+  it('sorted so that best deals come first [bad test]', async function () {
```
JST5000, line 123: nit: "bad test" might be interpreted as 'this test has poor coverage' instead of 'this test is expected to throw an error'. Maybe better wording would be `failure case` (Whatever you decide, apply it to the other test names :) )

JST5000 (changes requested):
A couple of refactors/questions, but the logic looks good!

mvadari:
This PR should be `test:` not `refactor:`

src/sugar/balances.ts
```diff
@@ -41,16 +41,19 @@ async function getBalances(
...
+    const xrpRequest: AccountInfoRequest = {
+      command: 'account_info',
+      account,
+      ledger_index: options.ledger_index ?? 'validated',
+      ledger_hash: options.ledger_hash,
+    }
+    const balance = await this.request(xrpRequest).then(
+      (response) => response.result.account_data.Balance,
+    )
+    xrpBalance.push({ currency: 'XRP', value: dropsToXrp(balance) })
```
natenichols, line 55: Why do we need an array for this. Is there a scenario where you can have two XRP balances?

src/sugar/orderbook.ts
```diff
@@ -37,32 +38,33 @@ interface OrderbookOptions {
...
+ * @param TakerGets - Specs of the currency account taking the offer receives.
  * @param options - Options to include for getting orderbook between payer and receiver.
  * @returns An object containing buy and sell objects.
  */
 // eslint-disable-next-line max-params -- Function needs 4 params.
 async function getOrderbook(
   this: Client,
-  takerPays: TakerAmount,
-  takerGets: TakerAmount,
+  TakerPays: TakerAmount,
```
mvadari, line 49: Why is this uppercase? It should be `camelCase`.

test/client/getOrderbook.ts
mvadari: This wasn't resolved - this should be `request.takerPays.issuer`, since the request takes a `takerPays` param. Same with `takerGets`

mvadari: Now I'm really confused about how this still passes the tests

mvadari: oh nvm, ignore - this is the `rippled` params.

mvadari (approved):
just one naming comment that needs to be fixed before merging, nothing to stop approval

natenichols (approved):

JST5000 (approved):
LGTM!

PR #1636, opened 2021-09-17
Add integration tests for Transaction methods

JST5000:

Title says it all

### Context of Change

Adding integration tests for xrpl.js Transaction methods.

### Type of Change

## Before / After

* Added an integration test for TxRequest
* Moved the integration.ts multisign test into it's own folder

## Test Plan

Integration tests pass


test/integration/integration.ts
```diff
@@ -29,49 +19,4 @@ describe('integration tests', function () {
   it('isConnected', function () {
     assert(this.client.isConnected())
   })
```
JST5000, line 21: Since this is the only test left, and it's super tiny, is there a better place we could put it instead so we can just get rid of the general 'integration.ts' file? @mvadari

test/integration/requests/multisign.ts
```diff
@@ -0,0 +1,72 @@
...
+import { setupClient, suiteClientSetup, teardownClient } from '../setup'
+import {
+  fundAccount,
+  ledgerAccept,
+  testTransaction,
+  verifySubmittedTransaction,
+} from '../utils'
+
+// how long before each test case times out
+const TIMEOUT = 20000
```
JST5000, line 19: @mvadari Should this TIMEOUT be moved to a constant in `integration/utils`?

mvadari:
This PR doesn't cover the other transaction methods

test/integration/requests/txRequest.ts
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable mocha/no-hooks-for-single-case -- Makes test setup consistent */
```
mvadari, line 1: This file should be called `tx.ts` - the `Request` part is redundant

mvadari (changes requested):
This PR is missing the other requests

test/integration/integration.ts
```diff
@@ -1,20 +1,10 @@
+/* eslint-disable mocha/no-hooks-for-single-case -- Keeps test setup consistent */
```
JST5000, line 1: Starting a thread here so we can discuss. 
@mvadari Which requests is this PR missing? (Going off of [this documentation](https://xrpl.org/transaction-methods.html), sign & sign_for are not included, submit is covered by individual transactions, tx_history is deprecated, and we don't seem to have tx_entry) - Are there other requests that should be added here that aren't included on this list?

mvadari: I change my mind, I think we do need a specific `submit` test. @natenichols is there a specific reason we didn't add a `transaction_entry` request?

test/integration/requests/multisign.ts
```diff
@@ -0,0 +1,72 @@
...
+    const accountSet: AccountSet = {
+      TransactionType: 'AccountSet',
+      Account: this.wallet.getClassicAddress(),
+      Domain: convertStringToHex('example.com'),
+    }
+    const accountSetTx = await client.autofill(accountSet, 2)
+    const signed1 = sign(signerWallet1, accountSetTx, true)
+    const signed2 = sign(signerWallet2, accountSetTx, true)
+    const combined = multisign([signed1, signed2])
+    const submitResponse = await client.submitSignedTransaction(combined)
```
mvadari, line 67: This test should test the `submit_multisigned` request, instead of the `submit` request

mvadari: https://xrpl.org/submit_multisigned.html

test/integration/integration.ts
JST5000: I chatted with Nathan about this, and AFAIK the reason is because we suggest people use tx instead of tx_entry in the documentation: https://xrpl.org/transaction_entry.html 
@natenichols Chime in if there's a different reason I'm forgetting!

test/integration/integration.ts
JST5000: Nevermind, I've returned the multisign test to integration.ts because the `submit_multisign` test is different (It won't work with the browser test because the browser doesn't use the rippled instance). For now integration.ts is basically just the browser test file.

test/integration/requests/multisign.ts
mvadari: The whole `this.timeout` could actually be moved into `suiteClientSetup`

test/integration/requests/multisign.ts
```diff
@@ -0,0 +1,78 @@
...
+  fundAccount,
+  ledgerAccept,
+  testTransaction,
+  verifySubmittedTransaction,
+} from '../utils'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('multisign', function () {
```
mvadari, line 23:
```suggestion
describe('submit_multisign', function () {
```

test/integration/requests/multisign.ts
```diff
@@ -0,0 +1,78 @@
+/* eslint-disable mocha/no-hooks-for-single-case -- Makes test setup consistent across files */
+import { assert } from 'chai'
+import _ from 'lodash'
+import { decode } from 'ripple-binary-codec/dist'
+
+import { AccountSet, Client, SignerListSet, Wallet } from 'xrpl-local'
+import { convertStringToHex } from 'xrpl-local/utils'
+import { multisign, sign } from 'xrpl-local/wallet/signer'
+
+import { SubmitMultisignedRequest, Transaction } from '../../../src'
```
mvadari, line 10:
```suggestion
import { SubmitMultisignedRequest, Transaction } from 'xrpl-local'
```

test/integration/requests/submit.ts
```diff
@@ -0,0 +1,41 @@
+import { assert } from 'chai'
+import _ from 'lodash'
+
+import { AccountSet } from 'xrpl-local'
+import { convertStringToHex } from 'xrpl-local/utils'
+
+import { SubmitRequest, SubmitResponse } from '../../../src'
```
mvadari, line 7:
```suggestion
import { SubmitRequest, SubmitResponse } from 'xrpl-local'
```

mvadari (changes requested):

test/integration/requests/multisign.ts
```diff
@@ -0,0 +1,110 @@
+/* eslint-disable mocha/no-hooks-for-single-case -- Makes test setup consistent across files */
```
mvadari, line 1: oh I really like this line-specific notation for linter errors

test/integration/requests/multisign.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+    const accountSetTx = await client.autofill(accountSet, 2)
+    const signed1 = sign(signerWallet1, accountSetTx, true)
+    const signed2 = sign(signerWallet2, accountSetTx, true)
+    const multisigned = multisign([signed1, signed2])
+    const multisignedRequest: SubmitMultisignedRequest = {
+      command: 'submit_multisigned',
+      tx_json: decode(multisigned) as unknown as Transaction,
+    }
+    const submitResponse: SubmitMultisignedResponse = await client.request(
```
mvadari, line 81: This should be automatically typed

test/integration/requests/multisign.ts
```diff
@@ -0,0 +1,110 @@
...
+      multisignedRequest,
+    )
+    await ledgerAccept(client)
+    assert.strictEqual(submitResponse.result.engine_result, 'tesSUCCESS')
+    await verifySubmittedTransaction(this.client, multisigned)
+
+    // The ledger returns an extra 'hash' field on 'submit' requests which is easily calculatable with a normal tx_blob
+    const tx_json = {
+      ...(decode(multisigned) as unknown as Transaction),
```
mvadari, line 90: you shouldn't need to type this - `assert` doesn't care about typing

test/integration/requests/submit.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+    const response: SubmitResponse = await this.client.request(submitRequest)
+    assert.equal(response.status, 'success')
+
+    await ledgerAccept(this.client)
+    await verifySubmittedTransaction(this.client, signedTx)
+
+    // The ledger returns an extra 'hash' field on 'submit' requests which is easily calculatable with a normal tx_blob
+    const tx_json = {
+      ...(decode(signedTx) as unknown as Transaction),
```
mvadari, line 49: ditto

test/integration/requests/multisign.ts
JST5000: Chai asserts do care about typing

mvadari (approved):

test/integration/requests/multisign.ts
```diff
@@ -0,0 +1,104 @@
...
+  before(suiteClientSetup)
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('submit_multisigned transaction', async function () {
+    const client: Client = this.client
+    const signerWallet1 = Wallet.generate()
+    await fundAccount(client, signerWallet1)
+    const signerWallet2 = Wallet.generate()
+    await fundAccount(client, signerWallet2)
```
mvadari, line 42: use `generateFundedWallet` instead

mukulljangid (approved):

PR #1637, opened 2021-09-18
rename browser library

natenichols:
Changes `ripple` to `xrpl`


mvadari:
Should this be a bigger PR that goes through and looks for all instances of `ripple` in the library?

natenichols:
It could be, but this is something that definitely needs to go into next beta.

mvadari (approved):

JST5000 (approved):
Well, that was straightforward :P

PR #1639, opened 2021-09-20
Autofills AccountDelete Fee w/ `reserve_inc_xrp`

natenichols:
Instead of hardcoding, we query the ledger to find the Owner Reserve.

### Context of Change
Reserves changed to 10/2 last night, and this is a straightforward field to autofill.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
Modified the mocked fixtures to require the AccountDelete autofill tests to retrieve from mock rippled.


src/sugar/autofill.ts
```diff
@@ -132,6 +131,19 @@ async function setNextValidSequenceNumber(
   tx.Sequence = data.result.account_data.Sequence
 }
 
+async function fetchAccountDeleteFee(client: Client): Promise<BigNumber> {
+  const response = await client.request({ command: 'server_info' })
+  console.log(response.result.info.validated_ledger)
```
JST5000, line 136: Should this log be there?

src/sugar/autofill.ts
```diff
@@ -132,6 +131,19 @@ async function setNextValidSequenceNumber(
...
+
+  const fee = response.result.info.validated_ledger?.reserve_inc_xrp
+
+  if (fee == null) {
+    return Promise.reject(new Error('Could not fetch Owner Reserve.'))
+  }
+
+  return new BigNumber(fee).multipliedBy(DROPS_PER_XRP)
```
mvadari, line 144: Why not just use `dropsToXrp` in `utils`?

src/sugar/autofill.ts
```diff
@@ -10,8 +10,7 @@ import { xrpToDrops } from '../utils'
 
 // 20 drops
```
mDuo13, line 11: This comment is incorrect. The LEDGER_OFFSET is actually used to set the default for `LastLedgerSequence`, basically, how many ledgers can be validated before a transaction expires. The default of 20 ledger versions is ~1 minute given typical ledger close times of ~3 seconds.

```suggestion
// Expire unconfirmed transactions after 20 ledger versions, approximately 1 minute, by default
```

src/sugar/autofill.ts
mDuo13: It probably should not.

src/sugar/autofill.ts
mDuo13: Better yet, use server_state instead to get the value in drops to begin with.

mDuo13 (changes requested):

src/sugar/autofill.ts
natenichols: No it definitely should not

src/sugar/autofill.ts
```diff
@@ -132,6 +130,17 @@ async function setNextValidSequenceNumber(
   tx.Sequence = data.result.account_data.Sequence
 }
 
+async function fetchAccountDeleteFee(client: Client): Promise<BigNumber> {
+  const response = await client.request({ command: 'server_state' })
```
mvadari, line 134: What's the difference between `server_state` and `server_info`?

mvadari (approved):
just one question

src/sugar/autofill.ts
mDuo13: Per [the documentation](https://xrpl.org/server_state.html) (emphasis added):

> The server_state command asks the server for various machine-readable information about the rippled server's current state. The response is almost the same as the server_info method, but uses units that are easier to process instead of easier to read. (For example, **XRP values are given in integer drops** instead of scientific notation or decimal values, and time is given in milliseconds instead of seconds.)

JST5000 (approved):

mDuo13 (approved):

PR #1640, opened 2021-09-20
Add Ledger Method Integration Tests

JST5000:

Title says it all

### Context of Change

Added tests for these methods: https://xrpl.org/ledger-methods.html#main-page-header 

### Type of Change

## Before / After

Just adding a new integration test file for ledger methods.

## Test Plan

Integration tests pass


test/integration/requests/ledger.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    const ledgerResponse: LedgerResponse = await this.client.request(
+      ledgerRequest,
+    )
+
+    verifySuccessfulResponse(ledgerResponse)
+    assert.ok(ledgerResponse.result.validated)
+  })
+
+  it('LedgerClosed', async function () {
```
JST5000, line 45: Should each of these be in a separate file? (Or does it make sense to group them like this?) @mvadari

mvadari: They should be in separate folders. The `test/integration/requests` folder should parallel the `src/models/requests` folder.

test/integration/requests/ledgerEntry.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  it('base', async function () {
+    const validatedLedgerResponse: LedgerDataResponse =
+      await this.client.request({
+        command: 'ledger_data',
+        ledger_index: 'validated',
+      })
+
+    verifySuccessfulResponse(validatedLedgerResponse)
+    const ledgerEntryIndex = validatedLedgerResponse.result.state[0].index
```
mukulljangid, line 32: Just use validated instead of getting the index

mukulljangid (approved):

test/integration/requests/ledger.ts
```diff
@@ -0,0 +1,33 @@
...
+import { LedgerRequest, LedgerResponse } from 'xrpl-local'
+
+import serverUrl from '../serverUrl'
+import { setupClient, suiteClientSetup, teardownClient } from '../setup'
+import { verifySuccessfulResponse } from '../utils'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('Ledger', function () {
```
mvadari, line 13:
```suggestion
describe('ledger', function () {
```

test/integration/requests/ledgerClosed.ts
```diff
@@ -0,0 +1,28 @@
...
+import { LedgerClosedRequest, LedgerClosedResponse } from 'xrpl-local'
+
+import serverUrl from '../serverUrl'
+import { setupClient, suiteClientSetup, teardownClient } from '../setup'
+import { verifySuccessfulResponse } from '../utils'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('LedgerClosed', function () {
```
mvadari, line 12:
```suggestion
describe('ledger_closed', function () {
```

mvadari (changes requested):

test/integration/requests/ledgerClosed.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  before(suiteClientSetup)
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const ledgerClosedRequest: LedgerClosedRequest = {
+      command: 'ledger_closed',
+    }
+    const ledgerClosedResponse: LedgerClosedResponse =
```
mvadari, line 24: this doesn't need to be typed, `assert` doesn't care

test/integration/requests/ledgerCurrent.ts
```diff
@@ -0,0 +1,40 @@
...
+  before(suiteClientSetup)
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const ledgerCurrentRequest: LedgerCurrentRequest = {
+      command: 'ledger_current',
+    }
+
+    const ledgerCurrentResponse: LedgerCurrentResponse =
```
mvadari, line 25: ditto

test/integration/requests/ledgerData.ts
```diff
@@ -0,0 +1,37 @@
...
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const ledgerDataRequest: LedgerDataRequest = {
+      command: 'ledger_data',
+      ledger_index: 'validated',
+    }
+
+    const ledgerDataResponse: LedgerDataResponse = await this.client.request(
```
mvadari, line 26: ditto

test/integration/requests/ledger.ts
```diff
@@ -0,0 +1,45 @@
...
+      command: 'ledger',
+      ledger_index: 'validated',
+    }
+
+    const ledgerResponse: LedgerResponse = await this.client.request(
+      ledgerRequest,
+    )
+
+    const expectedResponse: LedgerResponse = {
+      id: ledgerResponse.id,
```
mvadari, line 31: You could create a helper function to check requests that does an `assert` compare that strips both objects of IDs, so that you don't need this part.

mvadari: I think this happens in `verifySubmittedTransaction`

test/integration/requests/ledgerEntry.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    const ledgerEntryRequest: LedgerEntryRequest = {
+      command: 'ledger_entry',
+      index: ledgerEntryIndex,
+    }
+
+    const ledgerEntryResponse: LedgerEntryResponse = await this.client.request(
```
mvadari, line 39: ditto

test/integration/requests/ledger.ts
mvadari: even better, just add it to `verifySuccessfulResponse`

JST5000: The main reason I'm doing this is because it allows for a very straightforward assert.deepEquals which ensures the shape matches, instead of going line by line with individual asserts. (Places where I borrow from the previous request are just an easier way of saying "I care that the info is there, but not what's in it") - Does that make sense? (Maybe I'm missing an antipattern that this is falling into?)

mvadari (approved):

PR #1642, opened 2021-09-21
docs: document transaction models

natenichols:
Adds JSDoc style comments to document Transaction models.

### Type of Change

- [x] Documentation Updates
- [ ] Release


mvadari (commented):
Why are a bunch of unrelated files deleted/recreated in this PR?

src/models/transactions/checkCreate.ts
```diff
@@ -8,12 +8,30 @@ import {
...
+ * cashed by its intended destination. The sender of this transaction is the
+ * sender of the Check.
+ *
+ * @category Transaction Models
+ */
 export interface CheckCreate extends BaseTransaction {
   TransactionType: 'CheckCreate'
+  /** The unique address of the account that can cash the Check. */
   Destination: string
+  /** 	Maximum amount of source currency the Check is allowed to debit the
```
mvadari, line 21: What's with the extra spacing here?

docs/interfaces/AccountLinesResponse.html
```diff
@@ -1,342 +1 @@
-<!doctype html>
-<html class="default no-js">
```
mvadari, line 2: Why did all of these files become just one line?

mvadari:
Some notes after looking at the HTML in a browser:
* Any way to make the optional variables show up after the required ones?
* Is there still display work to do for the categories, to have them have different symbols or something? Because it looks like they haven't changed as of now.
* Looks like you need to add documentation to `BaseTransaction`, for all the bits that other transactions inherit. They appear as doc-less in the HTML.

Might not be worth doing at this point, but IMO all the non-transaction changes in this PR should be in a separate PR.

PR #1643, opened 2021-09-21
test: add integration tests for Account requests

mukulljangid:
Add integration tests for Account requests

### Type of Change

## Test Plan
Passes all the tests


test/integration/requests/accountChannels.ts
```diff
@@ -0,0 +1,30 @@
...
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const request: AccountChannelsRequest = {
+      id: 1,
+      command: 'account_channels',
+      account: 'rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh',
+      ledger_index: 'validated',
+    }
+    const response = await this.client.requestAll(request)
```
JST5000, line 26: Is there a reason you're using `requestAll` here instead of `request`?

test/integration/requests/accountChannels.ts
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable mocha/no-hooks-for-single-case -- Use of hooks is restricted when there is a single test case. */
+import { assert } from 'chai'
+import _ from 'lodash'
+
+import { AccountChannelsRequest } from '../../../src'
```
mvadari, line 5:
```suggestion
import { AccountChannelsRequest } from 'xrpl-local'
```
All imports should be from `xrpl-local` instead of from `../../../src`, because otherwise browser tests will fail

test/integration/requests/accountChannels.ts
mvadari: Agreed - this should be testing `request`, `requestAll` is used elsewhere

mvadari (changes requested):

mvadari:
This PR should be `test:` not `refactor:`

test/integration/requests/accountChannels.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { AccountChannelsRequest } from 'xrpl-local'
+
+import serverUrl from '../serverUrl'
+import { setupClient, suiteClientSetup, teardownClient } from '../setup'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('AccountChannels', function () {
```
mvadari, line 12:
```suggestion
describe('account_channels', function () {
```

mvadari: To clarify something about naming conventions: we want to use exactly what rippled does if it's just a descriptive string or something, but if someone's actually going to be interacting with it in a JS way (such as a function param), then it should be camelCase.

test/integration/requests/accountChannels.ts
```diff
@@ -0,0 +1,46 @@
...
+          'C8BFA74A740AA22AD9BD724781589319052398B0C6C817B88D55628E07B7B4A1',
+        ledger_index: 150,
+        validated: true,
+      },
+      status: 'success',
+      type: 'response',
+    }
+    assert.equal(response.status, expected.status)
+    assert.equal(response.type, expected.type)
+    assert.deepEqual(
```
mvadari, line 41: ++

mvadari:
why is the lint tests a commit in this PR?

test/integration/requests/accountChannels.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  it('base', async function () {
+    const request: AccountChannelsRequest = {
+      command: 'account_channels',
+      account: 'rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh',
+      ledger_index: 'validated',
+    }
+    const response = await this.client.request(request)
+    const expected = {
+      id: 5,
```
mvadari, line 27: if the ID is just a placeholder, make it `0` - that's what the fixtures do.

test/integration/requests/accountCurrencies.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { AccountCurrenciesRequest } from 'xrpl-local'
+
+import serverUrl from '../serverUrl'
+import { setupClient, suiteClientSetup, teardownClient } from '../setup'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('AccountCurrencies', function () {
```
mvadari, line 12:
```suggestion
describe('account_currencies', function () {
```
you get the idea, I'm gonna stop doing every single one of these

test/integration/requests/accountInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+        ledger_index: 582,
+        validated: true,
+      },
+      status: 'success',
+      type: 'response',
+    }
+    assert.equal(
+      _.has(response.result.account_data, 'Balance'),
+      _.has(expected.result.account_data, 'Balance'),
+    )
```
mvadari, line 54: What does this do? Why do you need `has`? The balance should be the same every time.

test/integration/requests/accountInfo.ts
```diff
@@ -0,0 +1,73 @@
...
+    assert.equal(response.status, expected.status)
+    assert.equal(response.type, expected.type)
+    assert.equal(response.result.validated, expected.result.validated)
+    assert.deepEqual(
+      _.omit(response.result.account_data, [
+        'PreviousTxnID',
+        'PreviousTxnLgrSeq',
+        'Sequence',
+        'Balance',
```
mvadari, line 63: Why is balance omitted?

test/integration/requests/accountChannels.ts
```diff
@@ -0,0 +1,46 @@
...
+        ledger_index: 150,
+        validated: true,
+      },
+      status: 'success',
+      type: 'response',
+    }
+    assert.equal(response.status, expected.status)
+    assert.equal(response.type, expected.type)
+    assert.deepEqual(
+      _.omit(response.result, ['ledger_hash', 'ledger_index']),
```
mvadari, line 42: you should at least check type for the stuff you're omitting

src/errors.ts
```diff
@@ -20,7 +20,10 @@ class XrplError extends Error {
     this.name = this.constructor.name
     this.message = message
     this.data = data
-    Error.captureStackTrace(this, this.constructor)
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- `captureStackTrace` can be null in
```
mvadari, line 23: Where did this come from? This isn't related to this PR

test/integration/requests/accountTx.ts
```diff
@@ -0,0 +1,113 @@
...
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const request: AccountTxRequest = {
+      command: 'account_tx',
+      account: 'rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh',
+      ledger_index: 'validated',
+    }
+    const response = await this.client.requestAll(request)
```
mvadari, line 25:
```suggestion
    const response = await this.client.request(request)
```

test/integration/requests/gatewayBalances.ts
```diff
@@ -0,0 +1,48 @@
...
+  before(suiteClientSetup)
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const request: GatewayBalancesRequest = {
+      command: 'gateway_balances',
+      account: 'rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh',
+      ledger_index: 'validated',
+      strict: true,
```
mvadari, line 24: It doesn't really matter for this, but the base request should technically be with no additional variables.

mvadari (approved):
Please address comments before merging

src/errors.ts
mukulljangid: Yes, I'll change this!

test/integration/index.ts
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable import/export -- Tells webpack which files exist. */
+export * from './integration'
+export * from './transactions/signerListSet'
+export * from './transactions/payment'
+export * from './transactions/offerCreate'
+export * from './transactions/offerCancel'
+export * from './transactions/signerListSet'
+export * from './requests/utility'
```
mvadari, line 8: don't forget to also add your tests to this list

test/integration/index.ts
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable import/export -- Tells webpack which files exist. */
+export * from './integration'
```
mvadari, line 2: This file needs to be at the bottom for browser tests (it's a weird thing that needs to be resolved, but isn't a priority right now)

test/localIntegrationRunner.html
```diff
@@ -17,12 +17,7 @@
       mocha.ui('bdd')
     </script>
 
-    <script src="../testCompiledForWeb/offerCancel.js"></script>
-    <script src="../testCompiledForWeb/offerCreate.js"></script>
-    <script src="../testCompiledForWeb/payment.js"></script>
-    <script src="../testCompiledForWeb/signerListSet.js"></script>
-    <script src="../testCompiledForWeb/utility.js"></script>
-    <script src="../testCompiledForWeb/integration.js"></script>
+    <script src="../testCompiledForWeb/index.js"></script>
```
JST5000, line 21: Did you mean to delete these integration tests? (I think the expected output for this file should be to have all the integration tests in here as <script> ... </script> files)

webpack.config.js
```diff
@@ -120,30 +120,9 @@ function webpackForTest(testFileName) {
...
-    const subdirPaths = fs.readdirSync(path.join(dir, subdir))
-    subdirPaths.forEach((filename) => {
-      tests.push(webpackForTest(`./${path.join(dir, subdir, filename)}`))
-    })
-  })
-  return tests.map(
-    (test) => (env, argv) => Object.assign({}, getDefaultConfiguration(), test),
-  )
-}
-
```
JST5000, line 146: Same here, I think this file is desireable - Maybe you need to rebase on 2.0 again? (Not sure)

webpack.config.js
```diff
@@ -160,5 +139,5 @@ module.exports = [
     }
     return config
   },
-  ...webpackIntegrationTests(),
+  (env, argv) => webpackForTest('./test/integration/index.ts'),
```
JST5000, line 142: Ditto

JST5000 (changes requested):
The localTestRunner along with webpack.config.js seem to be accidentally removing the new browser integration setup, but other than that looks good!

JST5000 (approved):
Since the webpack changes are expected to avoid a crash happening with browser tests, this looks good to me!

PR #1644, opened 2021-09-21
Add Path and Orderbook integration tests

JST5000:

Title says it all

### Context of Change

Adding tests for these methods: https://xrpl.org/path-and-order-book-methods.html 

### Type of Change

## Before / After

Added integration tests for each type of request.

## Test Plan

Integration tests pass


mukulljangid (approved):

test/integration/requests/bookOffers.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { BookOffersRequest, BookOffersResponse } from 'xrpl-local'
+
+import serverUrl from '../serverUrl'
+import { setupClient, suiteClientSetup, teardownClient } from '../setup'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('BookOffers', function () {
```
mvadari, line 12:
```suggestion
describe('book_offers', function () {
```

mvadari (changes requested):

src/models/methods/ripplePathFind.ts
```diff
@@ -28,5 +28,11 @@ export interface RipplePathFindResponse extends BaseResponse {
     alternatives: PathOption[]
     destination_account: string
     destination_currencies: string[]
+    destination_amount: Amount
```
mvadari, line 31: whoops
this is why we do integration tests lol

mvadari (approved):

PR #1646, opened 2021-09-21
Add Payment Channel Tests

JST5000:

Title says it all

### Context of Change

This ends up just being ChannelVerify because AuthorizeChannel is handled offline with Signer, and ChannelVerify is the only other PaymentChannel method here: https://xrpl.org/payment-channel-methods.html 

### Type of Change

## Before / After

Added a test for ChannelVerify

## Test Plan

Integration tests pass


mvadari (changes requested):
This test should check the shape of the response and make sure it matches the object. See the utility tests for examples.

test/integration/requests/channelVerify.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { ChannelVerifyRequest, ChannelVerifyResponse } from 'xrpl-local'
+
+import serverUrl from '../serverUrl'
+import { setupClient, suiteClientSetup, teardownClient } from '../setup'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('ChannelVerify', function () {
```
mvadari, line 12:
```suggestion
describe('channel_verify', function () {
```

natenichols (approved):

mvadari (approved):

natenichols (approved):

PR #1649, opened 2021-09-22
fix: run browser tests on all integration tests

mvadari:

This PR webpacks all the integration test files for testing, and runs all of the tests on the browser.

### Context of Change

Because we started using multiple files for integration tests, the browser tests did not pick up those files, and continued to only run the tests on the one file.

### Type of Change

## Test Plan

CI passes. All the integration tests run on the browser.

## Future Tasks
This is not the best code, and needs to be revamped a bit before we officially publish 2.0. But it's good enough for now.
* The script tags in `localIntegrationRunner.html` should be dynamically added
* The files in `testCompiledForWeb` should be properly nested in folders
* Browser tests should print better debugging info in the CLI

### DIFF
--- a/test/localIntegrationRunner.html
+++ b/test/localIntegrationRunner.html
@@ -17,6 +17,11 @@
       mocha.ui('bdd')
     </script>
 
+    <script src="../testCompiledForWeb/offerCancel.js"></script>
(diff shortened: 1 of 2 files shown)


natenichols (approved):

JST5000 (approved):
LGTM!

PR #1663, opened 2021-09-24
fix: resolve browser issue with Error.captureStackTrace

mvadari:

Title says it all.

### Context of Change

@mDuo13 was having an issue with browser code, where it would fail because of the error `Uncaught TypeError: Error.captureStackTrace is not a function`. This PR should fix that.

### Type of Change

## Test Plan

CI passes. @mDuo13 will check if this fixes the problem.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -20,7 +20,10 @@ class XrplError extends Error {
     this.name = this.constructor.name
     this.message = message
     this.data = data
-    Error.captureStackTrace(this, this.constructor)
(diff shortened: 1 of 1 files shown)


natenichols (approved):

mDuo13:
This fixes the `Error.captureStackTrace` issue, though I'm now seeing `Uncaught Error: Serialized transaction does not match original txJSON. See error.data` which I think points to a different underlying bug in `autofill`. That said, I think this part is good.

mDuo13 (approved):

PR #1671, opened 2021-09-25
fix: return when account is not defined

natenichols:
Changing to throwing when an error is returned caused an issue in the `generateFaucetWallet()` method. 

### Context of Change
Fixes issue by returning `undefined` when the account does not exist. This method is called before and after the faucet, so `undefined` is a valid result for an account that has not been created yet. 

### Type of Change
- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/wallet/generateFaucetWallet.ts
+++ b/src/wallet/generateFaucetWallet.ts
@@ -217,13 +217,12 @@ async function getAddressXrpBalance(
 ): Promise<string | undefined> {
   // Get all the account balances
   try {
-    const balances = await client.getBalances(address)
+    const balances = await client.request({
+      command: 'account_info',
+      account: address,
+    })
 
-    // Retrieve the XRP balance
-    const xrpBalance = balances.filter(
-      (balance) => balance.currency.toUpperCase() === 'XRP',
-    )
-    return xrpBalance[0].value
+    return balances.result.account_data.Balance
   } catch (err) {
     if (err instanceof Error) {
       throw new XRPLFaucetError(
--- a/src/wallet/generateFaucetWallet.ts
+++ b/src/wallet/generateFaucetWallet.ts
@@ -54,10 +54,15 @@ async function generateFaucetWallet(
     }),
   )
   // Retrieve the existing account balance
-  const addressToFundBalance = await getAddressXrpBalance(
-    this,
-    fundWallet.classicAddress,
-  )
+  let addressToFundBalance: undefined | string
+  try {
+    addressToFundBalance = await getAddressXrpBalance(
+      this,
+      fundWallet.classicAddress,
+    )
+  } catch {
+    /* addressToFundBalance remains undefined */
+  }
 
   // Check the address balance is not undefined and is a number
   const startingBalance =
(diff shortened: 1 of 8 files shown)


intelliot:
Why is the method called before and after the faucet?

I suppose the alternative would be try/catch in the faucet method.

natenichols:
Its called before the faucet, because the faucet can also add balances to a wallet.

In that case, you'd wanna know the balance before and after the wallet has been funded.

natenichols:
The try/catch in `generateFaucetWallet` is prolly a better approach.

mvadari:
Can we not just have better error handling instead?

Also, while you're here, I think `getUrl` should throw an error if the provided URL isn't devnet or testnet, just in case someone accidentally tries to use it with standalone or something.

mvadari:
Can you add an integration test or two to this PR?

natenichols:
> Can you add an integration test or two to this PR?

Ah, so the integrations tests go in this PR 😄 

I'll write something up

mvadari:
> > Can you add an integration test or two to this PR?
> 
> Ah, so the integrations tests go in this PR 😄
> 
> I'll write something up

I mean, they don't have to, but IMO every bug fix should be accompanied with a test that checks for that bug

natenichols:
Added some integration tests.

src/wallet/generateFaucetWallet.ts
```diff
@@ -209,7 +214,7 @@ async function processSuccessfulResponse(
 async function getAddressXrpBalance(
   client: Client,
   address: string,
-): Promise<string> {
+): Promise<string | undefined> {
```
mvadari, line 217: Is the `| undefined` still needed here?

src/wallet/generateFaucetWallet.ts
```diff
@@ -292,7 +304,7 @@ function getFaucetUrl(client: Client): FaucetNetwork | undefined {
     return FaucetNetwork.Devnet
   }
 
-  return undefined
+  throw new XRPLFaucetError('Faucet URL is not defined or inferrable.')
```
mvadari, line 307: Thoughts on adding a special error message for standalone if they're trying to connect to localhost?

test/integration/faucet.ts
```diff
@@ -0,0 +1,63 @@
+import assert from 'assert'
```
mvadari, line 1: nit: IMO this file should be called `generateFaucetWallet`

test/integration/faucet.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import _ from 'lodash'
+
+import { Client, isValidClassicAddress, isValidXAddress } from 'xrpl-local'
+// how long before each test case times out
+const TIMEOUT = 60000
+// This test is reliant on external networks, and as such may be flaky.
+describe('generateFaucetWallet', function () {
+  this.timeout(TIMEOUT)
+  it('submit generates a testnet wallet', async function () {
```
mvadari, line 11: nit: add a blank line between tests/between the timeout and the first test

mvadari (approved):

mukulljangid (approved):

src/wallet/generateFaucetWallet.ts
natenichols: I don't see why we would, I think this message covers that case.

PR #1673, opened 2021-09-27
test: Integration tests for Account Transactions

mukulljangid:
Add Integration tests for Account Transactions

### Type of Change

## Test Plan
Passes all the tests

### DIFF
--- a/test/integration/transactions/accountDelete.ts
+++ b/test/integration/transactions/accountDelete.ts
@@ -1,0 +1,28 @@
+/* eslint-disable mocha/no-hooks-for-single-case -- Use of hooks is restricted when there is a single test case. */
+import _ from 'lodash'
+
+import { AccountDelete } from 'xrpl-local/models/transactions'
+
+import serverUrl from '../serverUrl'
+import { setupClient, suiteClientSetup, teardownClient } from '../setup'
+import { getMasterAccount, testTransaction } from '../utils'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('AccountDelete', function () {
+  this.timeout(TIMEOUT)
+
+  before(suiteClientSetup)
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const tx: AccountDelete = {
+      TransactionType: 'AccountDelete',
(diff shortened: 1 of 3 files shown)


test/integration/transactions/accountDelete.ts
```diff
@@ -0,0 +1,34 @@
...
+describe('AccountDelete', function () {
+  this.timeout(TIMEOUT)
+
+  before(suiteClientSetup)
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const wallet = await generateFundedWallet(this.client)
+    const promises: Array<Promise<void>> = []
```
natenichols, line 21: Make a comment here that explains why we need to skip 256 ledgers ahead.

natenichols (approved):

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "xrpl",
-  "version": "2.0.0-beta.0",
+  "version": "2.0.0-beta.1",
```
JST5000, line 3: Should this be changing in this PR?

test/integration/transactions/accountDelete.ts
```diff
@@ -0,0 +1,34 @@
...
+
+describe('AccountDelete', function () {
+  this.timeout(TIMEOUT)
+
+  before(suiteClientSetup)
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const wallet = await generateFundedWallet(this.client)
```
JST5000, line 20: nit: Using a name like `wallet2` might make this easier to differentiate from `this.wallet`

JST5000 (commented):
Code looks good, just confused about whether the package-lock should be changing.

mvadari:
> Code looks good, just confused about whether the package-lock should be changing.

That change was resolved in #1671 anyways.

package-lock.json
natenichols: This is changing because @mvadari  and I forgot to run `npm` on the PR that changed that `version`

It's autogenerated and shouldn't be a problem.

JST5000 (approved):

PR #1678, opened 2021-09-28
refactor: separate out webpacking for tests

natenichols:
`npm i` was taking a long time to run because of webpacking browser tests. This splits building browser tests into its own script.

### Context of Change
Previously, we were webpacking everything in one command. This meant that `npm i` took a long time to run, because in addition to webpacking `xrpl`, it also had to webpack browser tests. This PR separates out the webpacking into two commands `npm run build:browserTests`, which builds the browser tests, and `npm build:web`, which builds the `xrpl` library. `npm run build:web` runs on installation, and `npm run build:browserTests` only runs when `test:browser` script is run.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


test/webpack.config.js
```diff
@@ -0,0 +1,79 @@
...
+    cache: true,
+    externals: [
+      {
+        'xrpl-local': 'xrpl',
+        net: 'null',
+      },
+    ],
+    entry: testFileName,
+    output: {
+      library: match[1].replace(/-/g, '_'),
```
JST5000, line 23: What is match[0] here?

JST5000 (approved):
Moving this over seems like a straightforward change, although I'm still a bit confused about how the webpack.config.js works.

test/localIntegrationRunner.html
```diff
@@ -17,7 +17,7 @@
       mocha.ui('bdd')
     </script>
 
-    <script src="../testCompiledForWeb/index.js"></script>
```
mvadari, line 20: can we put this file in `test/integration`?

test/webpack.config.js
mvadari: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

mvadari: It's a regex thing

test/webpack.config.js
```diff
@@ -0,0 +1,79 @@
+'use strict'
```
mvadari, line 1: can this also go in `test/browser`?

mvadari (approved):
left a couple minor comments

mvadari:
> I'm still a bit confused about how the webpack.config.js works.

Honestly, I don't really understand how webpacking works either. It's a pretty big rabbit hole.

test/webpack.config.js
natenichols: It could, but it'd prolly fit better in `integration`. I'm okay with leaving it in `test` for now.

PR #1680, opened 2021-09-29
test: Add Integration tests for Escrow Transactions

mukulljangid:
Add Integration tests for Escrow Transactions

### Type of Change

## Test Plan
Passes all the tests

### DIFF
--- a/test/integration/transactions/escrowCancel.ts
+++ b/test/integration/transactions/escrowCancel.ts
@@ -1,0 +1,46 @@
+/* eslint-disable mocha/no-hooks-for-single-case -- Use of hooks is restricted when there is a single test case. */
+import _ from 'lodash'
+
+import { EscrowCancel, EscrowCreate, Wallet } from 'xrpl-local'
+
+import serverUrl from '../serverUrl'
+import { setupClient, suiteClientSetup, teardownClient } from '../setup'
+import {
+  fundAccount,
+  getEpochTime,
+  getSequence,
+  testTransaction,
+} from '../utils'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('EscrowCancel', function () {
+  this.timeout(TIMEOUT)
+
+  before(suiteClientSetup)
+  beforeEach(_.partial(setupClient, serverUrl))
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const wallet1 = Wallet.generate()
+    await fundAccount(this.client, wallet1)
+    const createTx: EscrowCreate = {
+      TransactionType: 'EscrowCreate',
+      Amount: '10000',
+      Account: this.wallet.getClassicAddress(),
+      Destination: wallet1.getClassicAddress(),
+      CancelAfter: getEpochTime() + 2,
+    }
+    console.log(createTx)
+    await testTransaction(this.client, createTx, this.wallet)
+    const cancelTx: EscrowCancel = {
+      TransactionType: 'EscrowCancel',
+      Account: this.wallet.getClassicAddress(),
+      Owner: this.wallet.getClassicAddress(),
+      OfferSequence: getSequence(),
+    }
+    console.log(cancelTx)
+    await testTransaction(this.client, cancelTx, this.wallet)
+  })
+})
--- a/test/integration/transactions/escrowCancel.ts
+++ b/test/integration/transactions/escrowCancel.ts
@@ -5,7 +5,7 @@ import { EscrowCancel, EscrowCreate } from 'xrpl-local'
 
 import serverUrl from '../serverUrl'
 import { setupClient, suiteClientSetup, teardownClient } from '../setup'
-import { generateFundedWallet, getSequence, testTransaction } from '../utils'
+import { generateFundedWallet, testTransaction } from '../utils'
 
 // how long before each test case times out
 const TIMEOUT = 20000
@@ -43,28 +43,34 @@ describe('EscrowCancel', function () {
     )[0].value
 
     // check that the object was actually created
-    assert.equal(
-      (
-        await this.client.request({
-          command: 'account_objects',
-          account: this.wallet.getClassicAddress(),
-        })
-      ).result.account_objects.length === 1,
-      true,
-    )
+    const accountObjects = (
+      await this.client.request({
+        command: 'account_objects',
+        account: this.wallet.getClassicAddress(),
+      })
+    ).result.account_objects
+
+    assert.equal(accountObjects.length, 1)
+
+    const sequence = (
+      await this.client.request({
+        command: 'tx',
+        transaction: accountObjects[0].PreviousTxnID,
+      })
+    ).result.Sequence
 
     const cancelTx: EscrowCancel = {
       TransactionType: 'EscrowCancel',
       Account: this.wallet.getClassicAddress(),
       Owner: this.wallet.getClassicAddress(),
-      OfferSequence: getSequence(),
+      OfferSequence: sequence,
     }
+
     await testTransaction(this.client, cancelTx, this.wallet)
 
     assert.equal(
-      (await this.client.getBalances(wallet1.getClassicAddress()))[0].value ===
-        initialBalanceWallet1,
-      true,
+      (await this.client.getBalances(wallet1.getClassicAddress()))[0].value,
+      initialBalanceWallet1,
     )
   })
 })
--- a/test/integration/transactions/escrowCancel.ts
+++ b/test/integration/transactions/escrowCancel.ts
@@ -1,15 +1,11 @@
+import { assert } from 'chai'
 import _ from 'lodash'
 
 import { EscrowCancel, EscrowCreate } from 'xrpl-local'
 
 import serverUrl from '../serverUrl'
 import { setupClient, suiteClientSetup, teardownClient } from '../setup'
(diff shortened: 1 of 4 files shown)


test/integration/transactions/escrowCancel.ts
```diff
@@ -0,0 +1,51 @@
...
+    await testTransaction(this.client, createTx, this.wallet)
+    await ledgerAccept(this.client)
+    const cancelTx: EscrowCancel = {
+      TransactionType: 'EscrowCancel',
+      Account: this.wallet.getClassicAddress(),
+      Owner: this.wallet.getClassicAddress(),
+      OfferSequence: getSequence(),
+    }
+    await testTransaction(this.client, cancelTx, this.wallet)
+  })
```
JST5000, line 50: Should we add a check here to see if the money actually moved? (For the cancel case, we want to make sure that it didn't move right?)

.eslintrc.js
```diff
@@ -0,0 +1,128 @@
+module.exports = {
+  root: true,
+
+  // Make ESLint compatible with TypeScript
```
natenichols, line 4: Why are `.eslintrc.js` and `.eslintignore` popping up. This is probably an issue with git, maybe some issues with an old merge.

test/integration/transactions/escrowCancel.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    // check that the object was actually created
+    assert.equal(
+      (
+        await this.client.request({
+          command: 'account_objects',
+          account: this.wallet.getClassicAddress(),
+        })
+      ).result.account_objects.length === 1,
+      true,
```
JST5000, line 53: You should probably check that `...account_objects.length` is equal to 1 instead of doing the comparison and checking if it's true. 

`assert.equal(
      (
        await this.client.request({
          command: 'account_objects',
          account: this.wallet.getClassicAddress(),
        })
      ).result.account_objects.length, 1)

test/integration/transactions/escrowCancel.ts
```diff
@@ -0,0 +1,70 @@
...
+      Account: this.wallet.getClassicAddress(),
+      Owner: this.wallet.getClassicAddress(),
+      OfferSequence: getSequence(),
+    }
+    await testTransaction(this.client, cancelTx, this.wallet)
+
+    assert.equal(
+      (await this.client.getBalances(wallet1.getClassicAddress()))[0].value ===
+        initialBalanceWallet1,
+      true,
```
JST5000, line 67: Same here, let the assert.equal do the job '===' is doing here.

test/integration/transactions/escrowCreate.ts
```diff
@@ -0,0 +1,42 @@
...
+  afterEach(teardownClient)
+
+  it('base', async function () {
+    const wallet1 = await generateFundedWallet(this.client)
+    const tx: EscrowCreate = {
+      Account: this.wallet.getClassicAddress(),
+      TransactionType: 'EscrowCreate',
+      Amount: '10000',
+      Destination: wallet1.getClassicAddress(),
+      FinishAfter: Math.round(Date.now() / 1000) - 0x386d4380 + 1000000,
```
JST5000, line 27: It'd be useful to have some variables explain why you chose these numbers. 
* Why divide by 1000?
* What does 0x386d4380 mean?
* Why 1000000?

test/integration/transactions/escrowCreate.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    // check that the object was actually created
+    assert.equal(
+      (
+        await this.client.request({
+          command: 'account_objects',
+          account: this.wallet.getClassicAddress(),
+        })
+      ).result.account_objects.length === 1,
+      true,
```
JST5000, line 39: Ditto - Let assert.equal do the equality check.

test/integration/transactions/escrowCancel.ts
mukulljangid: Actually, yes!

test/integration/transactions/escrowFinish.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    // check that the object was actually created
+    assert.equal(
+      (
+        await this.client.request({
+          command: 'account_objects',
+          account: this.wallet.getClassicAddress(),
+        })
+      ).result.account_objects.length === 1,
+      true,
```
JST5000, line 53: Ditto

test/integration/utils.ts
```diff
@@ -98,8 +99,14 @@ export async function testTransaction(
...
+
   // check that the transaction is on the ledger
   const signedTx = _.omit(response.result.tx_json, 'hash')
   await ledgerAccept(client)
   await verifySubmittedTransaction(client, signedTx as Transaction)
 }
+
+export function getSequence(): number {
+  return sequence
+}
```
JST5000, line 112: It might be better to return the transaction retrieved during `testTransaction` instead of having a secret listener which requires `testTransaction` to run implicitly.

test/integration/transactions/escrowCreate.ts
mukulljangid: I completely replaced it.

test/integration/utils.ts
mukulljangid: Done!

test/integration/transactions/escrowFinish.ts
```diff
@@ -0,0 +1,77 @@
...
+        ledger_index: 'validated',
+      })
+    ).result.ledger.close_time
+    const wallet1 = await generateFundedWallet(this.client)
+
+    const AMOUNT = 10000
+    const createTx: EscrowCreate = {
+      Account: this.wallet.getClassicAddress(),
+      TransactionType: 'EscrowCreate',
+      Amount: String(AMOUNT),
```
JST5000, line 34: This would probably be easier to read if you just make the amount a string from the start instead of converting :)

JST5000, line 34: For this moving it into a separate variable doesn't make it much more readable (Since you already have the "Amount:" in the object to tell you it's supposed to be an amount. So it would probably be better to just have the raw number there. (In string form if that's the proper type)

JST5000 (approved):
Just one nit, but looks good!

test/integration/transactions/escrowFinish.ts
natenichols: Can we just make this `Amount: "10000"`?

test/integration/transactions/escrowFinish.ts
```diff
@@ -0,0 +1,77 @@
...
+      Destination: wallet1.getClassicAddress(),
+      FinishAfter: CLOSE_TIME + 2,
+    }
+
+    await testTransaction(this.client, createTx, this.wallet)
+
+    const initialBalance = (
```
natenichols, line 41: This is a little hacky, I would prefer to do balances of `XRP` with an `account_info` request, but these are also integration tests, so its not a huge deal

test/integration/transactions/escrowCancel.ts
```diff
@@ -0,0 +1,76 @@
...
+      Amount: '10000',
+      Destination: wallet1.getClassicAddress(),
+      CancelAfter: CLOSE_TIME + 3,
+      FinishAfter: CLOSE_TIME + 2,
+    }
+
+    await testTransaction(this.client, createTx, this.wallet)
+
+    const initialBalanceWallet1 = (
+      await this.client.getBalances(wallet1.getClassicAddress())
```
natenichols, line 42: Same here.

test/integration/transactions/escrowFinish.ts
```diff
@@ -0,0 +1,77 @@
...
+      Account: this.wallet.getClassicAddress(),
+      Owner: this.wallet.getClassicAddress(),
+      OfferSequence: sequence,
+    }
+
+    await testTransaction(this.client, finishTx, this.wallet)
+
+    const expectedBalance = Number(initialBalance) + Number(dropsToXrp(AMOUNT))
+    assert.equal(
+      (await this.client.getBalances(wallet1.getClassicAddress()))[0].value,
```
natenichols, line 73: ditto, would prefer `account_info`

natenichols (changes requested):
IMO it is better to retreive XRP balances with `account_info`. I'm open to using `this.client.getBalances().filter(balance => balance.currency === "XRP')`, but using `[0]` clutters the code a bit IMO

test/integration/transactions/escrowFinish.ts
mukulljangid: Sure!

test/integration/transactions/escrowCancel.ts
```diff
@@ -0,0 +1,74 @@
...
+      TransactionType: 'EscrowCreate',
+      Amount: '10000',
+      Destination: wallet1.getClassicAddress(),
+      CancelAfter: CLOSE_TIME + 3,
+      FinishAfter: CLOSE_TIME + 2,
+    }
+
+    await testTransaction(this.client, createTx, this.wallet)
+
+    const initialBalanceWallet1 = await getXRPBalance(this.client, wallet1)
```
natenichols, line 41: Nice!

natenichols (approved):
LGTM 👍

PR #1686, opened 2021-10-01
feat: alias wallet.classicAddress as wallet.address

natenichols:
Aliases `wallet.classicAddress` as `wallet.address`. 

### Context of Change
This was feedback from @mDuo13, as `wallet.classicAddress` is frequently used, it is appropriate to alias it as `wallet.address` for brevity.

### Type of Change
- [ ] Documentation Updates
- [ ] Release


JST5000 (approved):

intelliot (approved):
We kinda wanted to "force" exchanges/wallets to support X-addresses so that users would never have to enter a separate destination tag. But that remains a long way off -- maybe something for xrpl.js 3.0... It's fine to alias this in the meantime, but someday, there might be a custodial wallet where the `address` is actually an X-address (with a tag included in it). Not sure if our `Wallet` class will handle "custodial" wallet use cases. 🤔

PR #1690, opened 2021-10-04
test: subscription standalone integration tests

JST5000:

Add integration tests for subscriptions that can be tested in standalone node.

### Context of Change

Subscription types can be found here: https://xrpl.org/subscription-methods.html 
Of these, the following recieved test cases:
   * 'transaction'
   * 'transactions_proposed'
   * 'ledger'
   * 'path_find'

These were unable to be tested in standalone node: 
   * 'consensus'
   * 'manifests'
   * 'peer_status'
   * 'validations'
   * 'server'

### Type of Change

## Before / After

* Added tests for subscriptions that are testable in standalone mode
* Updated LedgerStream and PathFindStream to match the rippled response

## Test Plan

CI Passes


src/models/methods/subscribe.ts
```diff
@@ -33,16 +33,16 @@ interface BaseStream {
   type: string
 }
 
-export interface LedgerStream extends BaseStream {
```
mvadari, line 36: why does this not extend baseStream anymore?

src/models/methods/subscribe.ts
```diff
@@ -33,16 +33,16 @@ interface BaseStream {
...
-  type: 'ledgerClosed'
+export interface LedgerStream {
+  type?: 'ledgerClosed'
```
mvadari, line 37: Why is this type optional?

test/integration/requests/pathFind.ts
```diff
@@ -45,4 +52,57 @@ describe('path_find', function () {
...
+      id: 10,
+      alternatives: [],
+    }
+
+    const event = new Promise((resolve, reject) => {
+      const client: Client = this.client
+      client.on('path_find', (path) => {
+        assert.equal(path.type, 'path_find')
+        assert.deepEqual(_.omit(path, 'id'), _.omit(expectedStreamResult, 'id'))
+        resolve('success')
```
mvadari, line 98:
```suggestion
        resolve()
```

test/integration/requests/pathFind.ts
```diff
@@ -45,4 +52,57 @@ describe('path_find', function () {
...
+      })
+    })
```
mvadari, line 100: The better way to handle this is to use `done` instead of this promise structure. See some of the tests in `test/connection.ts`.

test/integration/requests/subscribe.ts
```diff
@@ -0,0 +1,177 @@
...
+// how long before each test case times out
+const TIMEOUT = 20000
+
+async function createTxHandlerTest(
+  client: Client,
+  wallet: Wallet,
+  subscriptionStream: StreamType,
+): Promise<unknown> {
+  const event = new Promise((resolve, reject) => {
+    client.on('transaction', (tx) => {
```
mvadari, line 29: ditto

mvadari (changes requested):
You should use `done` instead of returning a `Promise` with the `.on` handler

natenichols:
Will review once @mvadari's comments have been addressed, I agree with all of them.

test/integration/requests/pathFind.ts
JST5000: The reason that this function can't use the `done` structure is because I was using `await` in order to get access to the ledger's response, and I can't use a callback and await in the same test (It gives an error that the resolution is ambiguous because await makes the function async, which forces it to return a promise and mocha can't tell if it's supposed to finish when the promise resolves or when the callback happens). 

So to avoid that issue I wrapped the callback into a promise to clarify that the test ends when that promise returns.

Maybe the underlying problem is that I don't need to use `await`s?

src/models/methods/subscribe.ts
JST5000: The type is optional here because the response from a 'ledger' `SubscribeRequest` omits the type, but later updates on the `LedgerStream` include the type.

Here's the docs and quote: https://xrpl.org/subscribe.html#response-format
> Stream: ledger - Information about the ledgers on hand and current fee schedule. This includes the same fields as a ledger stream message, except that it omits the type and txn_count fields.


Would a better solution be to have a separate type that's just for the response from a ledger subscribe request?

src/models/methods/subscribe.ts
JST5000: This does not extend baseStream anymore because I wanted to make the `type` property optional and `type` is the only field in baseStream. (See below for why I wanted to make `type` optional)

src/models/methods/subscribe.ts
mvadari: That response is actually different from the actual stream. That is what the `subscribe` request returns, which is different than what the listener is listening to.

PR #1692, opened 2021-10-05
add Regular Key Pair support

khancode:

Adds Regular Key Pair support to `Wallet` class.

### Context of Change

Refactors `Wallet` constructor to add a `WalletOptions` param that contains:
- `classicAddress` that corresponds to its Regular Key Pair.
- `seed` used to derive the account keys.

### Type of Change

## Test Plan

Adds a test to verify `Wallet` constructor supports Regular Key Pair initialization.

### DIFF
--- a/src/wallet/index.ts
+++ b/src/wallet/index.ts
@@ -31,6 +31,13 @@ function hexFromBuffer(buffer: Buffer): string {
   return buffer.toString('hex').toUpperCase()
 }
 
+interface WalletOptions {
+  // The classicAddress corresponding to its Regular Key Pair.
+  classicAddress?: string
+  // The seed used to derive the account keys.
+  seed?: string
+}
+
 /**
  * A utility for deriving a wallet composed of a keypair (publicKey/privateKey).
  * A wallet can be derived from either a seed, mnemnoic, or entropy (array of random numbers).
@@ -39,12 +46,6 @@ function hexFromBuffer(buffer: Buffer): string {
 class Wallet {
   public readonly publicKey: string
   public readonly privateKey: string
-  /**
-   * This only is correct if this wallet corresponds to your
-   * [master keypair](https://xrpl.org/cryptographic-keys.html#master-key-pair). If this wallet represents a
-   * [regular keypair](https://xrpl.org/cryptographic-keys.html#regular-key-pair) this will provide an incorrect address.
-   * TODO: Add support for Regular Keys to Wallet (And their corresponding impact on figuring out classicAddress).
-   */
   public readonly classicAddress: string
   public readonly seed?: string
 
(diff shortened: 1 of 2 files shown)


src/wallet/index.ts
```diff
@@ -31,6 +31,13 @@ function hexFromBuffer(buffer: Buffer): string {
   return buffer.toString('hex').toUpperCase()
 }
 
+interface WalletOptions {
+  // The classicAddress corresponding to its Regular Key Pair.
+  classicAddress?: string
+  // The seed used to derive the account keys.
+  seed?: string
+}
```
JST5000, line 39: If this is only used in the normal wallet constructor, I'm not sure what the seed option would be used for (Since the constructor already has you set up your public and private keys, which should correspond to the Regular Key that you want to sign with)

Unless I'm missing something I think we probably can just include `classicAddress` as the only optional parameter to the Wallet constructor

src/wallet/index.ts
```diff
@@ -53,13 +54,17 @@ class Wallet {
    *
    * @param publicKey - The public key for the account.
    * @param privateKey - The private key used for signing transactions for the account.
-   * @param seed - (Optional) The seed used to derive the account keys.
+   * @param opts - (Optional) The seed and classicAddress (corresponding to its Regular Key Pair) for the account.
```
JST5000, line 57: `opts` seems unrelated to regular keys. Maybe `regularKeyOpts` would highlight that relationship?

JST5000 (changes requested):
It should be possible to use any of the formats for generating a wallet to derive your regular keypair (Ex. You may have a seed that represents your regular key, or you may have a mnemonic) - So all wallet generator methods should allow for that. (And should probably have a test case for when used with a regular key)

src/wallet/index.ts
khancode: `seed` has a separate use case from Regular Key. We want to store the generated `seed` after calling `Wallet.generate()`. Otherwise, the client wouldn't know what seed was used to generate the wallet.

src/wallet/index.ts
mvadari: We use `options` or `opts` for additional optional options, so that you don't need to worry about the order of all the optional stuff and you just include what you need to use.

mvadari (approved):
lgtm

src/wallet/index.ts
JST5000: In this case would that be the seed for the keypair that corresponds with the `classicAddress`? (Or would the `seed` correspond to the `publicKey` and `privateKey` provided in the constructor?

(The reason I'm asking is because if `seed` corresponds to the `classicAddress` that might defeat the point of the regular keypair, which is supposed to let you do everything without the master keypair)

JST5000:
> Are we adding those tests to a separate PR instead of adding them to this PR?

Yes, I'm working on integration tests for regular keys (Multisigning specifically). I'll link when it's ready. 
(Also I just learned that if you highlight a comment on Github and press "r" it let's you kinda reply to it 😄 )

PR #1693, opened 2021-10-05
Refactor: Rewrite transaction signing

mukulljangid:
Make signing and multi-signing easier to call and use.

### Context of Change
Remove `xrpl-local sign` and rename `signTransaction` to `sign`
Return `hash` along with `tx_blob` from `sign()`

### Type of Change

## Before / After
Output for `sign()` now includes `hash` for convenience.

## Test Plan
Passes CI tests.


src/wallet/index.ts
```diff
@@ -162,17 +162,17 @@ class Wallet {
...
+   * @param forMultisign - Specify true/false if this is a multisign tx request.
    * @returns A signed transaction.
    * @throws ValidationError if the transaction is already signed or does not encode/decode to same result.
    */
-  public signTransaction(
+  public sign(
     this: Wallet,
     transaction: Transaction,
-    multisignAddress?: string,
+    forMultisign?: boolean,
```
mvadari, line 172: Why was this changed? There should still be an address here.

src/wallet/index.ts
```diff
@@ -162,17 +162,26 @@ class Wallet {
...
+  public sign(
     this: Wallet,
     transaction: Transaction,
-    multisignAddress?: string,
+    forMultisign?: boolean | string,
   ): string {
+    let multisignAddress: boolean | string = false
+    if (
+      typeof forMultisign === 'string' &&
+      forMultisign.trim().startsWith('X')
```
natenichols, line 178: I dont think you need this `trim()` here.

src/wallet/index.ts
```diff
@@ -162,17 +162,26 @@ class Wallet {
...
    * @param transaction - A transaction to be signed offline.
-   * @param multisignAddress - Multisign only. An account address corresponding to the multi-signature being added. If this
-   * wallet represents your [master keypair](https://xrpl.org/cryptographic-keys.html#master-key-pair) you can get your account address
-   * with the Wallet.getClassicAddress() function.
+   * @param forMultisign - Specify true/false if this is a multisign tx request.
    * @returns A signed transaction.
    * @throws ValidationError if the transaction is already signed or does not encode/decode to same result.
    */
-  public signTransaction(
+  // eslint-disable-next-line max-lines-per-function --- checks a lot of stuff.
```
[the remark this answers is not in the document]
mukulljangid: Oh no, I just put that as a placeholder to see if the function was working.

PR #1697, opened 2021-10-05
Link to xrpl-announce (updated from ripple-lib-announce)

intelliot:

We renamed the Google Group to xrpl-announce. In the future, the same list may be used for releases of other client libraries as well. It's a small list (less than 60 subscribers) and not very noisy, so it should be okay to include release announcements about libraries in other languages, especially since many developers are polyglots. Of course, we hope the list will grow in the future.

### Context of Change

Historically, many developers asked for a way to get notified by email when a new version of the library is available. We created ripple-lib-announce for that purpose. Now, this library has been renamed to xrpl.js, so we renamed the mailing list to xrpl-announce.

### Type of Change

- [x] Documentation Updates

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,6 @@
-# ripple-lib Release History
+# xrpl.js (ripple-lib) Release History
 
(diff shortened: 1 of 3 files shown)


mvadari:
Can we add this link to the other libraries as well?

natenichols (approved):

intelliot:
@mvadari ok for me to click `Squash and merge` on this?

natenichols:
@intelliot go for it

PR #1698, opened 2021-10-06
fix: remove response.status from returned object

mvadari:

Title says it all. This PR also cleans up some integration tests as a result of this change.

### Context of Change

`response.status` is a confusing field - it just means that the API call was successful, not that the transaction was a success. It is unneeded after the response object is returned, because the client throws an error when it is not a success.

### Type of Change

## Test Plan

CI passes.


test/integration/index.ts
```diff
@@ -29,7 +29,7 @@ export * from './requests/ledgerClosed'
 export * from './requests/ledgerCurrent'
 export * from './requests/ledgerData'
 export * from './requests/ledgerEntry'
-export * from './requests/multisign'
+export * from './requests/submit_multisigned'
```
natenichols, line 32: should this be `submitMultisigned`?

natenichols (approved):

test/integration/index.ts
mvadari: yes, whoops

PR #1699, opened 2021-10-06
feat: Add parseAccountRootFlags

JST5000:

This takes a flag number and turns it into a nice interface of booleans.

### Context of Change

Maintaining a useful helper function from 1.x

### Type of Change

## Before / After

* Created an interface for AccountRootFlags
* Added the function to convert a number to the interface
* Added some tests for that function

## Test Plan

CI Passes


src/models/utils/flags.ts
```diff
@@ -25,6 +29,34 @@ import {
...
+    lsfGlobalFreeze: isSet(flags, AccountRootLedgerFlags.lsfGlobalFreeze),
+    lsfDefaultRipple: isSet(flags, AccountRootLedgerFlags.lsfDefaultRipple),
+    lsfDepositAuth: isSet(flags, AccountRootLedgerFlags.lsfDepositAuth),
+  }
+  return flagsInterface
+}
+
+function isSet(num: number, flag: AccountRootLedgerFlags): boolean {
+  return (num & flag) === flag
+}
```
khancode, line 58: Use the existing utility function `isFlagEnabled` to do this 🤠

src/models/ledger/accountRoot.ts
```diff
@@ -19,6 +19,18 @@ export default interface AccountRoot extends BaseLedgerEntry {
   TransferRate?: number
 }
 
+export interface AccountRootLedgerFlagsInterface {
+  lsfPasswordSpent: boolean
```
natenichols, line 23: These should be optional:

```ts
{
    lsfPasswordSpent: true
}
```

should be valid flags interface.

src/models/utils/flags.ts
```diff
@@ -25,6 +29,34 @@ import {
...
+ * Convert an AccountRoot Flags number into an interface for easy interpretation.
+ *
+ * @param flags - A number which is the bitwise and of all enabled AccountRootLedgerFlags.
+ * @returns An interface with all flags as booleans.
+ */
+// eslint-disable-next-line import/no-unused-modules -- Used by end user
+export function parseAccountRootFlags(
+  flags: number,
+): AccountRootLedgerFlagsInterface {
+  const flagsInterface = {
```
natenichols, line 42: Design question: Do we want to return every flag with either true/false, or do we want to only return flags set to true, with it implied that missing flags are false.

natenichols (changes requested):
Couple overarching questions, let me know what other reviewers think about these suggestions.

src/models/utils/flags.ts
khancode: Since they should be optional flags, I vote to only return flags set to `true`.

natenichols (changes requested):
I haven't run it, but I am 95% sure your code will return 
```ts
{
    lsfPasswordSpent: true
    lsfRequireDestTag: undefined
}
```

Do something like 
```
const flagsInterface = {}

if  (isFlagEnabled(flags, AccountRootLedgerFlags.lsfPasswordSpent))
    flagsInterface.lsfPasswordSpent = true;
    
...........

return flagsInterface
```

PR #1700, opened 2021-10-06
refactor: reorganize/rename flags interfaces

mukulljangid:
### Context of Change
Rename Flags interfaces to remove the huge word `Transaction` and add types (`asf`, `lsf`, `tf`) to it.
This will make it more obvious and easier to use.

### Type of Change

## Before / After
Rename AccountSetFlags to AccountSetAsfFlags and AccountSetTransactionFlags to AccountSetTfFlags.
Example: `AccountSetFlags` -> `AccountSetAsfFlags`
and `AccountSetTransactionFlags` -> `AccountSetTfFlags`

## Test Plan
Passes all CI tests.


JST5000 (approved):
LGTM! :)

src/models/ledger/accountRoot.ts
```diff
@@ -19,7 +19,7 @@ export default interface AccountRoot extends BaseLedgerEntry {
   TransferRate?: number
 }
 
-export enum AccountRootLedgerFlags {
+export enum AccountRootLsfFlags {
```
mvadari, line 22: This should just be `AccountRootFlags`

src/models/ledger/index.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable import/no-unused-modules -- Needs to export all ledger objects */
-import AccountRoot, { AccountRootLedgerFlags } from './accountRoot'
```
mvadari, line 2: ditto with the rest of these flags

mvadari (changes requested):
I think you might have misunderstood the renaming - for most of the flags, you're just renaming to `[Object]Flags` - no `lsf` or `tf`. The only place where you're including the flag type is with `AccountSet`, because there are 2 types of flags there.

PR #1703, opened 2021-10-06
docs: finalizes documentation for xrpl.js

natenichols:
Finalizes documentation for xrpl.js

### Context of Change
Note that this is only the JSDoc comments in the code, Documentation will be generated in a separate PR.

### Type of Change
- [x] Documentation Updates
- [ ] Release


natenichols:
> Why are there so many files being deleted?

Those are the `docs` directory. We're removing that in this PR, and then I'll tack on another to actually generate the docs.

PR #1709, opened 2021-10-07
fix: improve partial payment handling

mvadari:

This PR creates a new stream type for the client, `unsafeTransaction`. `unsafeTransaction` emits all transaction streams, while `transaction` now only emits the transaction streams that are not partial payments. It also adds to the `trace` a message that the transaction received is a partial payment.

### Context of Change

Better partial payment handling

### Type of Change

## Test Plan

CI passes. I had to fix an issue where one test modified an object that was used later, which made tests that should have been independent dependent.

### DIFF
--- a/src/client/partialPayment.ts
+++ b/src/client/partialPayment.ts
@@ -148,5 +148,7 @@ export function handleStreamPartialPayment(
     stream.warnings = warnings
 
     log('Partial payment received', JSON.stringify(stream))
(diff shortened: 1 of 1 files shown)


intelliot:
I think it would be simpler and make more sense for partial payments not to be considered unsafe or even a special case. Partial payments are not unsafe; using the "Amount" field is unsafe. It's safe to use the "delivered_amount" field.

PR #1750, opened 2021-10-18
build: webpack `bn.js` only once instead of many times

mvadari:

The webpacking was installing the big number library `bn.js` like 8 times for some reason. This PR makes it only installed once. This reduces the size of the library by 300kb (from 1.25mb to 972kb).

### Context of Change

make the browser build smaller

### Type of Change

## Test Plan

Still builds fine. Passes browser tests.


mvadari:
The high # of lines is because of `package-lock.json` - I deleted `package-lock.json` and reinstalled packages in the process of debugging. I can remove those changes from this PR if you want.

PR #1786, opened 2021-11-08
monorepo conversion

ledhed2222:

To make sense of this PR, look commit-by-commit starting with [this one](https://github.com/XRPLF/xrpl.js/pull/1786/commits/99c817ca8aec5eb1a00ea3e4015d3c8de81b2678).

What this PR does:

* Migrates to a JS monorepo structure. In this structure, all the published packages live in the `packages` directory. The top level contains all the common dev dependencies, CI config, and a common tsconfig.json file. `xrpl` is now found at `packages/xrpl`.
* Imports `ripple-address-codec`, `ripple-binary-codec`, `ripple-keypairs` as packages. All of the commit history for those repos are saved, but will look like they came from this PR. The original authors and dates are preserved.
* Hoists all dev dependencies to the top, so that things aren't re-installed for no reason. In order to accomplish this, several other changes had to happen:
  * `ripple-address-codec` was using `tslint` (an abandoned project IIRC) instead of `eslint`. Some packages conflict with our `eslint` config so `tslint` had to be removed in favor of the former. I added a basic `.eslintrc.js` file to that project which inherits from our shared config, but turns off all failing rules. A later task should be to address these linter errors.
  * `ripple-keypairs` was using `eslint`, but used several plugins that conflict with our shared config. These were all removed and the linter config for that project is now the shared config. Again, all failing rules were turned off. A later task should be to address them.
  * Since I was already doing this, I also changed `ripple-binary-codec` to use our shared config and turned off all failing rules.
  * Since I was already doing this, I made all the imported projects use the shared `@xrplf/prettier-config`.
  * `ripple-address-codec` and `ripple-binary-codec` use `jest` for testing, whereas `ripple-keypairs` and `xrpl` use `mocha`. There is a conflict between the `@types/jest` and `@types/mocha` packages in that the two clobber each other, meaning they cannot be used in the same project. To resolve this, I removed `@types/jest` and migrated the `ripple-address-codec` tests from TypeScript back to Javascript. `ripple-binary-codec`'s jest tests are already written in Javascript so there is no problem. To address this in the future, `ripple-address-codec` and `ripple-binary-codec`'s tests should be re-written in mocha and typescript.
* Uses a common CI workflow located at `.github/workflows/nodejs.yml`. This is the same workflow that we had for `xrpl`, but modified for the new monorepo structure. Where possible, `npm i` was replaced with `npm ci` for speed. Because we are now committed to GH actions, I removed all `.travis.yml` files.
* Includes a common `tsconfig.json` at the top-level that all subpckages inherit from. Uses typescript references to link all projects together. In the future we should hoist as many compiler options to the top as possible to ensure standardized typescript output.

### Context of Change

We want to switch to a monorepo because these 4 packages are tightly coupled. Ideally, they would have been one package in the first place and leveraged tree-shaking to handle different use cases, but since they are all already released with separate versions and user bases I think this is the best we can do.

### Type of Change

- [x] Documentation Updates
- [ ] Release

## Before / After

See above

## Test Plan

No tests were harmed in the creation of this PR! All existing tests pass.

## Future Tasks

- [ ] actually lint each of the new repos using our shared config. nearly everything is turned off just to make this migration happen
- [ ] migrate `ripple-address-codec` and `ripple-binary-codec` test suites to mocha/typescript. they are using jest/javascript now.
- [ ] further consolidate tsconfig.json settings by moving them to the top-level `tsconfig.json` file. Everything that was already shared is hoisted up there, but there are sure to be more opportunities to be DRY.
- [ ] migrate any issues, PRs, and other GitHub artifacts from these three repos here to this repo.
- [ ] forward the old repos to this one?


ledhed2222:
TODO 

- write a contributing.md doc explaining lerna, 

- determine why local dependencies are shown as local (IE `file:`) within package.json, which will make installations impossible

packages/ripple-binary-codec/src/types/st-object.ts
```diff
@@ -101,8 +101,8 @@ class STObject extends SerializedType {
 
     const xAddressDecoded = Object.entries(value).reduce((acc, [key, val]) => {
       let handled: JsonObject | undefined = undefined;
-      if (isValidXAddress(val)) {
-        handled = handleXAddress(key, val);
+      if (val && isValidXAddress(val.toString())) {
+        handled = handleXAddress(key, val.toString());
```
ledhed2222, line 105: typescript was updated in this change to a common, and more recent version. these changes are necessary to pass `tsc`.

packages/ripple-address-codec/src/index.test.ts
```diff
@@ -2,258 +2,272 @@ import {
   classicAddressToXAddress,
   xAddressToClassicAddress,
   isValidXAddress,
-  encodeXAddress
-} from './index'
+  encodeXAddress,
+} from "./index";
```
ledhed2222: i accidentally ran `eslint --fix` on this. i think we can just ignore these changes for now.

.github/workflows/nodejs.yml
```diff
@@ -23,9 +23,10 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm install
-    - run: npm run lint
+    - run: npm install -g npm@7
```
ledhed2222, line 26: using `npm` workspaces absolutely requires `npm 7` now

ledhed2222: FYI this is how facebook is handling this with lerna and github actions

mvadari:
> All of the commit history for those repos are saved, but will look like they came from this PR. The original authors and dates are preserved.
I assume this means that we don't want to squash-and-merge this PR like we do with all our other PRs.

mvadari (approved):

intelliot:
One thing to note is that the prior links to issues/PRs (and things like # <number>) will no longer point to the correct issues/PRs. I think this is fine but it is important to remember.

(Thanks @mvadari for pointing this out)

intelliot (approved):

PR #1798, opened 2021-11-15
initial pass at linting ripple-address-codec

ledhed2222:

did some of the easier linting fixes for `ripple-address-codec`. there's more left to do, but this takes a dent out of it for sure

### Type of Change

- [ ] Documentation Updates
- [ ] Release


packages/ripple-address-codec/src/utils.ts
```diff
@@ -19,32 +19,32 @@ export function seqEqual(arr1: Sequence, arr2: Sequence): boolean {
   return true
 }
 
-/**
- * Check whether a value is a sequence (e.g. array of numbers).
- *
- * @param val The value to check.
- */
+// .
```
JST5000, line 22: Why is there a period at the top here?

JST5000 (approved):
Small rule maybe out of place, but other than that looks good

packages/ripple-address-codec/src/utils.ts
mvadari: The linter adds periods to the end of lines - there's probably a space in this line, so the linter assumes there needs to be a period. The line should just be deleted.

mvadari (approved):

PR #1799, opened 2021-11-15
[Feature] Generate Wallet from secret numbers

nixer89:
## Generate Wallet from Secret Numbers

- added "fromSecretNumbers" method to Wallet instance
- added tests for new "fromSercretNumbers method

### Context of Change

This PR introduces a new feature. Many people are using XUMM. XUMM offers the recovery key for the user as [Secret Numbers](https://github.com/XRPLF/XRPL-Standards/issues/15)
To make it easy for developers to use their XUMM managed account for their development work, a new Method, "Wallet.fromSercretNumbers" is introduces. This method generates a new Wallet instance by the given Secret Numbers.

Alternatively, the developer had to first derive the family seed / secret key from the Secret Numbers and use the seed to initiate the Wallet instance.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Tests can be found in the index.ts of the /test/Wallet directory

## Future Tasks

The documentation has to be updated / generated. I did not run the "npm run docgen" target because it did not just added my method but changed lots of other things in the docs as well. So maybe someone else can run it.


packages/xrpl/src/Wallet/index.ts
```diff
@@ -112,7 +116,8 @@ class Wallet {
     privateKey: string,
     opts: {
       masterAddress?: string
-      seed?: string
+      seed?: string,
+      secretNumbers?: string
```
JST5000, line 120: Is secretNumbers needed here? (I don't see it used anywhere in the constructor)

* If it is needed, could you add `@param opts.secretNumbers` + a description to the documentation comment?

packages/xrpl/src/Wallet/index.ts
```diff
@@ -197,6 +202,35 @@ class Wallet {
     })
   }
 
+  /**
+   * Derives a wallet from secret numbers.
+   *
+   * @param secretNumbers - A string consisting of 8 times 6 numbers (whitespace delimited) used to derive a wallet.
+   * @param opts - (Optional) Options to derive a Wallet.
+   * @param opts.masterAddress - Include if a Wallet uses a Regular Key Pair. It must be the master address of the account.
```
JST5000, line 210: Please add `@param opts.algorithm` + description (Should be a description in the other methods with it)

packages/xrpl/src/Wallet/index.ts
```diff
@@ -197,6 +202,35 @@ class Wallet {
...
+    secretNumbers: Array<string> | string,
+    opts: { masterAddress?: string, algorithm?: ECDSA } = {},
+  ): Wallet {
+    let numbersArray:Array<string> = [];
+
+    if (typeof secretNumbers === 'string') {
+      numbersArray = Utils.parseSecretString(secretNumbers);
+    } else if (Array.isArray(secretNumbers)) {
+      numbersArray = secretNumbers;
+    }
```
JST5000, line 224: Since some users might be using javascript and pass in something unexpected, it's probably worth adding an exception if the type doesn't match. 
```suggestion
    } else {
         throw new ValidationError("Unable to parse the given secret numbers.")
    }
```

packages/ripple-keypairs/src/index.ts
```diff
@@ -25,7 +25,7 @@ function generateSeed(
   )
   const entropy = options.entropy ? options.entropy.slice(0, 16) : brorand(16)
   const type = options.algorithm === 'ed25519' ? 'ed25519' : 'secp256k1'
-  return addressCodec.encodeSeed(entropy, type)
+  return addressCodec.encodeSeed(Buffer.from(entropy), type)
```
ledhed2222, line 28: @nixer89 this is what fixed the build issue

packages/ripple-address-codec/src/xrp-codec.ts
```diff
@@ -202,14 +202,22 @@ export function encodeAccountID(bytes: Buffer): string {
   return codecWithXrpAlphabet.encode(bytes, opts)
 }
 
+/* eslint-disable import/no-unused-modules ---
+ * unclear why this is aliased but we should keep it in case someone else is
+ * importing it with the aliased name */
 export const encodeAddress = encodeAccountID
+/* eslint-enable import/no-unused-modules */
```
ledhed2222, line 209: FYI added this to suppress a linter warning.

packages/ripple-address-codec/src/xrp-codec.ts
```diff
@@ -202,14 +202,22 @@ export function encodeAccountID(bytes: Buffer): string {
...
 export function decodeAccountID(accountId: string): Buffer {
   const opts = { versions: [ACCOUNT_ID], expectedLength: 20 }
   return codecWithXrpAlphabet.decode(accountId, opts).bytes
 }
 
+/* eslint-disable import/no-unused-modules ---
+ * unclear why this is aliased but we should keep it in case someone else is
+ * importing it with the aliased name */
 export const decodeAddress = decodeAccountID
+/* eslint-enable import/no-unused-modules */
```
ledhed2222, line 220: FYI added this to suppress a linter warning.

packages/xrpl/package.json
```diff
@@ -27,7 +27,8 @@
     "ripple-address-codec": "^4.1.1",
     "ripple-binary-codec": "^1.1.3",
     "ripple-keypairs": "^1.0.3",
-    "ws": "^8.2.2"
+    "ws": "^8.2.2",
+    "xrpl-secret-numbers": "^0.3.3"
```
ledhed2222, line 31: should this be included in the monorepo?

JST5000: Maybe wait to bring it up to the top level until another package needs it?

ledhed2222: no i mean, should `xrpl-secret-numbers`'s source be in this monorepo. not sure where else it's used or how big it is, but that might make sense

packages/xrpl/src/Wallet/index.ts
nixer89: indeed not needed. removed

packages/xrpl/src/Wallet/index.ts
nixer89: added param + description

packages/xrpl/package.json
nixer89: I am sure the XRPL-Labs team uses this lib in their XUMM app. I think @WietseWind is the maintainer?
For now I would leave it as "external" import and would not add it to the monorepo. but maybe Wietse thinks this would be a good idea :)

WietseWind: @ledhed2222 It's a fairly small package, and the (non dev) dependencies overlap with `ripple-keypairs` (brorand, ripple-keypairs)

package-lock.json
```diff
@@ -1,18158 +1,7 @@
 {
   "name": "xrpl.js",
-  "lockfileVersion": 2,
```
JST5000, line 3: Was it intentional to delete 18,000 lines from the package-lock? (maybe a rebase is required?)

nixer89: before making my latest changes I have merged the develop branch into my "secret numbers" branch. When I check the history of the package-lock.json file of the develop branch, there are also some big changes which might have caused this:

https://github.com/XRPLF/xrpl.js/commit/9f19d771d24fa186ae23d586f380725f53a59059#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519

https://github.com/XRPLF/xrpl.js/commit/2b424276344b2aa8b8b76d621500f4d9e1436663#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519

mvadari: I believe this is happening because you're using a different version of npm. I'm on v7.20.3.

nixer89: I just updated to npm 7.24.x and the package-lock.json was not changed after "npm ci"
I still  believe it is because I have merged the develop branch into my branch

packages/xrpl/package.json
ledhed2222: that would be a good reason to maintain it within the monorepo :)

IMO, the criteria for whether it should be part of the monorepo:

1) is the purpose/scope of the package overlapping with the monorepo's? in this case i think it might be
2) does it depend on other packages w/in this monorepo, potentially requiring work on it and `ripple-keypairs` simultaneously.

package-lock.json
ledhed2222: `npm ci` doesn't change the package-lock, instead it reads from the package-lock for a faster install with deterministic package resolution. if the package-lock needs to be written, as in this case, you need `npm i`

ledhed2222: check out my documentation in `CONTRIBUTING.md` for how to add/remove packages in this repo

packages/xrpl/src/Wallet/index.ts
```diff
@@ -197,6 +198,38 @@ class Wallet {
...
+    opts: { masterAddress?: string; algorithm?: ECDSA } = {},
+  ): Wallet {
+    let numbersArray: string[] = []
+
+    if (typeof secretNumbers === 'string') {
+      numbersArray = Utils.parseSecretString(secretNumbers)
+    } else if (Array.isArray(secretNumbers)) {
+      numbersArray = secretNumbers
+    } else {
+      throw new ValidationError('Unable to parse the given secret numbers.')
```
ledhed2222, line 222: unnecessary use of `let` instead of `const` is a pet peeve of mine lol. you can rewrite this like the following to not require a let:


```javascript
const numbersArray: string[] = []

if (typeof secretNumbers === 'string') {
  numberArray.concat(Utils.parseSecretString(secretNumbers))
} else if (Array.isArray(secretNumbers)) {
  numbersArray.concat(secretNumbers)
} ....
```

JST5000 (dismissed):
LGTM

packages/xrpl/src/Wallet/index.ts
```diff
@@ -197,6 +198,38 @@ class Wallet {
...
+    }
+
+    const entropy = Utils.secretToEntropy(numbersArray)
+
+    return Wallet.fromEntropy(entropy, {
+      algorithm: opts.algorithm ?? ECDSA.secp256k1,
```
ledhed2222, line 228: should use `DEFAULT_ALGORITHM` as a fallback here, not `secp256k1`

ledhed2222: you can also just forward the `opts` dict entirely:

```javascript
return Wallet.fromEntropy(entropy, opts)
```

packages/xrpl/src/Wallet/index.ts
ledhed2222: alternatively, there's a typescript syntax that lets you use `let` but also tell typescript that you expect the variable to never _actually_ be nullish. IMO this is strange to look at, so i prefer the above with `const` but another option:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

```javascript
let numbersArray!: string []

if (typeof secretNumbers === 'string') {
  numbersArray = Utils.parseSecretString(secretNumbers)
} ....
```

packages/xrpl/src/Wallet/index.ts
nixer89: I decided to fallback to `secp256k1` because, the only wallet using "Secret Numbers" so far is XUMM. And they use `secp256k1` and not `ed25519`.

So if someone wants to create a xrpl.js "Wallet.fromSecretNumbers" he always have to define the algorithm `secp256k1` or else he ends up with a different address.

packages/xrpl/src/Wallet/index.ts
nixer89: used your first solution! :)

packages/xrpl/src/Wallet/index.ts
JST5000: That makes sense, could you add a comment with that context?

nixer89: Yes, will do tomorrow 🙂

ledhed2222: Sorry I don't agree. I think that XUMM in this case should override the default. We want this lib to be useful to anyone and we want people to default to ed25519

nixer89: I meant it the other way around: Users create a wallet in XUMM. This process gives them the "Secret Numbers" via `secp256k1`.

Now this XUMM user is also a developer. So he wants to build on XRPL, using xrpl.js and his Secret Numbers he got from XUMM. If `ed25519` is the default, the user always has to specify the algorithm when using the ".fromSecretNumbers" method.

What is the reason to always default to ed25519?

JST5000: The reason to default to ed25519 is that it's apparently got better performance. (See comment on the two algorithms here: https://xrpl.org/cryptographic-keys.html#signing-algorithms)

packages/xrpl/src/Wallet/index.ts
Realetive: @nixer89 , first solution will not change original `numbersArray` — `.concat()` does not change the existing arrays, but instead returns a new array. Example with `let` is right.

mvadari:
Should this PR be updated now?

intelliot:
@nixer89 will you be working on this?


(Alternatively - please let us know if this is something that you would like to ask RippleX to pick up instead)

mvadari:
> @nixer89 will you be working on this?
> 
> (Alternatively - please let us know if this is something that you would like to ask RippleX to pick up instead)

Isn't this already implemented?

ckniffen:
> > @nixer89 will you be working on this?
> > (Alternatively - please let us know if this is something that you would like to ask RippleX to pick up instead)
> 
> Isn't this already implemented?

I know it was added to `xrpl-p@ 2.0`.

mvadari:
> I know it was added to `xrpl-py@2.0`.

Whoops, I'm confusing Python and JS

packages/xrpl/test/wallet/index.test.ts
```diff
@@ -297,6 +297,86 @@ describe('Wallet', function () {
...
+    // TODO: Uncomment this test when the `deriveKeypair` fix is merged
+    // it('derives a wallet using algorithm ed25519', function () {
+    //   const algorithm = ECDSA.ed25519
+    //   const wallet = Wallet.fromSecretNumbers(secretNumbersString, {
+    //     algorithm,
+    //   })
+
+    //   assert.equal(wallet.publicKey, publicKeyED25519)
+    //   assert.equal(wallet.privateKey, privateKeyED25519)
+    // })
```
JST5000, line 357: This test fails because of this issue: https://github.com/XRPLF/xrpl.js/pull/2376

Commented with a TODO to re-add the test once that PR is resolved. (Open to other ideas about what's the best way to handle this :) )

ckniffen:
Ideally I would like to see this as a function separate from `Wallet` class.  Putting so much on the `Wallet` makes it impossible to tree shake.

JST5000:
> Ideally I would like to see this as a function separate from `Wallet` class. Putting so much on the `Wallet` makes it impossible to tree shake.

@ckniffen I think changing this to not be part of `Wallet` should be part of a bigger set of breaking changes to refactor how we handle Wallet generation as a whole. With our current setup, I think it makes sense to add this to the `Wallet` class because that's where all generators are currently. 

Do you think it's worth having this be off the `Wallet` class before a more sweeping change?

packages/xrpl/.eslintrc.js
```diff
@@ -34,6 +34,13 @@ module.exports = {
   plugins: [],
   extends: ['@xrplf/eslint-config/base'],
   rules: {
+    // Maintain existing line endings
```
xrpl-grants, line 37: What made this start becoming a problem?  This should be moved to a new PR.

ckniffen (approved):
As long as the other PR gets merged first and the tests are corrected I am good

jonathanlei (approved):

PR #1806, opened 2021-11-17
merge develop into master

ledhed2222:
 title says it all, but covers past merges into develop which moved the `docs` folder to the top, updated `HISTORY` files to account for latest release.

### Type of Change

- [ ] Documentation Updates
- [x] Release


intelliot:
There should not be a need to merge develop into master. We should be able to just fast forward master to match develop. It's effectively just a rename

JST5000 (approved):

JST5000:
> There should not be a need to merge develop into master. We should be able to just fast forward master to match develop. It's effectively just a rename

I think this might be referring to something different than you think - this is to move some docs, update the release steps, and update history.md (Not to remove the `develop` branch) - Or maybe I'm misinterpretting your commment?

intelliot:
> title says it all

This is a bad overview because the title definitely does not say it all

intelliot:
> I think this might be referring to something different than you think - this is to move some docs, update the release steps, and update history.md (Not to remove the `develop` branch) - Or maybe I'm misinterpretting your commment?

Indeed, you are misinterpreting my comment.

Also, in the future, PRs should explain the reason *why* - i.e. why do you want to merge develop into master? What's the motivation and what's the benefit?

We can explain next week

PR #1822, opened 2021-11-22
Fix Windows contributors being unable to build and test the xrpl mono repo

JST5000:

Two relevant changes:
1. Added OS specific build commands (So `npm build` works for both Windows and Linux-type machines)
2. Updated xrpl-js's tests to use @nixer89's `xrpl-local` fix so unit tests can run on windows

### Context of Change

@nixer89 was trying to contribute with a Windows machine and was running into problems building and testing their code.

### Type of Change

## Before / After

* Added `run-script-os` dependency to allow for OS-dependent build instructions 
   * This allows us to dynamically use `copy` for Windows instead of `cp` in ripple-binary-codec's build script
* Added `xrpl-local` as a dev dependency for `xrpl` instead of a manual file path (which only works on Linux-esque OS's)

## Test Plan

Manually build and run tests in Windows and Mac OS's to see if it works, and check that CI continues passing.


packages/xrpl/test/client.ts
```diff
@@ -1,6 +1,5 @@
 import { assert } from 'chai'
 import _ from 'lodash'
-
```
mvadari, line 3: Why is this change necessary?

JST5000: Linter was unhappy - it's a result of running --fix

nixer89 (approved):
Changes looks good to me. I can confirm that "npm run test" now works on my windows machine:


```
lerna success run Ran npm script 'test' in 4 packages in 30.1s:
lerna success - ripple-address-codec
lerna success - ripple-binary-codec
lerna success - ripple-keypairs
lerna success - xrpl
```

mvadari (approved):

PR #1826, opened 2021-11-29
rename: branch develop -> main; master -> old-master

### LINKED ISSUE #1824: Remove the develop branch and just use a single `main` branch @xrpl.js
With the recent change to a mono-repo structure, having both a `develop` branch and a `master` branch is awkward. The reason to have both is that `master` is supposed to reflect the last release, while `develop` reflects the current state of development efforts. With a mono-repo though, since development may happen across multiple packages simultaneously, merging develop into master may NOT reflect the latest releases. 

Ex. I might update `xrpl` with a small change, then write a patch for `ripple-address-codec` which I want to release. Master would then have both of these changes, even though I only released `ripple-address-codec`. 

So, we should probably remove develop, and just work directly with `master` since we're not really getting the benefits of a separate 'release' branch. 

Additionally, since we're changing the branch convention anyway, we may as well adopt the new default of calling `master` "`main`". 

One implication of this will be that people with existing PR's will likely have to change the target of their PR to point at `main` instead of `develop`, but that shouldn't be very hard.

intelliot:

Updates the code now that we've moved to just using `main` instead of `master` and `develop` branches.

Fixes #1824

### Context of Change

Having a mono-repo makes it very awkward to have both `master` and `develop`, and since we're changing branches, we may as well rename the default branch to be `main` to follow industry best practice.

### Type of Change

## Before / After

* Updated Github Actions to point to `main` and `1.x`
* Updated CodeQL to point to `main`
* Updated broken links to point to the correct branch
* Ran `npm run docgen` to regenerate the docs

## Test Plan

Manually clicked all the links with `develop` or `master` in them.


JST5000 (approved):
LGTM (Went though and updated the remaining references after our meeting, so it'd be good to get other eyes on this as well)

natenichols (approved):

PR #1830, opened 2021-11-30
fix: add workaround for rippled UNLModify encoding bug

mvadari:

There's a bug in rippled that means that UNLModify pseudotransactions aren't being encoded properly (the account isn't being included in the encoded pseudo-tx). This PR adds a workaround for that issue.

### Context of Change

We had reports of data ingestion issues of ledgers on mainnet starting a couple weeks ago, due to the NegativeUNL amendment being approved. 

### Type of Change

## Test Plan

CI passes. Had to fix a TS issue to get complication to work.


packages/ripple-binary-codec/src/types/st-object.ts
```diff
@@ -125,8 +127,11 @@ class STObject extends SerializedType {
       const associatedValue = field.associatedType.from(
         xAddressDecoded[field.name],
       )
-
-      bytes.writeFieldAndValue(field, associatedValue)
+      if ((associatedValue as unknown as Bytes).name === 'UNLModify') {
+        isUnlModify = true
+      }
```
ledhed2222, line 132: nbd but could just be

```javascript
const isUnlModify = (associatedValue as unknown as Bytes).name === 'UNLModify'
```

packages/ripple-binary-codec/src/serdes/binary-serializer.ts
```diff
@@ -145,9 +149,12 @@ class BinarySerializer {
...
    * @param value length encoded value to write to BytesList
    */
-  public writeLengthEncoded(value: SerializedType): void {
+  public writeLengthEncoded(
+    value: SerializedType,
+    isUnlModifyWorkaround = false,
+  ): void {
     const bytes = new BytesList()
-    value.toBytesSink(bytes)
+    if (!isUnlModifyWorkaround) value.toBytesSink(bytes)
```
ledhed2222, line 157: i'm surprised the linter lets us do this?

shouldn't allow omitting braces. w/e for this PR

however if think we should comment what this workaround is all about

packages/ripple-binary-codec/src/types/st-object.ts
mvadari: This actually doesn't work - it's not this field that requires the `isUnlModify`, it's the `Account` field that needs to know this

ledhed2222: woof.

ledhed2222: you might want to comment that since this could be a refactoring trap later

natenichols (approved):

packages/ripple-binary-codec/src/types/st-object.ts
JST5000: The way that works definitely took me a while to understand

JST5000 (approved):
LGTM

khancode (approved):
LGTM

ledhed2222 (approved):

PR #1854, opened 2021-12-02
Updating CONTRIBUTING with the correct steps

JST5000:

When using lerna with a protected branch, there's different steps required to publish.

### Context of Change

These were the steps used in the latest release of xrpl packages.

### Type of Change

- [X] Documentation Updates

## Before / After

* Updated release steps with the work around we have to use for a protected branch based on this thread: https://github.com/lerna/lerna/issues/1957#issuecomment-476894990

## Test Plan

N/A


khancode (approved):
LGTM, thanks for updating this! 😀

mvadari (approved):

PR #1857, opened 2021-12-02
fix: encoding issue with encoding non-traditional currency codes

mvadari:

This PR fixes an issue where some tokens with non-traditional currency codes weren't being encoded properly; they were being encoded as `XRP` for some reason. 

### Context of Change

Issue with data ingestion from the ledger (calculated tx hash wasn't equal to the rippled-calculated hash)

### Type of Change

## Test Plan

Added test. CI passes.

### DIFF
--- a/packages/ripple-binary-codec/test/hash.test.js
+++ b/packages/ripple-binary-codec/test/hash.test.js
@@ -69,6 +69,7 @@ describe('Currency', function () {
     expect(Currency.from('X8P').toJSON()).toBe('X8P')
     expect(Currency.from('USD').toJSON()).toBe('USD')
   })
+
   test('can be constructed from a Buffer', function () {
     const xrp = new Currency(Buffer.alloc(20))
     expect(xrp.iso()).toBe('XRP')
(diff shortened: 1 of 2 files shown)


natenichols (approved):

packages/ripple-binary-codec/test/hash.test.js
```diff
@@ -77,6 +78,12 @@ describe('Currency', function () {
     const currency = '015841551A748AD2C1F76FF6ECB0CCCD00000000'
     expect(Currency.from(currency).toJSON()).toBe(currency)
   })
+
+  test('Can handle other non-standard currency codes', () => {
+    const currency = '0000000000414C6F676F30330000000000000000'
```
JST5000, line 83: This is not quite right - For non-standard currency codes the 1st byte CANNOT be 0x00 - https://xrpl.org/currency-formats.html#nonstandard-currency-codes

mvadari: People are doing it anyways and we have no means of stopping them

natenichols: The encoding for non non-standard currency codes is whatever rippled says it is

In this case, rippled says that this should work like this, so it should work like this.

Because rippled doesnt actually enforce how non-standard currency codes work, we cant enforce how this works.

JST5000 (approved):

PR #1883, opened 2021-12-23
fix: `txnNotFound` error

mukulljangid:
Add validation for RippledError when a tx has not been validated and the lastLedgerSequence hasn't passed.

### Context of Change
I came across 2 transactions where the `submitAndWait` function failed to get the status of the transaction when the transaction is still in queue and tx command is not able to find it reason being it is not processed and hence doesn't reflect on rippled databases.

image

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Now, if the transaction is not found, we wait for it to show up(assuming it is going to be processed) or wait for the LastLedgerSequence to pass.

## Test Plan
Added test to demonstrate how it is reproduced. Passes CI tests.

### DIFF
--- a/packages/xrpl/src/sugar/submit.ts
+++ b/packages/xrpl/src/sugar/submit.ts
@@ -1,3 +1,5 @@
+/* eslint-disable max-params -- waitForFinalTxOutcome needs to display and do with more information to work perfectly. */
+/* eslint-disable max-lines-per-function -- waitForFinalTxOutcome has a lot of computations to do and conditions to check. */
 import { decode, encode } from 'ripple-binary-codec'
 
 import type { Client, SubmitRequest, SubmitResponse, Wallet } from '..'
@@ -81,10 +83,15 @@ async function submitAndWait(
     )
   }
 
-  await submitRequest(this, signedTx, opts?.failHard)
+  const response = await submitRequest(this, signedTx, opts?.failHard)
 
   const txHash = hashes.hashSignedTx(signedTx)
-  return waitForFinalTransactionOutcome(this, txHash, lastLedger)
+  return waitForFinalTransactionOutcome(
+    this,
+    txHash,
+    lastLedger,
+    response.result.engine_result,
+  )
 }
 
 // Helper functions
@@ -121,6 +128,7 @@ async function waitForFinalTransactionOutcome(
   client: Client,
   txHash: string,
   lastLedger: number,
+  submissionResult: string,
 ): Promise<TxResponse> {
   await sleep(LEDGER_CLOSE_TIME)
 
@@ -138,12 +146,18 @@ async function waitForFinalTransactionOutcome(
       if (message === 'txnNotFound') {
         if (lastLedger < (await client.getLedgerIndex())) {
           throw new XrplError(
-            `The latest ledger sequence ${latestLedger} is greater than the transaction's LastLedgerSequence (${lastLedger}).`,
+            `The latest ledger sequence ${latestLedger} is greater than the transaction's LastLedgerSequence (${lastLedger}).
+            Preliminary result: ${submissionResult}`,
           )
         }
-        return waitForFinalTransactionOutcome(client, txHash, lastLedger)
+        return waitForFinalTransactionOutcome(
+          client,
+          txHash,
+          lastLedger,
+          submissionResult,
+        )
       }
-      throw new Error(message)
+      throw new Error(`Preliminary result: ${submissionResult}.\n ${message}`)
     })
 
   if (txResponse.result.validated) {
@@ -151,11 +165,17 @@ async function waitForFinalTransactionOutcome(
   }
 
   if (lastLedger > latestLedger) {
-    return waitForFinalTransactionOutcome(client, txHash, lastLedger)
+    return waitForFinalTransactionOutcome(
+      client,
+      txHash,
+      lastLedger,
+      submissionResult,
+    )
   }
 
   throw new XrplError(
-    `The latest ledger sequence ${latestLedger} is greater than the transaction's LastLedgerSequence (${lastLedger}).`,
+    `The latest ledger sequence ${latestLedger} is greater than the transaction's LastLedgerSequence (${lastLedger}).
+    Preliminary result: ${submissionResult}`,
   )
 }
 
--- a/packages/xrpl/src/sugar/submit.ts
+++ b/packages/xrpl/src/sugar/submit.ts
@@ -134,6 +134,13 @@ async function waitForFinalTransactionOutcome(
 
   const latestLedger = await client.getLedgerIndex()
 
+  if (lastLedger < latestLedger) {
+    throw new XrplError(
+      `The latest ledger sequence ${latestLedger} is greater than the transaction's LastLedgerSequence (${lastLedger}).
+      Preliminary result: ${submissionResult}`,
+    )
+  }
+
   const txResponse = await client
     .request({
       command: 'tx',
@@ -144,12 +151,6 @@ async function waitForFinalTransactionOutcome(
       // eslint-disable-next-line @typescript-eslint/consistent-type-assertions,@typescript-eslint/no-unsafe-member-access -- L131
       const message = error.data.error as string
       if (message === 'txnNotFound') {
-        if (lastLedger < (await client.getLedgerIndex())) {
-          throw new XrplError(
-            `The latest ledger sequence ${latestLedger} is greater than the transaction's LastLedgerSequence (${lastLedger}).
-            Preliminary result: ${submissionResult}`,
-          )
-        }
         return waitForFinalTransactionOutcome(
           client,
           txHash,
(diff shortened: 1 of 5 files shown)


packages/xrpl/test/integration/requests/submit.ts
```diff
@@ -74,4 +76,25 @@ describe('submit', function () {
 
     assert.deepEqual(submitResponse, expectedResponse)
   })
+
+  // Reproduces the 'txnNotFound' case
+  it('submit & quickly check status without waiting', async function () {
```
mvadari, line 81: Does this actually replicate the situation? It seems that it's not even using the new code that you wrote.

packages/xrpl/src/sugar/submit.ts
```diff
@@ -122,15 +122,25 @@ async function waitForFinalTransactionOutcome(
...
-    transaction: txHash,
-  })
-  if (txResponse.result.validated) {
+  const txResponse = await client
+    .request({
+      command: 'tx',
+      transaction: txHash,
+    })
+    .catch((error) => {
+      const castedError = error.data as { error: string }
```
mvadari, line 131:
```suggestion
      const castedError = error.data.error as string
```
Does this not work?

mukulljangid: yes it works

packages/xrpl/test/integration/requests/submit.ts
mukulljangid: removed.

packages/xrpl/src/sugar/submit.ts
```diff
@@ -122,10 +122,21 @@ async function waitForFinalTransactionOutcome(
...
+    .request({
+      command: 'tx',
+      transaction: txHash,
+    })
+    .catch(async (error) => {
+      // this type of error does not have its interface, type-assert and extract the value we need.
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions,@typescript-eslint/no-unsafe-member-access -- L131
+      const message = error.data.error as string
+      if (message === 'txnNotFound') {
+        return waitForFinalTransactionOutcome(client, txHash)
```
natenichols, line 135: This will never actually hit the condition where we check if we're past latestLedgerSequence.

natenichols:
```
const txResponse = await client
    .request({
      command: 'tx',
      transaction: txHash,
    })
    .catch((error) => {
      if (message !== 'txnNotFound') {
            Promise.reject(error)
      }
      return null
    })

  if (txResponse != null && txResponse.result.validated) {
    return txResponse
  }
```

natenichols: If the transaction never gets validated it will just loop infinitely

packages/xrpl/src/sugar/submit.ts
```diff
@@ -119,29 +120,42 @@ async function submitRequest(
...
+    .request({
+      command: 'tx',
+      transaction: txHash,
+    })
+    .catch(async (error) => {
+      // this type of error does not have its interface, type-assert and extract the value we need.
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions,@typescript-eslint/no-unsafe-member-access -- L131
+      const message = error.data.error as string
+      if (message === 'txnNotFound') {
+        if (lastLedger > (await client.getLedgerIndex())) {
```
natenichols, line 139: This should be the other way around.
```ts
lastLedger < (await client.getLedgerIndex()
```

packages/xrpl/src/sugar/submit.ts
```diff
@@ -119,29 +120,42 @@ async function submitRequest(
...
 
-  const txLastLedger = txResponse.result.LastLedgerSequence
-  if (txLastLedger == null) {
-    throw new XrplError('LastLedgerSequence cannot be null')
-  }
-  const latestLedger = await client.getLedgerIndex()
-
-  if (txLastLedger > latestLedger) {
-    return waitForFinalTransactionOutcome(client, txHash)
+  if (lastLedger > latestLedger) {
```
natenichols, line 153: Same

packages/xrpl/src/sugar/submit.ts
natenichols: We want to throw when the current index is greater than the lastLedgerSequence

mukulljangid: yes

packages/xrpl/src/sugar/submit.ts
mukulljangid: not this, here we are making the call again, only if the last ledger seq is greater

packages/xrpl/src/sugar/submit.ts
```diff
@@ -119,29 +120,42 @@ async function submitRequest(
...
+    return waitForFinalTransactionOutcome(client, txHash, lastLedger)
   }
 
   throw new XrplError(
-    `The latest ledger sequence ${latestLedger} is greater than the transaction's LastLedgerSequence (${txLastLedger}).`,
+    `The latest ledger sequence ${latestLedger} is greater than the transaction's LastLedgerSequence (${lastLedger}).`,
   )
```
mvadari, line 159: Why are we erroring in 2 places?

mvadari (commented):
Can we take care of this issue while we're here? https://github.com/XRPLF/xrpl-py/pull/325

packages/xrpl/src/sugar/submit.ts
mukulljangid: CI passed right now

packages/xrpl/src/sugar/submit.ts
natenichols: Ah, you're right, because this one is recursing. Good catch.

natenichols (approved):
I'm okay w/ this. We should have @mvadari review as well.

packages/xrpl/src/sugar/submit.ts
```diff
@@ -119,29 +120,42 @@ async function submitRequest(
...
+    .request({
+      command: 'tx',
+      transaction: txHash,
+    })
+    .catch(async (error) => {
+      // this type of error does not have its interface, type-assert and extract the value we need.
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions,@typescript-eslint/no-unsafe-member-access -- L131
+      const message = error.data.error as string
+      if (message === 'txnNotFound') {
+        if (lastLedger < (await client.getLedgerIndex())) {
```
mvadari, line 139: This should use `latestLedger`

packages/xrpl/src/sugar/submit.ts
mvadari: Not a CI issue, a flow issue. We should only be erroring once for one issue. We should either be erroring for `lastLedgwrSequence` passing above or here, not both

packages/xrpl/src/sugar/submit.ts
mukulljangid: latest is the the value that is being computed on the RHS.

mukulljangid: so if last<latest, which means lastLedgerSeq has passed, we throw an error.

packages/xrpl/src/sugar/submit.ts
mukulljangid: my bad. let me see

packages/xrpl/src/sugar/submit.ts
mvadari: Yes, why is it being recalculated instead of just using the existing calculated value?

mukulljangid: got rid of it

packages/xrpl/src/sugar/submit.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable max-params -- waitForFinalTxOutcome needs to display and do with more information to work perfectly. */
```
mvadari, line 1: You should disable this just for the one function that needs it, not the whole file

packages/xrpl/src/sugar/submit.ts
```diff
@@ -119,29 +127,49 @@ async function submitRequest(
...
-  if (txLastLedger > latestLedger) {
-    return waitForFinalTransactionOutcome(client, txHash)
+  const txResponse = await client
+    .request({
+      command: 'tx',
+      transaction: txHash,
+    })
+    .catch(async (error) => {
+      // this type of error does not have its interface, type-assert and extract the value we need.
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions,@typescript-eslint/no-unsafe-member-access -- L131
```
mvadari, line 151: What is this explanation? I don't understand it

packages/xrpl/src/sugar/submit.ts
```diff
@@ -116,32 +122,53 @@ async function submitRequest(
...
-  }
+  const latestLedger = await client.getLedgerIndex()
 
-  const txLastLedger = txResponse.result.LastLedgerSequence
-  if (txLastLedger == null) {
-    throw new XrplError('LastLedgerSequence cannot be null')
+  if (lastLedger < latestLedger) {
+    throw new XrplError(
+      `The latest ledger sequence ${latestLedger} is greater than the transaction's LastLedgerSequence (${lastLedger}).
+      Preliminary result: ${submissionResult}`,
```
mvadari, line 139: This should use string concatenation so you don't have the `\n` in the error message

mvadari (approved):

DEa1188:
> ## High Level Overview of Change
> Add validation for RippledError when a tx has not been validated and the lastLedgerSequence hasn't passed.
> 
> 
> ### Context of Change
> I came across 2 transactions where the `submitAndWait` function failed to get the status of the transaction when the transaction is still in queue and tx command is not able to find it reason being it is not processed and hence doesn't reflect on rippled databases.
> 
> image
> 
> 
> 
> ### Type of Change
> 
> 
> 
> - [X] Bug fix (non-breaking change which fixes an issue)
> - [ ] New feature (non-breaking change which adds functionality)
> - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
> - [ ] Refactor (non-breaking change that only restructures code)
> - [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
> - [ ] Documentation Updates
> - [ ] Release
> 
> ## Before / After
> Now, if the transaction is not found, we wait for it to show up(assuming it is going to be processed) or wait for the LastLedgerSequence to pass.
> 
> 
> ## Test Plan
> Added test to demonstrate how it is reproduced. Passes CI tests.
> 
> 
>

PR #1892, opened 2022-01-06
Update: Add hash type to base transaction interface (common.ts)

interc0der:
## Add hash & date type to base transaction interface (common.ts)

- Simple additions to the base transaction interface within ./packages/xrpl/src/models/transaction/common.ts

### Context of Change

All transaction types include a unique hash such that it can be identified and validated on chain. All transaction hashes are present in string type.

All transactions include a date - UNIX timestamp (in seconds). This may have been formerly known as executed_time.

https://xrpl.org/transaction-basics.html#identifying-transactions

### Type of Change

[x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/xrpl/src/models/transactions/common.ts
+++ b/packages/xrpl/src/models/transactions/common.ts
@@ -160,6 +160,10 @@ export interface BaseTransaction {
    */
   TxnSignature?: string
   /**
+   * A unix timestamp (in seconds) marking the time of the transaction
+   */
+  date?: Number
+  /**
    * Every signed transaction has a unique "hash" that identifies it. 
    * The transaction hash can be used as a "proof of payment" to verify the final status.
    */
--- a/packages/xrpl/src/models/transactions/common.ts
+++ b/packages/xrpl/src/models/transactions/common.ts
@@ -163,7 +163,7 @@ export interface BaseTransaction {
    * Every signed transaction has a unique "hash" that identifies it. 
    * The transaction hash can be used as a "proof of payment" to verify the final status.
    */
-     hash?: string
+  hash?: string
 }
 
 /**
--- a/packages/xrpl/src/models/transactions/common.ts
+++ b/packages/xrpl/src/models/transactions/common.ts
@@ -159,6 +159,11 @@ export interface BaseTransaction {
    * account it says it is from.
    */
   TxnSignature?: string
+  /**
+   * Every signed transaction has a unique "hash" that identifies it. 
+   * The transaction hash can be used as a "proof of payment" to verify the final status.
+   */
+     hash?: string
 }
 
 /**


JST5000:
Having chatted with @mDuo13 about this, it seems that the `hash` and `date` fields are not canonical parts of Transaction objects, but rather are added in afterwards when requesting info about them. 

That means that those fields are not information that gets serialized with the transaction in `rippled`. 

Instead, the fields are added when information is requested about the transaction (Ex. on the `tx` command)
https://github.com/XRPLF/xrpl.js/blob/7041a0bbd3e3f0d1e7316f85fd791a718cbfdf5d/packages/xrpl/src/models/methods/tx.ts#L41

It seems the code is missing the `date` field, but does have the hash field noted there. I'm not entirely sure which other requests will result in a `date` field being added.

JST5000 (changes requested):
Detailed comment above - This does seem to be missing, but the edits should happen on the `Request` objects involving transactions instead of the Transaction model itself (Because the Transaction model is trying to mirror the canonical `rippled` format)

intelliot:
> the hash and date fields are not canonical parts of Transaction objects, but rather are added in afterwards when requesting info about them. That means that those fields are not information that gets serialized with the transaction in rippled.

Incidentally, this is why some fields start with a lowercase letter, while others start with a capital letter. The fields with names that start with a capital letter are stored directly on the ledger. If it starts with a lowercase letter, then it's a derived field: not directly stored, but calculated/determined at request time.

interc0der:
Thank you for deliberating on this issue. 

I had encounter the missing typing on the `Subscribe` method. 
Not the `tx` method.  Your update to the `TxResponse` typing is still appropriate, but it may not be a comprehensive fix. Take a look at the `Subscribe` method.

So if the `hash` and `date` are not serialized within transactions in rippled, how is this information added to the transaction objects within a stream? Since it is not canonical to the transaction object, these must be getting added on each stream output. 

If you track the `TransactionStream` within the`Subscribe` method, it refers back to the base transaction typing which I had modified. 

Sample Initialization
```
     await api.connect();
     await api.request({
         command: 'subscribe',
         streams: [
           'transactions'
          ]
     })

     api.on('transaction', (tx:TransactionStream) => {
         console.log("Transaction Recieved")
         console.log(tx)
     })

```
Sample Output
```
Transaction Recieved
{
  engine_result: 'tecUNFUNDED_OFFER',
  engine_result_code: 103,
  engine_result_message: 'Insufficient balance to fund created offer.',
  ledger_hash: '7AA05F32AB4E1E7F4FB7E99209D6DE8B1A899CC96F7402FF97569E7B8E549107',
  ledger_index: 68822626,
  meta: {
    AffectedNodes: [ [Object] ],
    TransactionIndex: 64,
    TransactionResult: 'tecUNFUNDED_OFFER'
  },
  status: 'closed',
  transaction: {
    Account: 'r4AZpDKVoBxVcYUJCWMcqZzyWsHTteC4ZE',
    Fee: '15',
    Flags: 0,
    LastLedgerSequence: 68822628,
    Sequence: 98566330,
    SigningPubKey: '024E30BA54A70F4298A854B15554A5448305FDB866292C5C4543025D4218D1E94A',
    TakerGets: '3481759904',
    TakerPays: {
      currency: 'CNY',
      issuer: 'rJ1adrpGS3xsnQMb9Cw54tWJVFPuSdZHK',
      value: '20235.8492956083'
    },
    TransactionType: 'OfferCreate',
    TxnSignature: '3044022031CAF3118A1EFEC7FA229ACF1D1D527F94A9272CFCA364E6F0870B28C22480E502205427F56BAD6F544703A4A20F8BE6774C7CEAA86BD1A344854552076CDDC97719',
    date: 694751890,
    hash: '2EABCD6168FD00CE313DD872303C7E86BD55EC09D5E264F65B3C58DBC0F3645F',
    owner_funds: '0'
  },
  type: 'transaction',
  validated: true
}

```

JST5000:
Nice catch :) 

As a current workaround, the typing in TypeScript doesn't affect the data at runtime, so if you need to access those fields you can cast it to `Any` and just access them (The data should be there). 

I also opened #1893 to make the changes we discussed. (I was hoping to just add it to this ticket, but I couldn't figure out how to do that unfortunately :/ )

interc0der:
Thanks @JST5000 + @mDuo13  + @intelliot.
I will close this `PR`. All comments and change should move over to [#1893](https://github.com/XRPLF/xrpl.js/pull/1893)

PR #1893, opened 2022-01-06
Add missing fields to TxSubscription and Tx requests

JST5000:

Adds non-canonical fields which are returned on requests, but are not normally part of transactions on the ledger.

### Context of Change

See #1892 

### Type of Change

## Before / After

* Before `date` and `hash` were missing from these models even though the requests/subscriptions returned them.

## Test Plan

CI Passes

### DIFF
--- a/packages/xrpl/src/models/methods/subscribe.ts
+++ b/packages/xrpl/src/models/methods/subscribe.ts
@@ -275,7 +275,7 @@ export interface TransactionStream extends BaseStream {
     date?: number
     /**
      * Every signed transaction has a unique "hash" that identifies it.
-     * The transaction hash can be used as a "proof of payment" to verify the final status.
+     * The transaction hash can be used to look up its final status, which may serve as a "proof of payment"
      */
     hash?: string
   }
--- a/packages/xrpl/src/models/methods/subscribe.ts
+++ b/packages/xrpl/src/models/methods/subscribe.ts
@@ -269,7 +269,14 @@ export interface TransactionStream extends BaseStream {
   meta?: TransactionMetadata
   /** The definition of the transaction in JSON format. */
   transaction: Transaction & {
-    date?: number
+    /**
+     * A unix timestamp (in seconds) marking the time of the transaction
+     */
+    date?: Number
+    /**
+     * Every signed transaction has a unique "hash" that identifies it.
+     * The transaction hash can be used as a "proof of payment" to verify the final status.
+     */
     hash?: string
   }
   /**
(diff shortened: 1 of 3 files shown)


mvadari (approved):
Thanks for catching this!

packages/xrpl/src/models/methods/subscribe.ts
```diff
@@ -268,7 +268,17 @@ export interface TransactionStream extends BaseStream {
    */
   meta?: TransactionMetadata
   /** The definition of the transaction in JSON format. */
-  transaction: Transaction
+  transaction: Transaction & {
+    /**
+     * A unix timestamp (in seconds) marking the time of the transaction
```
intelliot, line 273: Are you sure this is a unix timestamp? I would expect "Ripple time" instead.

https://xrpl.org/basic-data-types.html#specifying-time

intelliot (changes requested):

packages/xrpl/src/models/methods/subscribe.ts
JST5000: Oops - just checked by looking at the time of a transaction from a few seconds ago and it is indeed "Ripple time"

interc0der: My mistake. I am not on "Ripple time"

JST5000: Most code isn't :P

packages/xrpl/src/models/methods/subscribe.ts
```diff
@@ -268,7 +268,17 @@ export interface TransactionStream extends BaseStream {
...
   /** The definition of the transaction in JSON format. */
-  transaction: Transaction
+  transaction: Transaction & {
+    /**
+     * This number measures the number of seconds since the "Ripple Epoch" of January 1, 2000 (00:00 UTC)
+     */
+    date?: number
+    /**
+     * Every signed transaction has a unique "hash" that identifies it.
+     * The transaction hash can be used as a "proof of payment" to verify the final status.
```
intelliot, line 278:
```suggestion
     * The transaction hash can be used to look up its final status, which may serve as a "proof of payment"
```

packages/xrpl/src/models/methods/subscribe.ts
```diff
@@ -336,7 +346,17 @@ export interface OrderBookStream extends BaseStream {
...
   meta: TransactionMetadata
-  transaction: Transaction | ModifiedOfferCreateTransaction
+  transaction: (Transaction | ModifiedOfferCreateTransaction) & {
+    /**
+     * This number measures the number of seconds since the "Ripple Epoch" of January 1, 2000 (00:00 UTC)
+     */
+    date?: number
+    /**
+     * Every signed transaction has a unique "hash" that identifies it.
+     * The transaction hash can be used as a "proof of payment" to verify the final status.
```
intelliot, line 356:
```suggestion
     * The transaction hash can be used to look up its final status, which may serve as a "proof of payment"
```

intelliot (approved):
minor suggestion but it's ok to take it or leave it

PR #1906, opened 2022-01-27
modifiedNodes bug fix

sampleblood:
1. AffectedNodes may have not only ModifiedNode, but also createNode and DeletedNode, then crash


mvadari:
This and #1907 are doing the same thing - do you have a preference on which PR we use?

mvadari:
Closing because this is identical to #1907

Issue #1919, opened 2022-02-02
Binary decode transactions with "unsafe" currency codes


Silkjaer:
Decoding certain objects with "unsafe" currency codes returns an error.

One example is transaction `64CE88666FB8C22BF0025C6E78E41B9E8A6F8D469C2A7101152C919CC3A91118` that is a TrustSet for a currency code `0000000000000000000000005852500000000000`.

Since the currency code decodes to "XRP", the binary decoder fails with "Disallowed currency code: to indicate the currency XRP you must use 20 bytes of 0s".

This makes it impossible to decode historical data from rippled using ripple-binary-codec.

This seem to be introduced after 1.1.3, possibly with #1786.

mvadari:
Relevant commit: https://github.com/ripple/ripple-binary-codec/pull/137

PR #1920, opened 2022-02-02
Remove exception from decoding issued currencies which evaluate to 'XRP'

JST5000:

Removed exception raised when decoding an issued currency which evaluated to `XRP`

### Context of Change

In response to issue https://github.com/XRPLF/xrpl.js/issues/1919
The error makes it impossible to use ripple-binary-codec to decode ledger history since dodge `XRP` issued currencies have been added to the ledger at some points.

### Type of Change

## Before / After

* Removed raised exception
* Revised test cases to use new chai syntax
* Updated test to reflect the new expected behavior

## Test Plan

CI passes


packages/ripple-binary-codec/test/tx-encode-decode.test.js
```diff
@@ -22,7 +26,7 @@ describe('encoding and decoding tx_json', function () {
   test('can encode tx_json without Amount or Fee', function () {
     const encoded = encode(tx_json)
     const decoded = decode(encoded)
-    expect(tx_json).toEqual(decoded)
```
mvadari, line 25: Why was this changed?

packages/ripple-binary-codec/test/tx-encode-decode.test.js
```diff
@@ -48,7 +52,30 @@ describe('encoding and decoding tx_json', function () {
...
+      Fee: '10',
+      SigningPubKey:
+        '023076CBB7A61837F1A23D4A3DD7CE810B694992EB0959AB9D6F4BB6FED6F8CC26',
+      TxnSignature:
+        '304502202D0CD77D8E765E3783C309CD663723B18406B7950A348A6F301492916A990FC70221008A76D586111205304F10ADEFDFDDAF804EF202D8CD1E492DC6E1AA8030EA1844',
+      Account: 'rPtfQWdcdhuL9eNeNv5YfmekSX3K7vJHbG',
+    }
+    const encoded =
+      '1200142200000000240000001363D5071AFD498D00000000000000000000000000005852500000000000585E1F3BD02A15D6185F8BB9B57CC60DEDDB37C168400000000000000A7321023076CBB7A61837F1A23D4A3DD7CE810B694992EB0959AB9D6F4BB6FED6F8CC267447304502202D0CD77D8E765E3783C309CD663723B18406B7950A348A6F301492916A990FC70221008A76D586111205304F10ADEFDFDDAF804EF202D8CD1E492DC6E1AA8030EA18448114FAFD540634705D7B1B9C8C31BFE2CC40326EB404'
+    const decoded = decode(encoded)
```
mvadari, line 77: Does rippled ban currencies that are equivalent to XRP? If that's true, then this should be fine. If not, then this should fail.

packages/ripple-binary-codec/test/tx-encode-decode.test.js
JST5000: When running it was saying that `chai has no attribute toEqual`, so I think the syntax may have changed to `.deep.equal`.

packages/ripple-binary-codec/test/tx-encode-decode.test.js
JST5000: Well we've got one on ledger which has this encoded value, so I think it doesn't (or at least historically didn't) which leads me to believe this should be fine.

mvadari: It wasn't XRP and shouldn't be XRP. It was `0x0000000000000000000000005852500000000000`

packages/ripple-binary-codec/test/tx-encode-decode.test.js
mvadari: This isn't chai - this is jest. Don't worry about changing the whole structure of the tests in this PR.

packages/ripple-binary-codec/test/tx-encode-decode.test.js
```diff
@@ -1,3 +1,4 @@
+const { expect } = require('chai')
```
mvadari, line 4: This shouldn't be here.

packages/ripple-binary-codec/test/tx-encode-decode.test.js
JST5000: Oh, you mean does rippled accept **encoded** `xrp` codes. Let me check.

JST5000: Submitting an issued currency with code `0x0000000000000000000000005852500000000000` results in temBAD_CURRENCY (While sending an issued currency with code `XRP` results in `Invalid native currency`) So they both error out, but give different messages. (Note that changing the 25 to 24 results in a successful issued currency trust set, so it's specifically if it decodes to XRP that creates the temBAD_CURRENCY as far as I can tell)

JST5000: So I think the answer is: Yes, rippled bans currencies that are equivalent to XRP

packages/ripple-binary-codec/test/tx-encode-decode.test.js
JST5000: Ah, I'm not sure why that ended up in there in the 1st place, but it was the root of the other issues I was running into. Reverting the changes to the test syntax.

packages/ripple-binary-codec/test/tx-encode-decode.test.js
JST5000: That makes sense

packages/ripple-binary-codec/test/tx-encode-decode.test.js
```diff
@@ -48,7 +52,30 @@ describe('encoding and decoding tx_json', function () {
...
+      SigningPubKey:
+        '023076CBB7A61837F1A23D4A3DD7CE810B694992EB0959AB9D6F4BB6FED6F8CC26',
+      TxnSignature:
+        '304502202D0CD77D8E765E3783C309CD663723B18406B7950A348A6F301492916A990FC70221008A76D586111205304F10ADEFDFDDAF804EF202D8CD1E492DC6E1AA8030EA1844',
+      Account: 'rPtfQWdcdhuL9eNeNv5YfmekSX3K7vJHbG',
+    }
+    const encoded =
+      '1200142200000000240000001363D5071AFD498D00000000000000000000000000005852500000000000585E1F3BD02A15D6185F8BB9B57CC60DEDDB37C168400000000000000A7321023076CBB7A61837F1A23D4A3DD7CE810B694992EB0959AB9D6F4BB6FED6F8CC267447304502202D0CD77D8E765E3783C309CD663723B18406B7950A348A6F301492916A990FC70221008A76D586111205304F10ADEFDFDDAF804EF202D8CD1E492DC6E1AA8030EA18448114FAFD540634705D7B1B9C8C31BFE2CC40326EB404'
+    const decoded = decode(encoded)
+    expect(expectedTx).deep.equal(decoded)
```
JST5000, line 78: This test now fails due to the decode issue. I'll need to rebase onto the fix in order to merge this.

packages/ripple-binary-codec/test/hash.test.js
```diff
@@ -50,10 +50,8 @@ describe('Hash256', function () {
 })
 
 describe('Currency', function () {
-  test('Will throw an error for dodgy XRP ', function () {
-    expect(() =>
-      Currency.from('0000000000000000000000005852500000000000'),
-    ).toThrow()
+  test('Decoding allows dodgy XRP without throwing', function () {
+    Currency.from('0000000000000000000000005852500000000000')
```
mvadari, line 54: This test isn't checking anything

package.json
```diff
@@ -30,7 +30,7 @@
     "@xrplf/prettier-config": "^1.5.0",
     "assert": "^2.0.0",
     "buffer": "^6.0.2",
-    "chai": "^4.3.4",
```
mvadari, line 33: This should be undone

packages/ripple-binary-codec/test/hash.test.js
JST5000: If the test throws it fails right? (I just want to make sure that the decoding happens without any exceptions being raised)

packages/ripple-binary-codec/src/types/currency.ts
```diff
@@ -25,11 +25,6 @@ function isIsoCode(iso: string): boolean {
 
 function isoCodeFromHex(code: Buffer): string | null {
   const iso = code.toString()
-  if (iso === 'XRP') {
-    throw new Error(
```
natenichols, line 29: This should still throw. If someone tries to encode `XRP` as a 3 letter currency code, it should fail.

packages/ripple-binary-codec/test/hash.test.js
```diff
@@ -50,10 +50,8 @@ describe('Hash256', function () {
 })
 
 describe('Currency', function () {
-  test('Will throw an error for dodgy XRP ', function () {
-    expect(() =>
```
natenichols, line 54: Same here, this test should pass, because it's behavior we want.

packages/ripple-binary-codec/test/tx-encode-decode.test.js
natenichols: Then yeah, we should still throw here.

packages/ripple-binary-codec/src/types/currency.ts
mvadari: I think it's fine if it doesn't throw - rippled will catch it

packages/ripple-binary-codec/test/hash.test.js
mvadari: You should still check the result in a test - otherwise it's not a complete test

mvadari (approved):

natenichols (approved):

packages/ripple-binary-codec/src/types/currency.ts
mvadari, line 29: This should probably `return null` instead of erroring

JST5000: I think that's still unexpected. The ledger has defenses up now at least to make it harder to scam people by having fake XRP on the ledger, and for folks doing historical checks it seems better to just be accurate rather than force them to know XRP becomes null.

packages/ripple-binary-codec/test/hash.test.js
alloynetworks: I disagree. The codec should not pass judgement. It is not middleware.

PR #1921, opened 2022-02-02
[ripple-binary-codec] fix ISO when parsing currency code

khancode:

Fixes ISO when parsing currency code.

### Context of Change

Storm importer reported a bug where ripple-binary-codec encode/decode wasn't properly working.

Root cause is the parsing logic inside Currency constructor doesn't set ISO field correctly.

### Type of Change

## Test Plan

Added tests to verify the bug is resolved.

### DIFF
--- a/packages/ripple-binary-codec/src/types/currency.ts
+++ b/packages/ripple-binary-codec/src/types/currency.ts
@@ -87,12 +87,13 @@ class Currency extends Hash160 {
     super(byteBuf ?? Currency.XRP.bytes)
     const code = this.bytes.slice(12, 15)
 
-    if (this.bytes[0] !== 0) {
-      this._iso = null
-    } else if (/^0*$/.test(this.bytes.toString('hex'))) {
+    if (/^0*$/.test(this.bytes.toString('hex'))) {
       this._iso = 'XRP'
-    } else {
+      // eslint-disable-next-line no-control-regex
+    } else if (/0{24}[\x00-\x7F]{6}0{10}/gm.test(this.bytes.toString('hex'))) {
       this._iso = isoCodeFromHex(code)
+    } else {
+      this._iso = null
     }
   }
 
--- a/packages/ripple-binary-codec/src/types/currency.ts
+++ b/packages/ripple-binary-codec/src/types/currency.ts
@@ -1,10 +1,11 @@
 import { Hash160 } from './hash-160'
 import { Buffer } from 'buffer/'
 
+const XRP_REGEX = /^0{40}$/
 const ISO_REGEX = /^[A-Z0-9]{3}$/
 const HEX_REGEX = /^[A-F0-9]{40}$/
 // eslint-disable-next-line no-control-regex
-const STANDARD_FORMAT_HEX_REGEX = /0{24}[\x00-\x7F]{6}0{10}/
+const STANDARD_FORMAT_HEX_REGEX = /^0{24}[\x00-\x7F]{6}0{10}$/
 
 /**
  * Convert an ISO code to a currency bytes representation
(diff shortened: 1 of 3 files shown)


packages/ripple-binary-codec/src/types/currency.ts
```diff
@@ -87,12 +87,13 @@ class Currency extends Hash160 {
...
     const code = this.bytes.slice(12, 15)
 
-    if (this.bytes[0] !== 0) {
-      this._iso = null
-    } else if (/^0*$/.test(this.bytes.toString('hex'))) {
+    if (/^0*$/.test(this.bytes.toString('hex'))) {
       this._iso = 'XRP'
-    } else {
+      // eslint-disable-next-line no-control-regex
+    } else if (/0{24}[\x00-\x7F]{6}0{10}/gm.test(this.bytes.toString('hex'))) {
```
mvadari, line 93: Can the regex be a variable at the top of the file with the other regexes?

packages/ripple-binary-codec/src/types/currency.ts
```diff
@@ -87,12 +87,13 @@ class Currency extends Hash160 {
     super(byteBuf ?? Currency.XRP.bytes)
     const code = this.bytes.slice(12, 15)
 
-    if (this.bytes[0] !== 0) {
-      this._iso = null
-    } else if (/^0*$/.test(this.bytes.toString('hex'))) {
+    if (/^0*$/.test(this.bytes.toString('hex'))) {
```
mvadari, line 90: `this.bytes.toString('hex')` is now used twice - can you make a variable for it so it doesn't need to be recomputed?

packages/ripple-binary-codec/src/types/currency.ts
```diff
@@ -87,12 +87,13 @@ class Currency extends Hash160 {
     super(byteBuf ?? Currency.XRP.bytes)
     const code = this.bytes.slice(12, 15)
```
mvadari, line 88: This is now only used in the `else if` case - it can probably just be computed in that section

mvadari (approved):
a few minor things, but lgtm. Thanks!

packages/ripple-binary-codec/src/types/currency.ts
```diff
@@ -86,13 +88,14 @@ class Currency extends Hash160 {
   constructor(byteBuf: Buffer) {
     super(byteBuf ?? Currency.XRP.bytes)
     const code = this.bytes.slice(12, 15)
+    const hex = this.bytes.toString('hex')
 
-    if (this.bytes[0] !== 0) {
-      this._iso = null
-    } else if (/^0*$/.test(this.bytes.toString('hex'))) {
+    if (/^0*$/.test(hex)) {
```
JST5000, line 92: Should this check for exactly 40 zeros? (I believe currently this will match the empty string, and I don't know if that's correct)

khancode: Ah, you're right. It does match for empty string. Maybe empty string also means to use XRP? @mvadari

JST5000: Seems handled elsewhere (Specifically bytesFromRepresentation)

JST5000 (approved):
LGTM

packages/ripple-binary-codec/src/types/currency.ts
JST5000: Empty string is an invalid currency code regardless. (We should only be allowing 3 character strings, or 40 character strings representing bytes. 0-2 character strings and 4-39 character strings are incorrectly formatted and should raise an exception)

mvadari: We can change this to check for a regex of 40 0s

mvadari: There could be an issue if you have a currency code of `"000"` (which is a valid currency code, but is not XRP)

khancode: I changed it to requiring 40 0s.

PR #1928, opened 2022-02-10
Update reference docs

intelliot:

Links to the source code from the documentation don't work. This (hopefully) fixes that.

For future PRs - When merging PRs that update the docs, it may be necessary to use the `Create a merge commit` option in GitHub. `Create a merge commit` will preserve the PR's commits, ensuring that the docs links work.

(NOT `Squash and merge`, which replaces prior commits with a new commit which has a new hash - introducing the risk that the docs could point to a commit that no longer exists)

### Context of Change

Reference docs are generated by typedoc and hosted on js.xrpl.org

### Type of Change

- [x] Documentation Updates

### Test Plan

I manually tested a random one of the "Defined in" links, and it worked: https://github.com/XRPLF/xrpl.js/blob/1e30b4cb/packages/xrpl/src/Wallet/signer.ts#L30


intelliot:
Agree - updated the PR description to be a little more accurate. I can squash this one if you prefer. I am neutral on that aspect :)

JST5000 (approved):
LGTM!

PR #1937, opened 2022-03-11
[Small Type error Fix]TrustSet LimitAmount Type changed

efraim9:
This PR changes Type on TrustSet interface property LimitAmount from Amount (string/object) to IssuedCurrencyAmount(object)

The reason why I made this PR is because I encountered an error while working with data from the TrustSet transaction in the TypeScript environment. 

If you tried to access data from the LimitAmount object on TrustSet transaction, like value, issuer, or currency. You would get a type error that this property does not exist on type Amount (type string). This is because TrustSet transaction was using universal Amount type and not specific type like IssuedCurrencyAmount.

According to [xrpl.org](https://xrpl.org/trustset.html#trustset) docs, TrustSet's LimitAmount property should be an Object, not string, which makes me even more sure with this change.

I tested this change locally and it fixed the type issue.

### DIFF
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -5,6 +5,7 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 ## Unreleased
 
 ### Fixed
+* Type of TrustSet transaction edited, specifically LimitAmount property type (fixed typescript issue)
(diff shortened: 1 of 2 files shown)


mvadari (approved):
Please update the changelog too.

efraim9:
Is this change sufficient?

JST5000 (approved):
Looks good to me :)

Issue #1871, opened 2021-12-16
`deriveAddress()` is missing @xrpl.js


mDuo13:
There _should_ be a way to derive a (classic) address from just a public key. The function to do this, `deriveAddress(pubkey)`, exists in the address codec and was in ripple-lib 1.x, but appears to have been unintentionally removed from xrpl.js 2.0.

ckniffen:
Closed with #1939

PR #1939, opened 2022-03-16
Export deriveAddress

JST5000:

Title says it all

### Context of Change

In response to #1871 

### Type of Change

## Before / After

Added deriveAddress to the utils folder exports.

## Test Plan

None

### DIFF
--- a/packages/xrpl/src/utils/derive.ts
+++ b/packages/xrpl/src/utils/derive.ts
@@ -20,4 +20,4 @@ function deriveXAddress(options: {
   return classicAddressToXAddress(classicAddress, options.tag, options.test)
 }
 
-export { deriveKeypair, deriveXAddress }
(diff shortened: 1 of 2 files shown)


mvadari (approved):
lgtm, but I suspect there'll be some docs changes as well.

packages/xrpl/src/utils/derive.ts
```diff
@@ -20,4 +20,5 @@ function deriveXAddress(options: {
   return classicAddressToXAddress(classicAddress, options.tag, options.test)
 }
 
-export { deriveKeypair, deriveXAddress }
+// eslint-disable-next-line import/no-unused-modules - deriveAddress is used externally
```
mvadari, line 23: This error will go away if you export it further up as well (at the `src/utils` level)

khancode (approved):

PR #1940, opened 2022-03-16
feat: add sidechain-specific RPCs/sugar

mvadari:

This PR:
* Adds `federator_info` request and response objects
* Adds a sugar util method to create a cross-chain payment transaction from a normal payment transaction

### Context of Change

xrpl.js sidechain support

### Type of Change

## Test Plan

Added tests for the helper method. I'm not sure how to test the RPC call and object, since standalone rippled isn't a federator. CI passes.


packages/xrpl/src/models/methods/federatorInfo.ts
```diff
@@ -0,0 +1,78 @@
...
+export interface FederatorInfoRequest extends BaseRequest {
+  command: 'federator_info'
+}
+
+/**
+ * Response expected from a {@link FederatorInfoRequest}.
+ *
+ * @category Responses
+ */
+export interface FederatorInfoResponse extends BaseResponse {
```
JST5000, line 20: Should these values have docstrings/comments explaining what they are used for like we do with other requests?

packages/xrpl/src/utils/createXchainPayment.ts
```diff
@@ -0,0 +1,40 @@
...
+ * signed, then it will need to be re-signed. There must be no more than 2
+ * memos, since one memo is used for the sidechain destination account.
+ * @param destAccount - the destination account on the sidechain.
+ * @returns A cross-chain payment transaction, where the mainchain door account
+ * is the `Destination` and the destination account on the sidechain is encoded
+ * in the memos.
+ * @throws XrplError - if there are more than 2 memos.
+ * @category Utilities
+ */
+export default function createXchainPayment(
```
JST5000, line 21: Naming wise I think it's better to explicitly write "Cross" rather than "X" since X is actually meaningful sometimes in our libraries (Ex. xAddress)

Also, "chain" should be capitalized.

I'd recommend to rename this function to `createCrossChainPayment`

packages/xrpl/src/utils/createXchainPayment.ts
```diff
@@ -0,0 +1,40 @@
+import { XrplError } from '../errors'
+import { Payment } from '../models'
+import { Memo } from '../models/common'
+
+import { convertStringToHex } from './stringConversion'
+
+/**
+ * Computes the complete list of every balance that changed in the ledger
+ * as a result of the given transaction.
```
JST5000, line 9: Should this function work for both Mainnet -> Sidechain payments and Sidechain -> Mainnet payments?

packages/xrpl/src/models/methods/federatorInfo.ts
mvadari: Yes - however, those don't exist yet.

packages/xrpl/src/utils/createXchainPayment.ts
khancode: I agree. Let's use "Cross" instead to avoid ambiguity.

khancode (approved):
Just need to change the "X" wording to "Cross". Otherwise, LGTM!

JST5000 (approved):
LGTM

PR #1942, opened 2022-03-26
Update connection.ts to always clear the heartbeat interval

nitowa:

Update `connection.ts` in the function `disconnect()` to always call `clearHeartbeatInterval()`

### Context of Change

In some cases when `connect()` fails and `disconnect()` is called immediately, a residual timed interval is left running. This patch makes sure any heartbeat intervals are reliably cleared on disconnect.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


JST5000:
@nitowa The change looks good to me. Could you add a bullet to the Unreleased Fixes section of HISTORY.md mentioning this change? - https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/HISTORY.md#fixed

mvadari:
Does this resolve #1825?

mvadari (approved):

khancode:
> Does this resolve #1825?

Tested the fix and looks like it doesn't.

Issue #1897, opened 2022-01-17
 Resolve TokenId


develoQ:
Can we resolve TokenID of NFToken object?

like:
`anyFunc(000B013A95F14B0044F78A264E41713C64B5F89242540EE208C3098E00000D65)`

return
```
{
    Flags: 12,
    TransferFee: 0.0314,
    Issuer: rNCFjv8Ek5oDrNiMJ3pw6eLLFtMjZLJnf2,
    Taxon: 146999694,
    Sequence: 3429,
}
```

JST5000:
I don't believe there's any rippled request which can do that unfortunately. 

Closest thing I could find in the XLS 20d spec is this description of how to search for NFTokens:

> To search for a specific NFToken, the first step is to locate the NFTokenPage, if any, that should contain that NFToken. For that do the following:
> 
> Compute the NFTokenPageID using the account of the owner and the TokenID of the token, as described above. Then search for the ledger entry whose identifier is less than or equal to that value. If that entry does not exist or is not an NFTokenPage, the NFToken is not held by the given account.

https://github.com/XRPLF/XRPL-Standards/discussions/46

(One way to accomplish that yourself is to have a database which watches the ledger and updates whenever you see a TokenMint, which you can then search for TokenIds in whichever manner you want)

yyolk:
I would like to add this to xrpl-py too (if it doesn't exist yet 😀) 


if I'm not mistaken OP wants to translate the encoded id into the unpacked map ![image](https://user-images.githubusercontent.com/134478/150440128-fe20aa9e-579f-4010-9912-aec1e6194d95.jpeg)

JST5000:
Oh shoot, sorry for misinterpreting! 

That makes sense :)

develoQ:
> if I'm not mistaken OP wants to translate the encoded id into the unpacked map

Yes, that's correct.

develoQ:
I don't know how to decode the encoded address `95F14B0044F78A264E41713C64B5F89242540EE2` to `rNCFjv8Ek5oDrNiMJ3pw6eLLFtMjZLJnf2`.

ihomp:
+1, would be nice to have it in the lib.

ildaruz:
> I don't know how to decode the encoded address `95F14B0044F78A264E41713C64B5F89242540EE2` to `rNCFjv8Ek5oDrNiMJ3pw6eLLFtMjZLJnf2`.

[...]

PR #1961, opened 2022-04-07
Add parseNFTokenID and tests

JST5000:

Add a utility method to parse NFTokenIDs into a readable format.

### Context of Change

In response to #1897 

### Type of Change

## Test Plan

Added unit tests for this new feature.


ihomp:
looks good! :) 
/ Bithomp

packages/xrpl/src/index.ts
```diff
@@ -7,6 +7,8 @@ export * from './models'
 
 export * from './utils'
 
+export * from './sugar'
```
mvadari, line 10: The `sugar` folder shouldn't be exported directly, it should only be exported via methods on the client.

mvadari (changes requested):
See comments

packages/xrpl/src/utils/parseNFTokenID.ts
```diff
@@ -0,0 +1,94 @@
+/* eslint-disable @typescript-eslint/no-magic-numbers -- Doing hex string parsing. */
+import BigNumber from 'bignumber.js'
+import { encodeAccountID } from 'ripple-address-codec'
+
+import { XrplError } from '../errors'
+
+/**
+ * All information encoded within an NFTokenID.
+ */
+interface NFTokenID {
```
mvadari, line 10: Shouldn't this be in the `models` folder?

packages/xrpl/src/utils/parseNFTokenID.ts
```diff
@@ -0,0 +1,94 @@
...
+ * |    |
+ * |    `---> TransferFee: 1337.0 bps or 13.37%
+ * |
+ * `---> Flags: 11 -> lsfBurnable, lsfOnlyXRP and lsfTransferable
+ *
+ * @param tokenID - A hex string which identifies an NFToken on the ledger.
+ * @throws XrplError when given an invalid tokenID.
+ * @returns a decoded tokenID with all information encoded within.
+ */
+function parseNFTokenID(tokenID: string): NFTokenID {
```
mvadari, line 72: Are there plans to have a method that goes in the opposite direction?

JST5000: I don't think the reverse matters, you'd need to know everything about the NFT which practically means you either minted the token, or just looked it up, and the minter can already get the NFTokenID I believe.

JST5000:
@mvadari I'm still exporting NFTokenID from the parseNFTokenID file because you always need them together, does that make sense? (Or is it better to export from NFTokenID in the models folder?)

packages/xrpl/src/utils/parseNFTokenID.ts
mvadari: The grantees I've talked to made it sound like it's kind of a pain to dig out the NFTokenID from the depths of the transaction metadata - so if we could write the reverse function (or even just a method to do the digging of the metadata) I think that would also be super helpful. IIRC the only piece of information that you don't know at the time of sending the mint tx is the sequence #.

mvadari:
> @mvadari I'm still exporting NFTokenID from the parseNFTokenID file because you always need them together, does that make sense? (Or is it better to export from NFTokenID in the models folder?)

I think in that case the way we've done it in this repo previously is to define the response of the method as the object instead of creating a new interface (e.g. https://js.xrpl.org/classes/Client.html#getBalances)

packages/xrpl/src/utils/parseNFTokenID.ts
mvadari: If you want to do that in a separate PR that's fine

JST5000:
> I think in that case the way we've done it in this repo previously is to define the response of the method as the object instead of creating a new interface (e.g. https://js.xrpl.org/classes/Client.html#getBalances)

I think it's worth having it's own type because that allows for cleaner explanations for what each field represents. Especially if we're planning on making an "encode" function later on as well.

ckniffen (approved):

mvadari:
> I think it's worth having it's own type because that allows for cleaner explanations for what each field represents. Especially if we're planning on making an "encode" function later on as well.

Hmm, I felt like it made it more confusing when there were a bunch of random types floating around instead of just an object.

But if we do want to keep the type, then, you're right, the type shouldn't go in `models`.

mvadari (approved):

Issue #1872, opened 2021-12-16
deriveXAddress derives different addresses for `tag: false` and `tag: null` @xrpl.js


mDuo13:
I believe this is incorrect behavior. Both `false` and `null` are intended to derive an X-address without a tag, but `null` seems to derive the same result as the tag `0`.

```
> xrpl.deriveXAddress({publicKey: "ED02C98225BD1C79E9A4F95C6978026D300AFB7CA2A34358920BCFBCEBE6AFCD6A", tag: false, test: false})
'X7FbrqVEqdTNoX5qq94rTdarGjeVYmkxi8A1TKAJUnyLL9g'
> xrpl.deriveXAddress({publicKey: "ED02C98225BD1C79E9A4F95C6978026D300AFB7CA2A34358920BCFBCEBE6AFCD6A"})
'X7FbrqVEqdTNoX5qq94rTdarGjeVYmqohnQV3vMEprV2A73'
> xrpl.deriveXAddress({publicKey: "ED02C98225BD1C79E9A4F95C6978026D300AFB7CA2A34358920BCFBCEBE6AFCD6A", tag: null})
'X7FbrqVEqdTNoX5qq94rTdarGjeVYmqohnQV3vMEprV2A73'
> xrpl.deriveXAddress({publicKey: "ED02C98225BD1C79E9A4F95C6978026D300AFB7CA2A34358920BCFBCEBE6AFCD6A", tag: false})
'X7FbrqVEqdTNoX5qq94rTdarGjeVYmkxi8A1TKAJUnyLL9g'
> xrpl.deriveXAddress({publicKey: "ED02C98225BD1C79E9A4F95C6978026D300AFB7CA2A34358920BCFBCEBE6AFCD6A", tag: 0})
'X7FbrqVEqdTNoX5qq94rTdarGjeVYmqohnQV3vMEprV2A73'
```

## Environment
xrpl.js 2.0.4
Using Node.js v17.1.0 (not using TypeScript)

ckniffen:
Closed with #1964.

PR #1964, opened 2022-04-13
Fix deriveXAddress null behavior

JST5000:

`deriveXAddress` should have the same behavior for `null` as it does for `false`. 

### Context of Change

In response to #1872 

### Type of Change

## Test Plan

Added a unit test for it.

### DIFF
--- a/packages/ripple-address-codec/src/index.ts
+++ b/packages/ripple-address-codec/src/index.ts
@@ -43,8 +43,10 @@ function encodeXAddress(
   if (tag > MAX_32_BIT_UNSIGNED_INT) {
     throw new Error('Invalid tag')
   }
-  const theTag = tag === false ? 0 : tag
-  const flag = tag === false ? 0 : 1
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Passing null is a common js mistake
+  const theTag = tag === false || tag === null ? 0 : tag
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Passing null is a common js mistake
+  const flag = tag === false || tag === null ? 0 : 1
   /* eslint-disable no-bitwise ---
    * need to use bitwise operations here */
   const bytes = Buffer.concat([
--- a/packages/ripple-address-codec/src/index.ts
+++ b/packages/ripple-address-codec/src/index.ts
@@ -43,8 +43,7 @@ function encodeXAddress(
   if (tag > MAX_32_BIT_UNSIGNED_INT) {
     throw new Error('Invalid tag')
   }
-  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Passing null is a common js mistake
-  const theTag = tag === false || tag == null ? 0 : tag
+  const theTag = tag || 0
   // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Passing null is a common js mistake
   const flag = tag === false || tag == null ? 0 : 1
   /* eslint-disable no-bitwise ---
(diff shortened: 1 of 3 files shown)


packages/ripple-address-codec/src/index.ts
```diff
@@ -43,8 +43,10 @@ function encodeXAddress(
   if (tag > MAX_32_BIT_UNSIGNED_INT) {
     throw new Error('Invalid tag')
   }
-  const theTag = tag === false ? 0 : tag
-  const flag = tag === false ? 0 : 1
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Passing null is a common js mistake
+  const theTag = tag === false || tag === null ? 0 : tag
```
mvadari, line 47:
```suggestion
  const theTag = tag === false || tag == null ? 0 : tag
```

JST5000: Prettier won't let me add the parentheses, but I can add the `==` change to check for `undefined` as well.

packages/ripple-address-codec/src/index.ts
```diff
@@ -43,8 +43,10 @@ function encodeXAddress(
   if (tag > MAX_32_BIT_UNSIGNED_INT) {
     throw new Error('Invalid tag')
   }
-  const theTag = tag === false ? 0 : tag
-  const flag = tag === false ? 0 : 1
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Passing null is a common js mistake
+  const theTag = tag === false || tag == null ? 0 : tag
```
ckniffen, line 47: Should undefined be thrown in for good measure?

JST5000: I think doing `== null` instead of `=== null` does the undefined check as well.

packages/ripple-address-codec/src/index.ts
```diff
@@ -43,8 +43,10 @@ function encodeXAddress(
...
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Passing null is a common js mistake
+  const flag = tag === false || tag == null ? 0 : 1
```
khancode, line 49: We can simplify this to:
```suggestion
  const theTag = tag || 0
  const flag = tag ? 1 : 0
```

khancode (changes requested):

mvadari (approved):

packages/xrpl/test/utils/deriveXAddress.ts
```diff
@@ -22,4 +22,17 @@ describe('deriveXAddress', function () {
...
     )
   })
+
+  it('does not include tag when null', function () {
+    assert.equal(
+      deriveXAddress({
+        publicKey:
+          'ED02C98225BD1C79E9A4F95C6978026D300AFB7CA2A34358920BCFBCEBE6AFCD6A',
+        // @ts-expect-error -- Assessing null behavior
+        tag: null,
```
mvadari, line 32:
```suggestion
        tag: undefined,
```
This might help with the TS error

packages/ripple-address-codec/src/index.ts
JST5000: The 1st simplification works, but the 2nd doesn't because if the tag is `0` we need that to end up with `flag = 1` 
("There is a tag, and its value is 0")

packages/xrpl/test/utils/deriveXAddress.ts
JST5000: Unfortunately it did not, but because we're now trying to have `undefined` match as well I added a test case for it :)

ckniffen (approved):

packages/ripple-address-codec/src/index.ts
khancode: Oh, then modify the 2nd one to be
```TypeScript
const flag = tag || 0 ? 1 : 0
```

khancode (approved):
LGTM

PR #1975, opened 2022-04-27
fix(ripple-keypairs): make bytesToHex work with typed arrays (Uint8Array)

PiDelport:

`bytesToHex` was producing invalid output when the input bytes were contained in a typed array such as `Uint8Array` instead of a generic `Array`.

This change fixes that, and should also make it work correctly with a wider range of `Iterable` or `ArrayLike` input types.

### Context of Change

We encountered this issue while writing XRPL signing code for [nautilus-wallet](https://github.com/ntls-io/nautilus-wallet), which uses `Uint8Array` to work with bytes.

### Type of Change

## Before

```node
> bytesToHex(new Uint8Array([1,10,20,100]))
'101464'
```
## After

```node
> bytesToHex(new Uint8Array([1,10,20,100]))
'010A1464'
```

## Test Plan

I added test coverage for `bytesToHex` based on the existing `hexToBytes` tests, and added a case for `Uint8Array` input.

### DIFF
--- a/packages/ripple-keypairs/test/utils-test.js
+++ b/packages/ripple-keypairs/test/utils-test.js
@@ -11,4 +11,12 @@ describe('utils', () => {
   it('hexToBytes - DEADBEEF', () => {
     assert.deepEqual(utils.hexToBytes('DEADBEEF'), [222, 173, 190, 239])
   })
+
+  it('bytesToHex - DEADBEEF', () => {
+    assert.deepEqual(utils.bytesToHex([222, 173, 190, 239]), 'DEADBEEF')
+  });
+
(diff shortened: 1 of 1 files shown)


JST5000:
Thanks for the addition! 

Could you add a bullet for your change in the `ripple-keypairs` HISTORY.md file? (You'll need to add a ## Unreleased header :)

Ex. 
```
#ripple-keypairs Release History

## Unreleased
* Extended bytesToHex to work with typed arrays like UInt8Array
```

JST5000 (approved):
Looks good to me :)

ckniffen (approved):

PR #1977, opened 2022-04-27
fix(ripple-keypairs): hexToBytes should produce [] for empty input, not [0]

PiDelport:

`hexToBytes` uses the `BN` library, but `BN.toArray` intentionally produces `[0]` rather than `[]` for a zero buffer, even when requesting output of zero length. This may make sense for BigNum buffers, but not for byte strings.

This change fixes that by special-casing zero-length input to `[]`.

### Context of Change

- Related to https://github.com/XRPLF/xrpl.js/pull/1975

### Type of Change

## Before / After

- Before: `hexToBytes('')` produced `[0]`
- After: `hexToBytes('')` produces `[]`

## Test Plan

I added a unit test case.

### DIFF
--- a/packages/ripple-keypairs/HISTORY.md
+++ b/packages/ripple-keypairs/HISTORY.md
@@ -1,5 +1,8 @@
 # ripple-keypairs Release History
 
+## Unreleased
+- `hexToBytes` now produces empty output for empty input, rather than `[0]`.
+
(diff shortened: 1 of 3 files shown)


ckniffen (approved):

JST5000 (approved):
Looks good to me :)

Issue #1926, opened 2022-02-09
TypeError: Cannot read property 'LedgerEntryType' of undefined


yashutanna:
There seems to be an error that is thrown when retrieving the transactions for ledger number 69578408. Not really sure what / where this originated as subsequent and previous ledgers are no issue. 

I am using Ripple node version 1.8.4
ripple-lib version 1.10.0

Error is: 

TypeError: Cannot read property 'LedgerEntryType' of undefined
    at /var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/parse/settings.js:31:84
    at Array.filter (<anonymous>)
    at getAccountRootModifiedNode (/var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/parse/settings.js:31:49)
    at parseFlags (/var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/parse/settings.js:40:18)
    at parseSettings (/var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/parse/settings.js:72:30)
    at Object.parseTransaction [as default] (/var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/parse/transaction.js:78:11)
    at parseTransactionWrapper (/var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/parse/ledger.js:34:41)
    at apply (/var/worker/node_modules/lodash/lodash.js:478:17)
    at wrapper (/var/worker/node_modules/lodash/lodash.js:5337:16)
    at Array.map (<anonymous>)
    at parseTransactions (/var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/parse/ledger.js:48:36)
    at Object.parseLedger (/var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/parse/ledger.js:73:8)
    at RippleAPI.<anonymous> (/var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/ledger.js:24:25)
    at Generator.next (<anonymous>)
    at fulfilled (/var/worker/node_modules/@project/common/node_modules/ripple-lib/dist/npm/ledger/ledger.js:5:58)
    at runMicrotasks (<anonymous>)

PR #1907, opened 2022-01-27
Update settings.ts

sampleblood:
bug fix 

AffectedNodes may have not only ModifiedNode, but also createNode and DeletedNode, then crash

PR #1979, opened 2022-04-28
Add test for parseSettings on AccountSet with deleted node

JST5000:

Title says it all.

### Context of Change

Follow up to #1907 to add a test for the change.

### Type of Change

## Test Plan

I manually reverted the changes from #1907 to see if this test case fails, then re-added the change and checked that the test now passes. (The test case uses real rippled data from the ledger mentioned in #1926 where this issue was originally raised, so it should parse it successfully).

### DIFF
--- a/test/fixtures/rippled/account-set-with-deleted-node.json
+++ b/test/fixtures/rippled/account-set-with-deleted-node.json
@@ -1,0 +1,91 @@
+{
+    "id": 1,
+    "result": {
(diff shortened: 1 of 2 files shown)


JST5000:
Note that this PR is for ripple-lib 1.x, NOT xrpl.js

ckniffen (approved):

khancode (approved):

PR #1981, opened 2022-04-30
Fix browser infinite disconnect if exception raised during 'connected'

JST5000:

When an exception is raised during the websocket connection, it can cause a disconnect with no error code, which can then cause an infinite series of reconnect + errors.

### Context of Change

In response to #1825 

### Type of Change

## Before / After
Before:
(Errors keep repeating infinitely)
image

After:
(Error happens once with a more descriptive error message)
image

## Test Plan

Manually reproduced using @mDuo13's steps to reproduce and verified it only disconnected once after the changes.

### DIFF
--- a/packages/xrpl/src/client/connection.ts
+++ b/packages/xrpl/src/client/connection.ts
@@ -463,7 +463,7 @@ export class Connection extends EventEmitter {
          * Error code 1011 represents an Internal Error according to
          * https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code
          */
-        const internalErrorCode = 1011
+        const internalErrorCode = 1012
         this.emit('disconnected', internalErrorCode)
       } else {
         this.emit('disconnected', code)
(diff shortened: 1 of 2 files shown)


khancode (approved):
LGTM 👌

packages/xrpl/src/client/connection.ts
```diff
@@ -448,9 +449,34 @@ export class Connection extends EventEmitter {
       )
       this.ws.removeAllListeners()
       this.ws = null
-      this.emit('disconnected', code)
-      // If this wasn't a manual disconnect, then lets reconnect ASAP.
-      if (code !== INTENTIONAL_DISCONNECT_CODE) {
+
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- undefined can be passed in.
+      if (code === undefined) {
```
ckniffen, line 454: Will typescript allow you to add undefined as a possible parameter values to line 439?

JST5000: Turns out yep! - I previously tried to just change `code` to an optional value which failed because optional values have to be at the end. I didn't try setting both `code` and `reason` to optional though, which works! (oops :P)

ckniffen (approved):

PR #1992, opened 2022-05-03
Add fromMnemonic using same decode method as `wallet_propose` (rfc1751)

JST5000:

Add a way to generate a wallet from RFC1751 mnemonics like are generated in the deprecated rippled function `wallet_propose`. 

### Context of Change

In response to an end user issue where they depended upon `wallet_propose` and so couldn't use xrpl.js without this backwards compatibility. 

### Type of Change

## Before / After

* Added `Wallet.fromRFC1751Mnemonic` which generates a new wallet from `wallet_propose`'s `master_key` field.
* Also fixed a small typo (bip38 -> bip39)

## Test Plan

Added unit tests to ensure it decodes correctly for both `secp256k1` keys and `ed25519` keys.


packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -0,0 +1,376 @@
+/*
+    rfc1751.js : Converts between 128-bit strings and a human-readable
+    sequence of words, as defined in RFC1751: "A Convention for
+    Human-Readable 128-bit Keys", by Daniel L. McDonald.
+    Ported from rfc1751.py / Python Cryptography Toolkit (public domain).
+    Copied from https://github.com/bip32/bip32.github.io/blob/master/js/rfc1751.js which
+    is part of the public domain.
+*/
+
+var rfc1751_wordlist = [ "A", "ABE", "ACE", "ACT", "AD", "ADA", "ADD",
```
mvadari, line 10: Never use vars

mvadari, line 10: Consider moving this variable to a separate JSON

mvadari, line 10: `camelCase`

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -0,0 +1,376 @@
...
+    let res = ''
+    for (var i = 0; i < key.length; i++) {
+        res += binary[key[i] >> 4] + binary [key[i] & 0x0f];
+    }
+    return res;
+}
+
+function _extract(key, start, length): number {
+    const k = key.substring(start, start+length);
+    var acc = 0;
```
mvadari, line 267: ditto - should be `let`

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -0,0 +1,376 @@
...
+   "WISE", "WISH", "WITH", "WOLF", "WONT", "WOOD", "WOOL", "WORD",
+   "WORE", "WORK", "WORM", "WORN", "WOVE", "WRIT", "WYNN", "YALE",
+   "YANG", "YANK", "YARD", "YARN", "YAWL", "YAWN", "YEAH", "YEAR",
+   "YELL", "YOGA", "YOKE" ];
+
+var binary = ['0000', '0001', '0010', '0011', '0100', '0101', '0110', '0111',
+              '1000', '1001', '1010', '1011', '1100', '1101', '1110', '1111'];
+
+function _key2bin(key) {
+    let res = ''
```
mvadari, line 258: Consider writing this with a `reduce` instead

packages/xrpl/src/Wallet/rfc1751.ts
mvadari, line 267: ditto with the `reduce`

packages/xrpl/src/Wallet/rfc1751.ts
JST5000: I tried it out, but it ended up being harder to read for the same lines of code 

```
  let res2 = key.reduce((prev, curr, index, arr) => {
    return prev + binary[curr >> 4] + binary[curr & 0x0f]
  }, '')
```

packages/xrpl/src/Wallet/rfc1751.ts
JST5000: Apparently strings don't have reduce, but even with it runs into the same issue - it's a bit harder to read for the same code length, so I don't think it's worth it.

ckniffen (approved):

packages/xrpl/src/Wallet/index.ts
```diff
@@ -166,7 +169,35 @@ class Wallet {
...
-   * Derives a wallet from a mnemonic.
+   * Derives a wallet from a RFC1751 mnemonic, which is how `wallet_propose` encodes mnemonics.
+   *
+   * @param mnemonic - A string consisting of words (whitespace delimited) used to derive a wallet.
+   * @param opts - (Optional) Options to derive a Wallet.
+   * @param opts.masterAddress - Include if a Wallet uses a Regular Key Pair. It must be the master address of the account.
+   * @param opts.algorithm - The digital signature algorithm to generate an address for.
+   * @returns A Wallet derived from a mnemonic.
+   */
+  public static fromRFC1751Mnemonic(
```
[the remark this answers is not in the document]
JST5000: I suppose it shouldn't be :)

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -0,0 +1,173 @@
...
+ *Human-Readable 128-bit Keys", by Daniel L. McDonald.
+ *Ported from rfc1751.py / Python Cryptography Toolkit (public domain).
+ *Copied from https://github.com/bip32/bip32.github.io/blob/master/js/rfc1751.js which
+ *is part of the public domain.
+ */
+
+import rfc1751Words from './rfc1751Words.json'
+
+// prettier-ignore
+const rfc1751WordList = rfc1751Words.wordList
```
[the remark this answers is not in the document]
JST5000: Nothing now - good catch!

packages/xrpl/src/Wallet/rfc1751Words.json
```diff
@@ -0,0 +1,245 @@
+{
```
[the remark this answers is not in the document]
JST5000: I didn't know that, fixed :)

ckniffen (approved):

ckniffen (approved):

packages/xrpl/src/Wallet/index.ts
```diff
@@ -166,19 +170,60 @@ class Wallet {
...
    *
    * @param mnemonic - A string consisting of words (whitespace delimited) used to derive a wallet.
    * @param opts - (Optional) Options to derive a Wallet.
-   * @param opts.derivationPath - The path to derive a keypair (publicKey/privateKey) used for mnemonic-to-seed conversion.
    * @param opts.masterAddress - Include if a Wallet uses a Regular Key Pair. It must be the master address of the account.
+   * @param opts.derivationPath - The path to derive a keypair (publicKey/privateKey). Only used for bip39 conversions.
+   * @param opts.useRFC1751 - If true, this interprets the mnemonic as a rippled RFC1751 mnemonic like `wallet_propose`
+   *                          generates in rippled.
+   * @param opts.algorithm - Only used if opts.useRFC1751Mnemonic is true. Allows the mnemonic to generate its
+   *                         secp256k1 seed, or its ed25519 seed.
```
mvadari, line 207: I don't love that there are 2 options for this - it feels like there should be one option with 3 possible values: bip39, RFC1751-with-secp, RFC1751-with-ed25519 (obviously with better naming conventions). Happy to discuss.

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -0,0 +1,178 @@
...
+ *Ported from rfc1751.py / Python Cryptography Toolkit (public domain).
+ *Copied from https://github.com/bip32/bip32.github.io/blob/master/js/rfc1751.js which
+ *is part of the public domain.
+ */
+
+import rfc1751Words from './rfc1751Words.json'
+
+const rfc1751WordList: string[] = rfc1751Words
+
+// prettier-ignore
```
mvadari, line 17: Why is this ignored?

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -0,0 +1,178 @@
...
+function keyToBinary(key: number[]): string {
+  let res = ''
+  for (const num of key) {
+    res += binary[num >> 4] + binary[num & 0x0f]
+  }
+  return res
+}
+
+/**
+ * Converts a substring of an encoded secret to it's numeric value.
```
mvadari, line 36:
```suggestion
 * Converts a substring of an encoded secret to its numeric value.
```

packages/xrpl/src/Wallet/rfc1751Words.json
```diff
@@ -0,0 +1,243 @@
+[ "A", "ABE", "ACE", "ACT", "AD", "ADA", "ADD",
```
mvadari, line 1: It annoys me that these lines aren't even but that's really not worth fixing, it would be a pain

packages/xrpl/src/index.ts
```diff
@@ -11,4 +11,6 @@ export * from './errors'
 
 export { default as Wallet } from './Wallet'
 
+export { keyToRFC1751Mnemonic, rfc1751MnemonicToKey } from './Wallet/rfc1751'
```
mvadari, line 14: Why is this exported?

Would it be better to export them as static methods on the wallet?

JST5000: I don't currently have any use for turning a secret into a mnemonic, but I figured if we're including the decode through `Wallet.fromMnemonic`, we should also offer the encode. (Especially since `rippled` has a slightly different implementation than RFC1751 elsewhere, so it'd require one line of custom code to use other solutions which is a pain)

mvadari: Wouldn't it be better to have a `.toMnemonic` method on the wallet then?

packages/xrpl/src/Wallet/rfc1751.ts
JST5000: Prettier wants this to be one entry per line, which is much less readable. I added a comment explaining that.

packages/xrpl/src/Wallet/index.ts
JST5000: I think regardless you end up with awkwardness (Since `derivationPath` is an optional param which is only relevant for bip39)

This also allows us to set the default to match `wallet_propose` which defaults to secp256k1 as opposed to our usual default of ed25519.

packages/xrpl/src/index.ts
JST5000: In order to match our interface, `.toMnemonic` should probably allow them to generate bip39 mnemonics as well, which I don't want to implement currently.

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -0,0 +1,194 @@
...
+ *Copied from https://github.com/bip32/bip32.github.io/blob/master/js/rfc1751.js which
+ *is part of the public domain.
+ */
+
+import rfc1751Words from './rfc1751Words.json'
+
+const rfc1751WordList: string[] = rfc1751Words
+
+// prettier-ignore - It's more readable to have this on two lines than one per line.
+const binary = [
```
mvadari, line 18: Looks like the linter made it one per line anyways

mvadari: nit: this should be
```suggestion
const _BINARY = [
```

JST5000: Adding the comment made the `// prettier-ignore` stop working, so I removed the comment

packages/xrpl/src/Wallet/index.ts
```diff
@@ -198,27 +244,30 @@ class Wallet {
...
+    if (opts.algorithm === ECDSA.ed25519) {
+      encodeAlgorithm = 'ed25519'
+    } else {
+      // Defaults to secp256k1 since that's the default for `wallet_propose`
+      encodeAlgorithm = 'secp256k1'
     }
-    const seed = generateSeed(options)
-    return Wallet.deriveWallet(seed, {
-      algorithm,
+    const encodedSeed = encodeSeed(seed, encodeAlgorithm)
```
mvadari, line 267: `encodeSeed` should take `ECDSA` too, not a string - then the previous bit becomes 
```
const encodeAlgorithm = opts.algorithm ?? ECDSA.secp256k1
```

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -0,0 +1,180 @@
...
+ */
+
+import rfc1751Words from './rfc1751Words.json'
+
+const rfc1751WordList: string[] = rfc1751Words
+
+// Added prettier-ignore to allow _BINARY to be on two lines, instead of one entry per line.
+
+// prettier-ignore
+const _BINARY = ['0000', '0001', '0010', '0011', '0100', '0101', '0110', '0111',
```
mvadari, line 20: nvm, JS doesn't use the `_` in the front (sorry) - this should just be `BINARY`

packages/xrpl/src/Wallet/index.ts
JST5000: That'd be a breaking change to the address-codec

mvadari: Whoops, didn't realize this was an address-codec method

mvadari (approved):
lgtm

PR #1998, opened 2022-05-06
Update dependabot.yml to ignore bip32

JST5000:

We want dependabot to stop recommending to upgrade bip32

### Context of Change

It's currently not worth updating bip32 for reasons explained in this issue: #1839 
 
### Type of Change

## Before / After

Just setting dependabot to ignore bip32 since we don't want to update it automatically going forward.

## Test Plan

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,6 @@ updates:
     interval: monthly
     time: "15:00"
   open-pull-requests-limit: 10
+  ignore:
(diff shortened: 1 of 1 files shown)


ckniffen (approved):

mvadari:
This shouldn't be a code change - you can just tell dependabot to ignore this major version in the PR (you can take a look at the commands to determine how to do that). 

IMO it shouldn't be a blanket ignore, it should just ignore this major version - because version 3.x might be more usable.

PR #2000, opened 2022-05-10
fix verification bug in Wallet.sign() when signing a non-XRP Payment amount with trailing zeros

khancode:

Fixed serialize/deserialize verification bug in `Wallet.sign()` when signing a non-XRP Payment with an amount that contains trailing insignificant zeros.

### Context of Change

When attempting to sign a transaction with an issued currency amount including insignificant trailing zeroes (such as "123.40") the sign function raises a mismatch error because the transaction deserializes the amount without trailing zeroes (such as "123.4").

This fix removes trailing zeros in non-XRP amounts when using `Wallet.sign()`

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added unit test to verify non-XRP amount with trailing zeros doesn't throw an error.


packages/xrpl/src/Wallet/index.ts
```diff
@@ -32,6 +33,7 @@ import { rfc1751MnemonicToKey } from './rfc1751'
 
 const DEFAULT_ALGORITHM: ECDSA = ECDSA.ed25519
 const DEFAULT_DERIVATION_PATH = "m/44'/144'/0'/0/0"
+const TRAILING_ZEROS_REGEX = /.*\..*[0]+$/u
```
ckniffen, line 36: Is a more optimized REGEX for this?  The checker seems to hate it

packages/xrpl/src/Wallet/index.ts
```diff
@@ -318,13 +320,26 @@ class Wallet {
...
     }
 
-    const txToSignAndEncode = { ...transaction }
+    // Remove trailing insignificant zeros for non-XRP amount
+    if (tx.TransactionType === 'Payment' && tx.Amount) {
+      if (
+        typeof tx.Amount !== 'string' &&
+        TRAILING_ZEROS_REGEX.exec(tx.Amount.value) !== null
+      ) {
+        tx.Amount = { ...tx.Amount }
```
JST5000, line 337: Why is this step necessary?

packages/xrpl/test/wallet/index.ts
```diff
@@ -589,6 +589,31 @@ describe('Wallet', function () {
...
+      }
+      const result = wallet.sign(payment)
+      const expectedResult = {
+        tx_blob:
+          '12000022800000002400000017201B008694F261D504625103A72000000000000000000000000000464F4F00000000002E099DD75FDD96EB4A603037844F964832FED86B68400000000000000C732102A8A44DB3D4C73EEEE11DFE54D2029103B776AA8A8D293A91D645977C9DF5F5447446304402206EBFC9B8061C3F82D521506CE62B6BBA99995B2175BFE0E1BC516775932AECEB0220172B9CE9C0EFB3F4870E19B79B4E817DD376E5785F034AB792708F92282C65F781145E7B112523F68D2F5E879DB4EAC51C6698A693048314FDB08D07AAA0EB711793A3027304D688E10C3648',
+        hash: '6235E5A3CC14DB97F75CAE2A4B5AA9B4134B7AD48D7DD8C15473D81631435FE4',
+      }
+
+      assert.deepEqual(result, expectedResult)
+    })
```
JST5000, line 616: Would it be possible to include a test with `value = 123.000` or something like that? (With only trailing zeros after the decimal?)

If it's treated like a string, that could result in a weird interpretation of `123.` instead of `123`. (I think your implementation covers that case correctly, but I think it's worth a test case)

packages/xrpl/src/Wallet/index.ts
```diff
@@ -318,13 +320,26 @@ class Wallet {
...
+        tx.Amount.value = new BigNumber(tx.Amount.value).toString()
+      }
+    }
+
+    const txToSignAndEncode = { ...tx }
```
JST5000, line 342: Same question - why is this copy necessary?

JST5000: Ah I didn't realize it mutated the parameter. Good catch :)

packages/xrpl/src/Wallet/index.ts
JST5000: But we're setting it equal to itself, so wouldn't it still modify the `transaction` param?

packages/xrpl/src/Wallet/index.ts
JST5000: You may be able to use a look-behind to check for the decimal place instead of `.*` - https://www.regular-expressions.info/lookaround.html

That combined with your `[0-9]` check may fix the issue?

ckniffen (approved):

packages/xrpl/src/Wallet/index.ts
khancode: If `tx.Amount` is an object, it still refers to `transaction.Amount` since it copied the reference to the object. Therefore, we need to do a deep copy of `tx.Amount` that won't reference `transaction.Amount` object.

packages/xrpl/src/Wallet/index.ts
```diff
@@ -377,6 +382,29 @@ class Wallet {
...
+  // eslint-disable-next-line class-methods-use-this -- Helper for organization purposes
+  private removeTrailingZeros(tx: Transaction): void {
+    if (
+      tx.TransactionType === 'Payment' &&
+      typeof tx.Amount !== 'string' &&
+      tx.Amount.value.includes('.') &&
+      tx.Amount.value.endsWith('0')
+    ) {
+      // eslint-disable-next-line no-param-reassign -- Required to update Transaction.Amount.value
+      tx.Amount = { ...tx.Amount }
```
JST5000, line 402: I'm pretty sure if you remove these two lines your code will work exactly the same.

packages/xrpl/src/Wallet/index.ts
```diff
@@ -377,6 +382,29 @@ class Wallet {
...
+  /**
+   * Remove trailing insignificant zeros for non-XRP Payment amount.
+   * This resolves the serialization mismatch bug when encoding/decoding a non-XRP Payment transaction
+   * with an amount that contains trailing insignificant zeros; for example, '123.4000' would serialize
+   * to '123.4' and cause a mismatch.
+   *
+   * @param tx - The transaction prior to signing.
+   */
+  // eslint-disable-next-line class-methods-use-this -- Helper for organization purposes
+  private removeTrailingZeros(tx: Transaction): void {
```
JST5000, line 394: This helper doesn't use `this` so should probably be outside the class

JST5000 (approved):
LGTM (One structural comment, but not a big deal either way)

PR #2006, opened 2022-05-23
Add detail and visibility to fundWallet documentation

JST5000:

Adds detail to the `fundWallet` function, and adds an alias to Wallet.fundWallet to increase the discoverability of the function.

### Context of Change

In response to some issues debugging how fundWallet works after the update to allow for custom faucet urls.

### Type of Change

- [X] Documentation Updates

## Before / After

* Adds a fundWallet alias to Wallet because that's a common spot to search for wallet functions
* Adds an error messages if `faucetHost` has `https://` or `wss://` as that's not the expected format for the function.
* Adds an example to the description of the function for how to use `faucetHost`
* Adds tests

## Test Plan

CI passes

### DIFF
--- a/packages/xrpl/src/Wallet/fundWallet.ts
+++ b/packages/xrpl/src/Wallet/fundWallet.ts
@@ -94,10 +94,7 @@ async function fundWallet(
 
   if (
     options?.faucetHost != null &&
-    (options.faucetHost.startsWith('ws://') ||
-      options.faucetHost.startsWith('wss://') ||
-      options.faucetHost.startsWith('http://') ||
-      options.faucetHost.startsWith('https://'))
+    /^(http|https|ws|wss):\/\//.test(options.faucetHost)
   ) {
     throw new XRPLFaucetError(
       `Invalid format for faucetHost. Should not include ws(s)/http(s) prefix. Received '${options.faucetHost}'.`,
--- a/packages/xrpl/src/Wallet/fundWallet.ts
+++ b/packages/xrpl/src/Wallet/fundWallet.ts
@@ -45,7 +45,8 @@ const MAX_ATTEMPTS = 20
  * @param options.faucetHost - A custom host for a faucet server. `fundWallet` will automatically
  * determine the correct server for devnet and testnet as long as `faucetHost` is null or undefined.
  * In other environments, you should provide the host using this option.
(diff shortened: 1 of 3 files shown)


packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -88,6 +92,17 @@ async function fundWallet(
     /* startingBalance remains '0' */
   }
 
+  if (
+    options?.faucetHost != null &&
+    (options.faucetHost.startsWith('ws://') ||
```
ckniffen, line 97: You could also do `/^(http|https|ws|wss):\/\//.test(options.faucetHost)`.

packages/xrpl/src/Wallet/index.ts
```diff
@@ -382,6 +383,34 @@ class Wallet {
     return classicAddressToXAddress(this.classicAddress, tag, isTestnet)
   }
 
+  /**
```
ckniffen, line 386: Would a `@see` suffice?

packages/xrpl/src/Wallet/index.ts
```diff
@@ -382,6 +383,34 @@ class Wallet {
...
+   * @param client - An XRPL client.
+   * @param options - See below.
+   * @param options.faucetHost - A custom host for a faucet server. `fundWallet` will automatically
+   * determine the correct server for devnet and testnet as long as `faucetHost` is null or undefined.
+   * In other environments, you should provide the host using this option.
+   * Here's an example of how to format `faucetHost`: `faucet.devnet.rippletest.net`
+   * @returns This wallet and that its new balance in XRP.
+   * @throws When either Client isn't connected or unable to fund wallet address.
+   */
+  public async fundWallet(
```
mvadari, line 401: I don't like `fundWallet` being a wallet method. IMO the wallet should be an object that doesn't have anything to do with a client (more specifically, with a network connection) - all it does it handle the keys and do things with the keys (e.g. sign and verify). When actually using a `Wallet` in production, you don't want `fundWallet` anywhere near it.

khancode: I agree. We should just keep `fundWallet` on `Client` and make sure the function is documented well on our docs to improve discoverability.

JST5000:
Closing since this doesn't seem to be something we want to do. If we change our mind we can re-open with the preferred changes.

Issue #2004, opened 2022-05-19
NFTokenMint URI should allow lowercase hex


mDuo13:
The `URI` field of the `NFTokenMint` field is arbitrary binary specified as hex. However, signing a transaction with a URI specified in lowercase hex fails with the error, "Serialized transaction does not match original txJSON."

This appears to be the same problem as #967 but with a new field.

Tested on xrpl.js 2.2.3 using Node.js 16.15.0.

Test case (note, the seed is to an account I funded on NFT-Devnet strictly for this test purpose and has no real value):

```js
'use strict'
const xrpl = require('xrpl')

const wallet = xrpl.Wallet.fromSeed('ss7FshSz5ozfG1vZEGJGotC5f9aWT')

const tx_upperhex = {
  "TransactionType": "NFTokenMint",
  "Account": wallet.address,
  "TransferFee": 314,
  "NFTokenTaxon": 0,
  "Flags": 8,
  "Fee": "10",
  "URI": "697066733A2F2F62616679626569676479727A74357366703775646D37687537367568377932366E6634646675796C71616266336F636C67747179353566627A6469",
  "Memos": [
        {
            "Memo": {
                "MemoType":
                  "687474703A2F2F6578616D706C652E636F6D2F6D656D6F2F67656E65726963",
                "MemoData": "72656E74"
            }
        }
    ]
}

const tx_lowerhex = {
  "TransactionType": "NFTokenMint",
  "Account": wallet.address,
  "TransferFee": 314,
  "NFTokenTaxon": 0,
  "Flags": 8,
  "Fee": "10",
  "URI": "697066733a2f2f62616679626569676479727a74357366703775646d37687537367568377932366e6634646675796c71616266336f636c67747179353566627a6469",
  "Memos": [
        {
            "Memo": {
                "MemoType":
                  "687474703a2f2f6578616d706c652e636f6d2f6d656d6f2f67656e65726963",
                "MemoData": "72656e74"
            }
        }
    ]
}
console.log("-------\nUppercase:\n")
console.log(wallet.sign(tx_upperhex))

console.log("-------\nLowercase:\n")
console.log(wallet.sign(tx_lowerhex))
```

mDuo13:
Note, I confirmed that xrpl-py handles the lowercase hex correctly (it generates identical binary blobs for both of the above formats)

khancode:
Fix has been deployed.

PR #2008, opened 2022-05-23
allow lowercase hex values for `NFTokenMint.URI`

khancode:

Allow lowercase hex values for `NFTokenMint.URI`.

### Context of Change

Signing a transaction with a `URI` specified in lowercase hex fails with the error, "Serialized transaction does not match original txJSON." This fix allows lowercase hex values to be accepted for `URI` in `NFTokenMint` transaction.

https://github.com/XRPLF/xrpl.js/issues/2004

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added a unit test to verify fix works.


packages/xrpl/src/Wallet/index.ts
```diff
@@ -440,7 +440,12 @@ class Wallet {
 
       return memo
     })
-    if (!_.isEqual(decoded, tx)) {
+
+    if (txCopy.TransactionType === 'NFTokenMint' && txCopy.URI) {
+      txCopy.URI = txCopy.URI.toUpperCase()
+    }
```
JST5000, line 446: I think this bug should be fixed on the decode side (So that lowercase URIs decode to lowercase URIs again). I think that because URIs can be case sensitive, so our decode should be case sensitive. (Client servers decide how to interpret the URI after the host - https://webmasters.stackexchange.com/questions/5157/is-it-ok-to-have-a-case-insensitive-uri)

ckniffen (approved):

JST5000 (changes requested):
(See above)

JST5000 (commented):
After talking about this more, it seems that the URI has to eventually be hex encoded, so we're requiring URIs to be submitted in hex when using our library. We should throw an error if it's not in hex to enforce this :)

packages/xrpl/src/Wallet/index.ts
khancode: We require `URI` to be in hex format because it needs to be binary at the protocol level. Here's Rome's explanation for it:

"At the protocol level, it needs to be binary. What form it started out as is immaterial.
So the requirement that it be hex is a detail of the serialization tools. There are two reasons the tools we have now generally require arbitrary binary blob fields to be specified in hex:
It's technologically simpler and more flexible. Our tools can't have bugs in this code because they simply don't have this code at all. :good-thinking:

Automatically encoding/decoding, say, UTF-8 means that you can't also accept hex without having ambiguous cases. ("What if I want to encode a message that happens to also be valid hex?") But if you only accept, say, UTF-8, then there are valid binary values people might want to use that are not valid UTF-8. The URI field is allowed to contain arbitrary data that's not actually a URI; this allows for very small NFTs to be stored actually on-chain. But some of those custom encodings/decodings would be tricky to put in place if you had to write a special serializer to get them into a transaction's binary because they don't produce valid UTF-8.

There are a bunch of potential edge cases in encoding/decoding UTF-8. For example, many UTF-8 characters (many ASCII characters, even) are not valid characters in a URI, but are valid in an IRI (internationalized resource identifier), for example. Should the tool encode these or raise an error?

Requiring hex for binary fields sidesteps these issues and allows each client to choose a way of encoding the data that makes sense for their own circumstances."

ckniffen (approved):

packages/xrpl/src/Wallet/index.ts
```diff
@@ -440,7 +440,12 @@ class Wallet {
 
       return memo
     })
-    if (!_.isEqual(decoded, tx)) {
+
+    if (txCopy.TransactionType === 'NFTokenMint' && txCopy.URI) {
```
ckniffen, line 445: Should this check also live in validateNFTokenMint?

khancode: Ah yes, I'll add it to the model check too.

packages/xrpl/src/Wallet/index.ts
```diff
@@ -440,7 +441,16 @@ class Wallet {
...
+      if (!HEX_REGEX.test(txCopy.URI)) {
+        throw new ValidationError('URI must be a hex value')
+      }
+      console.log('yo')
```
JST5000, line 449: You'll probably want to take this out :P

JST5000 (approved):
LGTM after above comments are resolved :)

ckniffen (approved):

Issue #1528, opened 2021-08-13
Signing an issued currency transaction with lowercase or symbol-containing "Standard" currency codes fails


mDuo13:
When signing a transaction involving an issued currency code that contains lower-case letters or [other allowed symbols](https://xrpl.org/currency-formats.html#standard-currency-codes), ripple-lib raises an error for a (de-)serialization mismatch.

# Steps to Reproduce

Clone https://github.com/XRPLF/xrpl-dev-portal/tree/issue_token_tutorial/content/_code-samples/issue-a-token

(Note, to run in Node.js, you need to uncomment some "require" lines towards the beginning of the file. Or you can just open the demo.html  file in your browser.)

Edit the currency code (line 123) from `"FOO"` to `"$$$"`, `"foo"`, or some other string that should be treated as 3-character "Standard" currency code but is not strictly uppercase alphanumeric.

Run the sample code (note, it depends on a submit-and-verify file in a sibling folder)

# Expected Result

Transaction is signed and returned successfully, the same as if the code was "FOO" or "999".

# Actual Result

```
Uncaught (in promise) extends: Serialized transaction does not match original txJSON. See `error.data`
    u https://unpkg.com/ripple-lib@1.9.8/build/ripple-latest-min.js:2
    <anonymous> https://unpkg.com/ripple-lib@1.9.8/build/ripple-latest-min.js:2
    g https://unpkg.com/ripple-lib@1.9.8/build/ripple-latest-min.js:2
    g https://unpkg.com/ripple-lib@1.9.8/build/ripple-latest-min.js:2
    default https://unpkg.com/ripple-lib@1.9.8/build/ripple-latest-min.js:2
    main file:///another/devel/xrpl-dev-portal/content/_code-samples/issue-a-token/issue-a-token.js:138
ripple-latest-min.js:2:21241
    main file:///another/devel/xrpl-dev-portal/content/_code-samples/issue-a-token/issue-a-token.js:191
    InterpretGeneratorResume self-hosted:1482
    AsyncFunctionNext self-hosted:692
```

From what I can tell, it gets (de-)serialized into the 160-bit hex equivalent of the same currency code, so that gets treated as a mismatch. 

[...]

PR #2009, opened 2022-05-25
Handle edge cases for standard currency code signing

JST5000:

Better handle encoding/decoding standard currency codes when they are lowercase and have allowed symbols.

### Context of Change

In response to #1528 

### Type of Change

## Before / After

* Allows `ripple-binary-codec` to encode/decode lowercase and symbolic standard currencies into standard form.
* Adds explicit errors to xrpl.js for trying to sign a transaction involving lowercase IssuedCurrencyAmounts to avoid impersonation scams.

## Test Plan

CI Passes


packages/xrpl/src/Wallet/index.ts
```diff
@@ -440,6 +447,39 @@ class Wallet {
...
+        // Then standardize the format of currency codes to the 40 byte hex string for comparison
+        if(amount.currency.length !== decodedCurrency.length) {
+          if(decodedCurrency.length == standard_currency_code_len) {
+            (decoded[key] as unknown as IssuedCurrencyAmount).currency = isoToHex(decodedCurrency)
+          } else {
+            txCopy[key].currency = isoToHex(txCopy[key].currency)
+          }
+        }
+      }
+    })
```
JST5000, line 481: Hmm... I don't really like how this implementation adds a performance hit to all signing in exchange for protecting against a rare edge case. 

Alternatives:
1. Enforce this at the `decode` level in `ripple-binary-codec` - That way there's no additional lookup required. (Before this, ripple-binary-codec didn't allow lowercase anyway, this could just be a better error message)
2. Add it as a check when creating an IssuedCurrencyAmount potentially?

Thoughts? @khancode @intelliot

packages/xrpl/test/wallet/index.ts
```diff
@@ -590,6 +591,93 @@ describe('Wallet', function () {
...
+    it('sign throws when a payment contains an issued currency like XRP', async function () {
+      const payment: Payment = { ...issuedCurrencyPayment }
+      payment.Amount = {
+        currency: 'xrp',
+        issuer: 'rnURbz5HLbvqEq69b1B4TX6cUTNMmcrBqi',
+        value: '123.40',
+      }
+      assert.throws(() => {
+        wallet.sign(payment)
+      }, /^The standard currency code provided \(xrp\).*XRP is not an issued currency./u)
```
intelliot, line 630: Why the `*` here?

packages/xrpl/src/Wallet/index.ts
intelliot: Regardless of the currency code, users have to trust the accounts that they open trustlines with. (Hence the name "trust lines".) "usd" could legitimately be used for US Dollars, if issuers and holders agreed to it and creditors understood the counterparty risk.

In https://github.com/XRPLF/xrpl.js/issues/1528 the problem is that the currency code codec is not symmetric: encoding "usd" yields one result, but decoding that result doesn't yield "usd". 3-letter uppercase currency codes made sense 5 years ago, but we are now in a world where even currencies like EUR are being issued with non-standard codes (`EURS`). Because of this, I think our libraries should change with the times.

We have two options:

1. Make the codec symmetric so that "usd" in becomes "usd" out. This is a breaking change unless we leave the existing behavior and add a new method/option for this.
2. Require that lowercase or symbol-containing "Standard" currency codes are provided in hex (similar to what we do for [NFTokenMint.URI](https://github.com/XRPLF/xrpl.js/pull/2008#discussion_r884002507)) - that is what we are currently doing, and we're improving the error messages. If we choose this option, then we should add a method to make it easy to convert to/from hex, and make sure this method is documented and used in tutorials.

JST5000: After talking through this a bit, I'm leaning towards making the codec symmetric and allowing lowercase standard currency codes in xrpl.js

The reason being that folks can already issue capitalized standard currencies to pretend to be legitimate (Ex. "USD"). So like @intelliot pointed out above - users already have to determine if the issued currency is trustworthy by checking the issuer since the currency code offers no protections (They can name it whatever they want besides XRP). 

So, to resolve #1528, I think we should just update the binary-codec to decode lowercase and allowed symbol standard currency codes. This fix should NOT include additional checks in xrpl.js like I implemented here to disallow lowercase standard currency codes.

@mDuo13 Does that resolution seem reasonable?

packages/ripple-binary-codec/test/hash.test.js
```diff
@@ -54,16 +54,20 @@ describe('Currency', function () {
...
-    )
+  test('Currency code with lowercase letters decodes to ISO code', () => {
+    expect(Currency.from('xRp').toJSON()).toBe('xRp')
+  })
+  test('Currency codes with symbols decodes to ISO code', () => {
+    expect(Currency.from('x|p').toJSON()).toBe('x|p')
   })
-  test('Currency codes with symbols decode to hex', () => {
-    expect(Currency.from('x|p').toJSON()).toBe(
-      '000000000000000000000000787C700000000000',
```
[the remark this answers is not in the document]
JST5000: It seems like `rippled` does include these symbols and lowercase letters in it's isoCharSet definition: https://github.com/ripple/rippled/blob/9d89d4c1885e13e3c0aa7cdb4a389c6fbfd3790a/src/ripple/protocol/impl/UintTypes.cpp#L35-L39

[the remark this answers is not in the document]
JST5000: Thanks for pointing that out though - I forgot to actually check what rippled's definition was :)

packages/xrpl/test/wallet/index.ts
JST5000: That was actually because I saw that pattern in other tests of just checking the key messages and not the middle. I shifted the check to be explicitly the same message :)

packages/ripple-binary-codec/test/hash.test.js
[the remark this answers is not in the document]
mvadari: Ok, if rippled supports that conversion then we should do it too

packages/xrpl/src/Wallet/index.ts
mvadari: I think it's fine to allow lowercase standard currency codes. We shouldn't be overly aggressive in what we limit, because there _are_ legitimate usecases that someone might want to do that.

packages/xrpl/HISTORY.md
```diff
@@ -18,6 +21,7 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 * Errors during reliable submission with no error message now properly show the preliminary result instead of a type error
 * Fixed serialize/deserialize verification bug in `Wallet.sign()` when signing a non-XRP Payment with an amount that contains trailing insignificant zeros
 * Allow lowercase hex values for `NFTokenMint.URI`
+* Now allows signing of symbolic standard-currency codes, and gives better errors for lowercase standard currencies.
```
mvadari, line 24: Is this supposed to be in the unreleased section?

JST5000: Ah good catch - I think this line is no longer needed at all :)

packages/xrpl/src/Wallet/index.ts
intelliot: Just curious - what is a legitimate use case where someone would want to do that?

ISO 4217 codes are all capital letters, AFAIK.

intelliot (approved):

ckniffen (approved):

packages/xrpl/src/Wallet/index.ts
mvadari: There are all sorts of tokens on the ledger right now with various names and capitalizations. I figured there's gotta be at least one that has 3 lower-case letters.

ckniffen: One I found that is 3 characters and has at least one lowercase is xSD.  Not to be confused with XSD which is also on the ledger.

intelliot: Agree that they exist and are definitely there. Not sure that they are a legitimate use case. Would any of them be fundamentally worse if they used all capital letters instead?

ckniffen (approved):

intelliot:
Released in xrpl.js 2.3.1 / ripple-binary-codec 1.4.2

intelliot:
Related: https://github.com/XRPLF/xrpl-dev-portal/pull/1417

PR #2010, opened 2022-05-26
feat(source-maps): Generate source-maps for browser build

ckniffen:

Improve debugging for users loading in library via unpkg or jsdelivr

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/packages/xrpl/webpack.config.js
+++ b/packages/xrpl/webpack.config.js
@@ -23,6 +23,7 @@ function getDefaultConfiguration() {
       path: path.join(__dirname, 'build/'),
       filename: `xrpl.default.js`,
     },
+    devtool: 'source-map',
     plugins: [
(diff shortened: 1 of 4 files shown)


packages/xrpl/webpack.config.js
```diff
@@ -45,7 +46,12 @@ function getDefaultConfiguration() {
       }),
     ],
     module: {
-      rules: [],
+      rules: [
+        {
+          test: /\.js$/,
```
khancode, line 51: Should we be checking for `.ts` files too? Or does this check the post-processed TS files that are converted to JS?

ckniffen: The webpack config points at the dist folder as its entrypoint which contains only .js source code.  This rule allows the source maps of transpiled typescript to be reflected in the final sourcemaps.

khancode (approved):
LGTM 👌

intelliot (approved):

PR #2023, opened 2022-06-09
signer_lists interface change to match the node response

LukaAvbreht:

Moved signer_list field from `AccountInfoResponse` to  `AccountRoot` to match the response from node 

### Context of Change

Interface change for response typings

### Type of Change


JST5000:
Good catch! - If you remove the `SignerList` import in `accountRoot.ts` it should resolve the unit test error you're running into.

mvadari:
This is actually incorrect - the `account_info` request does return the `signer_lists` field in the `account_data` field, but if you search by ledger ID using the `ledger_entry` request, you don't get the `signer_lists` field (because it's in a separate object).

So instead of modifying the `AccountRoot` LedgerEntry model, the `account_info` command should really read something like:
```js
account_data: AccountRoot & { signer_lists?: SignerList[] }
```

mvadari:
Closed in favor of #2295

PR #2024, opened 2022-06-10
Add NFT Devnet as an inferable faucet 

JST5000:

Lets fundWallet infer the NFT Devnet faucet when the client is connected to NFT Devnet.

### Context of Change

It's not easy to find the url for the NFT Devnet, and it's really common to want to generate a new wallet for testing purposes on the NFT Devnet, so this seems like a nice quality of life change. (If NFT Devnet turns off, their client won't be pointed to it anyway, and we can then remove this at any point)

### Type of Change

## Test Plan

CI Passes

### DIFF
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -3,6 +3,8 @@
 Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xrpl-announce) for release announcements. We recommend that xrpl.js (ripple-lib) users stay up-to-date with the latest stable release.
 
 ## Unreleased
+### Added
+* When connected to nft-devnet, Client.fundWallet now defaults to using the nft-devnet faucet instead of requiring specification.
 
 ## 2.3.0 (2022-06-02)
 ### Added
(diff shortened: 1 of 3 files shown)


JST5000:
> It was a purposeful choice by @ledhed2222 not to add this faucet to the library, because it's not a permanent network, so we'd have to remove it at some point. Perhaps we can just add it as a note to the documentation instead.

Had a brief chat with him about it. 

The trade offs are:
* If we add it in, we'll want to remove it when we remove support for the NFT Devnet
* We're already allowing folks to define their own faucet URLs, so we could just let them do that for NFT Devnet (As we are now)
* By adding it though we simplify the developer experience (Hunting down the URL for the faucet requires inspecting an element on the faucet page currently, if we were to add docs for the NFT Devnet to the faucet page, we'd still be requiring them to visit a specific page to get basic functionality for automated testing)
* In the case we turn off the NFT Devnet, the inferred faucet will error (or point to a dead link), but that's the same thing that'll happen currently if folks are defining the URL themselves with the custom URL method. (So I don't think it affects this decision)

Based on that, my thought is that this is overall worth it because:
1. We'll be very aware when the NFT Devnet is being turned off, so also removing the inference seems fine. 
2. Since generating a wallet from the faucet is a core part of almost every automated test, this seems like an important quality of life change.

(I also think we should add Hooks V2 inference because it also seems to be a longer-standing test network with a hard-to-find faucet URL that we want to allow easy experimentation with)

ckniffen (approved):

khancode (approved):

intelliot (approved):
in addition to this, I'd be in favor of revamping the faucets page and adding NFT-Devnet info on xrpl.org https://github.com/XRPLF/xrpl-dev-portal

PR #2025, opened 2022-06-14
Update currency code parsing: lower case chars & some special chars

WietseWind:
`ripple-binary-codec` would turn specific three char currency codes into 40 char HEX currency format. Which is great, except the rules applied to determine if something is a valid three character currency code were off.

### Problem description

When encoding a transaction, `ripple-binary-codec` would change a currency code like `Ab!` into the HEX representation, while the token `Ab!` is a valid three char currency code, and can be issued, traded, sent, etc. just fine on the XRPL (rippled accepts this, as per docs (see below)).

### Old (incorrect) binary codec regex:
```
const ISO_REGEX = /^[A-Z0-9]{3}$/
```

This regex assumes exactly three upper case or numeric chars. 

### New (correct) binary codec regex:
```
const ISO_REGEX = /^[A-Z0-9?!@#$%^&*<>(){}[\]|]{3}$/i
```

This regex matches the docs & the behaviour of rippled:
> Currency codes must be **exactly 3 ASCII characters** in length. The following characters are permitted: **all uppercase and lowercase** letters, digits, **as well as the symbols** ?, !, @, #, $, %, ^, &, *, <, >, (, ), {, }, [, ], and |.

Source: https://xrpl.org/currency-formats.html#standard-currency-codes

### DIFF
--- a/packages/ripple-binary-codec/src/types/currency.ts
+++ b/packages/ripple-binary-codec/src/types/currency.ts
@@ -2,7 +2,7 @@ import { Hash160 } from './hash-160'
 import { Buffer } from 'buffer/'
 
 const XRP_HEX_REGEX = /^0{40}$/
-const ISO_REGEX = /^[A-Z0-9]{3}$/
+const ISO_REGEX = /^[A-Z0-9?!@#$%^&*<>(){}[\]|]{3}$/i
 const HEX_REGEX = /^[A-F0-9]{40}$/
 // eslint-disable-next-line no-control-regex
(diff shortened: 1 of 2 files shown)


mvadari:
There's already an in-progress PR regarding this issue: https://github.com/XRPLF/xrpl.js/pull/2009

WietseWind:
Thanks @mvadari, I didn't know. Seems to be about the same thing indeed. If it's already WIP & this PR isn't needed, feel free to close this one.

JST5000:
Closing as I believe #2009 fixes this :)

PR #2026, opened 2022-06-14
add ExpandedSignerList amendment support

khancode:

Adds [ExpandedSignerList](https://xrpl.org/known-amendments.html#expandedsignerlist) amendment support.

### Context of Change

- Expands the maximum signer list size to 32 entries.
- Adds an optional field `WalletLocator` to `SignerEntry`.
- [rippled PR](https://github.com/ripple/rippled/pull/4097)

### Type of Change

- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Updates
- [ ] Release

## Test Plan

Adds unit & integration tests.

## Future Tasks
Merge this PR once the amendment goes live.


ckniffen (approved):

packages/xrpl/src/models/common/index.ts
```diff
@@ -61,10 +61,31 @@ interface PathStep {
...
+     * A multi-signature is only valid if the sum weight of the signatures provided meets
+     * or exceeds the signer list's SignerQuorum value.
+     */
     SignerWeight: number
+    /**
+     * An arbitrary 256-bit (32-byte) field that can be used to identify the signer, which
+     * may be useful for smart contracts, or for identifying who controls a key in a large
+     * organization.
+     */
+    WalletLocator?: string
```
JST5000, line 88: I think something like `SignerMetadata` would be a better name here because it allows arbitrary data. (And although we think the most common use case will be for identifying the signer, I don't think `WalletLocator` accurately conveys the dynamic data type)

packages/xrpl/src/models/transactions/signerListSet.ts
```diff
@@ -20,13 +20,15 @@ export interface SignerListSet extends BaseTransaction {
...
+   * more than 32 members. No address may appear more than once in the list, nor
    * may the Account submitting the transaction appear in the list.
    */
   SignerEntries: SignerEntry[]
 }
 
-const MAX_SIGNERS = 8
+const MAX_SIGNERS = 32
+
+const HEX_WALLET_LOCATOR_REGEX = /^[0-9A-Fa-f]{64}$/u
```
JST5000, line 31: This matches 64 bytes, but the HEX_WALLET_LOCATOR is supposed to be 32 bytes.

JST5000: Nevermind, thought you needed 1 hex per byte, but you need 2. Oops :)

packages/xrpl/src/models/common/index.ts
khancode: @JST5000 `WalletLocator` is the field name that the XRPL expects in the transaction - https://xrpl.org/signerlist.html#signer-entry-object

JST5000: Ah that makes sense, my bad :)

JST5000 (approved):
LGTM

intelliot:
It is possible to rename the field to `SignerMetadata` since the actual data is in a binary format, and clients can interpret (map) / encode any field names they want. But given the code in rippled and on the docs, keeping the `WalletLocator` name is certainly the path of least resistance

khancode:
> It is possible to rename the field to `SignerMetadata` since the actual data is in a binary format, and clients can interpret (map) / encode any field names they want. But given the code in rippled and on the docs, keeping the `WalletLocator` name is certainly the path of least resistance

@intelliot I believe the better path is to use the same parameter name, `WalletLocator`, that's on the docs to avoid confusion. AFAIK, we've been consistent with our models containing the same field names as what's on the docs. If users do get confused by the name, they can read the JSDocs or the official docs to understand its usage. In addition, we can propose the name change `SignerMetadata` to rippled team and see if they can refactor the name and publish the change. Not sure if this would have to be in a different amendment or if they can modify the amendment?

JST5000:
> @intelliot I believe the better path is to use the same parameter name, `WalletLocator`, that's on the docs to avoid confusion. AFAIK, we've been consistent with our models containing the same field names as what's on the docs. If users do get confused by the name, they can read the JSDocs or the official docs to understand its usage.

Re: the first part, just wanted to add +1 to staying consistent with rippled's naming/interface.

intelliot:
The rippled team is aware of this and there's no need to change it. It was discussed in the PR you linked to, https://github.com/ripple/rippled/pull/4097

There's nothing strictly canonical about the JSON format: only the binary really matters at a protocol level. But given that rippled is currently the only implementation, I agree that we should just use the same names unless there is a very good reason not to.

intelliot (approved):

intelliot:
1) Can we add a caveat in the release notes to say that the limit is 8 unless the [ExpandedSignerList](https://xrpl.org/known-amendments.html#expandedsignerlist) amendment is active?
2) What error should users expect if they attempt to set 9+ signers on a network where `ExpandedSignerList` is *not* active? It may be useful to document this in the release notes (or elsewhere) as well.

PR #2046, opened 2022-07-15
Fix Wallet.fromMnemonic 

### LINKED ISSUE #2042: Library generates same wallet with any mnemonic generated
return xrpl.Wallet.fromMnemonic(secret, {
          mnemonicEncoding: 'rfc1751',
          algorithm: 'ecdsa-secp256k1',
        })
    }

// light maze exit awful hire ramp trigger will response offer throw hobby
// glue belt debris word submit clump provide organ waste group plunge glory
//frgrfge reer rere reer frerf ergre reer reter erert erret erwre ertrt

SAME Wallet address generated for all these mnemonics :rGzzsoiXsM3RmHU9Fu9eDQ6hoNoouyQiNT

These mnemonics were created using xrpl-accountlib.generate.mnemonic({strength:128}) 
( https://www.npmjs.com/package/xrpl-accountlib )

### LINKED ISSUE #2044: xrpl.Wallet.fromMnemonic doesn't work with react native android [BigInt Error]
Hi,
xrpl.Wallet.fromMnemonic does not work with react-native [android]. we debugged the root cause and found the 
`swap128()` to be the culprit.
An alternate implementation seems to work adding here for other people who may be facing this issue.
Earlier
```js
function swap128(buf) {
     const result = Buffer.alloc(16);
     result.writeBigUInt64LE(buf.readBigUInt64BE(0), 8);
     result.writeBigUInt64LE(buf.readBigUInt64BE(8), 0);
     return result;
}
```
Working Code.

```js
function swap128(buf) {
  // Interprets buffer as an array of (two, in this case) 64-bit numbers and swaps byte order in-place.
  const buf2 = buf.swap64();

  // Swap the two 64-bit numbers since our buffer is 128 bits.
  return Buffer.concat([buf2.slice(8, 16), buf2.slice(0, 8)], 16);
}
```

Thanks
Himanshu
Spend the Bits.

JST5000:

Updates Wallet.fromMnemonic to have better validation checks, work with react-native on android, and 

### Context of Change

In response to a couple issues raised surrounding the function #2042 and #2044. (Exact changes made as a result are listed below)

### Type of Change

## Before / After

* There are now validation checks on whether an encoding is valid in BOTH bip39 and rfc1751 encodings (So should not be possible to end up in a situation like #2042)
* Lowercase words now work with rfc1751 encoding (Fixes #2042) 
* It wasn't working in react-native for Android (Fixes #2044) 
* Added more tests for common mistakes people could make (And ensured that they properly error)

## Test Plan

CI Passes

### DIFF
--- a/packages/xrpl/src/Wallet/rfc1751.ts
+++ b/packages/xrpl/src/Wallet/rfc1751.ts
@@ -177,10 +177,14 @@ function bufferToArray(buf: Buffer): number[] {
  * @returns A buffer containing the same data with reversed endianness
  */
 function swap128(buf: Buffer): Buffer {
-  const result = Buffer.alloc(16)
-  result.writeBigUInt64LE(buf.readBigUInt64BE(0), 8)
-  result.writeBigUInt64LE(buf.readBigUInt64BE(8), 0)
-  return result
+  // Interprets buffer as an array of (two, in this case) 64-bit numbers and swaps byte order in-place.
+  const reversedBytes = buf.swap64()
+
+  // Swap the two 64-bit numbers since our buffer is 128 bits.
+  return Buffer.concat(
+    [reversedBytes.slice(8, 16), reversedBytes.slice(0, 8)],
+    16,
+  )
 }
 
 export { rfc1751MnemonicToKey, keyToRFC1751Mnemonic }
--- a/packages/xrpl/src/Wallet/rfc1751.ts
+++ b/packages/xrpl/src/Wallet/rfc1751.ts
@@ -138,7 +138,7 @@ function getSubKey(
   for (word of sublist) {
     const idx = rfc1751WordList.indexOf(word.toUpperCase())
     if (idx === -1) {
-      throw TypeError(
+      throw new TypeError(
         `Expected an RFC1751 word, but received '${word}'. ` +
           `For the full list of words in the RFC1751 encoding see https://datatracker.ietf.org/doc/html/rfc1751`,
       )
--- a/packages/xrpl/src/Wallet/rfc1751.ts
+++ b/packages/xrpl/src/Wallet/rfc1751.ts
@@ -136,7 +136,13 @@ function getSubKey(
   const ch = [0, 0, 0, 0, 0, 0, 0, 0, 0]
   let word = ''
   for (word of sublist) {
-    const idx = rfc1751WordList.indexOf(word)
+    const idx = rfc1751WordList.indexOf(word.toUpperCase())
+    if (idx === -1) {
+      throw TypeError(
+        `Expected an RFC1751 word, but received '${word}'. ` +
+          `For the full list of words in the RFC1751 encoding see https://datatracker.ietf.org/doc/html/rfc1751`,
+      )
+    }
     const shift = (8 - ((bits + 11) % 8)) % 8
     const y = idx << shift
     const cl = y >> 16
--- a/packages/xrpl/test/wallet/index.ts
+++ b/packages/xrpl/test/wallet/index.ts
@@ -143,6 +143,41 @@ describe('Wallet', function () {
       assert.equal(wallet.seed, expectedSeed)
     })
 
+    it('throws an error when using an RFC1751 mnemonic for bip39', function () {
+      const algorithm = ECDSA.ed25519
+      const mnemonic =
+        'CAB BETH HANK BIRD MEND SIGN GILD ANY KERN HYDE CHAT STUB'
+      assert.throws(() => {
+        Wallet.fromMnemonic(mnemonic, {
+          mnemonicEncoding: 'bip39',
+          algorithm,
+        })
+      }, /^Unable to parse the given mnemonic using bip39 encoding$/u)
+    })
+
+    it('throws an error when using an bip39 mnemonic for RFC1751', function () {
+      const mnemonic =
+        'draw attack antique swing base employ blur above palace lucky glide clap pen use illegal'
+      assert.throws(() => {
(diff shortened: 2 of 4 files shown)


khancode (approved):
LGTM

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -171,10 +177,14 @@ function bufferToArray(buf: Buffer): number[] {
...
-  result.writeBigUInt64LE(buf.readBigUInt64BE(8), 0)
-  return result
+  // Interprets buffer as an array of (two, in this case) 64-bit numbers and swaps byte order in-place.
+  const reversedBytes = buf.swap64()
+
+  // Swap the two 64-bit numbers since our buffer is 128 bits.
+  return Buffer.concat(
+    [reversedBytes.slice(8, 16), reversedBytes.slice(0, 8)],
+    16,
+  )
```
khancode, line 187: Is this related to the fix?

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -136,7 +136,13 @@ function getSubKey(
   const ch = [0, 0, 0, 0, 0, 0, 0, 0, 0]
   let word = ''
   for (word of sublist) {
-    const idx = rfc1751WordList.indexOf(word)
+    const idx = rfc1751WordList.indexOf(word.toUpperCase())
```
intelliot, line 139: Instead of allowing words to be lowercase (or have lowercase letters), it may make sense to throw an error. I believe rippled RFC-1751 always provides mnemonics in uppercase. If the mnemonic is not provided in uppercase, it is likely that the caller is doing something wrong.

If they're not doing something wrong, then it is trivial for them to use `toUpperCase()` themselves - which helps to ensure that they know what they are doing.

```suggestion
    const idx = rfc1751WordList.indexOf(word)
```

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -136,7 +136,13 @@ function getSubKey(
...
+    if (idx === -1) {
+      throw new TypeError(
+        `Expected an RFC1751 word, but received '${word}'. ` +
+          `For the full list of words in the RFC1751 encoding see https://datatracker.ietf.org/doc/html/rfc1751`,
```
intelliot, line 143:
```suggestion
          `Words must be in uppercase and in accordance with RFC-1751: https://datatracker.ietf.org/doc/html/rfc1751`,
```

packages/xrpl/test/wallet/index.ts
```diff
@@ -143,6 +143,41 @@ describe('Wallet', function () {
...
+    })
+
+    it('throws an error when using an bip39 mnemonic for RFC1751', function () {
+      const mnemonic =
+        'draw attack antique swing base employ blur above palace lucky glide clap pen use illegal'
+      assert.throws(() => {
+        Wallet.fromMnemonic(mnemonic, {
+          mnemonicEncoding: 'rfc1751',
+        })
+      }, /^Expected an RFC1751 word, but received 'attack'\. For the full list of words in the RFC1751 encoding see https:\/\/datatracker\.ietf\.org\/doc\/html\/rfc1/u)
```
intelliot, line 165:
```suggestion
      }, /^Expected an RFC1751 word, but received 'attack'\. Words must be in uppercase and in accordance with RFC-1751: https:\/\/datatracker\.ietf\.org\/doc\/html\/rfc1/u)
```

intelliot:
On further thought - I don't actually care whether we allow lowercase. Take it or leave it :)

intelliot (approved):
A follow-up task related to this PR is to improve the docs and change the example in the docs so that it doesn't look like a mnemonic is the primary way to create a Wallet. Most wallets should use a family seed or secret numbers. bip39 mnemonics are supported (and we should confirm that our implementation is compatible with Ledger Nano and others). But RFC1751 mnemonics should be discouraged. (In fact, the option itself should be deprecated in a follow-up PR)

https://js.xrpl.org/classes/Wallet.html

packages/xrpl/test/wallet/index.ts
```diff
@@ -211,11 +246,11 @@ describe('Wallet', function () {
...
   describe('fromMnemonic', function () {
     const mnemonic =
-      'try milk link drift aware pass obtain again music stick pluck fold'
+      'assault rare scout seed design extend noble drink talk control guitar quote'
     const publicKey =
-      '0257B550BA2FDCCF0ADDA3DEB2A5411700F3ADFDCC7C68E1DCD1E2B63E6B0C63E6'
+      '035953FCD81D001CF634EB44A87940F3F98ADF2483D09C914BAED0539BE50F385D'
     const privateKey =
-      '008F942B6E229C0E9CEE47E7A94253DABB6A9855F4BA2D8A741FA31851A1D423C3'
+      '0013FC461CA5799F1357C8130AF703CBA7E9C28E072C6CA8F7DEF8601CDE98F394'
```
mvadari, line 253: Why were all these test cases changed? Were they no longer valid?

JST5000: Yeah, they were invalid bip39 encodings to begin with when I checked using this site and the bip39's `isValid` function: https://iancoleman.io/bip39/

mvadari: lmao

JST5000:
> On further thought - I don't actually care whether we allow lowercase. Take it or leave it :)

I think allowing lowercase is fine. If the encoding follows the RFC1751 standard I think that's indication enough of what they hope the output will be.

PR #2056, opened 2022-07-25
Make the unknown submitAndWait error more verbose

JST5000:

Provides more context when an unknown error happens during the submitAndWait function

### Context of Change

The integration test for `submitAndWait` is being flaky and giving this error, but it's not possible to debug because the message is undefined. This should allow us to debug it if it happens in the future, in addition to providing a better experience for folks running into unexpected exceptions using `submitAndWait`.

### Type of Change

## Before / After

Added more information to the exception message.

## Test Plan

CI Passes

### DIFF
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -3,6 +3,9 @@
 Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xrpl-announce) for release announcements. We recommend that xrpl.js (ripple-lib) users stay up-to-date with the latest stable release.
 
 ## Unreleased
+### Added
+* Made unexpected errors in `submitAndWait` more verbose to make them easier to debug.
+
 ### Fixed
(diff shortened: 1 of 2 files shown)


ckniffen (approved):

intelliot (approved):

packages/xrpl/HISTORY.md
```diff
@@ -3,6 +3,9 @@
 Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xrpl-announce) for release announcements. We recommend that xrpl.js (ripple-lib) users stay up-to-date with the latest stable release.
 
 ## Unreleased
+### Added
+* Made unexpected errors in `submitAndWait` more verbose to make them easier to debug.
```
mvadari, line 7: I would call this a fix not an "added" - I think of the "added" section as new features (aka things that require a minor version bump, not a patch)

PR #2073, opened 2022-08-31
feat: Export verify

JST5000:

Exports the `verify` function from `ripple-binary-codec` so it can be used in web apps.

### Context of Change

Trying to use the function in a webapp, but ripple-keypairs doesn't have webpacking.

### Type of Change

## Test Plan

CI Passes

### DIFF
--- a/packages/xrpl/src/utils/index.ts
+++ b/packages/xrpl/src/utils/index.ts
@@ -15,7 +15,7 @@ import {
   xAddressToClassicAddress,
 } from 'ripple-address-codec'
 import * as rbc from 'ripple-binary-codec'
-import { verify as verifySignature } from 'ripple-keypairs'
+import { verify as verifyKeypairSignature } from 'ripple-keypairs'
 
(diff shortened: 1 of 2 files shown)


mDuo13 (approved):
:tada:

ckniffen (approved):

packages/xrpl/src/utils/index.ts
```diff
@@ -194,6 +195,7 @@ export {
   deriveAddress,
   deriveXAddress,
   signPaymentChannelClaim,
+  verify,
```
mvadari, line 198: Is it not ambiguous to have a method just called `verify` in the utils?

JST5000: Ah, that is ambiguous. Good point.

JST5000: Renamed to `verifySignature`

ckniffen (approved):

PR #2083, opened 2022-09-02
feat: amm devnet to fundWallet and custom faucetPaths

connorjchen:

Allows inferencing for amm devnet. Also, custom faucetPaths are enabled to give users customizable endpoints instead of defaulting to accounts.

### Context of Change

Seems like a nice quality of life upgrade for folks building with AMM devnet, which is a test network which is planned to be long-standing, and involves a feature we’d like more folks to play around with.. (It’s hard to find the URL for the hooks v2 testnet faucet, and generating new test accounts is a nice feature for test code)

As we add more sidechains and different faucetNetworks may not always have /accounts as url to fundWallet, we need an optional custom path parameter to pass in

### Type of Change

- [x] Documentation Updates

## Test Plan

Ran linter, unit tests, integration tests


connorjchen:
For separate hooks v2 logic, created separate processHooksV2Response function - lot of redundant code after parsing initial data 

Options:
1. Leave my implementation
2. Create another function for the bottom part logic that repeats
3. Take out top parse part and add that into function that ends up calling processResponse

JST5000:
> Also. single browser test fails - unsure why

For context, the browser tests simply try to run the normal integration tests from a headless browser. So you can see the actual error message a bit above the "Browser test failed". (In this case it seems it can't generate a wallet using the Hooks faucet)

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines -- barely exceeds limit */
```
JST5000, line 1: This comment is likely to become incorrect over time (When the file extends, people won't update this line). Can't think of a better one though, so maybe fine?

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -18,10 +19,28 @@ interface FaucetWallet {
   balance: number
 }
 
+interface HooksV2FaucetWallet {
+  address: string
+  secret: string
+  xrp: number
+  hash: string
+  code: string
+}
```
JST5000, line 28: Would it be possible to have comments explaining what each of these fields is? (If you know)

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -90,7 +116,19 @@ async function fundWallet(
...
+    this,
+    postBody,
+    options?.faucetHost,
+    options?.faucetPath,
+  )
+
+  // Currently, hooks testnet v2 has no way of funding a given wallet
+  if (httpOptions.hostname === 'hooks-testnet-v2.xrpl-labs.com' && wallet) {
+    throw new XRPLFaucetError(
+      'Cannot fund passed in wallet with hooks testnet v2',
```
JST5000, line 129: Ideally the error message is descriptive enough that the additional comment isn't necessary :) (Also added in a work around that was described to me)

```suggestion
  if (httpOptions.hostname === 'hooks-testnet-v2.xrpl-labs.com' && wallet) {
    throw new XRPLFaucetError(
            'Currently the Hooks Testnet v2 faucet has no way of funding a given wallet. If you need to do that, you can create a new funded account and have it send a payment transaction to the existing account.',
```

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -179,6 +221,8 @@ async function onEnd(
       resolve,
       reject,
     )
+  } else if (client.url === 'wss://hooks-testnet-v2.xrpl-labs.com') {
```
JST5000, line 224: I don't think this would work for the JSON client because it doesn't use the `wss`, it uses `http`/`https`. Potentially look for `client.url.contains('hooks-testnet-v2.xrpl-labs.com')` (Not sure if contains is valid js, but that's the idea)

packages/xrpl/src/Wallet/fundWallet.ts
ckniffen: You could move each of the two processing functions to their own files to avoid the error.

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -320,11 +422,27 @@ function getFaucetHost(client: Client): FaucetNetwork | undefined {
...
 }
 
+/**
+ * Get the faucet pathname based on the faucet hostname.
+ *
+ * @param hostname - hostname.
+ * @returns A String with the correct path for the input hostname.
+ * If hostname undefined or cannot find (key, value) pair in {@link FaucetNetworkPaths}, defaults to '/accounts'
+ */
+function getFaucetPath(hostname: string | undefined): string {
```
JST5000, line 432:
```suggestion
function getDefaultFaucetPath(hostname: string | undefined): string {
```
(This only works for known addresses and ones following the default format, so I think it'd be clearer to add that to the function name)

connorjchen:
Must implement hooks v2 fund given wallet still - working with Wietsl on changing faucet to match other faucets

connorjchen:
may want to make helper functions for fund wallet tests

packages/xrpl/HISTORY.md
```diff
@@ -1,207 +1,244 @@
 # xrpl.js (ripple-lib) Release History
 
 Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xrpl-announce) for release announcements. We recommend that xrpl.js (ripple-lib) users stay up-to-date with the latest stable release.
+
```
mvadari, line 4: These HISTORY changes are unnecessary. If you want them, they should be in a separate PR, because otherwise they make this PR look massive.

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -42,22 +32,26 @@ const MAX_ATTEMPTS = 20
...
- * faucet host in devnet or testnet, you should provide the host using this
- * option.
+ * @param options.faucetHost - A custom host for a faucet server. On devnet, testnet, AMM devnet, NFT devnet, and HooksV2 testnet,
+ * `fundWallet` will attempt to determine the correct server automatically. In other environments, or if you would like to
+ * customize the faucet host in devnet or testnet, you should provide the host using this option.
+ * @param options.faucetPath - A custom path for a faucet server. On devnet, testnet, AMM devnet, NFT devnet, and HooksV2 testnet,
+ * `fundWallet` will attempt to determine the correct path automatically. In other environments, or if you would like to customize
+ * the faucet path in devnet or testnet, you should provide the path using this option.
+ * Ex: client.fundWallet(null,{'faucet.altnet.rippletest.net', '/accounts'}) specifies a request to
+ * 'faucet.altnet.rippletest.net/accounts' to fund a new wallet.
```
mvadari, line 42: These should be wrapped to be the length that the linter checks for (I think it's 80 or 88 characters)

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -144,15 +142,19 @@ async function returnPromise(
   })
 }
 
+// eslint-disable-next-line max-params -- helper function
 function getHTTPOptions(
   client: Client,
   postBody: Uint8Array,
   hostname?: string,
+  pathname?: string,
```
mvadari, line 150: These should be in an `options` so it's neater and easier to add more optional params in the future

packages/xrpl/HISTORY.md
```diff
@@ -17,6 +17,8 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 * Support for ExpandedSignerList amendment that expands the maximum signer list to 32 entries.
 * Add `cookie` and `data` to `ValidationStream` interface
 * Additional check for memos field format, provide more detailed error messages.
+* When connected to hooks v2 testnet or amm devnet, Client.fundWallet now defaults to using the faucet instead of requiring specification.
```
jonathanlei, line 20: I think those need to be moved to unreleased instead of previous release?

jonathanlei (approved):

packages/xrpl/test/wallet/fundWallet.ts
```diff
@@ -30,7 +33,30 @@ describe('Get Faucet host ', function () {
...
+    )
+  })
+
+  it('returns the correct faucetPath for undefined host', function () {
+    const expectedFaucetPath = '/accounts'
+
+    assert.strictEqual(getDefaultFaucetPath(undefined), expectedFaucetPath)
+  })
+
+  it('returns undefined if not a Testnet, Devnet, NFT-Devnet, or AMM devnet server URL', function () {
```
JST5000, line 59:
```suggestion
  it('throws if not connected to a known faucet host', function () {
```

JST5000:
LGTM after that one wording change on the test name :)

PR #2101, opened 2022-09-23
Make it possible to modify `definitions.json` at runtime

JST5000:

Make it easy to dynamically update `definitions.json`

### Context of Change

With Sidechains being developed and people working on new amendments, it'd be helpful to have the ability to dynamically update the `definitions.json` file to incorporate the new transactions and types that are desired without forking the xrpl.js repository.

### Type of Change

## Before / After

Before the `definitions.json` file was hardcoded and could only be updated by forking the repository.
Now you should be able to update the `DEFINITIONS` variable to modify the defined types at runtime.

## Test Plan

CI Passes (Added in new tests for updating the DEFINITIONS and then encoding/decoding)


JST5000:
@RichardAH Hey, not sure if you have any thoughts on the test cases which should go into this - I wasn't sure how to port over the Hooks Transactions as potential test cases. (If you have any documentation links I should read that might help me understand how to add a new transaction I'd appreciate it :) ) - Regardless, I'll keep tinkering with this over next week.

package-lock.json
```diff
@@ -68,7 +68,7 @@
         "webpack-cli": "^4.2.0"
       },
       "engines": {
-        "node": ">=10.0.0",
+        "node": ">=12.0.0",
```
mvadari, line 71: Why is this change here?

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -66,8 +66,9 @@ function serializeObject(object: JsonObject, opts: OptionObject = {}): Buffer {
   const filter = signingFieldsOnly
     ? (f: FieldInstance): boolean => f.isSigningField
     : undefined
-
-  coreTypes.STObject.from(object, filter).toBytesSink(bytesList)
+  ;(coreTypes.STObject as typeof STObject)
```
mvadari, line 69: What does the `;` do here?

mvadari, line 69: Why do all the types need to be casted now?

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -31,30 +33,70 @@ Sort code numbers are reused for fields of different types, but different fields
...
+### Transaction Types
 
 See https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/TxFormats.h
+
+# Defining Your Own Definitions
+
+If you're building your own side chain or amendment for the XRPL, you may need to create new definitions.
+
+To do that there are three high level things you need to do:
+1. Copy and modify the `definitions.json` file to have your definitions ([link](https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/enums/definitions.json)) as a baseline, or you can generate your own from rippled source code with [this tool](https://github.com/RichardAH/xrpl-codec-gen)
```
mvadari, line 69: We do not want to recommend that people edit their definitions file by hand - it's a pain and it's really easy to make mistakes.

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -31,30 +33,70 @@ Sort code numbers are reused for fields of different types, but different fields
...
+  // Should be the same as UInt32
+}
+
+const extendedCoreTypes = { ...coreTypes }
+extendedCoreTypes['NewType'] = NewType
+
+const newDefs = new DefinitionContents(newTypeDefs)
+DEFINITIONS.updateAll(newDefs, extendedCoreTypes)
+
+// From this point on, we should be able to serialize / deserialize Transactions with fields that have 'NewType' as their Type.
```
mvadari, line 100: Does this new API work if you have ripple-binary-codec installed, instead of having the source code locally?

I'm not a huge fan of this API - it requires you to dig into the depths of ripple-binary-codec, which is pretty complicated. Is it possible to instead just accept extra params into the top level methods, like `encode(blob, newDefs, extendedCoreTypes)`?

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -70,22 +72,34 @@ export class Bytes {
...
 
+  /**
+   * Add a new name value pair to the BytesLookup. Does not work properly if either name or value already
+   * exist within the BytesLookup.
+   *
+   * @param name - A human readable name for the field.
+   * @param value - The numeric value for the field.
+   */
+  add(name: string, value: number): void {
+    this[name] = new Bytes(name, value, this.ordinalWidth)
```
mvadari, line 87: `this.ordinalWidth` doesn't ever appear to be initialized.

packages/ripple-binary-codec/src/enums/README.md
mvadari: You could then call those same methods in `encode`, instead of asking the user to do so.

packages/ripple-binary-codec/src/enums/index.ts
ckniffen: It comes from the constructor and is a required parameter.

packages/ripple-binary-codec/src/binary.ts
JST5000: The `;` is used to prevent the interpreter from treating the casted object as a parameter of an `undefined` function like so: (This is what the code becomes with prettier otherwise - it can't tell automatically where the variable declaration ends if it's a statement I believe)
```
  const filter = signingFieldsOnly
    ? (f: FieldInstance): boolean => f.isSigningField
    : undefined(coreTypes.STObject as typeof STObject)
```

JST5000: Before I believe they were `any` - This specific instance looks at a single type which is in `coreTypes` which we know more specifically than the BaseClass. (So we cast it down to the level we need for the more specific `from` function)

JST5000:
Closing in favor of #2127 which implements this as a runtime parameter instead of a global variable.

PR #2102, opened 2022-09-26
fix: corrected snippet to show regular key works

connorjchen:

Code snippet for regular wallet did not do as intended. Fixed one liner to make it so tx was signed by regular key wallet and not original wallet as a normal tx would.

### Context of Change

Snippet was misleading (comments stated one thing, code did another)

### Type of Change

- [x] Documentation Updates

## Before / After

Edited one line in snippet to sign with reg key rather than wallet

## Test Plan

Ran snippet in separate window


JST5000 (approved):
LGTM

PR #2119, opened 2022-10-27
feat: allow `verifyTransaction` to verify transaction objects, not just strings

mvadari:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/packages/xrpl/test/wallet/index.ts
+++ b/packages/xrpl/test/wallet/index.ts
@@ -1,5 +1,5 @@
 import { assert } from 'chai'
-import { decode } from 'ripple-binary-codec/dist'
+import { decode } from 'ripple-binary-codec'
 import { NFTokenMint, Payment, Transaction } from 'xrpl-local'
 import ECDSA from 'xrpl-local/ECDSA'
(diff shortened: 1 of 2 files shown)


packages/xrpl/test/wallet/index.ts
```diff
@@ -949,6 +949,16 @@ describe('Wallet', function () {
...
+
+    it('returns true when verifying a deserialized Transaction object', function () {
+      const wallet = new Wallet(publicKey, privateKey)
+      const decodedTransaction = decode(
+        prepared.signedTransaction,
+      ) as unknown as Transaction
+      const isVerified: boolean = wallet.verifyTransaction(decodedTransaction)
+
+      assert.equal(isVerified, true)
+    })
```
JST5000, line 961: Should there also be a test that fails for an invalid transaction? (Ex. Improper use of string in a number field?)

mvadari: I think that's a broader scope than this PR - we don't do _any_ checks on the transaction data right now.

jonathanlei (approved):
LGTM

PR #2120, opened 2022-10-27
docs: code snippets for multisigning

pdp2121:

Add quality of life code snippets for multisigning.

### Type of Change

- [x] Documentation Updates
- [ ] Release

### DIFF
--- a/packages/xrpl/snippets/src/multisigning.ts
+++ b/packages/xrpl/snippets/src/multisigning.ts
@@ -61,9 +61,13 @@ async function multisigning(): Promise<void> {
       )
     }
   } else {
-    console.log('The multisigned transaction was rejected by rippled. Here's the response from rippled:')
+    console.log(
+      "The multisigned transaction was rejected by rippled. Here's the response from rippled:",
+    )
     console.log(submitResponse)
   }
+
+  await client.disconnect()
 }
 
 void multisigning()
--- a/packages/xrpl/snippets/src/multisigning.ts
+++ b/packages/xrpl/snippets/src/multisigning.ts
@@ -57,7 +57,7 @@ async function multisigning(): Promise<void> {
     console.log(submitResponse)
     if (submitResponse.result.tx_json.Signers) {
       console.log(
-        `Signers contain ${submitResponse.result.tx_json.Signers.length} entries`,
+        `The transaction had ${submitResponse.result.tx_json.Signers.length} signatures`,
       )
     }
   } else {
--- a/packages/xrpl/snippets/src/multisigning.ts
+++ b/packages/xrpl/snippets/src/multisigning.ts
@@ -12,7 +12,7 @@ async function multisigning(): Promise<void> {
   await client.connect()
   /*
    * This wallet creation is for demonstration purposes.
-   * in practice, users generally will not have all keys in one spot,
+   * In practice, users generally will not have all keys in one spot,
    * hence, users need to implement a way to get signatures.
    */
   const { wallet: wallet1 } = await client.fundWallet()
--- a/packages/xrpl/snippets/src/multisigning.ts
+++ b/packages/xrpl/snippets/src/multisigning.ts
@@ -61,7 +61,8 @@ async function multisigning(): Promise<void> {
       )
     }
   } else {
-    console.log('There was an error multi-signing this transaction.')
+    console.log('The multisigned transaction was rejected by rippled. Here's the response from rippled:')
+    console.log(submitResponse)
   }
 }
 
--- a/packages/xrpl/snippets/src/multisigning.ts
+++ b/packages/xrpl/snippets/src/multisigning.ts
@@ -53,7 +53,7 @@ async function multisigning(): Promise<void> {
   const submitResponse = await client.submit(multisignedTx)
 
   if (submitResponse.result.engine_result === 'tesSUCCESS') {
-    console.log('Transaction multisigned successfully:')
+    console.log('The multisigned transaction was accepted by the ledger:')
     console.log(submitResponse)
     if (submitResponse.result.tx_json.Signers) {
       console.log(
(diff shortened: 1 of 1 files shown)


packages/xrpl/snippets/src/multisigning.ts
```diff
@@ -0,0 +1,68 @@
...
+  console.log('SignerListSet constructed successfully:')
+  console.log(signerListResponse)
+
+  const accountSet: AccountSet = {
+    TransactionType: 'AccountSet',
+    Account: walletMaster.classicAddress,
+    Domain: convertStringToHex('example.com'),
+  }
+  const accountSetTx = await client.autofill(accountSet, 2)
+  console.log('Transaction set successfully:')
```
JST5000, line 48: `Transaction set successfully:` is a bit confusing - I'm not sure what you mean by `set` - I think this rewording might work better
```suggestion
  console.log('AccountSet transaction is ready to be multisigned:')
```

packages/xrpl/snippets/src/multisigning.ts
```diff
@@ -0,0 +1,68 @@
...
+  console.log(accountSetTx)
+  const { tx_blob: tx_blob1 } = wallet1.sign(accountSetTx, true)
```
JST5000, line 50:
```suggestion
  # Individually signing the transaction with each signer. This would normally be handled by your application 
  # e.g. by sending push notifications to each signer to confirm if this transaction should be sent by signing it
  const { tx_blob: tx_blob1 } = wallet1.sign(accountSetTx, true)
```

packages/xrpl/snippets/src/multisigning.ts
```diff
@@ -0,0 +1,68 @@
...
+  const { tx_blob: tx_blob2 } = wallet2.sign(accountSetTx, true)
+  const multisignedTx = multisign([tx_blob1, tx_blob2])
```
JST5000, line 52:
```suggestion
  const multisignedTx = multisign([tx_blob1, tx_blob2])
  console.log('Successfully multisigned the transaction:')
  console.log(multisignedTx)
```

packages/xrpl/snippets/src/multisigning.ts
```diff
@@ -0,0 +1,68 @@
...
+  const submitResponse = await client.submit(multisignedTx)
+
+  if (submitResponse.result.engine_result === 'tesSUCCESS') {
+    console.log('Transaction multisigned successfully:')
```
JST5000, line 56:
```suggestion
    console.log('The multisigned transaction was accepted by the ledger:')
```

packages/xrpl/snippets/src/multisigning.ts
```diff
@@ -0,0 +1,68 @@
...
+    console.log(submitResponse)
+    if (submitResponse.result.tx_json.Signers) {
+      console.log(
+        `Signers contain ${submitResponse.result.tx_json.Signers.length} entries`,
```
JST5000, line 60:
```suggestion
        `The transaction had ${submitResponse.result.tx_json.Signers.length} signatures`,
```

packages/xrpl/snippets/src/multisigning.ts
```diff
@@ -0,0 +1,68 @@
...
+      )
+    }
+  } else {
+    console.log('There was an error multi-signing this transaction.')
```
JST5000, line 64:
```suggestion
    console.log('The multisigned transaction was rejected by rippled. Here's the response from rippled:')
    console.log(submitResponse)
```

packages/xrpl/snippets/src/multisigning.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
```
connorjchen, line 65: Make sure to add `await client.disconnect()` at the end of test alike to all other code snippets

packages/xrpl/snippets/src/multisigning.ts
pdp2121: Could you elaborate more on the real use case?

JST5000: When you're doing multisigning in practice, you generally don't have all the keys in one spot. 

If you're self-custodying the keys - you might have a hardware wallet or two, along with a key in a wallet like XUMM and maybe one you wrote down somewhere just in case. Getting those keys to sign the transaction will have some sort of delay and can require a decent amount of work to set up. 

If you are using multisigning to have multiple people control an account (ex. a company wanting to decentralize control over their treasury wallet) then you'd need to contact those people whenever you want to sign the transaction, which will have a human delay. (And you'd need some sort of notification system to let them know what transaction to sign)

So in this snippet we're just assuming we have control over the keys in the code, but in practice I think normally there's some in-between steps that you have to go through to get signatures on the multisigned transaction. (Does that add context?)

pdp2121: Yes, it makes sense to me. However, it is still unclear to me how we can replicate practical multisign in one snippet. Is it necessary in this case?

JST5000: Ah sorry - I just meant to have a comment explaining that if they wanted to turn this snippet into something more practical they'd have to go implement a way to get signatures instead of assuming they'd already have them.

For a snippet we definitely don't have to go implement that ourselves :)

pdp2121: Got it. I added a comment to notice users about this scenario.

JST5000 (approved):
LGTM :)

packages/xrpl/snippets/src/multisigning.ts
```diff
@@ -0,0 +1,78 @@
+import {
+  multisign,
+  Client,
+  AccountSet,
+  convertStringToHex,
+  SignerListSet,
+} from '../../dist/npm'
```
JST5000, line 7: Oh wait - should this import from `xrpl` instead of `dist/npm`?

JST5000: (Can check what the other snippets do)

pdp2121: Others also import from that same `dist/npm` repo so I dont think it's an issue :)

PR #2127, opened 2022-11-03
feat: modifiable `definitions.json` values at runtime via encode/decode parameter

JST5000:

Allows for encoding/decoding transactions which are using alternate values for the information that's normally stored in  `definitions.json`. 

### Context of Change

Supersedes #2101 

Changes the implementation from modifying a global variable to being an optional parameter to encode/decode since a single app may have to interface with both a sidechain and mainchain, so they need the ability to dictate which types to use. 

### Type of Change

- [X] Documentation Updates

## Before / After

Before: `definitions.json` was hardcoded into the library
After: You can now create your own version of `definitions.json` and use it throughout to encode/decode transactions.

## Test Plan

Added unit tests for encoding/decoding transactions using the new type definitions. (Showing that we can add new versions)

## Future Tasks
Still TODO:
- [x] Make xrpl.js's signing interface allow this as an optional parameter
- [X] Update the docs I added explaining how to set this up to reflect the new implementation method
- [ ] Add tests for using modified types with `encodeForSigning`, `encodeForSigningClaim`, and `encodeForMultisigning`


packages/ripple-binary-codec/src/binary.ts
```diff
@@ -16,26 +16,44 @@ import * as bigInt from 'big-integer'
...
  *
  * @param bytes hex-string to construct BinaryParser from
+ * @param customDefinitions Rippled definitions used to parse the values of transaction types and such.
+ *                          Can be customized for sidechains and amendments.
  * @returns A BinaryParser
  */
-const makeParser = (bytes: string): BinaryParser => new BinaryParser(bytes)
+const makeParser = (
+  bytes: string,
+  customDefinitions: DefinitionContents = DEFINITIONS,
```
mvadari, line 25: nit (to apply elsewhere as well):
```suggestion
  definitions: DefinitionContents = DEFINITIONS,
```
The definitions aren't always custom.

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -16,26 +16,44 @@ import * as bigInt from 'big-integer'
...
+ * @param customDefinitions Rippled definitions used to parse the values of transaction types and such.
+ *                          Can be customized for sidechains and amendments.
  * @returns JSON
  */
-const binaryToJSON = (bytes: string): JsonObject => readJSON(makeParser(bytes))
+const binaryToJSON = (
+  bytes: string,
+  customDefinitions: DefinitionContents = DEFINITIONS,
+): JsonObject =>
+  readJSON(makeParser(bytes, customDefinitions), customDefinitions)
```
mvadari, line 56: ditto

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -1,22 +1,24 @@
 # Definitions
 
-## Types
+## Overview of Terms
+
+### Types
```
mvadari, line 5: Why is this change made? It doesn't look like there are any other subheadings under `Overview of Terms`

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -70,22 +71,34 @@ export class Bytes {
...
     Object.entries(types).forEach(([k, v]) => {
-      this[k] = new Bytes(k, v, ordinalWidth)
-      this[v.toString()] = this[k]
+      this.add(k, v)
     })
   }
 
+  /**
+   * Add a new name value pair to the BytesLookup. Does not work properly if either name or value already
+   * exist within the BytesLookup.
```
mvadari, line 80:
```suggestion
   * Add a new name value pair to the BytesLookup. Does not work properly if either name or value already
   * exist within the BytesLookup.
```
If so, there should be additional checks for if `name` or `value` already exist, to throw a nicer error.

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -119,46 +134,141 @@ function buildField([name, info]: [string, FieldInfo]): FieldInstance {
...
+
   fromString(value: string): FieldInstance {
     return this[value] as FieldInstance
   }
 }
+/**
+ * Stores the various types and fields for rippled to be used to encode/decode information later on.
+ *
+ */
+class DefinitionContents {
```
mvadari, line 167: IMO this class should go in a separate file

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -119,46 +134,141 @@ function buildField([name, info]: [string, FieldInfo]): FieldInstance {
...
+   * For an example of the input format see `definitions.json`
+   * To generate a new definitions file from rippled source code, use this tool: https://github.com/RichardAH/xrpl-codec-gen
+   *
+   * See the definitions.test.js file for examples of how to create your own updated definitions.json.
+   *
+   * @param enums - A json encoding of the core types, transaction types, transaction results, transaction names, and fields.
+   * @param types - A list of type objects with the same name as the fields defined.
+   *              You can use the coreTypes object if you are not adding new types.
+   */
+  constructor(
```
mvadari, line 187: This class should have an alternate constructor that takes in a file location and converts it into a `DefinitionsContents`

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -119,46 +134,141 @@ function buildField([name, info]: [string, FieldInfo]): FieldInstance {
...
+    enums: DefinitionsData,
+    types: Record<string, typeof SerializedType>,
+  ) {
+    // Helps catch errors early in JavaScript code.
+    if (types == undefined) {
```
mvadari, line 192: IMO `types` should be allowed to be undefined. If it's undefined, that means that you should use `coreTypes`. Most folks won't have any new types.

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -119,46 +134,141 @@ function buildField([name, info]: [string, FieldInfo]): FieldInstance {
...
+    )
+    this.field = new FieldLookup(
+      enums.FIELDS as Array<[string, FieldInfo]>,
+      enums.TYPES,
+    )
+    this.transactionNames = Object.entries(enums.TRANSACTION_TYPES)
+      .filter(([_key, value]) => value >= 0)
+      .map(([key, _value]) => key)
+
+    this.addedDataTypes = {} // Filled in via associateTypes
```
mvadari, line 220: nit:
```suggestion
    this.dataTypes = {} // Filled in via associateTypes
```

packages/xrpl/src/Wallet/index.ts
```diff
@@ -318,6 +321,7 @@ class Wallet {
     this: Wallet,
     transaction: Transaction,
     multisign?: boolean | string,
+    customDefinitions: DefinitionContents = DEFINITIONS,
```
mvadari, line 324:
```suggestion
    customDefinitions: DefinitionContents = undefined,
```

packages/ripple-binary-codec/src/enums/README.md
JST5000: I forget :P - Doesn't seem necessary.

packages/ripple-binary-codec/src/enums/index.ts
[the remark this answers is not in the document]
JST5000: I think `RippledDefinitions` makes sense

packages/ripple-binary-codec/src/enums/index.ts
JST5000: I agree that it'd be better if `types` defaulted to `coreTypes` - I'm struggling to do that without introducing a circular dependency between `coreTypes` -> `BinaryParser` (in order to initialize some of the types in coreTypes) -> `RippledTypes` (which wants `coreTypes` as a default parameter, and is used in BinaryParser to decode field values) 

I'll look at it again tomorrow and see if I can crack how to do it

packages/xrpl/src/Wallet/index.ts
JST5000: Kept the default values at the low level (`binary.ts` and `binary-parser.ts` primarily) and then removed the need for it in the higher level functions (like `encode` and `decode`).

packages/ripple-binary-codec/src/enums/index.ts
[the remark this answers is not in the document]
mvadari: Actually, perhaps a better (more general) name would be `XrplDefinitions` - future-proofing for if there's another XRPL node

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -81,13 +106,19 @@ function serializeObject(object: JsonObject, opts: OptionObject = {}): Buffer {
  *
  * @param transaction Transaction to serialize
  * @param prefix Prefix bytes to put before the serialized object
+ * @param definitions Custom rippled types to use instead of the default. Used for sidechains and amendments.
  * @returns A Buffer with the serialized object
  */
 function signingData(
   transaction: JsonObject,
   prefix: Buffer = HashPrefix.transactionSig,
+  definitions: RippledDefinitions = DEFAULT_DEFINITIONS,
```
mvadari, line 115:
```suggestion
  definitions?: RippledDefinitions,
```

packages/ripple-binary-codec/src/enums/rippled-definitions.ts
```diff
@@ -0,0 +1,267 @@
...
+  LEDGER_ENTRY_TYPES: Record<string, number>
+  FIELDS: (string | FieldInfo)[][]
+  TRANSACTION_RESULTS: Record<string, number>
+  TRANSACTION_TYPES: Record<string, number>
+}
+
+/*
+ * @brief: Serialize a field based on type_code and Field.nth
+ */
+function fieldHeader(type: number, nth: number): Buffer {
```
mvadari, line 21: All these other methods/classes that aren't `RippledDefinitions` should be in a separate file

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -1,164 +1,30 @@
...
-/*
- * @brief: All valid transaction types
+import { RippledDefinitions, FieldInstance, Bytes } from './rippled-definitions'
+/**
+ * By default, coreTypes from the `types` folder is where known type definitions are initialized to avoid import cycles.
  */
-export const TRANSACTION_TYPES = Object.entries(enums.TRANSACTION_TYPES)
-  .filter(([_key, value]) => value >= 0)
-  .map(([key, _value]) => key)
+const DEFAULT_DEFINITIONS = new RippledDefinitions(enums, {})
```
mvadari, line 6: Why doesn't `DEFAULT_DEFINITIONS` have types?

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -4,19 +4,19 @@
 
 These are the [types](https://xrpl.org/serialization.html#type-list) associated with a given Serialization Field. Each type has an arbitrary [type_code](https://xrpl.org/serialization.html#type-codes), with lower codes sorting first.
 
-## Ledger Entry Types
+### Ledger Entry Types
```
mvadari, line 7: These extra `#`s aren't needed

packages/xrpl/src/models/transactions/transaction.ts
```diff
@@ -1,8 +1,10 @@
 /* eslint-disable complexity -- verifies 19 tx types hence a lot of checks needed */
 /* eslint-disable max-lines-per-function -- need to work with a lot of Tx verifications */
+/* eslint-disable import/no-cycle -- Flags does a type import of Transactions which which is not a real cycle */
```
mvadari, line 3: It shouldn't be a cycle if it's a type import

packages/xrpl/src/models/transactions/transaction.ts
```diff
@@ -1,8 +1,10 @@
...
 
 import _ from 'lodash'
 import { encode, decode } from 'ripple-binary-codec'
+import { type XrplDefinitions } from 'ripple-binary-codec/dist/enums'
```
mvadari, line 7:
```suggestion
import { type XrplDefinitions } from 'ripple-binary-codec'
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -20,7 +21,7 @@ import {
   PaymentChannelClaimFlagsInterface,
   PaymentChannelClaimFlags,
 } from '../transactions/paymentChannelClaim'
-import type { Transaction } from '../transactions/transaction'
+import { type Transaction } from '../transactions/transaction'
```
mvadari, line 24: Why was this changed? Shouldn't this get around the cycle issue?

packages/xrpl/src/sugar/submit.ts
```diff
@@ -70,18 +75,25 @@ async function submitAndWait(
...
     // A wallet to sign a transaction. It must be provided when submitting an unsigned transaction.
     wallet?: Wallet
+    // Custom rippled types to use instead of the default. Used for sidechains and amendments.
+    definitions?: XrplDefinitions
   },
 ): Promise<TxResponse> {
   const signedTx = await getSignedTx(this, transaction, opts)
 
-  const lastLedger = getLastLedgerSequence(signedTx)
+  const lastLedger = getLastLedgerSequence(signedTx, opts?.definitions)
```
mvadari, line 84: Why does this method need `definitions`?

packages/xrpl/src/sugar/submit.ts
```diff
@@ -95,23 +107,25 @@ async function submitAndWait(
...
   const signedTxEncoded =
     typeof signedTransaction === 'string'
       ? signedTransaction
-      : encode(signedTransaction)
+      : encode(signedTransaction, definitions)
   const request: SubmitRequest = {
     command: 'submit',
     tx_blob: signedTxEncoded,
-    fail_hard: isAccountDelete(signedTransaction) || failHard,
+    fail_hard: isAccountDelete(signedTransaction, definitions) || failHard,
```
mvadari, line 128: ditto

packages/xrpl/src/utils/hashes/hashLedger.ts
```diff
@@ -5,6 +5,7 @@
 
 import BigNumber from 'bignumber.js'
 import { decode, encode } from 'ripple-binary-codec'
+import { type XrplDefinitions } from 'ripple-binary-codec/dist/enums'
```
mvadari, line 8:
```suggestion
import { type XrplDefinitions } from 'ripple-binary-codec'
```

packages/xrpl/src/utils/index.ts
```diff
@@ -15,6 +15,11 @@ import {
   xAddressToClassicAddress,
 } from 'ripple-address-codec'
 import * as rbc from 'ripple-binary-codec'
+import {
+  XrplDefinitions,
+  DEFAULT_DEFINITIONS,
+} from 'ripple-binary-codec/dist/enums'
+import { coreTypes } from 'ripple-binary-codec/dist/types'
```
mvadari, line 22: These imports shouldn't be needed with the import on line 17

packages/ripple-binary-codec/src/enums/index.ts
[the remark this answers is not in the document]
JST5000: Updated

packages/ripple-binary-codec/src/enums/index.ts
JST5000: Chatted with @ckniffen about this and decided that having us import the json directly seems to add a lot of complexity without much value add. Importing a json file seems like a fine thing to ask the user to do as an input.

JST5000:
Noting here: @ckniffen noticed that the typescript definitions in xrpl.js need a way to be updated. Looking at using this method as a way to extend the type definition of `Transaction` to include any new TransactionTypes people define for themselves: https://stackoverflow.com/questions/43952198/how-to-override-or-extend-a-libary-type-definition-in-typescript

packages/xrpl/src/sugar/submit.ts
JST5000: In order to decode the transaction - if it happens to be a new type, then `decode` would fail otherwise.

packages/xrpl/src/sugar/submit.ts
JST5000: Same issue - without the optional definitions it's possible to not have enough information to decode a new transaction type.

packages/xrpl/src/models/utils/flags.ts
JST5000: This fixed a different cycle that was happening between Flags and Transaction (there were multiple :P)

JST5000: Nevermind - reverting this fixed the circular import - nice catch!

packages/ripple-binary-codec/src/binary.ts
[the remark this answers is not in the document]
JST5000: Even for just a single variable?

packages/ripple-binary-codec/src/enums/index.ts
JST5000: Circular dependency problem again - can't import `../types` in the same file that `DEFAULT_DEFINITIONS` is initialized since `DEFAULT_DEFINITIONS` is needed to form some of the types.

mvadari: Do we _need_ the types in XrplDefinitions? IMO it'll be pretty rare that someone needs to create a new type.
Perhaps instead of needing to add all the types, the `types` field accepts only extra types.

packages/xrpl/src/Wallet/index.ts
```diff
@@ -337,11 +345,17 @@ class Wallet {
...
     }
 
-    removeTrailingZeros(tx)
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Only modifies Payments
+    removeTrailingZeros(tx as Transaction)
 
     const txToSignAndEncode = { ...tx }
 
     txToSignAndEncode.SigningPubKey = multisignAddress ? '' : this.publicKey
+    if (tx.SigningPubKey && txToSignAndEncode.SigningPubKey === '') {
```
ckniffen, line 354: Was this validation needed to added custom defintions?

ckniffen (approved):

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -53,8 +57,92 @@ See:
...
+# Defining Your Own Definitions
+
+If you're building your own sidechain or writing an amendment for the XRPL, you may need to create new XRPL definitions.
+
+To do that there are a couple things you need to do:
+
+1. Generate your own `definitions.json` file from rippled source code using [this tool](https://github.com/RichardAH/xrpl-codec-gen) (The default `definitions.json` for mainnet can be found [here](https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/enums/definitions.json))
+2. Create new SerializedType classes for any new Types (So that encode/decode behavior is defined)
+
+- For examples of how to implement that you can look at objects in the `types` folder, such as `Amount`, `UInt8`, or `STArray`.
```
jonathanlei, line 75: would it be helpful to link to the folder?

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -81,13 +106,19 @@ function serializeObject(object: JsonObject, opts: OptionObject = {}): Buffer {
  *
  * @param transaction Transaction to serialize
  * @param prefix Prefix bytes to put before the serialized object
+ * @param definitions Custom rippled types to use instead of the default. Used for sidechains and amendments.
  * @returns A Buffer with the serialized object
  */
 function signingData(
   transaction: JsonObject,
   prefix: Buffer = HashPrefix.transactionSig,
+  definitions?: XrplDefinitions,
```
mvadari, line 115: This should go in an `opts` dict, for better future-proofing

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -102,13 +133,18 @@ interface ClaimObject extends JsonObject {
  * Serialize a signingClaim
  *
  * @param claim A claim object to serialize
+ * @param definitions Custom rippled types to use instead of the default. Used for sidechains and amendments.
  * @returns the serialized object with appropriate prefix
  */
-function signingClaimData(claim: ClaimObject): Buffer {
+function signingClaimData(
+  claim: ClaimObject,
+  definitions: XrplDefinitions = DEFAULT_DEFINITIONS,
```
mvadari, line 141: ditto

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -123,18 +159,24 @@ function signingClaimData(claim: ClaimObject): Buffer {
  *
  * @param transaction transaction to serialize
  * @param signingAccount Account to sign the transaction with
+ * @param definitions Custom rippled types to use instead of the default. Used for sidechains and amendments.
  * @returns serialized transaction with appropriate prefix and suffix
  */
 function multiSigningData(
   transaction: JsonObject,
   signingAccount: string | AccountID,
+  definitions: XrplDefinitions = DEFAULT_DEFINITIONS,
```
mvadari, line 168: ditto

packages/xrpl/src/Wallet/index.ts
JST5000: No, I'm honestly not sure why I added this. Maybe it was meant to be part of another PR but I was on the wrong branch? I'll remove it for now.

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -81,13 +106,19 @@ function serializeObject(object: JsonObject, opts: OptionObject = {}): Buffer {
  *
  * @param transaction Transaction to serialize
  * @param prefix Prefix bytes to put before the serialized object
+ * @param opts.definitions Custom rippled types to use instead of the default. Used for sidechains and amendments.
  * @returns A Buffer with the serialized object
  */
 function signingData(
   transaction: JsonObject,
   prefix: Buffer = HashPrefix.transactionSig,
+  opts: { definitions?: XrplDefinitions } = {},
```
mvadari, line 115: A note about doing this (I realize it's done in other parts of the code too): in Python, if you make a param equal to a mutable object, like a dict, then mutate the param, it uses that same (already-mutated) object for every other call to that function. That's why you often see paradigms like this:
```py
def f(param=None):
  if param is None:
    param = {}
  ...
```
Not sure if this is also true in JS, but if it is, it's probably be a good idea to avoid doing that to future-proof the code.

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -16,26 +20,43 @@ import * as bigInt from 'big-integer'
...
  * @param parser BinaryParser object
+ * @param definitions rippled definitions used to parse the values of transaction types and such.
+ *                          Can be customized for sidechains and amendments.
  * @returns JSON for the bytes in the BinaryParser
  */
-const readJSON = (parser: BinaryParser): JsonObject =>
-  (parser.readType(coreTypes.STObject) as STObject).toJSON()
+const readJSON = (
+  parser: BinaryParser,
+  definitions: XrplDefinitions = DEFAULT_DEFINITIONS,
```
mvadari, line 42: Why is there a default `definitions` provided in some methods but not all?

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -53,8 +57,92 @@ See:
...
+
+3. Import your `definitions.json` file and `coreTypes` from the `types` folder, then use them to construct your own `XrplDefinitions` object.
+4. Pass the `XrplDefinitions` object whenever you `encode` or `decode` a transaction.
+5. If you added any new transaction types, you should create an `interface` for the transaction that extends `BaseTransaction` from the `xrpl` repo to use it with the functions on `Client` (See the below example of adding a new transaction type)
+
+## Example of adding a new Transaction type
+
+```
+// newTypeDefs is where you can import your custom defined definitions.json file
+const newTypeDefs = require('./new-type.json')
```
mvadari, line 85: The name `newTypeDefs` is a little confusing - it makes me thing that it's new types. Maybe it should be named something like `newDefinitions` instead.

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -53,8 +57,92 @@ See:
...
+const { XrplDefinitions, coreTypes, Client } = require('xrpl')
+
+const newDefs = new XrplDefinitions(newTxDefinitions, coreTypes)
```
mvadari, line 88: Could `coreTypes` be provided by default? IMO 99.9% of people don't need to define new types, so this just adds additional friction.

packages/ripple-binary-codec/src/binary.ts
JST5000: That's a super unexpected behavior. Let me check if it is in JS as well.

packages/ripple-binary-codec/src/enums/README.md
JST5000: Renamed to `newDefinitionsJson` to avoid it conflicting too hard with `newDefs` which is the object we use throughout the code to replace the types (since `newDefinitions` and `newDefs` are basically the same meaning)

packages/ripple-binary-codec/src/binary.ts
JST5000: I removed explicit parameters from all but the bottom functions. So if they only use functions which have `DEFAULT_DEFINITIONS` defined, then I just had them use the default for that function. (This was in response to another comment earlier where it seemed better to avoid over-specifying which definitions to use at the top level)

I'm honestly not sure what's better. My thought process is that it probably makes sense to avoid explicit defaults across package boundaries (So having explicit default of `DEFAULT_DEFINITIONS` within the binary codec, but not having those within xrpl.js functions) in order to allow the lower-level package to update that default independently of updating xrpl.js. So maybe I should add the explicit default for all the functions within the binary codec which use this param?

packages/ripple-binary-codec/src/enums/README.md
JST5000: I'd love to, but I for the life of me can't figure out how to do that and resolve the circular dependency problems :/

Importing `coreTypes` to use it as a default param to `XrplDefinitions` creates 3 circular dependencies:
```
1) packages/ripple-binary-codec/dist/enums/index.js > packages/ripple-binary-codec/dist/enums/xrpl-definitions.js > packages/ripple-binary-codec/dist/types/index.js
2) packages/ripple-binary-codec/dist/enums/index.js > packages/ripple-binary-codec/dist/enums/xrpl-definitions.js > packages/ripple-binary-codec/dist/types/index.js > packages/ripple-binary-codec/dist/types/amount.js > packages/ripple-binary-codec/dist/serdes/binary-parser.js
3) packages/ripple-binary-codec/dist/enums/index.js > packages/ripple-binary-codec/dist/enums/xrpl-definitions.js > packages/ripple-binary-codec/dist/types/index.js > packages/ripple-binary-codec/dist/types/st-array.js > packages/ripple-binary-codec/dist/types/st-object.js
```

In order:
1. This one is caused by attempts to work around the other two circular dependencies. (We import `DEFAULT_DEFINITIONS` into the `coreTypes` file in order to ensure we assign it the `coreTypes` since the circular dependencies often result in `coreTypes` being `undefined` when `DEFAULT_DEFINITIONS` is defined.)
2. The `BinaryParser` relies on knowing how to decode all the types from `DEFAULT_DEFINITIONS, and received that as a global. That is then needed for `Amount`, which is a type in `coreTypes`, which then becomes a default parameter for `XrplDefinitions` and thus `DEFAULT_DEFINITIONS`.
3. Same flow as above, where `STObject` is the type in `coreTypes` this time that needs `DEFAULT_DEFINITIONS`. 

I tried for a couple days, and another couple hours today, but couldn't figure out how to untangle the mess :/ 
(There's this suggestion of making a central import / export file that everything uses so you can manually specify the module resolution order, but that also feels like a bad answer: https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de 

I can make the tests pass while _including_ the circular dependency, but it's incredibly hacky: https://github.com/XRPLF/xrpl.js/pull/2178

Basically with the circular dependency we can't tell what order the modules will resolve, and we're guaranteed to get an undefined somewhere since the definitions happens after the imports regardless of which path they follow. This means either 
1.`coreTypes` will be undefined on first load of `DEFAULT_DEFINITIONS`, in which case we need to add the coreTypes again with `associateTypes`.
2. `DEFAULT_DEFINITIONS` will be undefined on first load of `coreTypes`, so we can't `associateTypes`. But that means that when we do load `DEFAULT_DEFINITIONS` coreTypes will be there, so it will be defined in the constructor. 

While doing this passes tests, I can't forsee what other bugs/race conditions it will introduce, so I really would not prefer to do that.

JST5000: Update, I found a way to get the desired behavior. Basically by adding a subclass that imports both `coreTypes` and `XrplDefinitions` (renamed to `XrplDefinitionsBase`) that has a constructor with default value. For end users they can use the new `XrplDefinitions` object which subclasses `XrplDefinitionsBase`.

(Separating `coreTypes` from the `XrplDefinitions` object ended up not making sense because the core issue is that `BinarySerializer` needs to know how to go from random binary to objects, which it does through a combination of fields and types - so you need both.)

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -1,164 +1,34 @@
...
-)
-const TransactionType = new BytesLookup(
-  enums.TRANSACTION_TYPES,
-  TRANSACTION_TYPE_WIDTH,
-)
-const TransactionResult = new BytesLookup(
-  enums.TRANSACTION_RESULTS,
-  TRANSACTION_RESULT_WIDTH,
-)
-const Field = new FieldLookup(enums.FIELDS as Array<[string, FieldInfo]>)
```
JST5000, line 155: All the Field methods are copied unchanged into `field.ts` here: https://github.com/XRPLF/xrpl.js/blob/pass-through-encode/packages/ripple-binary-codec/src/enums/field.ts

(With the last line `const Field` being defined elsewhere in this file (`enums/index.ts`) using `DEFAULT_DEFINITIONS` which is designed to be equivalent to previous behavior.)

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -1,164 +1,34 @@
...
- * @brief: All valid transaction types
- */
-export const TRANSACTION_TYPES = Object.entries(enums.TRANSACTION_TYPES)
-  .filter(([_key, value]) => value >= 0)
-  .map(([key, _value]) => key)
-
-const TYPE_WIDTH = 2
-const LEDGER_ENTRY_WIDTH = 2
-const TRANSACTION_TYPE_WIDTH = 2
-const TRANSACTION_RESULT_WIDTH = 1
```
JST5000, line 16: These constants have been moved to `constants.ts` so they can be imported by the split out files that used to make up `enums/index.ts`. https://github.com/XRPLF/xrpl.js/blob/pass-through-encode/packages/ripple-binary-codec/src/enums/constants.ts

packages/ripple-binary-codec/src/enums/index.ts
```diff
@@ -1,164 +1,34 @@
...
+const DEFAULT_DEFINITIONS = new XrplDefinitionsBase(enums, {})
 
-  from(value: Bytes | string): Bytes {
-    return value instanceof Bytes ? value : (this[value] as Bytes)
-  }
-
-  fromParser(parser): Bytes {
-    return this.from(parser.readUIntN(this.ordinalWidth).toString())
-  }
-}
```
JST5000, line 85: `Bytes` and `ByteLookup` were added to `bytes.ts`: https://github.com/XRPLF/xrpl.js/blob/pass-through-encode/packages/ripple-binary-codec/src/enums/bytes.ts

* I added an error message when you try to add multiple bytes to the lookup with the same name because it results in invalid state. (It results in an ambiguous decoding since it could become either of the `named` values) - Tests failed when I did this by accident trying to figure out how it worked. - There's an error for both sides since the values are a 2-way lookup.

packages/ripple-binary-codec/src/enums/bytes.ts
```diff
@@ -0,0 +1,75 @@
+import { BytesList, BinaryParser } from '../binary'
```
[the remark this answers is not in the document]
JST5000: Unfortunately it doesn't seem possible (took a crack at it, but it was really hard to put things back so the diff was clean, and also seemed like it would cause the circular import issues again). 

I explicitly highlighted the areas that are mostly moving code along with a link to the changed code so you can more directly compare what happened by alt-tabbing between the highlight and changed code. (Along with explicitly calling out the little changes I made during the transition)

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -16,26 +20,43 @@ import * as bigInt from 'big-integer'
...
  *
  * @param bytes hex-string to construct BinaryParser from
+ * @param definitions rippled definitions used to parse the values of transaction types and such.
+ *                          Can be customized for sidechains and amendments.
  * @returns A BinaryParser
  */
-const makeParser = (bytes: string): BinaryParser => new BinaryParser(bytes)
+const makeParser = (
+  bytes: string,
+  definitions?: XrplDefinitionsBase,
```
mvadari, line 29: Does this not need `definitions?: XrplDefinitionsBase = undefined`?

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -53,8 +57,92 @@ See:
...
 See https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/TxFormats.h
+
+# Defining Your Own Definitions
+
+If you're building your own sidechain or writing an amendment for the XRPL, you may need to create new XRPL definitions.
+
+To do that there are a couple things you need to do:
+
+1. Generate your own `definitions.json` file from rippled source code using [this tool](https://github.com/RichardAH/xrpl-codec-gen) (The default `definitions.json` for mainnet can be found [here](https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/enums/definitions.json))
+2. Create new SerializedType classes for any new Types (So that encode/decode behavior is defined)
```
mvadari, line 73: Add a note that this is likely not necessary, and that they map to the `ST...` classes in rippled

packages/ripple-binary-codec/src/enums/README.md
mvadari: Was this change pushed?

packages/ripple-binary-codec/src/binary.ts
mvadari: ditto for all the similar methods

JST5000: Isn't that implied by the optional parameter?

packages/ripple-binary-codec/src/enums/README.md
JST5000: Apparently not - updated :)

JST5000: Oh I see - I updated one snippet, but there were too - oops!

packages/ripple-binary-codec/src/binary.ts
mvadari: Why is it undefined in some places and equal to DEFAULT_DEFINITIONS in others? It doesn't seem super uniform where it's included and where it's not.

RichardAH:
Hi xrpljs devs

I've written an RPC method for rippled that generates xrpl.js compatible definitions based on the internals of that specific rippled instance.
PR is here: https://github.com/XRPLF/rippled/pull/4402

When xrpl.js connects to a server it should first attempt to query:
```json
{"command":"server_definitions"}
```
If the node doesn't support the API, then it will return an error and xrpl.js should continue using its default definitions.
If the node *does* support the API then it will return the following:
```json
{
    "result": {
        "FIELDS": [
            [
                "Generic",
                {
                    "isSerialized": false,
                    "isSigningField": false,
                    "isVLEncoded": false,
                    "nth": 0,
<...>
            "UInt96": 20,
            "Unknown": -2,
            "Validation": 10003,
            "Vector256": 19
        },
        "hash": "D4A1424FAFB7E36115FCD3FF1EE1E7A72EBFACAB6659AFC018D6FAB5FAEE74DE"
    },
    "status": "success",
    "type": "response"
}
```

Next time the client connects it should remember the hash returned last time and provide that with the API call.
```json
{"command":"server_definitions", "hash":"D4A1424FAFB7E36115FCD3FF1EE1E7A72EBFACAB6659AFC018D6FAB5FAEE74DE"}'
```

If the hash does not match then it will respond as though the hash had not been provided (with the full enumerated definitions.)
If the hash does match then it will send the hash back as a confirmation response:
```json
{
    "result": {
        "hash": "D4A1424FAFB7E36115FCD3FF1EE1E7A72EBFACAB6659AFC018D6FAB5FAEE74DE"
    },
    "status": "success",
    "type": "response"
}
```

If implemented in sidechains and in xrpl.js this can solve all definitions problems. Essentially it allows the client and server to handshake how they map serialized fields to json fields.

RichardAH:
This is a node you can test server_definitions on wss://hooks-testnet-v3.xrpl-labs.com

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -53,8 +57,94 @@ See:
...
+# Defining Your Own Definitions
+
+If you're building your own sidechain or writing an amendment for the XRPL, you may need to create new XRPL definitions.
+
+To do that there are a couple things you need to do:
+
+1. Generate your own `definitions.json` file from rippled source code using [this tool](https://github.com/RichardAH/xrpl-codec-gen) (The default `definitions.json` for mainnet can be found [here](https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/enums/definitions.json))
+2. Create new SerializedType classes for any new Types (So that encode/decode behavior is defined).
+
+- Note: This is very rarely required. The SerializedType classes correspond to "ST..." classes in Rippled.
```
mvadari, line 75:
```suggestion
2. Create new SerializedType classes for any new Types (So that encode/decode behavior is defined). The SerializedType classes correspond to "ST..." classes in Rippled. Note: This is very rarely required. 
```

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -53,8 +57,94 @@ See:
...
+
+3. Import your `definitions.json` file and `coreTypes` from the `types` folder, then use them to construct your own `XrplDefinitions` object.
+4. Pass the `XrplDefinitions` object whenever you `encode` or `decode` a transaction.
+5. If you added any new transaction types, you should create an `interface` for the transaction that extends `BaseTransaction` from the `xrpl` repo to use it with the functions on `Client` (See the below example of adding a new transaction type)
+
+## Example of adding a new Transaction type
+
+```
+// newDefinitionsJson is where you can import your custom defined definitions.json file
+const newDefinitionsJson = require('./new-type.json')
```
mvadari, line 87:
```suggestion
const newDefinitionsJson = require('./new-definitions.json')
```

packages/ripple-binary-codec/src/enums/README.md
```diff
@@ -53,8 +57,94 @@ See:
...
+## Example of adding a new serializable Type
+
+```
+const { XrplDefinitions } = require('../dist/coretypes')
+
+// coreTypes is the default list of serialized Types that are defined in xrpl.js
+const { coreTypes } = require('../dist/types')
+
+// newDefinitionsJson is where you can import your custom defined definitions.json file
+const newDefinitionsJson = require('./fixtures/new-type.json')
```
mvadari, line 126:
```suggestion
const newDefinitionsJson = require('./fixtures/new-definitions.json')
```

packages/ripple-binary-codec/src/enums/xrpl-definitions-base.ts
```diff
@@ -0,0 +1,112 @@
...
+   */
+  constructor(
+    enums: DefinitionsData,
+    types: Record<string, typeof SerializedType>,
+  ) {
+    // Helps catch errors early in JavaScript code.
+    if (types == undefined) {
+      throw new TypeError(
+        'You passed in an undefined `types` parameter, but `types` must be defined since it contains logic for encoding/decoding transaction data.' +
+          ' If you have NOT added/modified any data types, you can import and use `coreTypes` from the types folder.',
```
mvadari, line 51: super nit: the space should be at the end of the first line instead of at the beginning of the second line

packages/ripple-binary-codec/src/enums/xrpl-definitions-base.ts
```diff
@@ -0,0 +1,112 @@
...
+      enums.TRANSACTION_RESULTS,
+      TRANSACTION_RESULT_WIDTH,
+    )
+    this.field = new FieldLookup(
+      enums.FIELDS as Array<[string, FieldInfo]>,
+      enums.TYPES,
+    )
+    this.transactionNames = Object.entries(enums.TRANSACTION_TYPES)
+      .filter(([_key, value]) => value >= 0)
+      .map(([key, _value]) => key)
```
mvadari, line 74: Why are you trying to exclude `Invalid` here? Seems like it just adds complexity.

packages/ripple-binary-codec/src/enums/xrpl-definitions.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  type DefinitionsData,
+  XrplDefinitionsBase,
+} from './xrpl-definitions-base'
+import { coreTypes } from '../types'
+import { SerializedType } from '../types/serialized-type'
+
+export class XrplDefinitions extends XrplDefinitionsBase {
```
mvadari, line 8: Shouldn't the class itself also have a docstring?

packages/ripple-binary-codec/src/index.ts
```diff
@@ -56,11 +71,16 @@ function encodeForSigning(json: object): string {
...
  */
-function encodeForSigningClaim(json: object): string {
+function encodeForSigningClaim(
+  json: object,
+  definitions?: XrplDefinitionsBase,
+): string {
   assert.ok(typeof json === 'object')
-  return signingClaimData(json as ClaimObject)
+  const definitionsOpt = definitions ? { definitions } : undefined
+  return signingClaimData(json as ClaimObject, definitionsOpt)
```
mvadari, line 83: Does it not work like this?
```suggestion
  return signingClaimData(json as ClaimObject, { definitions })
```

packages/ripple-binary-codec/src/serdes/binary-parser.ts
```diff
@@ -13,8 +17,13 @@ class BinaryParser {
    * Initialize bytes to a hex string
    *
    * @param hexBytes a hex string
+   * @param definitions Rippled definitions used to parse the values of transaction types and such.
+   *                          Can be customized for sidechains and amendments.
    */
-  constructor(hexBytes: string) {
+  constructor(
+    hexBytes: string,
+    readonly definitions: XrplDefinitionsBase = DEFAULT_DEFINITIONS,
```
mvadari, line 25: Why is this readonly?

packages/xrpl/src/Wallet/signer.ts
```diff
@@ -6,12 +6,17 @@ import {
   encode,
   encodeForSigning,
   encodeForSigningClaim,
+  XrplDefinitionsBase,
```
mvadari, line 9: Shouldn't this be `XrplDefinitions`?

packages/ripple-binary-codec/src/index.ts
ckniffen: No because then it will always be an object just sometimes with no properties.

mvadari: Yeah, that should also be acceptable by the code.

packages/ripple-binary-codec/src/enums/xrpl-definitions-base.ts
JST5000: That's how it was before: https://github.com/XRPLF/xrpl.js/blob/8fe87c0e867752a7eb8663d18defc33329784511/packages/ripple-binary-codec/src/enums/index.ts#L9-L11

My guess is that it's because Invalid doesn't have a TypeScript class defined for it, while all the other types do.

packages/ripple-binary-codec/src/index.ts
JST5000: As written defining Definitions is non-optional, but defining your own _is_ optional. In order to allow `{ undefined }` I'd have to set the default value inside the function instead of as part of the parameter, which I think is worse for readability since it requires adding an extra check and setting the default of `DEFAULT_DEFINITIONS` twice. 

Current signingClaimData implementation for reference: 
```
function signingClaimData(
  claim: ClaimObject,
  opts: { definitions: XrplDefinitionsBase } = {
    definitions: DEFAULT_DEFINITIONS,
  },
): Buffer {
  const num = bigInt(String(claim.amount))
  const prefix = HashPrefix.paymentChannelClaim
  const types = opts.definitions.getAssociatedTypes()
  const channel = types.Hash256.from(claim.channel).toBytes()
  const amount = types.UInt64.from(num).toBytes()

  const bytesList = new BytesList()

  bytesList.put(prefix)
  bytesList.put(channel)
  bytesList.put(amount)
  return bytesList.toBytes()
}
```

Additional logic to allow `{ undefined }`:
```
function signingClaimData(
  claim: ClaimObject,
  opts: { definitions?: XrplDefinitionsBase } = {
    definitions: DEFAULT_DEFINITIONS,
  },
): Buffer {
let types = undefined
if(opts.definitions === undefined) {
  types = DEFAULT_DEFINITIONS.getAssociatedTypes()
} else {
  types = opts.definitions.getAssociatedTypes()
}
...
}
```

packages/ripple-binary-codec/src/serdes/binary-parser.ts
JST5000: Huh, I guess I thought it had to be since I saw so many readonly's in the other parts of this change - oops! (I don't think it actually needs to be readonly)

packages/ripple-binary-codec/src/binary.ts
JST5000: I only included DEFAULT_DEFINITIONS in places where the underlying functions didn't know how to interpret undefined. In cases where the sub-functions had their own default, I just passed in undefined to allow the lower levels to specify details rather than enforcing the default of a higher function. 

This specifically matters so that an update to the binary-codec can improve upon the default without requiring an update to xrpl.js (Although I suppose you may still need an updated version in order to have the binary-codec dependency be updated)

Ex. `encode` (undefined) -> `serializeObject` (undefined) -> `.from` (lowest level, so defaults to `DEFAULT_DEFINITIONS`)

WietseWind:
> I'm neither a typescript expert nor an expert in / familiar with this repo, I don't think this should be assigned to me. Happy to answer general questions about the `server_definitions` API etc.

@intelliot I reviewed in his place :) Working really well, I compiled and embedded in some of our own libs & ran test cases on custom definitions: ✅. Tested with regular & multi signed transactions. 

Thanks @JST5000 et al.

sublimator:
Alright, alright, alright

WietseWind (approved):
👍

JST5000:
> (Note: as this is a monorepo, I assume this will be added to the parent lib. over time, post `ripple-binary-codec` release?)

@WietseWind Thanks for reviewing! The next step after this is to update the xrpl.js Client to keep track of the `definitions` for whichever network it is attached to :) (With the `server_definitions` command @RichardAH created, that can hopefully be fully automatic eventually!)

For context for anyone curious, the reason to set definitions on a per-client basis is that if you're using a sidechain, a single bit of code will likely need to talk to two networks at once, the side-chain, and then mainnet. So by attaching the `definitions` to the client you can handle that seamlessly - whenever you submit to one network, under the hood it can encode/decode using the proper implementations.

packages/xrpl/test/wallet/signer.test.ts
```diff
@@ -139,7 +139,9 @@ describe('Signer', function () {
   it('multisign runs successfully with tx_blobs', function () {
     const transactions = [multisignTxToCombine1, multisignTxToCombine2]
 
-    const encodedTransactions: string[] = transactions.map(encode)
+    const encodedTransactions: string[] = transactions.map((transaction) =>
```
mvadari, line 142: Why was this change made?

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -66,8 +86,9 @@ function serializeObject(object: JsonObject, opts: OptionObject = {}): Buffer {
   const filter = signingFieldsOnly
     ? (f: FieldInstance): boolean => f.isSigningField
     : undefined
-
-  coreTypes.STObject.from(object, filter).toBytesSink(bytesList)
+  ;(coreTypes.STObject as typeof STObject)
```
mvadari, line 89: Can this cast be returned to how it was previously?

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -123,18 +151,26 @@ function signingClaimData(claim: ClaimObject): Buffer {
...
 function multiSigningData(
   transaction: JsonObject,
   signingAccount: string | AccountID,
+  opts: { definitions: XrplDefinitionsBase } = {
+    definitions: DEFAULT_DEFINITIONS,
+  },
 ): Buffer {
   const prefix = HashPrefix.transactionMultiSig
-  const suffix = coreTypes.AccountID.from(signingAccount).toBytes()
+  const suffix = opts.definitions
```
mvadari, line 165: ditto

packages/ripple-binary-codec/src/enums/xrpl-definitions-base.ts
```diff
@@ -0,0 +1,119 @@
...
+   * @param enums - A json encoding of the core types, transaction types, transaction results, transaction names, and fields.
+   * @param types - A list of type objects with the same name as the fields defined.
+   *              You can use the coreTypes object if you are not adding new types.
+   */
+  constructor(
+    enums: DefinitionsData,
+    types: Record<string, typeof SerializedType>,
+  ) {
+    // Helps catch errors early in JavaScript code.
+    if (types == undefined) {
```
mvadari, line 55: This if statement should be taken out now, right?

packages/ripple-binary-codec/src/index.ts
```diff
@@ -112,4 +134,8 @@ export {
   decodeQuality,
   decodeLedgerData,
   TRANSACTION_TYPES,
+  XrplDefinitions,
+  XrplDefinitionsBase,
```
mvadari, line 138: Shouldn't this class not be exported, since we just want people to use `XrplDefinitions`?

packages/ripple-binary-codec/test/fixtures/definitions-with-diff-payment.json
```diff
@@ -0,0 +1,2337 @@
+{
```
mvadari, line 1: A thought - it's really hard to parse these definitions.json files and figure out what the diff is with the existing definitions.json. Could we instead just use the existing definitions.json, and modify it in the JS code, instead of having these massive fixtures that are hard to understand?

packages/ripple-binary-codec/test/definitions.test.js
```diff
@@ -0,0 +1,71 @@
+const { encode, decode } = require('../dist')
```
mvadari, line 1: This file should be in TypeScript - e.g. `definitions.test.ts`

packages/ripple-binary-codec/test/definitions.test.js
```diff
@@ -0,0 +1,71 @@
...
+const { XrplDefinitions } = require('../dist/enums/xrpl-definitions')
```
mvadari, line 2: This should be importable directly from `../src/enums/xrpl-definitions` if you use TS

packages/xrpl/test/wallet/signer.test.ts
JST5000: This change was made because the function signature changed, so the implicit syntax doesn't work anymore. (Before it was a single parameter, single output, but since there's now an additional optional parameter it needs an explicit call)

packages/ripple-binary-codec/src/binary.ts
JST5000: Unfortunately not - the `STObject` `.from` method is unique because it has dynamic parsing, whereas the other entries in `coreTypes` are concrete implementations so don't need to know about `definitions.json` when decoding.

That means that in order to access the special `.from` we have to cast down to `typeof STObject` instead of `typeof SerializedType`

packages/ripple-binary-codec/src/binary.ts
JST5000: This one is revertable :)

packages/ripple-binary-codec/src/enums/xrpl-definitions-base.ts
JST5000: Yep, good catch :)

packages/ripple-binary-codec/test/definitions.test.js
JST5000: Why should this file be in TypeScript? (The rest of the test files are js for the binary codec)

packages/ripple-binary-codec/test/fixtures/definitions-with-diff-payment.json
JST5000: That seems like a good suggestion :) - Done.

packages/ripple-binary-codec/test/definitions.test.js
justinr1234: I’m not sure in JS how, but yes we should try to avoid using dist

mvadari: It looks like the existing tests don't use dist anymore even in JS: https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/test/binary-json.test.js#L2
I assume these changes were made in your Jest PR and thus weren't there when Jackson originally started this PR.

packages/ripple-binary-codec/test/definitions.test.js
JST5000: There's a decent amount of config changes required to make ts files work with our setup (We have to update our eslint settings, change the tsconfig, update our run script in npm)

After spending about 30 minutes trying to get it working, I don't think this is worth doing right now. I can make a follow up ticket for it, but for this test in particular it doesn't add much error checking because we're violating the type checks anyway in order to update the `definitions` mid-test.

packages/ripple-binary-codec/test/definitions.test.js
```diff
@@ -1,9 +1,6 @@
-const { encode, decode } = require('../dist')
-const { XrplDefinitions } = require('../dist/enums/xrpl-definitions')
-const newDefinitionsJson = require('./fixtures/new-definitions.json')
-const newFieldDefs = require('./fixtures/new-field.json')
+const { encode, decode, XrplDefinitions } = require('../src')
+const normalDefinitionsJson = require('./fixtures/normal-definitions.json')
```
mvadari, line 2: Why is this a separate fixture? Why not just pull from the one in `src`?

packages/ripple-binary-codec/test/signing-data-encoding.test.js
```diff
@@ -4,6 +4,10 @@ const {
   encodeForSigningClaim,
   encodeForMultisigning,
 } = require('../src')
+const { XrplDefinitions } = require('../src/enums/xrpl-definitions')
+
+// This changes the Payment TransactionType from being 0 to being 30 (aka 0x001F when encoded)
+const definitions = require('./fixtures/definitions-with-diff-payment.json')
```
mvadari, line 10: ditto here with not using the hard-to-parse fixtures - it'll be easier to understand if you just modify the normal definitions.json

packages/ripple-binary-codec/test/definitions.test.js
```diff
@@ -0,0 +1,103 @@
...
+  Flags: 0,
+  Sequence: 1,
+  TransactionType: 'Payment',
+}
+
+describe('encode and decode using new types as a parameter', function () {
+  test('can encode and decode a new TransactionType', function () {
+    const tx = Object.assign({}, txJson, {
+      TransactionType: 'NewTestTransaction',
+    })
```
mvadari, line 19: nit: this would be a little cleaner like this:
```suggestion
    const tx = {...txJson, TransactionType: 'NewTestTransaction'}
```

mvadari (approved):

PR #2129, opened 2022-11-08
add amount field to faucet

jonathanlei:

related to DGE [649](https://ripplelabs.atlassian.net/jira/software/projects/DGE/boards/524?selectedIssue=DGE-649)

### Context of Change
Add amount field to allow custom amount to be passed in for faucet transactions. 

### Type of Change
- [X ] New feature (non-breaking change which adds functionality)

## Test Plan
changed integration test, unit tests and integration tests passed.

### DIFF
--- a/packages/xrpl/src/Wallet/fundWallet.ts
+++ b/packages/xrpl/src/Wallet/fundWallet.ts
@@ -52,6 +52,7 @@ const MAX_ATTEMPTS = 20
  * and that wallet's balance in XRP.
  * @throws When either Client isn't connected or unable to fund wallet address.
  */
+// eslint-disable-next-line max-params, max-lines-per-function -- this function needs to display and do with more information.
 async function fundWallet(
   this: Client,
   amount?: string | '1000',
--- a/packages/xrpl/src/Wallet/fundWallet.ts
+++ b/packages/xrpl/src/Wallet/fundWallet.ts
@@ -40,6 +40,7 @@ const MAX_ATTEMPTS = 20
  * ```
  *
  * @param this - Client.
+ * @param amount - A custom amount to fund, if undefined or null, the default amount will be 1000.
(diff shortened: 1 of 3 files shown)


packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -40,6 +40,7 @@ const MAX_ATTEMPTS = 20
  * ```
  *
  * @param this - Client.
+ * @param amount - A custom amount to fund, if undefined or null, the default amount will be 1000.
```
JST5000, line 43: I think amount should be added to the `options` object as an optional parameter. (Currently this would actually be a breaking change since it changes the order of parameters, so someone who had code like
`client.fundWallet(wallet)` would not longer match the code signature)

jonathanlei: changed

ckniffen (approved):

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -76,6 +79,7 @@ async function fundWallet(
     new TextEncoder().encode(
       JSON.stringify({
         destination: walletToFund.classicAddress,
+        xrpAmount: options?.amount,
```
JST5000, line 82: What happens if the options pass in an invalid amount? (One that goes beyond how much the Testnet should be able to fund?) - Probably don't need to make it a test, but curious about the edge case inputs.

jonathanlei: faucet repo has checks on the amounts - if input is invalid it will return an error
https://github.com/xpring-eng/testnet-faucet/blob/336ad485a51fbf63a9d5c335979c68a9886af563/index.js#L135

JST5000 (approved):
LGTM

JST5000:
Oh, can you update the HISTORY.md file in the xrpl package?

jonathanlei:
HISTORY.md updated now

PR #2131, opened 2022-11-09
ci: cache dependencies

mvadari:

This PR adds caching of node_modules to testing. This shaves about a minute from the time to run CI.

### Context of Change

Everyone likes speedy CI

### Type of Change

## Test Plan

Works in CI.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -25,14 +25,13 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: Setup npm version 8
-        id: npm8
+      - name: Setup npm version 7
         run: |
           npm i -g npm@7 --registry=https://registry.npmjs.org
 
       - name: Cache node modules
         id: cache-nodemodules
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         env:
           cache-name: cache-node-modules
         with:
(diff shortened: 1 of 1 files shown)


justinr1234 (approved):

.github/workflows/nodejs.yml
```diff
@@ -18,37 +18,79 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Setup npm version 8
+        id: npm8
+        run: |
+          npm i -g npm@7 --registry=https://registry.npmjs.org
```
JST5000, line 71: This says it's setting up version 8 then installs version 7

.github/workflows/nodejs.yml
```diff
@@ -97,14 +160,35 @@ jobs:
...
```
JST5000, line 172: ditto

.github/workflows/nodejs.yml
```diff
@@ -97,14 +160,35 @@ jobs:
...
+
+      - name: Cache node modules
+        id: cache-nodemodules
+        uses: actions/cache@v2
```
JST5000, line 176: This is using `actions/cache@v2`, while the other additions are using `@v3`, is that intentional?

JST5000 (approved):
LGTM

PR #2143, opened 2022-11-19
Add vite setup steps to README

JST5000:

Outlines how to set up a Vite project that wants to use xrpl.js

### Context of Change

Follow up to issues people were having using Vite with xrpl.js

### Type of Change

- [X] Documentation Updates


ckniffen (approved):

pdp2121 (approved):

README.md
```diff
@@ -169,6 +169,71 @@ If you want to use `xrpl.js` with React Native you will need to install shims fo
...
+  },
+  optimizeDeps: {
+    esbuildOptions: {
+
+        define: {
+          global: 'globalThis',
+        },
+        plugins: [
+            NodeGlobalsPolyfillPlugin({
+                process: true,
```
mpint, line 200: We ended up not needing `process` here

README.md
```diff
@@ -169,6 +169,71 @@ If you want to use `xrpl.js` with React Native you will need to install shims fo
...
+export default defineConfig({
+  plugins: [react()],
+  define: {
+    'process.env': {}
+  },
+  optimizeDeps: {
+    esbuildOptions: {
+
+        define: {
+          global: 'globalThis',
```
mpint, line 196: this is in our config but predates any changes made for xrpl.js

README.md
```diff
@@ -169,6 +169,71 @@ If you want to use `xrpl.js` with React Native you will need to install shims fo
...
+      ]
+  }
+},
+resolve: {
+  alias: {
+    'events': 'events',
+    'crypto': 'crypto-browserify',
+    'stream': 'stream-browserify',
+    'http': 'stream-http',
+    'https': 'https-browserify',
```
mpint, line 219: we also needed `ws: 'xrpl/dist/npm/client/WSWrapper',` here

README.md
JST5000: It seems necessary when trying it out in my toy example (a new vite project + basic xrpl.js code to fund a wallet)

README.md
JST5000: The code I have seems to run fine without `process` as well in the toy example - @ckniffen do you happen to remember what adding `process: true` does here?

JST5000: (If it isn't necessary I'll do a follow up PR to remove it)

JustAnotherDevv:
It doesn't solve the problem for me, still can't use the xrpl.js with Vite. Was this actually tested by anyone?

ckniffen:
Ripple has an internal project that is using this setup.

> It doesn't solve the problem for me, still can't use the xrpl.js with Vite. Was this actually tested by anyone?

It is being used by an internal project at ripple.  What error are you getting? There could be something we missed as we don't use every method in that app and our dependencies might be different.

JustAnotherDevv:
Just created new project using react + vite just to confirm that it's still happening. It's literally the most basic setup - I have followed instructions for everything except for the fact that I used react+js template instead of ts template.

Line `const client = new xrpl.Client("wss://s.altnet.rippletest.net:51233");`
gives error:
`Cannot read properties of undefined (reading 'Client')`

my package.json file:
```
{
  "name": "vite-xrp-test",
  "private": true,
  "version": "0.0.0",
  "scripts": {
    "dev": "vite",
    "build": "vite build",
    "preview": "vite preview"
  },
  "dependencies": {
    "crypto-browserify": "^3.12.0",
    "events": "^3.3.0",
    "https-browserify": "^1.0.0",
    "react": "^17.0.2",
    "react-dom": "^17.0.2",
    "stream-browserify": "^3.0.0",
    "stream-http": "^3.2.0",
    "xrpl": "^2.6.0"
  },
  "devDependencies": {
    "@esbuild-plugins/node-globals-polyfill": "^0.1.1",
    "@vitejs/plugin-react": "^1.0.7",
    "rollup-plugin-polyfill-node": "^0.11.0",
    "vite": "^2.8.0"
  }
}
```

PR #2149, opened 2022-11-22
add submitAndWaitBatch sugar function

khancode:

Adds a new sugar function `submitAndWaitBatch()` that's an extension of `submitAndWait()` where it takes in an array of transactions with options and asynchronously submits them for each different Sender account.

### Context of Change

Technical Design:
- Similar to `submitAndWait()` but it takes in an array of transactions.
- Each array item has two properties: transaction and opts.
- A promise is created for each different Sender account so that transactions can be submitted asynchronously.
- Therefore, if there are multiple transactions with the same Sender account, then they will be submitted synchronously to ensure their sequence numbers are correctly incremented in each submission.
- TODO: add a flag to check if Mode 1 or 2 is enabled. For now, Mode 1 is enabled by default.
- The mode flag determines what happens after a transaction submission fails for a Sender account:
  - Mode 1: Sender account won't send its remaining transactions and put them in `result.unsubmitted` list.
  - Mode 2: Sender account will continue to submit its remaining transactions.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Added new integ tests.

TODO: add a flag to check if Mode 1 or 2 is enabled. For now, Mode 1 is enabled by default.


mvadari:
Seems like this doesn't handle errors very elegantly - if one transaction fails, you lose all information about the other transactions.

Also, `autofill` will need to give transactions consecutive sequence numbers, instead of identical ones.

khancode:
> Seems like this doesn't handle errors very elegantly - if one transaction fails, you lose all information about the other transactions.
> 
> Also, `autofill` will need to give transactions consecutive sequence numbers, instead of identical ones.

I addressed your comments ✅

packages/xrpl/test/integration/submitAndWaitBatch.test.ts
```diff
@@ -0,0 +1,481 @@
...
+      Destination: this.wallet.classicAddress,
+      Amount: '1000',
+    }
+    const txList = [
+      {
+        transaction: paymentTx,
+        opts: { wallet: this.wallet },
+      },
+      {
+        transaction: invalidAccountDeleteTx,
```
justinr1234, line 143: What happens if you submit multiple transactions with a failed one in the middle of the order?

packages/xrpl/src/sugar/submit.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable max-depth -- necessary */
```
mvadari, line 1: max-depth seems solvable with helper methods

packages/xrpl/src/sugar/submit.ts
```diff
@@ -72,24 +90,77 @@ async function submitAndWait(
...
+        wallet?: Wallet
+      }
+    }>
+  > = new Map()
+  for (const tx of transactions) {
+    const account = tx.transaction.Account
+    if (!batchMap.has(account)) {
+      batchMap.set(account, [])
+    }
+    batchMap.get(account)?.push(tx)
```
mvadari, line 149: The performance can be improved slightly if this is set up as an if-else

packages/xrpl/src/sugar/submit.ts
```diff
@@ -176,6 +247,94 @@ async function waitForFinalTransactionOutcome(
...
+    lastLedger,
+    response.result.engine_result,
+  )
+}
+
+// eslint-disable-next-line max-params, max-lines-per-function -- necessary
+async function submitAndWaitBatchHelper(
+  client: Client,
+  account: string,
+  batchMap: Map<
```
mvadari, line 286: It would be a little neater if `submitAndWaitBatchHelper` only took `batchMap.get(account)` as a param, instead of the whole map

packages/xrpl/src/sugar/submit.ts
```diff
@@ -176,6 +247,94 @@ async function waitForFinalTransactionOutcome(
...
+): Promise<void> {
+  const transactions = batchMap.get(account)
+  if (transactions == null) {
+    throw Error(`transaction is undefined`)
+  }
+  for (let idx = 0; idx < transactions.length; idx++) {
+    const tx = transactions[idx]
+    try {
+      // eslint-disable-next-line no-await-in-loop -- necessary
+      const txResponse = await submitAndWaitHelper(
```
mvadari, line 307: Most of the advantage of a batch submit and wait method comes from submitting multiple transactions from one account. This structure means that you don't really save any time if you just want to batch a bunch of transactions from one account.

justinr1234 (approved):

mvadari (changes requested):
Most of the advantage of a batch submit and wait method comes from submitting multiple transactions from one account, since in prod, most people don't have access to keys for more than one account. This structure means that you don't really save any time if you just want to batch a bunch of transactions from one account, which kind of defeats the purpose of this method.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -57,6 +57,50 @@ async function autofill<T extends Transaction>(
...
+ * @param transactions - A {@link Transaction} in JSON format
+ * Only used for multisigned transactions.
+ * @returns The autofilled transaction.
+ */
+async function autofillBatch<T extends Transaction>(
+  this: Client,
+  transactions: Array<{ transaction: T; signersCount?: number }>,
+): Promise<T[]> {
+  // Sender Account to Sequence map
+  const map: Map<string, number> = new Map()
```
mvadari, line 76:
```suggestion
  const accountSequenceMap: Map<string, number> = new Map()
```
I think a descriptive name is actually useful here.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -57,6 +57,50 @@ async function autofill<T extends Transaction>(
...
+    const { transaction, signersCount } = txn
+    const account = transaction.Account
+
+    if (transaction.Sequence == null) {
+      // eslint-disable-next-line max-depth -- necessary
+      if (map.has(account)) {
+        transaction.Sequence = map.get(account)
+      } else {
+        // eslint-disable-next-line no-await-in-loop -- necessary
+        await setNextValidSequenceNumber(this, transaction)
```
mvadari, line 88: It doesn't look like you actually set the sequence number here.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -57,6 +57,50 @@ async function autofill<T extends Transaction>(
...
+  for (const txn of transactions) {
+    const { transaction, signersCount } = txn
+    const account = transaction.Account
+
+    if (transaction.Sequence == null) {
+      // eslint-disable-next-line max-depth -- necessary
+      if (map.has(account)) {
+        transaction.Sequence = map.get(account)
+      } else {
+        // eslint-disable-next-line no-await-in-loop -- necessary
```
mvadari, line 87: Seems like you could avoid this by putting everything inside this for loop in a helper method and `Promise.all`ing on that

packages/xrpl/src/sugar/autofill.ts
khancode: It sets the sequence number inside `setNextValidSequenceNumber` - [source](https://github.com/XRPLF/xrpl.js/blob/76b73e16a97e1a371261b462ee1a24f1c01dbb0c/packages/xrpl/src/sugar/autofill.ts#L140)

packages/xrpl/test/integration/submitAndWaitBatch.test.ts
khancode: For a single account, the failed one and subsequent transactions (they will also fail) will be returned in the `result.error` list.

mvadari: The subsequent ones won't fail if you're using tickets, which should be recommended in the docs for this method.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -88,6 +88,50 @@ async function autofill<T extends Transaction>(
...
+ *
+ * @param this - A client.
+ * @param transactions - A {@link Transaction} in JSON format
+ * Only used for multisigned transactions.
+ * @returns The autofilled transaction.
+ */
+async function autofillBatch<T extends Transaction>(
+  this: Client,
+  transactions: Array<{ transaction: T; signersCount?: number }>,
+): Promise<T[]> {
```
mvadari, line 105: Will this typing work with different types of transactions in the same array?

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -88,6 +88,50 @@ async function autofill<T extends Transaction>(
...
+ * `lastLedgerSequence` according to the current state of the server this Client
+ * is connected to. It also converts all X-Addresses to classic addresses and
+ * flags interfaces into numbers.
+ *
+ * @param this - A client.
+ * @param transactions - A {@link Transaction} in JSON format
+ * Only used for multisigned transactions.
+ * @returns The autofilled transaction.
+ */
+async function autofillBatch<T extends Transaction>(
```
mvadari, line 102: A thought - is it possible to have this be an alternate typing to the standard `autofill` method using typescript generics?

intelliot:
notes
- can use Tickets to allow batched transactions to succeed independently
  - Tickets do not improve performance in general
  - they provide some safety in the sense that an invalid transaction won't prevent other transactions from succeeding
- Mayukha indicated this would be useful and she would use it
- valid transactions will succeed. errors should be handled and corrected.

mvadari:
> * Mayukha indicated this would be useful and she would use it

I would use it in Python for https://github.com/XRPLF/xbridge-cli

sublimator:
> can use Tickets to allow batched transactions to succeed independently
> Tickets do not improve performance in general
> they provide some safety in the sense that an invalid transaction won't prevent other > transactions from succeeding

Way back in the day (2013) ripple-lib had a transaction manager that would handle transaction [re]signing and Sequence'ing. It would even plug Sequence gaps iirc (per Arthur' design) via AccountSet noop transactions (some fun bugs as you might imagine). For various reasons, both compelling and questionable, it was removed in the ill-fated (and some would argue (both then and now) ill-considered) move to RippleAPI. 

It is interesting that 10 years on this is still somewhat of an issue. Having to uses Tickets seems somewhat "expensive".

mvadari:
> Having to uses Tickets seems somewhat "expensive".

How so?

sublimator:
> Having to uses Tickets seems somewhat "expensive".
> How so?

Well you have to create and delete (consume) Ticket objects in the ledger?
Tickets were in development but canned for a while when I was at ripple, so I don't know if that's something that's changed, but it seems unlikely ?

intelliot:
[Tickets](https://xrpl.org/tickets.html) were added by the TicketBatch amendment and enabled on Mainnet in Nov 2021. They add some overhead as objects are created and then consumed, but that's not a particularly high cost compared to many other operations. While I don't think the use of Tickets is necessary for this PR, it would also be perfectly fine to use them if desired.

sublimator:
I have nothing against tickets per se, just surprised a bit that txn
submission is still not fully solved

mvadari:
> [Tickets](https://xrpl.org/tickets.html) were added by the TicketBatch amendment and enabled on Mainnet in Nov 2021. They add some overhead as objects are created and then consumed, but that's not a particularly high cost compared to many other operations. While I don't think the use of Tickets is necessary for this PR, it would also be perfectly fine to use them if desired.

I don't think it's worth the complexity of using tickets _in_ the library (you'd need to get a ticket transaction somehow signed by the wallet), the docs for this function should just strongly suggest that the dev use tickets to build the transactions that are being processed by `submitAndWait`.

sublimator:
It's tricky. You want to make it easy, but at the same time ...

khancode:
Work has been deprioritized, if anyone wants to pick this back up then feel free to copy the changes from this PR.

PR #2155, opened 2022-12-03
Deprecate mnemonic

jonathanlei:

deprecate Wallet.fromMnemonic since it's deprecated from Rippled and not recommended way to generate wallet. 

### Type of Change

- [ X] Bug fix (non-breaking change which fixes an issue)
- [X] Documentation Updates


packages/xrpl/src/Wallet/index.ts
```diff
@@ -201,6 +190,10 @@ class Wallet {
   /**
    * Derives a wallet from a bip39 or RFC1751 mnemonic (Defaults to bip39).
    *
+   * @deprecated since version 2.6.1.
+   * Will be deleted in version 3.0.0.
+   * Currently this implementation does not provide better reliability/Security, please use other methods to generate wallet.
```
JST5000, line 195:
```suggestion
   * This representation is currently deprecated in rippled. 
   * You should use another method to represent your keys such as a seed or public/private keypair.
```

JST5000: Reliability + security aren't the main reasons we are deprecating the feature so don't want to misrepresent the problem with `fromMnemonic`.

docs/classes/Wallet.html
```diff
@@ -341,14 +341,18 @@ <h5><code class="tsd-tag ts-flagOptional">Optional</code> master<wbr/>Address<sp
...
 <li class="tsd-description">
 <div class="tsd-comment tsd-typography"><p>Derives a wallet from a bip39 or RFC1751 mnemonic (Defaults to bip39).</p>
 
+<h3>Deprecated</h3><p>since version 2.6.1.
+Will be deleted in version 3.0.0.
+Currently this implementation does not provide better reliability/Security, please use other methods to generate wallet.</p>
```
JST5000, line 354: Are the generated docs necessary? (I think they'll be made when we do a release anyway, so may be worth removing changes to this file)

If they're important to keep - this is still showing the old message.

jonathanlei: good catch - I think we need to keep it because I manually updated the generate wallet examples, but the deprecation message should be updated now

JST5000 (approved):
LGTM

ckniffen (approved):

Issue #2150, opened 2022-11-23
Cannot delete EmailHash field.


develoQ:
The EmailHash field cannot be deleted in an AccountSet transaction.

The Domain field can be deleted by setting it to an empty string.

I think the EmailHash field should be able to be deleted by setting it to an empty string as well.

```js
import { Client, Wallet } from "xrpl";

const main = async () => {
  const client = new Client("wss://s.devnet.rippletest.net:51233/");
  const wallet = Wallet.generate()
  await client.connect();

  await client.fundWallet(wallet)
  await client.submitAndWait({
    TransactionType: 'AccountSet',
    Account: wallet.classicAddress,
    EmailHash: '',
  }, { wallet })
  await client.disconnect();
};

main();
```

```sh
Error: Invalid Hash length 0
    at Hash128.Hash [as constructor] (/Users/tequ/project/xrpl-test/node_modules/ripple-binary-codec/src/types/hash.ts:14:13)
    at new Hash128 (/Users/tequ/project/xrpl-test/node_modules/ripple-binary-codec/src/types/hash-128.ts:12:5)
    at Function.Hash.from (/Users/tequ/project/xrpl-test/node_modules/ripple-binary-codec/src/types/hash.ts:29:14)
    at /Users/tequ/project/xrpl-test/node_modules/ripple-binary-codec/src/types/st-object.ts:127:52
    at Array.forEach (<anonymous>)
    at Function.STObject.from (/Users/tequ/project/xrpl-test/node_modules/ripple-binary-codec/src/types/st-object.ts:126:12)
    at serializeObject (/Users/tequ/project/xrpl-test/node_modules/ripple-binary-codec/src/binary.ts:70:22)
    at signingData (/Users/tequ/project/xrpl-test/node_modules/ripple-binary-codec/src/binary.ts:90:10)
    at encodeForSigning (/Users/tequ/project/xrpl-test/node_modules/ripple-binary-codec/src/index.ts:49:10)
    at computeSignature (/Users/tequ/project/xrpl-test/node_modules/xrpl/src/Wallet/index.ts:530:31)
```

pdp2121:
Thanks for addressing this issue! We are working on fixing it. Will keep you updated!

PR #2156, opened 2022-12-06
fix: unset hash fields using AccountSet

pdp2121:

Fix [issue#2150](https://github.com/XRPLF/xrpl.js/issues/2150).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

User can unset a hash field using an empty string with AccountSet similarly to other fields.


packages/ripple-binary-codec/test/hash.test.js
```diff
@@ -1,7 +1,31 @@
...
+  test('can be unset', function () {
+    const h1 = Hash128.from('')
+    expect(h1.toJSON()).toBe('')
+  })
+  test('can be compared against another', function () {
+    const h1 = Hash128.from('100000000000000000000000000000000')
+    const h2 = Hash128.from('200000000000000000000000000000000')
+    const h3 = Hash128.from('000000000000000000000000000000003')
+    expect(h1.lt(h2)).toBe(true)
+    expect(h3.lt(h2)).toBe(true)
```
mvadari, line 18: Can this also test `.gt`?

jonathanlei (commented):
LGTM

packages/ripple-binary-codec/src/types/hash-128.ts
```diff
@@ -9,8 +9,25 @@ class Hash128 extends Hash {
   static readonly ZERO_128: Hash128 = new Hash128(Buffer.alloc(Hash128.width))
 
   constructor(bytes: Buffer) {
+    if (bytes && bytes.byteLength === 0) {
```
ckniffen, line 12: I think this should happen in Hash similar to how it is done in xrpl-py.

pdp2121: Added a PR for xrpl-py:
https://github.com/XRPLF/xrpl-py/pull/508

ckniffen (approved):

Issue #1818, opened 2021-11-19
Feature request: derive public key from private key @ripple-keypairs


khancode:
We should add a method such as `derivePublicKey` that takes a private key and derives a public key.

Original issue: https://github.com/ripple/ripple-keypairs/issues/168

EvanTedesco:
I put together a draft pull request to address this issue https://github.com/XRPLF/xrpl.js/pull/2158.  I wonder if `derivePublicKey` is a misnomer assuming what should actually be returned is the `accountId`.  It should probably be called `deriveAccountIdFromPrivate` I think.

In any case I made the change and I am happy to rename it if it is decided `deriveAccountIdFromPrivate` is preferred.

PR #2158, opened 2022-12-08
add derivePublicKey function

EvanTedesco:

adds `derivePublicKey(privateKey)` to key-pairs API. 

### Context of Change

addresses https://github.com/XRPLF/xrpl.js/issues/1818. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Before it was not immediately clear how to get the accountId from a private key.  This change adds a dedicated method to the key-pairs API.

### DIFF
--- a/packages/ripple-keypairs/src/index.ts
+++ b/packages/ripple-keypairs/src/index.ts
@@ -155,6 +155,11 @@ function deriveNodeAddress(publicKey): string {
   return deriveAddressFromBytes(accountPublicBytes)
 }
 
+function derivePublicKey(privateKey): string {
+  const { publicKey } = deriveKeypair(privateKey)
+  return deriveAddress(publicKey)
+}
+
 const { decodeSeed } = addressCodec
 
 export = {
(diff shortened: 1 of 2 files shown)


mvadari (commented):
Should this PR just return the public key instead of the account ID? A user could just run `deriveAddress` on the public key themselves.

jonathanlei:
also not sure if the PR solves the intention of the [issue](https://github.com/ripple/ripple-keypairs/issues/168) - "The deriveKeypair method takes a seed and derives the public and private key. We should add a method, perhaps derivePublicKey, that takes a private key and derives a public key. One complication is that we support two different algorithms: ecdsa-secp256k1 and ed25519. Given a private key, I don't think there is a way to know whether it is meant to be used as a secp256k1 key or an ed25519 one. So this would need to be determined by another parameter passed into the method (along with the private key data)." I am working on fixing decodeSeed to try the two algorithms and determine the right one.

EvanTedesco:
Ok I see what you mean.  My apologies for the confusion.

PR #2168, opened 2022-12-16
Update linter peer dependencies

JST5000:

Fixes a peer dependency mismatch and addresses a couple old Dependabot issues. 

### Context of Change

Updating dependencies

### Type of Change

## Before / After

Updated several packages associated with linting.

## Test Plan

CI Passes and can build locally


packages/xrpl/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "pretty": true,
-    "target": "es6",
+    "target": "es2021",
```
ckniffen, line 5: Do we want to update to such a new standard?

JST5000: es2017 is the minimum recommendation. I can use that instead? (Not sure what's best)

ckniffen: Yeah just wanted to warn.  It could be fine.  I just have never targeted 2021.

mvadari: Will this affect people who are using a lower standard? We want to make sure we're backwards compatible

JST5000: This in fact could be a non-backwards compatible change, so I'm undoing it: https://stackoverflow.com/questions/73387445/what-happens-if-tsconfig-targets-and-libs-are-mixed-among-dependencies-for-t

Seems the error I was getting which prompted the upgrade was a trickle down error from my other configuration changes, since it seems to have disappeared since I moved the `@types/node` import to the top level and made it match the earliest version of node we support.

package.json
```diff
@@ -21,25 +21,25 @@
     "@types/create-hash": "^1.2.2",
     "@types/lodash": "^4.14.136",
     "@types/mocha": "^9.0.0",
-    "@types/node": "^17.0.14",
+    "@types/node": "^12.12.31",
```
mvadari, line 24: Why is this being downgraded?

packages/ripple-address-codec/.eslintrc.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
     'jsdoc/require-throws': 'off',
     'jsdoc/require-jsdoc': 'off',
     'jsdoc/require-param': 'off',
+    'jsdoc/check-examples': 'off',
```
mvadari, line 42: Why is this turned off now?

packages/ripple-binary-codec/HISTORY.md
```diff
@@ -1,6 +1,7 @@
 # ripple-binary-codec Release History
 
 ## Unreleased
+- Small typing refactor in Shamap to match newer version of TypeScript.
```
mvadari, line 4: Does this affect the exported methods, or it just internal? If it's just internal, then it doesn't need to be in the changelog - it's effectively just a refactor.

packages/xrpl/snippets/src/claimPayChannel.ts
```diff
@@ -4,7 +4,8 @@ import {
   PaymentChannelCreate,
   PaymentChannelClaim,
   hashes,
-} from '../../dist/npm'
+} from '../../src'
+// Replace this import with "from 'xrpl'" if you copy and paste this code
```
mvadari, line 8: Should this be tested differently in CI and actually use `xrpl`?

packages/xrpl/src/models/transactions/checkCash.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unnecessary-condition -- JavaScript users can always pass in undefined */
```
mvadari, line 1: Is this only called in one place? If so, it should just disable that one line

package.json
JST5000: `@types/node` should be in sync with the oldest version of node that we support (which is currently 12). https://stackoverflow.com/questions/42035263/relationship-between-the-version-of-node-js-and-the-version-of-types-node

packages/ripple-address-codec/.eslintrc.js
JST5000: This rule isn't supported in the new version of `eslint` - ripple-address-codec:   145:1  warning  This rule cannot yet be supported for ESLint 8; you should either downgrade to ESLint 7 or disable this rule. The possibility for ESLint 8 support is being tracked at https://github.com/eslint/eslint/issues/14745  jsdoc/check-examples

packages/ripple-binary-codec/HISTORY.md
JST5000: Just internal, shouldn't affect anything

package.json
JST5000: I also forgot until just now that we're deprecating support for node 12 since it's no longer being supported. (So moving the types to v14)

packages/xrpl/src/models/transactions/checkCash.ts
JST5000: There were two, but they were close together so I did a disable + enable pair to cover them more closely.

packages/xrpl/snippets/src/claimPayChannel.ts
JST5000: That could be good since we want these to be copy + pastable. The only time that should cause problems is if we make breaking changes since the interface for the snippets would be incorrect, but that's decently rare. I'll open a ticket for it.

packages/ripple-binary-codec/src/shamap.ts
```diff
@@ -159,20 +159,22 @@ class ShaMapInner extends ShaMapNode {
...
-    } else if (existing instanceof ShaMapLeaf) {
-      const newInner = new ShaMapInner(this.depth + 1)
-      newInner.addItem(existing.index, undefined, existing)
-      newInner.addItem(index, item, leaf)
-      this.setBranch(nibble, newInner)
-    } else if (existing instanceof ShaMapInner) {
-      existing.addItem(index, item, leaf)
-    } else {
-      throw new Error('invalid ShaMap.addItem call')
+    if (index !== undefined) {
```
jonathanlei, line 162: was this change from dependabot?

JST5000: No, I made these changes - this was to deal with the new version of something not recognizing that `assert.ok(index !== undefined)` means that `index` is ok to use later on.

JST5000: (The Dependabot PR is separate & I'll close it once this gets merged :) )

intelliot (approved):

jonathanlei (approved):

package.json
```diff
@@ -21,25 +21,25 @@
...
+    "@types/node": "^14.14.31",
     "@types/puppeteer": "5.4.6",
     "@types/ws": "^8.2.0",
-    "@typescript-eslint/eslint-plugin": "^4.30.0",
-    "@typescript-eslint/parser": "^4.0.0",
-    "@xrplf/eslint-config": "^1.6.0",
+    "@typescript-eslint/eslint-plugin": "^5.28.0",
+    "@typescript-eslint/parser": "^5.0.0",
+    "@xrplf/eslint-config": "^1.9.0",
     "@xrplf/prettier-config": "^1.5.0",
```
mvadari, line 30: Shouldn't this be updated too?

ckniffen (approved):

justinr1234:
@SocketSecurity ignore jest@29.4.2
@SocketSecurity ignore jest-cli@29.4.2

justinr1234 (approved):
Why all the linting changes? Are those part of the eslint config itself?

JST5000:
> Why all the linting changes? Are those part of the eslint config itself?

Yep, they were part of the updated linting config (The 1.9.1 one)

ckniffen (approved):

PR #2181, opened 2023-01-06
chore: rename test files

justinr1234:
## High Level Overview of Change

Rename tests in preparation of [Jest PR](https://github.com/XRPLF/xrpl.js/pull/2170)

PR #2182, opened 2023-01-06
Revert "chore: rename test files"

justinr1234:
Reverts XRPLF/xrpl.js#2181


JST5000 (approved):
Why is the revert needed?

PR #2186, opened 2023-01-25
Swap create-hash for hash.js sha256

mankins:

Swap create-hash sha-256 implementation for hash.js which works in more environments.

### Context of Change

I was getting an error similar to https://github.com/crypto-browserify/createHash/issues/30 and found that swapping out the implementation fixed the referenced error. I see hash.js is used in other parts of the xrpl ecosystem too.

### Type of Change

## Before / After

There should be no change to existing running code. This allows the code to be run in more ecosytems (such as Tauri + Sveltekit where I'm running this).

## Test Plan

Existing tests pass.


packages/ripple-address-codec/package.json
```diff
@@ -28,5 +28,10 @@
   "prettier": "@xrplf/prettier-config",
   "engines": {
     "node": ">= 10"
+  },
+  "devDependencies": {
```
ckniffen, line 32: Why were these dependencies added?

mankins: @ckniffen `npm run test` didn't work without them. I can back them out if this doesn't belong here.

ckniffen: I think they should be added via lerna when doing `npm install`. @justinr1234 is this correct?

mankins: I verified that if I do `npm install` in the project root this is taken care of. Consequently I reverted the additional `devDependencies`.

justinr1234: Yes you can follow instructions here: https://futurestud.io/tutorials/lerna-install-dependencies-for-a-specific-package

`lerna add hash.js --scope=ripple-address-codec`

But I believe the effect might be similar.

justinr1234:
@mankins The issue is this: https://github.com/crypto-browserify/cipher-base/issues/11

So changing to a package that works well in the browser is a great solution.

justinr1234:
Also it seems like you probably need to add the polyfill for `stream` anyways. 

In webpack that would look like:

```
resolve: {
        alias: {
            stream: 'stream-browserify'
        }
    },
```

These polyfills are typically used by other libraries as well, such as [web3.js](https://web3js.readthedocs.io/en/v1.8.1/)

packages/ripple-address-codec/package.json
mankins: I did the above with no changes to the local setup, so I think this is resolved.

mankins:
> Also it seems like you probably need to add the polyfill for `stream` anyways.

You mean on my app, right? I haven't needed that yet, but good to know.

Let me know if I should do anything else here.

mankins:
> @mankins did you try adding those polyfills configuration in our docs?

Some context: I was doing this for `ripple-address-codec` not xrpl.js. This is now a mono repo for some npm packages, right? (or am I in the wrong place?)

I'm not sure which polyfills in your docs you were referring to (the stream one above? how does that relate to `ripple-address-codec`? I added polyfills via vite (not webpack), but was able to solve this by changing the dependency. It's possible I didn't do that right so if I need to I can try again.

> My hesitation with making this switch is it does not eliminate `create-hash` as a dependency it mere adds another one. Below are all the other places `create-hash` will still be required.

I'm seeing `hash.js` already in use in [ripple-keypairs](https://github.com/XRPLF/xrpl.js/search?q=hash.js).

I could also expand the PR and change out the rest if that would get this merged.

ckniffen:
> I'm seeing `hash.js` already in use in [ripple-keypairs](https://github.c> I could also expand the PR and change out the rest if that would get this merged.

Good point.  I didn't not realize that.  In that case can you update ripple-keypairs's version to be `^1.0.7` as well>

mankins:
> In that case can you update ripple-keypairs's version

Ok, updated.

packages/ripple-address-codec/src/xrp-codec.ts
```diff
@@ -161,7 +161,7 @@ const ED25519_SEED = [0x01, 0xe1, 0x4b]
 
 const codecOptions = {
   sha256(bytes: Uint8Array): Buffer {
-    return createHash('sha256').update(Buffer.from(bytes)).digest()
+    return Buffer.from(hashjs.sha256().update(bytes).digest())
```
sublimator, line 164: when I originally made this library, the hash function was configurable iirc 
"bring your own hash" :)

hashjs will be a bit of a perf drop for some use cases ?

sublimator:
Was ///years/// ago
2015

packages/ripple-address-codec/src/xrp-codec.ts
sublimator: Strange, createHash should use crypto when running on node, then use the "browser" versions in the browser. 
Maybe you are using it in a strange environment and are missing some polyfills that some build tools like webpack usually automatically bundle.

mankins: I think it's because it's compiled for Tauri ... you're right though I should try to get that working again. I had seen this as a quick route without downsides, but maybe that's not the case in practice.

mankins:
I'm going to close this, as it's stale and I got the library working with more polyfilling as suggested above. 

Would be cool to make `createHash` configurable as suggested above, but that will likely be another pr.

PR #2191, opened 2023-01-29
[WIP] Fature: add custom definitions.json import

WietseWind:
If you want to use this lib. on e.g. a side chain, where different (new) transaction types, fields, etc. are supported, you can load a [custom definitions.json](https://github.com/RichardAH/xrpl-codec-gen).

Loading a custom definitions.json file with the setDefinitions method needs to happen before any of the other methods of this lib. are called.

```js
// First only import the definitions methods
const { setDefinitions } = require('ripple-binary-codec/dist/enums/definitions')
setDefinitions(require('./my-custom-definitions.json'))

const { encode, decode } = require('ripple-binary-codec')

// Now enjoy your custom definitions:
const encoded = encode({ TransactionType: 'SomeNewTxtype' })

console.log(encoded)
```

This implementation relies on the fact that imports (require, ...) are only loading & executing imported code once. An object for that import is created, and this object will now be referenced whenever data/methods are referred to by other routines. If overwriting a var in the method scope, all calls relying on the data in that scope will point to the new data.

So: by setting (overwriting) the definitions object before other methods are called, the moment other methods are called basing const output on the overwritten object, the const output will be generated with the new, imported data.

### Warning

This works when using `require`. In modern js 'import' this approach doesn't work. A better solution would be to natively allow for a custom definitions import, results in breaking changes.

### DIFF
--- a/packages/ripple-binary-codec/README.md
+++ b/packages/ripple-binary-codec/README.md
@@ -105,3 +105,20 @@ Use `--coverage` to generate and display code coverage information:
     npm test --coverage
 
 This tells jest to output code coverage info in the `./coverage` directory, in addition to showing it on the command line.
+
+## Custom definitions
+If you want to use this lib. on e.g. a side chain, where different (new) transaction
+types, fields, etc. are supported, you can load a
+[custom `definitions.json`](https://github.com/RichardAH/xrpl-codec-gen).
+
+Loading a custom `definitions.json` file with the `setDefinitions` method needs
+to happen **before** any of the other methods of this lib. are called.
+
+```js
+// First only import the definitions methods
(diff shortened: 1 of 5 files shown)


ckniffen:
@WietseWind Have you had a chance to looks at this implementation? #2127.

WietseWind:
@ckniffen Perfect, so it's work in progress 👍  Then I'll be hands off with this workaround.  Encode/decode passing works.

ckniffen:
I think the library could benefit from both techniques.  @JST5000

mvadari:
> I think the library could benefit from both techniques. @JST5000

IMO the approach in #2127 is better, because it avoids potential confusion that may come from modifying global variables.

PR #2202, opened 2023-02-06
fix: load osano first in docs site

justinr1234:

The osano script needs to load before other scripts (GTM) to be able to detect them.


JST5000 (approved):
LGTM

ckniffen (approved):

PR #2205, opened 2023-02-14
Add a helper to get the NFTokenID after minting

JST5000:

Add `getNFTokenID` to parse the metadata after minting an NFT to get the NFTokenID.

### Context of Change

Right now it's very complicated to the get the NFTokenID after minting an NFT, which makes it hard to use the other transactions related to that NFT.

### Type of Change

## Before / After

* Added `getNFTokenID` to the utils folder
* Also added an integration test case for NFTokenMint

## Test Plan

CI passes


packages/xrpl/src/models/transactions/metadata.ts
```diff
@@ -29,6 +29,42 @@ interface DeletedNode {
 
 export type Node = CreatedNode | ModifiedNode | DeletedNode
 
+/* eslint-disable @typescript-eslint/no-unnecessary-condition -- Typeguard requires checking if undefined */
```
justinr1234, line 32: Is there a way to use explicit casting to avoid disabling the rules?

packages/xrpl/src/utils/getNFTokenID.ts
```diff
@@ -0,0 +1,86 @@
+/* eslint-disable @typescript-eslint/consistent-type-assertions -- Necessary for parsing metadata */
```
justinr1234, line 1: Can we avoid global disable like this

packages/xrpl/src/utils/getNFTokenID.ts
```diff
@@ -0,0 +1,86 @@
...
+import { flatMap } from 'lodash'
```
justinr1234, line 2: `import flatMap from 'lodash/flatMap'`

packages/xrpl/src/utils/getNFTokenID.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @returns The NFTokenID for the minted NFT.
+ * @throws if meta is not TransactionMetadata.
+ */
+export default function getNFTokenID(
+  meta: TransactionMetadata,
+): string | undefined {
+  /* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- Provides a nicer error for js users */
+  if (meta.AffectedNodes === undefined) {
+    throw new TypeError(`Unable to parse the parameter given to getNFTokenID. 
+    'meta' must be the metadata from an NFTokenMint transaction. Received ${JSON.stringify(
```
justinr1234, line 32: I think you can pass the data in the second parameter instead of stringifying it

packages/xrpl/src/utils/getNFTokenID.ts
```diff
@@ -0,0 +1,86 @@
...
+    if (isModifiedNode(node)) {
+      return (
+        node.ModifiedNode.LedgerEntryType === 'NFTokenPage' &&
+        Boolean(node.ModifiedNode.PreviousFields?.NFTokens)
+      )
+    }
+    return false
+  })
+
+  /* eslint-disable @typescript-eslint/no-unnecessary-condition -- Doing conditional type checking */
```
justinr1234, line 63: seems like there should be a way to avoid this

packages/xrpl/src/utils/getNFTokenID.ts
```diff
@@ -0,0 +1,86 @@
...
+  const previousTokenIDSet = new Set(
+    flatMap(affectedNodes, (node) =>
+      (
+        ((node as ModifiedNode).ModifiedNode?.PreviousFields
+          ?.NFTokens as NFToken[]) ?? []
+      ).map((token) => token.NFToken.NFTokenID),
+    ).filter((id) => id),
```
justinr1234, line 70: nit: `!!id`

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  beforeEach(async () => {
+    testContext = await setupClient(serverUrl)
+  })
+  afterEach(async () => teardownClient(testContext))
+
+  it(
+    'get NFTokenID',
+    async function () {
+      const client = new Client('wss://s.altnet.rippletest.net:51233/')
```
justinr1234, line 32: You can just pass a different url to `setupClient` if you want and it also gives you a funded wallet.

packages/xrpl/src/utils/getNFTokenID.ts
JST5000: Added, although we have an eslint rule which recommends doing `Boolean(id)` instead of `!!id` - they accomplish the same thing but Boolean is more explicit. (So I used the `Boolean(id)` syntax)

packages/xrpl/src/utils/getNFTokenID.ts
JST5000: Seems like that's not the case for TypeError - I get a TypeScript error when trying to just pass in a second argument to augment the message -
```
src/utils/getNFTokenID.ts:33:7 - error TS2554: Expected 0-1 arguments, but got 2.

33       meta,
         ~~~~
```

packages/xrpl/src/models/transactions/metadata.ts
JST5000: HasOwnProperty works for this file - https://stackoverflow.com/a/73810861

packages/xrpl/src/utils/getNFTokenID.ts
ckniffen: Have the serialized last to make it easier to read.  Maybe just `Recieved: ` and remove `instead.`

packages/xrpl/src/utils/getNFTokenID.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+    if (isModifiedNode(node)) {
+      return (
+        node.ModifiedNode.LedgerEntryType === 'NFTokenPage' &&
+        Boolean(node.ModifiedNode.PreviousFields?.NFTokens)
+      )
+    }
+    return false
+  })
+  /* eslint-disable @typescript-eslint/consistent-type-assertions -- Necessary for parsing metadata */
```
ckniffen, line 61: We should find a way to eliminate all of these disabled rules.

ckniffen: These linting errors don't show up in the explorer and I am pretty sure we have `@typescript-eslint/no-unnecessary-condition` turned on.

JST5000: I believe currently the rule is set to "No Type Assertions", we could override it to do "as" type assertions. (I don't think type assertions are unreasonable)

Alternatively the thought could be that having the "No Type Assertions" rule has a more positive impact on the code quality which is worth the cost of occasionally adding the linting directive. (There are about 15 files in the `xrpl` package with this rule disabled with directives)

I don't have a strong preference either way.

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
JST5000: `setupClient` currently relies on owning the network in order to fund the account - it uses logic like the faucet to try and send money from the account that starts with all the money. Currently for this test though I have to use testnet because our standalone node doesn't support NFT transactions. (There's a separate ticket to update our standalone docker container)

So for now instead of updating `setupClient` to have 2 separate ways of creating a client & wallet, I'm just going to use `testnet` for this test. The real fix imo will be updating the docker container so that all transactions can use standalone mode normally :)

packages/xrpl/src/utils/getNFTokenID.ts
ckniffen: :( Not a fan of that rule.

ckniffen (approved):

packages/xrpl/src/utils/getNFTokenID.ts
JST5000: @ckniffen I made this PR to remove it :) - https://github.com/XRPLF/xrpl.js/pull/2220

jonathanlei (approved):

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
justinr1234: @JST5000 sorry I don’t follow. It seems your code is equivalent to setupClient and setupClient accepts whatever network you want. Can you show me at the code level what’s different?

JST5000: Resolved offline :)

PR #2211, opened 2023-02-16
Update xrpl.js to use the definition.json changes in ripple-binary-codec

JST5000:

Splitting up the [original PR here](https://github.com/XRPLF/xrpl.js/pull/2127) into two sections, this one focused on the xrpl.js interface for using the updated definitions.

### Context of Change

The changes to the codec are simpler, whereas the interface is still up for debate, so wanted to separate the concerns.

### Type of Change

## Before / After

Added the ability to specify the types desired for `encode` and `decode` when serializing transactions using high level functions like `submitAndWait` in xrpl.js

## Test Plan

CI Passes


JST5000:
Didn't quite work to get my changes split up - opened #2214 which actually worked :)

PR #2214, opened 2023-02-17
Add definitions.json changes to xrpl package

JST5000:

The user interface for defining custom definitions needs more work, so I split up #2127 to focus on the binary-codec, with this one focusing on the interface.

### Context of Change

With sidechains and many amendments being worked on, we need a better way to define which types & transactions we're able to serialize/deserialize.

### Type of Change

## Before / After

* Added an optional parameter to high level functions to manually specify the XRPL Definitions to use.
* Added tests for modifying definitions

## Test Plan

CI Passes

## Future Tasks
* Integrate with `server_definitions` 
* Use the Client as the "global state" so users only have to specify the definitions once per connection.
* Automatically determine definitions when client is connected (defaulting to mainnet values)


packages/xrpl/src/sugar/submit.ts
```diff
@@ -177,31 +190,40 @@ async function waitForFinalTransactionOutcome(
 }
 
 // checks if the transaction has been signed
-function isSigned(transaction: Transaction | string): boolean {
-  const tx = typeof transaction === 'string' ? decode(transaction) : transaction
+function isSigned<T extends BaseTransaction = Transaction>(
+  transaction: T | string,
+  definitions?: InstanceType<typeof XrplDefinitionsBase>,
```
JST5000, line 195: Should not be optional since this directly uses decode, so we should specify our default one layer above the binary codec across the board.

packages/xrpl/src/sugar/submit.ts
```diff
@@ -214,28 +236,38 @@ async function getSignedTx(
...
-  return wallet.sign(tx).tx_blob
+  return wallet.sign(tx, false, definitions).tx_blob
 }
 
 // checks if there is a LastLedgerSequence as a part of the transaction
-function getLastLedgerSequence(
-  transaction: Transaction | string,
+function getLastLedgerSequence<T extends BaseTransaction = Transaction>(
+  transaction: T | string,
+  definitions?: InstanceType<typeof XrplDefinitionsBase>,
```
JST5000, line 252: Same for this one.

packages/xrpl/src/sugar/submit.ts
```diff
@@ -214,28 +236,38 @@ async function getSignedTx(
...
   // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- converts LastLedgSeq to number if present.
   return tx.LastLedgerSequence as number | null
 }
 
 // checks if the transaction is an AccountDelete transaction
-function isAccountDelete(transaction: Transaction | string): boolean {
-  const tx = typeof transaction === 'string' ? decode(transaction) : transaction
+function isAccountDelete<T extends BaseTransaction = Transaction>(
+  transaction: T | string,
+  definitions?: InstanceType<typeof XrplDefinitionsBase>,
```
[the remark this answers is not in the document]
JST5000: Nevermind (and for all above) - I think it makes the most sense to just have `encode` and `decode` be the only functions with default value of `DEFAULT_DEFINITIONS`.

justinr1234:
@coderabbit full review

justinr1234:
@coderabbitai full review

PR #2215, opened 2023-02-17
add memos to fundWallet

jonathanlei:
initial commit 
### Type of Change

- [X ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/packages/xrpl/src/Wallet/fundWallet.ts
+++ b/packages/xrpl/src/Wallet/fundWallet.ts
@@ -19,6 +19,17 @@ const INTERVAL_SECONDS = 1
 // Maximum attempts to retrieve a balance
 const MAX_ATTEMPTS = 20
 
+export const DEFAULT_MEMO = {
+  data: 'xrpl-js',
+  type: 'user-agent',
+  format: 'text',
+}
+interface memo {
+  data?: string
+  type?: string
+  format?: string
+}
+
 /**
  * Generates a random wallet with some amount of XRP (usually 1000 XRP).
  *
@@ -45,6 +56,8 @@ const MAX_ATTEMPTS = 20
  * Ex: client.fundWallet(null,{'faucet.altnet.rippletest.net', '/accounts'})
  * specifies a request to 'faucet.altnet.rippletest.net/accounts' to fund a new wallet.
  * @param options.amount - A custom amount to fund, if undefined or null, the default amount will be 1000.
+ * @param options.memo - A custom memo object to add for on-chain tracking purposes
+ * default memo is added as "xrpl-js".
  * @returns A Wallet on the Testnet or Devnet that contains some amount of XRP,
  * and that wallet's balance in XRP.
  * @throws When either Client isn't connected or unable to fund wallet address.
@@ -57,6 +70,7 @@ async function fundWallet(
     faucetHost?: string
     faucetPath?: string
     amount?: string
+    memo?: memo
   },
 ): Promise<{
   wallet: Wallet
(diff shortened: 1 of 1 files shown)


packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -19,6 +19,17 @@ const INTERVAL_SECONDS = 1
 // Maximum attempts to retrieve a balance
 const MAX_ATTEMPTS = 20
 
+export const DEFAULT_MEMO = {
+  data: 'xrpl-js',
+  type: 'user-agent',
```
intelliot, line 24: I think it would be good to write/start an [XLS-Standard](https://github.com/XRPLF/XRPL-Standards/discussions) for the format of this memo

jonathanlei: do you mean the type "user-agent"? and the information getting passed in?

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -57,6 +70,7 @@ async function fundWallet(
     faucetHost?: string
     faucetPath?: string
     amount?: string
+    memo?: memo
```
mvadari, line 73: Why do we want people to be able to specify their own memos? Isn't this just for tracking faucet use?

packages/xrpl/src/Wallet/fundWallet.ts
JST5000: Would the intent of that be to get more opinions on the inclusion of this? (And is there an example of what that might look like?)

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -19,6 +19,17 @@ const INTERVAL_SECONDS = 1
...
+export const DEFAULT_MEMO = {
+  data: 'xrpl-js',
+  type: 'user-agent',
+  format: 'text',
+}
+interface memo {
+  data?: string
+  type?: string
+  format?: string
+}
```
JST5000, line 31: I think this is missing the "Memo" field. 
Ex. "Memos": [
        {
            "Memo": {
                "MemoType": "687474703a2f2f6578616d706c652e636f6d2f6d656d6f2f67656e65726963",
                "MemoData": "72656e74"
            }
        }
    ],
```suggestion
interface memo {
  Memo: {
    data?: string
    type?: string
    format?: string
  }
}
```

mvadari: This type can be imported from `src/models`

jonathanlei: > This type can be imported from `src/models`

the models is wrong - faucet API takes in another format

jonathanlei: actually this fix needs more fixing - I think the type is wrong

jonathanlei: i'll figure it out later today

packages/xrpl/src/Wallet/fundWallet.ts
intelliot: Yes, the type and its meaning.

We should probably follow existing standards for user agent identification. One big gap that I see here is that `xrpl-js` does not indicate the version of the library used. It would be ideal to incorporate the version number, e.g. `xrpl-js/2.8.0`.

Also, why is it `xrpl-js` instead of `xrpl.js`?

The intent would be to establish a standard that would be used by other client libraries and by other user agents. So, for example, a wallet like XUMM could set a user-agent of `XUMM/<version>` (or whatever format the XLS defines).

It would be fine to get some opinions, but that is not the primary intent.

packages/xrpl/src/Wallet/fundWallet.ts
intelliot: The format of the custom memos used by dev advocates, the learning portal, etc. would be another good thing to define in an XLS spec.

PR #2216, opened 2023-02-17
feature `NetworkID`

dangell7:

https://github.com/XRPLF/rippled/pull/4370

### Context of Change

Added network id to the base transaction
Added network id to the `server_info` response
Added network id to the binary codec
Added network id to client and modify on `client.connect()`
Added network id to autofill when `rippled_version` is 1.11.0 or later or network is hooks testnet

### Type of Change

## Test Plan

No tests were written


ckniffen:
This library will already need to be updated due to the addition of 1.10.0 amendment featureXRPFees and the changes it makes to `server_info`.  I was thinking of the autofill changes before digging into your PR description.  I think that is the best way to go.  It will largely allow a peaceful upgrade.

dangell7:
> This library will already need to be updated due to the addition of 1.10.0 amendment featureXRPFees and the changes it makes to `server_info`. I was thinking of the autofill changes before digging into your PR description. I think that is the best way to go. It will largely allow a peaceful upgrade.
> 
> #2219

Yeah I agree, using the auto fill seems like the best approach.

mvadari:
@dangell7 are you planning on finishing this PR yourself or are you just updating the definitions.json?

dangell7:
> @dangell7 are you planning on finishing this PR yourself or are you just updating the definitions.json?

I was waiting for `XRPFees` because I thought it would effect this. I will finish this now though

mvadari:
> > @dangell7 are you planning on finishing this PR yourself or are you just updating the definitions.json?
> 
> I was waiting for `XRPFees` because I thought it would effect this. I will finish this now though

I think they can be implemented independently (especially since this change doesn't require an amendment).

dangell7:
@mvadari I have some thoughts on this. 

I added networkID to the autofill function. Technically in hooks amendment there is a fee endpoint that is different than `server_info`. I think the XRPFees is a prequel to that. That is why I chose to add an additional call instead of piggy back off of the `server_info`call used for setFee.

However my thoughts are that the moment you are connected to the ws/rpc, you are connected with a NetworkID. The autofill adds the fields to the transaction but those fields are constantly adjusting based on different inputs and the Network ID doesn't change. 

So does it make sense to be in autofill or should we do a `server_info` call on connection of the api?

dangell7:
@mvadari I made this as a [second option](https://github.com/XRPLF/xrpl.js/pull/2217/commits/5ca2a849b882e8c93522b1693b54df3de4d0fd89). The sdk doesnt emit the connected message until after the server_info response is received.

dangell7:
Regardless of the option we choose, the integration/browser tests will fail because `autofill` is adding a field (NetworkID) to the transaction that does not exist on the standalone. I think we will need to update the standalone with this PR if we want to include `NetworkID` in the autofill.

mvadari:
This PR should not be merged in until the rippled PR is in `develop` (ideally not until it's in a new rippled release).

dangell7:
> This PR should not be merged in until the rippled PR is in `develop` (ideally not until it's in a new rippled release).

Wouldn't it still fail for mainnet until the amendment was passed and all servers upgraded?

I think we would need to check if the amendment is enabled before adding the `NetworkID` to the transaction in the autofill but that sounds overkill.

mvadari:
> > This PR should not be merged in until the rippled PR is in `develop` (ideally not until it's in a new rippled release).
> 
> Wouldn't it still fail for mainnet until the amendment was passed and all servers upgraded?
> 
> I think we would need to check if the amendment is enabled before adding the `NetworkID` to the transaction in the autofill but that sounds overkill.

There's no amendment, there's no change for mainnet. The only change is in testnets, and the only testnet that has this feature right now is the hooks one. So if we merge this in now, then it screws over everyone working on the Testnet or Devnet.

intelliot:
I think you can change this PR so that when xrpl.js connects to a server that doesn't use a NetworkID (including Mainnet, Testnet, Devnet, etc) it doesn't set the NetworkID field. That way, it wouldn't break anything, and also doesn't need to wait for anything.

dangell7:
Usage: https://github.com/Transia-RnD/xls-playground/blob/main/ts/test/networkID.test.ts

```
client.networkID = await client.getNetworkID();
```

mvadari:
> @mvadari What kind of tests do you want to see here?

You can add unit tests using the mocked rippled setup that we have, and integration tests can wait for when the PR is merged and the Docker container is updated.

ckniffen:
@dangell7 what was the reason for closing this?

dangell7:
@ckniffen I closed this because I was making lots of changes and releasing a custom branch at Transia. I wasn't sure exactly what I was going to do or how I was going to release on my fork.

pdp2121:
Hi @dangell7, could you provide some updates on this PR (plan to add tests, etc.)? Thank you!

dangell7:
> Hi @dangell7, could you provide some updates on this PR (plan to add tests, etc.)? Thank you!

I will push the tests today. No change in implementation.

pdp2121:
> @mvadari I have some thoughts on this.
> 
> I added networkID to the autofill function. Technically in hooks amendment there is a fee endpoint that is different than `server_info`. I think the XRPFees is a prequel to that. That is why I chose to add an additional call instead of piggy back off of the `server_info`call used for setFee.
> 
> However my thoughts are that the moment you are connected to the ws/rpc, you are connected with a NetworkID. The autofill adds fields to the transaction but those fields are constantly adjusting based on different inputs and the Network ID doesn't change.
> 
> So does it make sense to be in autofill or should we do a `server_info` call on connection of the api?
> 
> Also, is networkID a `required` field in the sdk? Making it optional like `Fee` and `Sequence` would allow users to use the sdk to connect to both versions of ripped. I also have that same question for the `server_state` model and test fixtures.

Hi @dangell7 , after some discussions, we think the approach to make `server_info` call on `Client.connect()` would be a preferred approach (`networkId` would be initialized to `undefined`, then its value would be set on connection, since `1` would be confused with mainnet and for version < 1.11.0 server_info might not return networkId). Also this approach would save one line of code for user (`client.networkID = await client.getNetworkID();`). Hope that makes sense to you! 

I can continue this PR to make modifications and add tests if you don't mind. Thanks!

dangell7:
> > @mvadari I have some thoughts on this.
> > I added networkID to the autofill function. Technically in hooks amendment there is a fee endpoint that is different than `server_info`. I think the XRPFees is a prequel to that. That is why I chose to add an additional call instead of piggy back off of the `server_info`call used for setFee.
> > However my thoughts are that the moment you are connected to the ws/rpc, you are connected with a NetworkID. The autofill adds fields to the transaction but those fields are constantly adjusting based on different inputs and the Network ID doesn't change.
> > So does it make sense to be in autofill or should we do a `server_info` call on connection of the api?
> > Also, is networkID a `required` field in the sdk? Making it optional like `Fee` and `Sequence` would allow users to use the sdk to connect to both versions of ripped. I also have that same question for the `server_state` model and test fixtures.
> 
> Hi @dangell7 , after some discussions, we think the approach to make `server_info` call on `Client.connect()` would be a preferred approach (`networkId` would be initialized to `undefined`, then its value would be set on connection, since `1` would be confused with mainnet and for version < 1.11.0 `server_info` might not return `networkId`). Also this approach would save one line of code for user (`client.networkID = await client.getNetworkID();`). Hope that makes sense to you!
> 
> I can continue this PR to make modifications and add tests if you don't mind. Thanks!

I like that approach as well however old servers wont respond with a network id hence the default.

dangell7:
I pushed my test/s. If you want to cherry pick this or just add to this PR I absolutely do not mind at all. I was just trying to keep things moving along and giving some of the devs who work on the experimental stuff early access though the sdk.

pdp2121:
> I pushed my test/s. If you want to cherry pick this or just add to this PR I absolutely do not mind at all. I was just trying to keep things moving along and giving some of the devs who work on the experimental stuff early access though the sdk.

Totally understood! Great work @dangell7 ! I can take it from here

pdp2121:
Hi @dangell7 , I don't have permission to push to your repo. Any suggestions on how to edit this? Thank you!

dangell7:
> Hi @dangell7 , I don't have permission to push to your repo. Any suggestions on how to edit this? Thank you!

I can invite you to the repo or you can create a new branch yourself if you'd like? I'm not sure either? I thought you should be able to push because the PR? lol

pdp2121:
> > Hi @dangell7 , I don't have permission to push to your repo. Any suggestions on how to edit this? Thank you!
> 
> I can invite you to the repo or you can create a new branch yourself if you'd like? I'm not sure either? I thought you should be able to push because the PR? lol

lol I think you'll need to set **Allow edits from maintainers** as instructed here:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

dangell7:
> > > Hi @dangell7 , I don't have permission to push to your repo. Any suggestions on how to edit this? Thank you!
> > 
> > 
> > I can invite you to the repo or you can create a new branch yourself if you'd like? I'm not sure either? I thought you should be able to push because the PR? lol
> 
> lol I think you'll need to set **Allow edits from maintainers** as instructed here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Unfortunately these commits are from an organization not a user. So I don't have access to that feature. :(

dangell7:
> > > > > Hi @dangell7 , I don't have permission to push to your repo. Any suggestions on how to edit this? Thank you!
> > > > 
> > > > 
> > > > I can invite you to the repo or you can create a new branch yourself if you'd like? I'm not sure either? I thought you should be able to push because the PR? lol
> > > 
> > > 
> > > lol I think you'll need to set **Allow edits from maintainers** as instructed here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork
> > 
> > 
> > Unfortunately these commits are from an organization not a user. So I don't have access to that feature. :(
> 
> hmm probably the next option would be adding me as a contributor to the forked repo then. Please let me know if that works. Thanks Denis

added

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -88,6 +93,32 @@ async function autofill<T extends Transaction>(
...
+    await client.setNetworkIDAndBuildVersion()
+  }
+  if (
+    client.networkID !== undefined &&
+    client.networkID > RESTRICTED_NETWORKS
+  ) {
+    // autofill transaction's networkID if either the network is hooks testnet or build version is >= 1.11.0
+    if (
+      (client.buildVersion &&
+        isEarlierRippledVersion(
```
jonathanlei, line 107: can this function be used anywhere else?

packages/xrpl/src/client/index.ts
```diff
@@ -668,6 +683,11 @@ class Client extends EventEmitter {
    */
   public getLedgerIndex = getLedgerIndex
 
+  /**
+   * @category Abstraction
+   */
+  public setNetworkIDAndBuildVersion = setNetworkIDAndBuildVersion
```
mvadari, line 689: Why does this need to be a public exported method?

pdp2121: This is the part that I'm not happy with. If I make it as a helper function inside `autofill`, ESLint will complain about race condition `require-atomic-updates` when I set `client.networkID = ...` and `client.buildVersion=...` but I agree that this shouldn't be a public exported method. Any ideas?

packages/xrpl/src/sugar/autofill.ts
pdp2121: Not at the moment, I made a generic version comparison since we might need it in the future for rules that only work from certain version or later similar to this

packages/xrpl/src/client/index.ts
pdp2121: I made it a helper function for now (which I think is more reasonable), but it will have a bunch of eslint disable statements

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -3,15 +3,22 @@ import { xAddressToClassicAddress, isValidXAddress } from 'ripple-address-codec'
...
 import { Transaction } from '../models/transactions'
 import { setTransactionFlagsToNumber } from '../models/utils/flags'
-import { xrpToDrops } from '../utils'
+import { isEarlierRippledVersion, xrpToDrops } from '../utils'
 
 import getFeeXrp from './getFeeXrp'
 
 // Expire unconfirmed transactions after 20 ledger versions, approximately 1 minute, by default
 const LEDGER_OFFSET = 20
+const RESTRICTED_NETWORKS = 1024
```
JST5000, line 19: I think it'd be useful to add a comment explaining what restrited networks means since I wouldn't know just from reading the name Ex. 
```suggestion
// Sidechains are expected to have network IDs above this. 
// Mainnet and testnet are exceptions. More context: https://github.com/XRPLF/rippled/pull/4370
const RESTRICTED_NETWORKS = 1024
```

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -88,6 +97,42 @@ async function autofill<T extends Transaction>(
   return Promise.all(promises).then(() => tx)
 }
 
+async function setNetworkID(client: Client, tx: Transaction): Promise<void> {
+  if (client.networkID === undefined) {
+    await setNetworkIDAndBuildVersion(client)
+  }
```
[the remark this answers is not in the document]
JST5000: Also, given that not having the `NetworkID` when you need it can cause all future transactions to fail - it might be worth adding retry logic if you send a `server_info` request at the start of `Client.connect()` and it doesn't get a response. (If it does get a response without a `NetworkID` - that just means it's talking to mainnet or a similar network and does not need to retry).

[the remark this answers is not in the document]
ckniffen: I agree with this except that if server info fails i feel that you probably don't have a valid connection to the socket.  That should be the fastest call if not one of the fastest calls.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -88,6 +97,42 @@ async function autofill<T extends Transaction>(
...
+}
+
+async function setNetworkIDAndBuildVersion(client: Client): Promise<void> {
+  const response: ServerInfoResponse = await client.request({
+    command: 'server_info',
+  })
+  // eslint-disable-next-line no-param-reassign, require-atomic-updates -- param reassign is safe, no race condition
+  client.networkID = response.result.info.network_id ?? undefined
+  // eslint-disable-next-line no-param-reassign, require-atomic-updates -- param reassign is safe, no race condition
+  client.buildVersion = response.result.info.build_version
```
JST5000, line 133: This would have a race condition if the network could actually change on a transaction by transaction basis. 

If you were sending a transaction to a sidechain & then sending another transaction to mainnet, you could end up with one of those updating the networkID _while_ the other was also being autofilled, and thus have the wrong networkID since it's shared state. (That's not a real scenario though because as it stands a Client only connects to one network)

(Setting the value once on `connect` fixes this - just noting for awareness :) )

pdp2121: I see what you meant but if a `client` connect to sidechain then send a transaction to mainnet then the networkId will be incorrect anyway (which is not a real scenario with the current setup anyway)

jonathanlei (approved):

justinr1234:
@pdp2121 can you merge main into your branch

packages/xrpl/src/sugar/autofill.ts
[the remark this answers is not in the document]
mvadari: Note: `server_info` will still return a `NetworkID` even if this feature hasn't been added to the network yet, and/or the `NetworkID` doesn't need to be included in the transaction (like with testnet).
You can see that with the current Testnet: https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Fs.altnet.rippletest.net%3A51233%2F&req=%7B%22id%22%3A1%2C%22command%22%3A%22server_info%22%7D

packages/xrpl/test/createMockRippled.ts
```diff
@@ -109,6 +128,8 @@ export default function createMockRippled(port: number): MockedWebSocketServer {
           mock.testCommand(conn, request)
         } else if (request.command in mock.responses) {
           conn.send(createResponse(request, mock.getResponse(request)))
+        } else if (request.command === 'server_info') {
+          server_info(conn, request)
```
mvadari, line 132: The mocks need to be set in the test itself, not here - this is going to add the mock to every test that uses `server_info`.

packages/xrpl/test/integration/onConnect.test.ts
```diff
@@ -12,13 +12,11 @@ describe('on handlers', function () {
     'on connect',
     async () => {
       const client = new Client(serverUrl)
-      return new Promise<void>(function (resolve) {
-        client.on('connected', function () {
-          client.removeAllListeners()
-          client.disconnect().then(() => resolve())
-        })
-        client.connect()
+      client.on('connected', function () {
```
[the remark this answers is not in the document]
pdp2121: It is because once connected client.connect() also calls server_info to retrieve networkId and buildVersion, which would takes longer than client.disconnect(), hence will cause the error since client has already been disconnected while connect() is still running so I have to move client.disconnected() outside of the on connected function

packages/xrpl/src/utils/isEarlierRippledVersion.ts
```diff
@@ -0,0 +1,66 @@
+/**
+ * Determines whether a rippled version (source) is earlier than another (target).
+ *
+ * @param source - The source rippled version.
+ * @param target - The target rippled version.
+ * @returns true if source is earlier, false otherwise.
+ */
+// eslint-disable-next-line max-lines-per-function, max-statements -- Disable for this utils functions.
+export default function isEarlierRippledVersion(
```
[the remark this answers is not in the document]
pdp2121: I made it a separate util in case there are features in the future that only apply to a certain build version or later similar to this one. It can certainly be a helper function inside autofill as well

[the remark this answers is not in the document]
pdp2121: Since we are going to remove buildVersion logic anyway then I think we should move this inside `autofill`

packages/xrpl/test/createMockRippled.ts
pdp2121: Agree. I can just add a `server_info` response to setupClient on connected instead of doing this.

packages/xrpl/test/mockRippledTest.test.ts
```diff
@@ -22,7 +22,7 @@ describe('mock rippled tests', function () {
     }
 
     await assertRejects(
-      testContext.client.request({ command: 'server_info' }),
+      testContext.client.request({ command: 'account_info' }),
```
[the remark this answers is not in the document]
pdp2121: I added a response for `server_info` in `setupClient()` so needed to switch to another command here for the rejection

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -252,6 +256,9 @@ export function validateBaseTransaction(common: Record<string, unknown>): void {
   ) {
     throw new ValidationError('BaseTransaction: invalid TxnSignature')
   }
+  if (common.NetworkID !== undefined && typeof common.NetworkID !== 'number') {
```
mvadari, line 259: There should be a unit test for this

packages/xrpl/test/setupClient.ts
```diff
@@ -30,6 +31,10 @@ async function setupMockRippledConnection(
     // We must have an error listener attached for reconnect errors
   })
 
+  context.client.on('connected', () => {
+    context.mockRippled?.addResponse('server_info', rippled.server_info.normal)
```
mvadari, line 35: Note: this fixture does not include a network ID, but all networks other than mainnet include a network ID.

packages/xrpl/test/client/autofill.test.ts
```diff
@@ -32,17 +34,112 @@ describe('client.autofill', function () {
...
+    testContext.mockRippled!.addResponse('ledger', rippled.ledger.normal)
+
+    const txResult = await testContext.client.autofill(tx)
+
+    assert.strictEqual(txResult.NetworkID, 1025)
+  })
+
+  it('ignores network ID if > 1024 but version is earlier than 1.11.0', async function () {
+    testContext.client.networkID = 1025
+    testContext.client.buildVersion = '1.10.1'
```
mvadari, line 88: Is it possible to write this test to use a different `server_info` mock instead of changing the values by hand?

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -88,6 +95,84 @@ async function autofill<T extends Transaction>(
   return Promise.all(promises).then(() => tx)
 }
 
+// eslint-disable-next-line max-lines-per-function, max-statements -- Disable for this helper functions.
```
mvadari, line 98: nit:
```suggestion
// eslint-disable-next-line max-lines-per-function, max-statements -- Disable for this helper function.
```

mvadari (approved):
lgtm but I think @justinr1234 should review the tests before this is merged

packages/xrpl/src/client/index.ts
```diff
@@ -223,6 +235,8 @@ class Client extends EventEmitter {
 
     this.feeCushion = options.feeCushion ?? DEFAULT_FEE_CUSHION
     this.maxFeeXRP = options.maxFeeXRP ?? DEFAULT_MAX_FEE_XRP
+    this.networkID = undefined
```
justinr1234, line 238: you don't need to explicitly set this

packages/xrpl/src/client/index.ts
```diff
@@ -223,6 +235,8 @@ class Client extends EventEmitter {
...
+    this.buildVersion = undefined
```
justinr1234, line 239: you don't need to explicitly set this either

packages/xrpl/src/client/index.ts
```diff
@@ -588,7 +602,13 @@ class Client extends EventEmitter {
    * @category Network
    */
   public async connect(): Promise<void> {
-    return this.connection.connect()
+    return this.connection.connect().then(async () => {
```
justinr1234, line 605: I think this shouldn't fail if the network id call doesn't work. I'd suggest adding a `catch` to log an error but still allow the connection to succeed

packages/xrpl/src/sugar/autofill.ts
justinr1234: I feel the comment is still unclear. Please explain what "restricted" means in the comment explicitly.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -88,6 +95,84 @@ async function autofill<T extends Transaction>(
...
+
+  return false
+}
+
+function txNeedsNetworkID(client: Client): boolean {
+  if (
+    client.networkID !== undefined &&
+    client.networkID > RESTRICTED_NETWORKS
+  ) {
+    // transaction needs networkID if either the network is hooks testnet or build version is >= 1.11.0
```
justinr1234, line 160: Please add this to the jsdoc string above the function instead of inline in the function

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -88,6 +95,84 @@ async function autofill<T extends Transaction>(
   return Promise.all(promises).then(() => tx)
 }
 
+// eslint-disable-next-line max-lines-per-function, max-statements -- Disable for this helper function.
+function isEarlierRippledVersion(source: string, target: string): boolean {
```
justinr1234, line 99: Please add a jsdoc to this explaining its purpose and example usage

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -88,6 +95,26 @@ async function autofill<T extends Transaction>(
...
 }
 
+function txNeedsNetworkID(client: Client): boolean {
+  if (
+    client.networkID !== undefined &&
+    client.networkID > RESTRICTED_NETWORKS
+  ) {
+    // transaction needs networkID if either the network is hooks testnet or build version is >= 1.11.0
+    if (
+      (client.buildVersion &&
```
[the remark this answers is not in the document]
justinr1234: Can you also make an issue in github and link the url in the comment

packages/xrpl/test/client/autofill.test.ts
```diff
@@ -32,17 +51,108 @@ describe('client.autofill', function () {
...
+      LastLedgerSequence,
+    }
+    testContext.mockRippled!.addResponse('ledger', rippled.ledger.normal)
+
+    const txResult = await testContext.client.autofill(tx)
+
+    assert.strictEqual(txResult.NetworkID, 1025)
+  })
+
+  it('ignores network ID if > 1024 but version is earlier than 1.11.0', async function () {
```
justinr1234, line 102: Can you add comments explaining why this is a test

packages/xrpl/test/client/autofill.test.ts
```diff
@@ -32,17 +51,108 @@ describe('client.autofill', function () {
...
+      LastLedgerSequence,
+    }
+    testContext.mockRippled!.addResponse('ledger', rippled.ledger.normal)
+
+    const txResult = await testContext.client.autofill(tx)
+
+    assert.strictEqual(txResult.NetworkID, undefined)
+  })
+
+  it('overrides network ID if > 1024 and version is later than 1.11.0', async function () {
```
justinr1234, line 84: Can you add comments explaining why this is a test

packages/xrpl/test/client/autofill.test.ts
```diff
@@ -32,17 +51,108 @@ describe('client.autofill', function () {
...
+      LastLedgerSequence,
+    }
+    testContext.mockRippled!.addResponse('ledger', rippled.ledger.normal)
+
+    const txResult = await testContext.client.autofill(tx)
+
+    assert.strictEqual(txResult.NetworkID, undefined)
+  })
+
+  it('override network ID for hooks testnet', async function () {
```
justinr1234, line 138: Can you add comments explaining why this is a test. Also it should be `overrides` I believe

packages/xrpl/test/client/autofill.test.ts
```diff
@@ -15,13 +15,32 @@ import {
 } from '../setupClient'
 import { assertRejects } from '../testUtils'
 
+const NetworkID = 1025
 const Fee = '10'
 const Sequence = 1432
 const LastLedgerSequence = 2908734
+const HOOKS_TESTNET_ID = 21338
```
justinr1234, line 22: Is this ID set in stone and will never change?

packages/xrpl/test/integration/onConnect.test.ts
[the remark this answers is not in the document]
justinr1234: @pdp2121 what error is caused? Can you give more information?

This leads me to believe your code needs to be enhanced to handle the case where the client is disconnected after you've already called server_info --- or that when your call to server_info is ready to be made, the client could already be disconnected.

I'm curious to know the error before we call this resolved

justinr1234 (changes requested):

packages/xrpl/test/client/autofill.test.ts
pdp2121: @mvadari What do you think? I can use the faucet to retrieve the ID for hooks testnet but it would make autofill a bit slower on the first call. I think the ID for each networks should remain unchanged but I might be wrong

mvadari: I don't think it changes, but if they switch to a v4 or something at some point then it might change then.
Perhaps a better solution is some sort of force-network-id flag? Or just tell people that if they're using the hooks testnet they'll have to add the network ID in themselves.

packages/xrpl/test/integration/onConnect.test.ts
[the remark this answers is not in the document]
pdp2121: `xrpl:     DisconnectedError: WebSocket is not open: readyState 2 (CLOSING)
xrpl: 
xrpl:       161 |     ws.send(message, (error) => {
xrpl:       162 |       if (error) {
xrpl:     > 163 |         reject(new DisconnectedError(error.message, error))
xrpl:           |                ^
xrpl:       164 |       } else {
xrpl:       165 |         resolve()
xrpl:       166 |       }`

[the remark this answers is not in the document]
pdp2121: The error is as above. Talked with @JST5000 about this and we decided to move `disconnect()` outside of `on('connected'...)`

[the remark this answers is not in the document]
pdp2121: In practice we would not put `disconnect()` inside connected listener anyway so everything inside connected will resolve first and the issue would not happen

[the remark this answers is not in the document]
justinr1234: It sounds like this is a bug in your code is what I'm saying. Where is this error being thrown from? Is it inside the new code you added to the .then ?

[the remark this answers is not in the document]
pdp2121: @justinr1234 "connected" will emit when the websocket is connected to server, and after that the code execute the `server_info` call. For the previous test to work (`disconnect()` happens immediately on "connected"), the `server_info` call has to finish its execution before the websocket emits "connected" message. Are you suggesting we only emit "connected" after the `server_info` call finish its execution? @ckniffen @mvadari @JST5000 please let me know what you think as well. Thanks!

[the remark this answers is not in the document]
JST5000: I would suggest that we only emit "connected" after the `server_info` call finishes it's connection. My understanding is that as soon as `connected` finishes, we can expect the user to create transactions and expect them to work. Since the information from `server_info` will be required to `autofill` a transaction for some networks, I believe you should have to wait until that's done beforehand.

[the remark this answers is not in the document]
pdp2121: That makes sense. Yea I can work on that. Thanks @justinr1234 @JST5000

justinr1234 (approved):

justinr1234:
@pdp2121 can you update with main

packages/xrpl/test/integration/onConnect.test.ts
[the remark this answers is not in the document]
justinr1234: It’s my understanding that server info isn’t necessary and is more of an optional thing. Unless I’m mistaken.

I am not saying we should wait for server info. The opposite, I’m suggesting we should allow the current behavior to continue to work

PR #2217, opened 2023-02-17
[WIP] feature `Hooks`

dangell7:

Add Hooks Amendment Support. A full read me can be found [here](https://xrpl-hooks.readme.io/docs/sethook-transaction)

### Type of Change

This PR adds the `SetHook` transaction and the utility functions for the Hooks Amendment. It also makes changes to the definitions. Finally this PR is rebased off, , which is required on the Hook network.

- HookParameters: https://xrpl-hooks.readme.io/docs/parameters
- Namespaces: https://xrpl-hooks.readme.io/docs/namespaces
- HookGrants: https://xrpl-hooks.readme.io/docs/grants
- HookOn: https://xrpl-hooks.readme.io/docs/hookon-field

- [ ] Documentation Updates
- [ ] Release

## Test Plan

I plan to write the standard tests matching a similar transaction like SignerListSet. There are 2 additional functions that needs to be written. CreateCode takes a binary string compiled from a wasm file and HookOn is a Hash256 compiled of different transaction types. Any additional files should include a test.

- [ ] Models
- [ ] Utils
- [ ] Integration

### DIFF
--- a/packages/ripple-binary-codec/src/enums/definitions.json
+++ b/packages/ripple-binary-codec/src/enums/definitions.json
@@ -48,7 +48,11 @@
     "Child": -2,
     "Nickname": 110,
     "Contract": 99,
-    "GeneratorMap": 103
+    "GeneratorMap": 103,
+    "Hook": 72,
+    "HookState": 118,
+    "HookDefinition": 68,
+    "EmittedTxn": 69
   },
   "FIELDS": [
     [
@@ -319,6 +323,16 @@
         "isSerialized": true,
         "isSigningField": true,
         "type": "UInt16"
+      }
+    ],
+    [
+      "NetworkID",
+      {
+        "nth": 1,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "UInt32"
       }
     ],
     [
@@ -892,16 +906,6 @@
       }
     ],
     [
-      "HookOn",
-      {
-        "nth": 16,
-        "isVLEncoded": false,
-        "isSerialized": true,
-        "isSigningField": true,
-        "type": "UInt64"
-      }
-    ],
-    [
       "HookInstructionCount",
       {
         "nth": 17,
@@ -1145,6 +1149,16 @@
       "Amendment",
       {
         "nth": 19,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "Hash256"
+      }
+    ],
+    [
+      "HookOn",
+      {
+        "nth": 20,
         "isVLEncoded": false,
         "isSerialized": true,
         "isSigningField": true,
@@ -1415,6 +1429,16 @@
       "NFTokenBrokerFee",
       {
         "nth": 19,
+        "isVLEncoded": false,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "Amount"
+      }
+    ],
+    [
+      "HookCallbackFee",
+      {
+        "nth": 20,
         "isVLEncoded": false,
         "isSerialized": true,
         "isSigningField": true,
@@ -1802,6 +1826,16 @@
       }
     ],
     [
+      "HookNamespaces",
+      {
+        "nth": 5,
+        "isVLEncoded": true,
+        "isSerialized": true,
+        "isSigningField": true,
+        "type": "Vector256"
+      }
+    ],
+    [
       "Paths",
       {
         "nth": 1,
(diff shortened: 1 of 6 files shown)


develoQ:
Not yet added to the documentation, no need to add an Invoke transaction?

dangell7:
> Not yet added to the documentation, no need to add an Invoke transaction?

Invoke is done through `HookOn` unless I misunderstood you. There are a few helper functions I will add to this PR hopefully tonight or tomorrow.

dangell7:
@develoQ Here is an example of how to use the sdk.

https://github.com/Transia-RnD/xls-playground/blob/main/ts/test/hooks.test.ts

khancode:
@dangell7 Great work on this so far! One suggestion is to add a test case for adding `HookParameters` to non-SetHook transactions since this is now supported in testnet v3.

For example:
```TypeScript	
{
   "NetworkID":21338,
   "TransactionType":"Payment",
   "Account":"rERZinA8bgHKx1Uu7goWwzJgRFiBx2o3Xg",
   "Amount":"10000000",
   "Destination":"r3GC3EBBCMbp17pi3NjmfVPAUM7UJuyWiE",
   "DestinationTag":160,
   "HookParameters":[
      {
         "HookParameter":{
            "HookParameterName":"ABCDEF12",
            "HookParameterValue":"12345678"
         }
      }
   ],
   "Fee":"50000"
}

File: packages/ripple-binary-codec/src/enums/definitions.json [hunk 1]
```diff
@@ -2211,10 +2249,12 @@
...
-    "temUNCERTAIN": -265,
-    "temUNKNOWN": -264,
-    "temSEQ_AND_TICKET": -263,
-    "temBAD_NFTOKEN_TRANSFER_FEE": -262,
+    "temHOOK_DATA_TOO_LARGE": -265,
+    "temHOOK_REJECTED": -264,
+    "temUNCERTAIN": -263,
+    "temUNKNOWN": -262,
+    "temSEQ_AND_TICKET": -261,
+    "temBAD_NFTOKEN_TRANSFER_FEE": -260,
```
mvadari, line 2257: This looks like a breaking change...

packages/xrpl/src/models/methods/serverInfo.ts
```diff
@@ -136,6 +136,10 @@ export interface ServerInfoResponse extends BaseResponse {
        * overall network's load factor.
        */
       load_factor?: number
+      /**
+       * The network id of this server
+       */
+      network_id?: number
```
mvadari, line 142: Why are all the network ID changes in here as well?

packages/xrpl/src/sugar/getFeeXrp.ts
```diff
@@ -43,3 +43,22 @@ export default async function getFeeXrp(
   // Round fee to 6 decimal places
   return new BigNumber(fee.toFixed(NUM_DECIMAL_PLACES)).toString(BASE_10)
 }
+
+/**
```
mvadari, line 47: Shouldn't this be a part of `autofill` instead of a separate method?

packages/xrpl/src/utils/hooks.ts
```diff
@@ -0,0 +1,128 @@
+/**
```
mvadari, line 1: What are the uses of these utils?

packages/xrpl/src/models/methods/serverInfo.ts
dangell7: This was added in because the hooks amendment includes the network id functionality.

packages/xrpl/src/sugar/getFeeXrp.ts
dangell7: Originally I was under the impression the `tx_blob` submitted to the fee rpc needed to be signed. Signing it in the autofill created a circular dependency. It does not need to be signed, only encoded. So I could move this into the `autofill`. However, I believe this does replace the current `Fee` request so if the current fee request is a sugar, then this should also be a sugar as well.

I can update the set fee in auto fill with this new estimate fee functionally.

packages/xrpl/src/utils/hooks.ts
dangell7: The purpose of these utils can be found here.

https://github.com/Transia-RnD/xls-playground/blob/main/ts/test/hooks.test.ts

I also have them in [this](https://github.com/Transia-RnD/xrpl-helpers/blob/main/ts/src/hooks/hooks.ts) repo here if we  dont want them in the sdk.

packages/xrpl/src/sugar/getFeeXrp.ts
mvadari: We shouldn't update the current functionality, because this new feature is not live yet (it's part of the Hooks amendment, right?), and we want this library to be useable on mainnet right now.

packages/xrpl/src/models/methods/serverInfo.ts
mvadari: We already have a separate PR for that (#2216), it shouldn't also be in this PR.

ckniffen: It becomes tricky because network_id is required for hooks testnet.  #2216 will need to be merged first.  Also the network_id code will not be hidden behind an amendment.

mvadari:
@dangell7 did you mean to put all these commits in this branch?

Issue #2212, opened 2023-02-16
NFTokenMint transaction does not support empty string in URI


crtag:
**Network: wss://s.altnet.rippletest.net:51233**

The following transaction body
```
{
        "TransactionType": 'NFTokenMint',
        "Account": wallet.classicAddress,
        "URI": xrpl.convertStringToHex(''),
        "Flags": 8,
        "TransferFee": 0,
        "NFTokenTaxon": 0 
}
```

causes the following error stack:
```
XrplError: The latest ledger sequence XXXXX is greater than the transaction's LastLedgerSequence (YYYYYY).
Preliminary result: temMALFORMED
    at \xrpl\node_modules\xrpl\dist\npm\sugar\submit.js:65:19
    at Generator.next (<anonymous>)
    at fulfilled (\xrpl\node_modules\xrpl\dist\npm\sugar\submit.js:5:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  data: undefined
}
```

Sample code:
```
const xrpl = require('xrpl');

const net = 'wss://s.altnet.rippletest.net:51233';
const opSeed = '';

const wallet = xrpl.Wallet.fromSeed(opSeed);
const client = new xrpl.Client(net);

client.connect().then(() => {
    console.log(`Connected to ${net}`)
    console.log(`Using address ${wallet.classicAddress}`)
    const transactionBlob = {
        "TransactionType": 'NFTokenMint',
        "Account": wallet.classicAddress,
        "URI": xrpl.convertStringToHex(''),
        "Flags": 8,
        "TransferFee": 0,
        "NFTokenTaxon": 0 
    };
      
    client.submitAndWait(transactionBlob, {wallet})
    .then(tx => console.log(tx))
    .catch(err => console.error(err))
    .finally(() => client.disconnect());

});
```

JST5000:
Thanks for sharing this, I was able to reproduce it locally. (It also seems that even `NFT: undefined` isn't allowed because it doesn't encode/decode to the same thing.) 

My initial impression would be that the client library should interpret an empty string URI as the same as the NFT not including the URI field at all. 

I believe rippled probably doesn't accept empty URI because it's a waste of space. (It's allocating space for a URI, but then not using it to store anything).

crtag:
> I believe rippled probably doesn't accept empty URI because it's a waste of space. (It's allocating space for a URI, but then not using it to store anything).

[...]

PR #2218, opened 2023-02-17
Issue 2212. Improve NFToken.URI handling for empty and undefined values

crtag:
https://github.com/XRPLF/xrpl.js/issues/2212
Add NFToken.URI validation for empty string
Improve NFToken.URI handling for its undefined value

### Context of Change
The discussion here: https://github.com/XRPLF/xrpl.js/issues/2212

### Type of Change


packages/xrpl/src/Wallet/index.ts
```diff
@@ -466,6 +466,22 @@ class Wallet {
       txCopy.URI = txCopy.URI.toUpperCase()
     }
 
+    if (
+      txCopy.TransactionType === 'NFTokenMint' &&
+      typeof txCopy.URI === 'string' &&
+      txCopy.URI === ''
+    ) {
+      throw new ValidationError('NFTokenMint: URI must not be empty string')
+    }
```
mvadari, line 475: This check shouldn't be happening here - I believe `validate` is already run on the transaction before this point.

packages/xrpl/src/Wallet/index.ts
```diff
@@ -466,6 +466,22 @@ class Wallet {
...
+
+    if (
+      txCopy.TransactionType === 'NFTokenMint' &&
+      txCopy.hasOwnProperty('URI') &&
+      txCopy.URI === undefined
+    ) {
+      delete txCopy.URI
+    }
```
mvadari, line 483: See my comment in #2212 about making this more general.

packages/xrpl/src/sugar/submit.ts
```diff
@@ -217,6 +217,10 @@ async function getSignedTx(
         (decode(transaction) as unknown as Transaction)
       : transaction
 
+  if (tx.TransactionType === 'NFTokenMint' && tx.hasOwnProperty('URI') && tx.URI == undefined) {
+    delete tx.URI
+  }
+
```
mvadari, line 223: This is unneeded if the logic is already happening in `sign`.

packages/xrpl/src/Wallet/index.ts
JST5000: Link to it for easier reference: https://github.com/XRPLF/xrpl.js/issues/2212#issuecomment-1434942602

packages/xrpl/src/models/transactions/transaction.ts
```diff
@@ -213,7 +213,7 @@ export function validate(transaction: Record<string, unknown>): void {
   if (
     !isEqual(
       decode(encode(tx)),
-      omitBy(tx, (value) => value == null),
+      decode(encode(omitBy(tx, (value) => value == null))),
```
JST5000, line 216: I think the point of this line of code before was that it didn't do `decode(encode(...))` - it needs to show the original is equivalent to what you're hoping to get. Maybe the change you wanted was to do this to the 1st value in this comparison, and then leave the 2nd value as `omitBy(tx, (value => value == null))` ?

crtag: @JST5000 to be honest I think this chunk of code should not be here, but rather a part of a test suite. Doesn't make sense as its intentions are to validate the logic, not the input data.

The way to overcome might be something like this you suggest?

```
  if (
    !isEqual(
      tx,
      omitBy(tx, (value) => value == null),
    )
  ) {
    throw new ValidationError(`Invalid Transaction: ${tx.TransactionType}`)
  }
```

mvadari: I'm also thinking that it may just be worth removing the `decode(encode(tx))` block - I'm not sure what value it adds.

packages/xrpl/src/models/transactions/transaction.ts
```diff
@@ -212,7 +211,7 @@ export function validate(transaction: Record<string, unknown>): void {
 
   if (
     !isEqual(
-      decode(encode(tx)),
+      tx,
```
JST5000, line 214: I think by remove the `decode(encode(tx))` block, Mayukha was referring to rmoving lines 212 -219 - @mvadari is that correct? (I think that makes sense to me as well if that's what you meant - we check this anyway when we submit, and the error isn't very verbose here)
```
if (
   !isEqual(
...
    throw new ValidationError(`Invalid Transaction: ${tx.TransactionType}`)
  }
```

crtag: @JST5000 the entire block as we noted above you mean?
Also, for failing checks - somehow I can not reproduce the same thing in the local environment (win10)

mvadari: > I think by remove the `decode(encode(tx))` block, Mayukha was referring to removing lines 212 -219 - @mvadari is that correct? (I think that makes sense to me as well if that's what you meant - we check this anyway when we submit, and the error isn't very verbose here)

Yes, that is what I meant.

crtag: @mvadari sorted it and linter + unit test failing due to decode/encode

crtag:
@mvadari @JST5000 anything else to be taken care of?

packages/xrpl/src/models/transactions/NFTokenMint.ts
```diff
@@ -86,8 +86,11 @@ export interface NFTokenMint extends BaseTransaction {
    *
    * This field must be hex-encoded. You can use `convertStringToHex` to
    * convert this field to the proper encoding.
+   *
+   * This field must not be an empty string. Omit it from the transaction or
+   * set to `undefined` value if you do not use it.
    */
-  URI?: string
+  URI?: string | null
```
mvadari, line 93: I don't think the type needs to be changed here - according to [this StackOverflow answer](https://stackoverflow.com/questions/65912702/can-an-optional-property-value-be-null), `?` is equivalent to adding ` | undefined` to a type.

crtag: This is done because of explicit null in the unit test as we have added it as something that dev can pass over as a valid value.

JST5000:
Sorry didn't have time to review today, I'll review on Monday.

crtag:
Quick question - the failing browser test seems like some sort of timeout on devnet?

ckniffen:
> Quick question - the failing browser test seems like some sort of timeout on devnet?

It looks like that.

JST5000 (approved):
LGTM

justinr1234 (approved):

jonathanlei (approved):

JST5000:
@crtag Thanks for making this PR & following up with it :)

PR #2220, opened 2023-02-17
Remove casting linting error from xrpl package

JST5000:

This linting error seemed extraneous more than it was useful.

### Context of Change

Part of #2205's discussion to make it more readable. 

### Type of Change

## Before / After

* Removed the @typescript-eslint/consistent-type-assertions rule from the xrpl package since type casting is a normal operation there, especially when dealing with `unknown` types like `PreviousFields`. 
* Also one nit rewording from "both" -> "All" since it was talking about 3 rules instead of 2.

## Test Plan

Linter passes

### DIFF
--- a/packages/xrpl/.eslintrc.js
+++ b/packages/xrpl/.eslintrc.js
@@ -62,6 +62,12 @@ module.exports = {
 
     'tsdoc/syntax': 'off',
     'jsdoc/require-description-complete-sentence': 'off',
+
+    // Type assertions are occassionally necessary when dealing with very complex types
+    '@typescript-eslint/consistent-type-assertions': [
+      'warn',
+      { assertionStyle: 'as' },
+    ],
   },
   overrides: [
     {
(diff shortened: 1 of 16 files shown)


justinr1234:
@JST5000 not sure I agree with removing this overall. I think in general we should have it in the codebase as it prevents issues. You should be required to explicitly explain why you're going against typescript (or find the proper solution). I think oftentimes, especially typescript beginners, people use assertions as a way to not have to deal with understanding the typing.

ckniffen (approved):

mvadari:
> @JST5000 not sure I agree with removing this overall. I think in general we should have it in the codebase as it prevents issues. You should be required to explicitly explain why you're going against typescript (or find the proper solution). I think oftentimes, especially typescript beginners, people use assertions as a way to not have to deal with understanding the typing.

I agree. It makes more sense to me to just disable the rules in the few places where it's necessary.

JST5000:
Fair enough :) - Closing since we've decided it's better to call out exceptions rather than removing the rule.

PR #2226, opened 2023-02-28
doc: Make the docs more robust with examples

justinr1234:

Improve documentation at https://js.xrpl.org by adding examples and tutorials.

### Type of Change

- [X] Documentation Updates
- [ ] Release


packages/xrpl/src/sugar/autofill.ts
```diff
@@ -23,6 +23,43 @@ interface ClassicAccountAndTag {
...
+ * const { Client } = require('xrpl')
+ *
+ * const client = new Client('wss://s.altnet.rippletest.net:51233')
+ *
+ * async function createAndAutofillTransaction() {
+ *   const transaction = {
+ *     TransactionType: 'Payment',
+ *     Account: 'rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh',
+ *     Destination: 'r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59',
+ *     Amount: '10'
```
justinr1234, line 38: Change this to drops

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -20,15 +20,103 @@ const INTERVAL_SECONDS = 1
...
  * @example
- * ```typescript
+ *
+ * Example 1: Fund a manually random generated wallet
+ * const { Client, Wallet } = require('xrpl')
+ *
+ * const client = new Client('wss://s.altnet.rippletest.net:51233')
+ *
+ * async function fundRandomWallet() {
+ *   const newWallet = new Wallet()
```
justinr1234, line 34: Change to `Wallet.generate()`

jonathanlei: are those comments the changes you plan to make? or you are asking us if you should make it

packages/xrpl/src/sugar/submit.ts
```diff
@@ -52,12 +52,73 @@ async function submit(
...
  *
+ * @example
+ *
+ * ```ts
+ * const { Client, Wallet } = require('xrpl')
+ * const client = new Client('wss://s.altnet.rippletest.net:51233')
+ *
+ * async function submitTransaction() {
+ *   const senderWallet = Wallet.generate()
+ *   const recipientWallet = Wallet.generate()
```
JST5000, line 63: As is this code won't work because you're trying to send money from an account which is newly generated, and doesn't have any funds yet. 

While for the presentation I think it's valuable to walk through the `Wallet.generate()` differences, I don't think we should do that in the core submit flow. (As there people just want it to work)

```suggestion
 *   const senderWallet = client.fundWallet()
 *   const recipientWallet = client.fundWallet()
```

packages/xrpl/src/sugar/submit.ts
```diff
@@ -52,12 +52,73 @@ async function submit(
...
+ *     console.error(`Failed to submit transaction: ${error}`)
+ *   }
+ * }
+ *
+ * submitTransaction()
+ * ```
+ *
+ * In this example, we create two new random wallets using the generate() method of the Wallet class,
+ * one for the sender and one for the recipient. We extract the publicKey property from the generated wallet
+ * objects to get the sender and recipient public keys.
```
JST5000, line 87:
```suggestion
 * In this example, we create two new test accounts, one for the sender and one for the recipient.
 * We then use them as the account sending the payment, and the recipient of the funds.
```

packages/xrpl/src/sugar/submit.ts
```diff
@@ -52,12 +52,73 @@ async function submit(
...
+ *     TransactionType: 'Payment',
+ *     Account: senderWallet.publicKey,
+ *     Destination: recipientWallet.publicKey,
+ *     Amount: '10'
+ *   }
+ *
+ *   try {
+ *     await client.autofill(transaction)
+ *     const signedTransaction = await Wallet.sign(transaction)
+ *     const result = await client.submitAndWait(signedTransaction)
```
JST5000, line 75: I'd recommend replacing this with the 1-line syntax because that's the recommended way to use this library imo. 
```suggestion
 *     await client.submitAndWait(signedTransaction, { wallet: senderWallet })
```

JST5000: (This will also cause a change in the below explanation of what this code does)

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -23,6 +23,43 @@ interface ClassicAccountAndTag {
...
+ * In this example, we create a transaction object that represents a payment of 10 XRP
+ * from the `rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh` account to the `r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59` account.
+ * We do not set the Fee or Sequence properties of the transaction object since `autofill()` will do that for us.
+ *
+ * We then pass the transaction object to the `autofill()` method of the `Client` class. The `autofill()` method will
+ * attempt to calculate the appropriate fee for the transaction based on the current state of the XRP Ledger and the sequence
+ * based on the last validated ledger specific to this account. It will also convert the X-Addresses to classic addresses.
+ *
+ * If the autofill is successful, the function logs the autofilled transaction object to the console. If there's an error
+ * autofilling the transaction, the function logs an error message to the console using console.error().
```
JST5000, line 61: I think the comments should explain the overall flow, and specific parts which are hard to understand, but don't need to repeat everything which the code is doing. 

```suggestion
 * Autofill helps fill in fields which should be included in a transaction, but can be determined automatically
 * such as `LastLedgerSequence` and `Fee`. If you override one of the fields `autofill` changes, your explicit
 * values will be used instead. 
```

packages/xrpl/src/client/index.ts
```diff
@@ -564,6 +564,50 @@ class Client extends EventEmitter {
...
+   *     console.log(`Failed to connect to server: ${error}`)
+   *   }
+   * }
+   *
+   * connectToServer()
+   * ```
+   *
+   * In this example, the connectToServer() function uses the async/await syntax to wait for the connection to
+   * be established before logging a message to the console.
+   *
```
JST5000, line 610: Simplifying to just use `await` syntax since that's preferred, and only commenting on the usage parts instead of re-describing the code.

```suggestion
   * Client.connect() establishes a connection between a Client object and the server.
   *
   * ```ts
   * const { Client } = require('xrpl')
   * const client = new Client('wss://s.altnet.rippletest.net:51233')
   *   try {
   *     await client.connect()
   *     console.log('Connected to server')
   *   } catch (error) {
   *     console.log(`Failed to connect to server: ${error}`)
   *   } 
   * ```
   * If you open a client connection, be sure to close it with `await client.disconnect()` 
   * before exiting your application.
```

packages/xrpl/src/Wallet/fundWallet.ts
justinr1234: Making changes

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -20,15 +20,103 @@ const INTERVAL_SECONDS = 1
...
+ * The walletData object returned by the generate() method contains the publicKey and secret values for the new wallet.
+ *
+ * The fundWallet() method of the Client class is then called with the publicKey value of the new wallet as the first
+ * argument, and an object that specifies the amount of XRP to send in the xrpAmount parameter. In this case, the
+ * xrpAmount parameter is set to '10', which is equivalent to 10 XRP.
+ *
+ * If the transaction is successful, the function logs a message to the console with the wallet public key and balance of the
+ * new wallet. If there's an error funding XRP to the wallet, the function logs an error message to the console using
+ * console.error().
+ *
```
JST5000, line 57: Similar to previous, this is the happy path we want them to follow & updated the description to include when you'd want to use the optional parameters. For this function in particular I think it's ok not to have the error handling because if the account doesn't get funded, everything else will break almost guaranteed.

```suggestion
 * Example 1: Fund a randomly generated wallet
 * const { Client, Wallet } = require('xrpl')
 *
 * const client = new Client('wss://s.altnet.rippletest.net:51233')
 * await client.connect()
 * const wallet = await client.fundWallet()
 *
 * Under the hood, this will use `Wallet.generate()` to create a new random wallet, then ask a testnet faucet
 * To send it XRP on ledger to make it a real account. If you'd like, you can also re-fill an existing account by
 * Passing in a Wallet you already have. 
```

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -20,15 +20,103 @@ const INTERVAL_SECONDS = 1
...
+ *       faucetPath: '/accounts'
+ *     })
+ *     console.log(`Sent 10 XRP to wallet: ${address}, balance: ${balance} XRP`)
+ *   } catch (error) {
+ *     console.error(`Failed to fund wallet: ${error}`)
+ *   }
+ * }
+ *
+ * fundWalletWithCustomfaucetHost(walletAddress)
+ * ```
```
JST5000, line 89: Updating with the proper types for the input Wallet and output of fundWallet.

```suggestion
 * Example 2: Fund wallet using a custom faucet host and known wallet address
 *
 * `fundWallet` will try to infer the url of a faucet API from the network your client is connected to. 
 * There are hardcoded default faucets for popular test networks like testnet and devnet. 
 * However, if you're working with a newer or more obscure network, you may have to specify the faucetHost
 * And faucetPath so `fundWallet` can ask that faucet to fund your wallet.
 * 
 * ```ts
 * const newWallet = Wallet.generate()
 * const { balance, wallet  } = await client.fundWallet(newWallet, {
 *       amount: '10',
 *       faucetHost: 'https://custom-faucet.example.com',
 *       faucetPath: '/accounts'
 *     })
 *     console.log(`Sent 10 XRP to wallet: ${address} from the given faucet. Resulting balance: ${balance} XRP`)
 *   } catch (error) {
 *     console.error(`Failed to fund wallet: ${error}`)
 *   }
 * }
 * ```
```

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -20,15 +20,103 @@ const INTERVAL_SECONDS = 1
...
+ *
+ * In this example, we call the `fundWallet()` method with no parameters to generate a new
+ * wallet and fund it with the XRP Ledger Testnet. The wallet variable contains an object that has publicKey
+ * and secret properties for the new wallet, and the balance variable contains the balance of the new wallet.
+ *
+ * If the transaction is successful, the function logs a message to the console with the publicKey value of the
+ * new wallet and the balance of the wallet. If there's an error funding XRP to the wallet, the function logs an
+ * error message to the console using console.error().
+ *
```
JST5000, line 119: Recommended moving this example to the top so they can find the happy path first, instead of starting with all the optional parameters.
```suggestion
```

packages/xrpl/src/Wallet/index.ts
```diff
@@ -121,7 +121,31 @@ class Wallet {
...
+   * const walletData = newWallet.generate()
+   *
+   * console.log(walletData.publicKey)
+   * console.log(walletData.secret)
+   * ```
+   *
+   * This will create a new wallet object and log the publicKey and secret values to the console.
+   * It's important to note that the secret value should be kept confidential and never shared with anyone.
+   * Anyone with access to the secret value can sign transactions and transfer XRP from the wallet, so it's
+   * crucial to keep it secure.
```
JST5000, line 148: The description of what a Wallet is seems better left to the Wallet class itself. Also I removed the logging part of the example because as the note suggested, we don't want them exposing their keys, and logging private keys to console is bad practice. If they want to see what just happened I think they'll be able to figure that part out.

```suggestion
   * `generate()` creates a new random Wallet. In order to make this a valid account on ledger, you must 
   * Send XRP to it. On test networks that can be done with "faucets" which send XRP to any account which asks
   * For it. You can call `client.fundWallet()` in order to generate credentials and fund the account on test networks. 
   *
   * @example
   * ```ts
   * const { Wallet } = require('xrpl')
   * const wallet = Wallet.generate()
   * ```
```

packages/xrpl/src/Wallet/index.ts
```diff
@@ -300,6 +324,58 @@ class Wallet {
...
+   * transaction object including `tx_blob` and `hash` to the console.
+   *
+   * `tx_blob` is a binary representation of a transaction on the XRP Ledger. It's essentially a byte array
+   * that encodes all of the data necessary to execute the transaction, including the source address, the destination
+   * address, the amount, and any additional fields required for the specific transaction type.
+   *
+   * `hash` is a unique identifier that's generated from the signed transaction data on the XRP Ledger. It's essentially
+   * a cryptographic digest of the signed transaction blob, created using a hash function. The signed transaction hash is
+   * useful for identifying and tracking specific transactions on the XRP Ledger. It can be used to query transaction
+   * information, verify the authenticity of a transaction, and detect any tampering with the transaction data.
```
JST5000, line 377: Updated example to use the proper type signatures for `sign` & initializing a Wallet.

```suggestion
   * ```ts
   * const { Client, Wallet } = require('xrpl')
   * const client = new Client('wss://s.altnet.rippletest.net:51233')
   *
   * async function signTransaction() {
   *   await client.connect()
   *   const { balance: balance1, wallet: wallet1 } = client.fundWallet()
   *   const { balance: balance2, wallet: wallet2 } = client.fundWallet()
   *
   *   const transaction = {
   *     TransactionType: 'Payment',
   *     Account: wallet1.address,
   *     Destination: wallet2.address,
   *     Amount: '10'
   *   }
   *
   *   try {
   *     await client.autofill(transaction)
   *     const { tx_blob: signed_tx_blob, hash} = await wallet1.sign(transaction)
   *     console.log(signed_tx_blob)
   *   } catch (error) {
   *     console.error(`Failed to sign transaction: ${error}`)
   *   }
   *   const result = await client.submit(signed_tx_blob)
   *   await client.disconnect()
   * }
   *
   * signTransaction()
   * ```
   * In order for a transaction to be validated, it must be signed by the account sending the transaction to prove
   * That the owner is actually the one deciding to take that action. 
   * 
   * In this example, we created, signed, and then submitted a transaction to testnet. You may notice that the
   * Output of `sign` includes a `tx_blob` and a `hash`, both of which are needed to submit & verify the results.
   * Note: If you pass a `Wallet` to `client.submit` or `client.submitAndWait` it will do signing like this under the hood.
   * 
   * `tx_blob` is a binary representation of a transaction on the XRP Ledger. It's essentially a byte array
   * that encodes all of the data necessary to execute the transaction, including the source address, the destination
   * address, the amount, and any additional fields required for the specific transaction type.
   *
   * `hash` is a unique identifier that's generated from the signed transaction data on the XRP Ledger. It's essentially
   * A cryptographic digest of the signed transaction blob, created using a hash function. The signed transaction hash is
   * Useful for identifying and tracking specific transactions on the XRP Ledger. It can be used to query transaction
   * Information, verify the authenticity of a transaction, and detect any tampering with the transaction data.
```

packages/xrpl/src/client/index.ts
```diff
@@ -564,6 +564,22 @@ class Client extends EventEmitter {
...
+   * ```ts
+   * const { Client } = require('xrpl')
+   * const client = new Client('wss://s.altnet.rippletest.net:51233')
+   *   try {
+   *     await client.connect()
+   *     console.log('Connected to server')
+   *   } catch (error) {
+   *     console.log(`Failed to connect to server: ${error}`)
+   *   }
```
JST5000, line 579: Simplifying to just the await because generally we don't presecribe having try / catch over it in example code.
```suggestion
   * await client.connect()
```

JST5000, line 579: Recommend re-using the disconnect example because it highlights that you should have both with real code.
```suggestion
   * const { Client } = require('xrpl')
   * const client = new Client('wss://s.altnet.rippletest.net:51233')
   * await client.connect()
   * // do something with the client
   * await client.disconnect()
```

packages/xrpl/src/sugar/submit.ts
```diff
@@ -52,12 +52,70 @@ async function submit(
...
+ * transaction blob is passed to the `submitAndWait()` method of the `Client` class.
+ *
+ * The `submitAndWait()` method submits the signed transaction blob to the XRP Ledger and waits for the transaction
+ * to be included in a validated ledger before resolving or rejecting the returned promise. If the transaction is
+ * successfully validated and included in a ledger, the promise resolves with the transaction result object. If there's
+ * an issue submitting or validating the transaction, the promise rejects with an error.
+ *
+ * If the promise resolves, the function logs a success message to the console that includes the transaction hash. If the
+ * promise rejects, the function logs an error message to the console using console.error().
+ *
```
JST5000, line 103:
```suggestion
 * In this example we submit a payment transaction between two newly created testnet accounts. 
 * 
 * Under the hood, `submit` will call `client.autofill` by default, and because we've passed in a `Wallet` it
 * Will also sign the transaction for us before submitting the signed transaction binary blob to the ledger. 
 * 
 * This is similar to `submitAndWait` which does all of the above, but also waits to see if the transaction has been validated.
```

jonathanlei (approved):

pdp2121 (approved):

PR #2247, opened 2023-03-28
Update getNFTokenID to properly handle binary blob

JST5000:

Update `getNFTokenID` to allow for binary blobs since that's a valid return from a `tx` request.

### Context of Change

When updating this test to work with the new standalone container, I realized that I had misunderstood the return type and it was more complicated than just the `meta` in JSON format.

Must be merged in after #2223 as it relies on that docker container for tests.

### Type of Change

## Before / After

* Migrates the integration tests to use standalone node instead of testnet
* Expands the function to accept a binary blob as well

## Test Plan

CI Passes


ckniffen (approved):

packages/xrpl/src/utils/getNFTokenID.ts
```diff
@@ -15,19 +16,36 @@ interface NFToken {
   }
 }
 
+/**
+ * Ensures that the metadata is in a deserialized format to parse.
+ *
+ * @param meta - the metadata from a `tx` method call. Can be in json format or binary format.
+ * @returns the metadata in a deserialized format.
+ */
+export function ensureDecodedMeta(
```
ckniffen, line 25: nit: this feels like a more generic utility.

mvadari, line 25: Why is this exported? It should just be a helper function, right?

JST5000: I needed it in my test and thought others might as well, but I realize now that most folks will know exactly which output they have (either a binary blob or json format) and so can explicitly do the casting themselves. Changed so it's not exported.

JST5000:
Test

JST5000:
Test 2

mvadari:
This PR needs to be rebased before it can be reviewed

JST5000:
> This PR needs to be rebased before it can be reviewed

I may not get to this for a bit, working on other things which have to be completed this week. Will ping again once this PR is ready to actually be reviewed.

ckniffen (approved):

packages/xrpl/src/utils/getNFTokenID.ts
```diff
@@ -15,19 +16,36 @@ interface NFToken {
...
  * Gets the NFTokenID for an NFT recently minted with NFTokenMint.
  *
- * @param meta - Metadata from the response to submitting an NFTokenMint transaction.
+ * @param meta - Metadata from the response to submitting and waiting for an NFTokenMint transaction or from a `tx` method call.
  * @returns The NFTokenID for the minted NFT.
  * @throws if meta is not TransactionMetadata.
  */
 export default function getNFTokenID(
-  meta: TransactionMetadata,
+  meta: TransactionMetadata | string | undefined,
```
mvadari, line 43: Why is this allowed to be `undefined`? That seems erroneous.

ckniffen: Agreed.  Undefined is strange.

JST5000: It's so that it can be used directly on the output of TxResponse.meta which is `TransactionMetadata | string | undefined` without needing custom logic to do typechecking/coercion. The `getNFTokenID` function will still throw if it receives undefined, but it'll do so with a consistent error instead of requiring any users of the function to write their own each time.

mvadari: IMO it's easy enough to do an `undefined` check.

JST5000: I feel like if you have to write the same error handler for every instance, it should just go in the function instead.

Especially since it's not obvious that the reason meta can be undefined is based on the transaction type you're looking up. So if they're searching for an NFTokenMint transaction, it should never be undefined. If they try searching up on xrpl.org how they should handle the undefined case for `meta` it's hard to get a clear answer using our docs as to what to do about it.

JST5000: Hmmm... Looking into it again I think you're right that it won't ever be undefined. 

It seems [way back when we did xrpl.js 2.0 I changed the type to be optional](https://github.com/XRPLF/xrpl.js/pull/1636/files#), but I'm guessing I did that because I confused the `tx` request with the `tx` field in a submitted Transaction. 

image

In a submitted transaction, the `meta` field is often undefined because the actual metadata is included as an additional field at the end. For a `tx` request though it seems that every transaction changes something on the ledger, so every `tx` response should have `meta` if it succeeds. 

Based on that, I think the best thing to do is to make `meta` not optional from a TxResponse, and then update this function to be the same (although I'm a bit worried that'd be a breaking change for TypeScript users because it'll give them errors of "could not be undefined" or something like that)

mvadari: Note: the "raw" view on the Explorer isn't actually the raw transaction, there's some minor mutations. So if you want to know what the raw transaction output looks like you'll need to look at the actual `tx` output.

mvadari: Bump on making this change to remove the `undefined` option

packages/xrpl/test/utils/getNFTokenID.test.ts
```diff
@@ -18,11 +18,4 @@ describe('getNFTokenID', function () {
       '0008125CBE4B401B2F62ED35CC67362165AA813CCA06316FFA766254000003EE'
     assert.equal(result, expectedNFTokenID)
   })
-
-  it('fails with nice error when given raw response instead of meta', function () {
```
mvadari, line 22: What's wrong with this test?

JST5000: Nothing - I was a bit hasty removing it. 

I removed the error check because I thought that there was no longer a case where `meta` could be undefined - but forgot about the case in JavaScript where this could happen. Then I saw the test fail and didn't read the title - I just thought it was checking for an error which couldn't happen anymore).

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
```diff
@@ -1,67 +1,87 @@
...
 
-        const accountNFTs = await client.request({
-          command: 'account_nfts',
-          account: wallet.address,
-        })
+      const txRequest: TxRequest = {
+        command: 'tx',
+        transaction: hashSignedTx(response.result.tx_blob),
+      }
+      const txResponse = await testContext.client.request(txRequest)
```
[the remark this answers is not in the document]
JST5000: To clarify are you saying that the `tx` request is unnecessary, or that the check that the `txResponse.meta.TransactionResult` is `tesSUCCESS` is unnecessary? 

I think it's ok to omit the `tesSUCCESS` check, although it's slightly different because it's making sure the metadata is up to date (which should always be the case), but the tx request in general is needed for the rest of the checks later on.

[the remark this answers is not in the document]
mvadari: Both - I thought the `testTransaction` method returned a response that you could use for the rest of the checks.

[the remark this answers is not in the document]
JST5000: `testTransaction` returns a `SubmitResponse` which is a bit different than a `TxResponse` - notably it doesn't have the meta field, which is needed to get the `NFTokenID`.

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
```diff
@@ -1,67 +1,87 @@
...
+        account: testContext.wallet.address,
+      })
+
+      const nftokenID =
+        getNFTokenID(txResponse.result.meta as TransactionMetadata) ??
+        'undefined'
+
+      const accountHasNFT = accountNFTs.result.account_nfts.some(
+        (value) => value.NFTokenID === nftokenID,
+      )
```
mvadari, line 63: Since the account was just created for this test shouldn't `account_nfts` only have one NFT in it?

JST5000: I'm reusing the `testContext.wallet` so this account may have other nfts from other nft-related tests, so checking for `some` gives resilience against that race condition.

mvadari: `testContext.wallet` is regenerated for every test

JST5000: I see - I think using `.some` is still fine regardless - I'd rather not depend on the account only having one nft when the function returns an array.

mvadari: I'd still rather double check that, since if there's somehow another NFT in that account that'd mean something was wrong with our test framework, but I'm not going to hold up this PR because of that if you want to keep it.

mvadari (approved):

justinr1234:
@JST5000 please fix merge conflicts

PR #2278, opened 2023-04-20
Add guard check for signing algorithm

wojake:
### Context of Change
You could put in a random string and it'll still work, we should be able to tell that the function generated a seed according to the right algorithm. Feel free to make changes to this PR since I have no clue if this works.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


justinr1234 (approved):

khancode (approved):

packages/xrpl/src/Wallet/index.ts
```diff
@@ -128,6 +128,11 @@ class Wallet {
    * @returns A new Wallet derived from a generated seed.
    */
   public static generate(algorithm: ECDSA = DEFAULT_ALGORITHM): Wallet {
+    if (algorithm !== DEFAULT_ALGORITHM || algorithm !== "ed25519" || algorithm !== "secp256k1") {
```
khancode, line 131:
```suggestion
    if (algorithm !== DEFAULT_ALGORITHM || algorithm !== "ed25519" || algorithm !== "ecdsa-secp256k1") {
```
"ecdsa-secp256k1" is the correct value according to https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/ECDSA.ts

khancode (changes requested):
Can you write unit tests for this and also change it to the correct value `ecdsa-secp256k1`

packages/xrpl/src/Wallet/index.ts
ckniffen: You might be able shorten it to `!Object.values(ECDSA).includes(algorithm as ECDSA)`

ckniffen:
Can you add a test where you pass the wrong value and check that it throws an exception?

mvadari:
Please update the changelog.

packages/xrpl/src/Wallet/index.ts
wojake: Thanks @ckniffen

packages/xrpl/test/wallet/index.test.ts
```diff
@@ -56,7 +56,13 @@ describe('Wallet', function () {
       assert.isTrue(wallet.privateKey.startsWith(ed25519KeyPrefix))
       assert.isTrue(wallet.classicAddress.startsWith(classicAddressPrefix))
     })
-
+    
+    it('generates a new wallet using an invalid/unknown algorithm', function () {
+      const algorithm = "test"
+      
+      assert.throws(() => { Wallet.generate(algorithm) }, /Invalid cryptographic signing algorithm/)
```
mvadari, line 63: You'll need to ignore the TypeScript check here for tests to pass

mvadari (approved):

packages/xrpl/test/wallet/index.test.ts
ckniffen, line 63: Need to add `//@ts-expect-error`.  This will allow the test to compile even though the type is wrong to simulate how it might work if using js.

sublimator: Or just cast it via `as unknown as ExpectedType`

ckniffen (approved):

khancode (approved):

PR #2280, opened 2023-04-21
Add NFT clio only API support

shawnxie999:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/packages/xrpl/src/models/common/index.ts
+++ b/packages/xrpl/src/models/common/index.ts
@@ -117,3 +117,22 @@ export interface NFTOffer {
   destination?: string
   expiration?: number
 }
+
+/**
+ * One NFToken that might be returned from either an {@link NFTInfoResponse}
+ * or an {@link NFTsByIssuerResponse}.
+ *
+ * @category Responses
+ */
+export interface NFToken {
+  nft_id: string
+  ledger_index: number
+  owner: string
+  is_burned: boolean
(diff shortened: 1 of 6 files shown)


packages/xrpl/src/models/common/index.ts
```diff
@@ -117,3 +117,22 @@ export interface NFTOffer {
...
+  nft_id: string
+  ledger_index: number
+  owner: string
+  is_burned: boolean
+  flags: number
+  transfer_fee: number
+  issuer: string
+  nft_taxon: number
+  nft_serial: number
+  uri?: string
```
ledhed2222, line 137: no longer optional

packages/xrpl/src/models/methods/nftHistory.ts
```diff
@@ -0,0 +1,117 @@
...
+  /** JSON object defining the transaction. */
+  tx?: Transaction & ResponseOnlyTxInfo
+  /** Unique hashed String representing the transaction. */
+  tx_blob?: string
+  /**
+   * Whether or not the transaction is included in a validated ledger. Any
+   * transaction not yet in a validated ledger is subject to change.
+   */
+  validated: boolean
+}
```
ledhed2222, line 75: is there not an object that represents a generic tx you can already use? what is being used for account_tx?

PR #2293, opened 2023-04-28
fix: remove tx serialization check

mvadari:

This PR removes the `decode(encode(tx)) == tx` check in the wallet signing process and moves the other validation checks that happened as a part of signing to `validate`.

### Context of Change

It requires a bunch of edge case checks and it's easy to get tripped up with an error, but it doesn't actually find bugs.

### Type of Change

## Test Plan

CI passes. Minorly fixed tests due to moving stuff around (one test was flat out bad).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,11 +1,15 @@
 {
     "editor.tabSize": 2,
     "cSpell.words": [
-        "Multisigned",
-        "Setf",
         "hostid",
+        "keypair",
+        "keypairs",
+        "multisign",
+        "multisigned",
+        "multisigning",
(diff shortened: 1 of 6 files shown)


JST5000 (approved):
LGTM

mDuo13 (approved):
Agreed. This check has caused numerous false positives over the years and I've never seen it flag a legitimate issue.

mvadari:
> Agreed. This check has caused numerous false positives over the years and I've never seen it flag a legitimate issue.

Yeah, I just ran into one of those false positives, which frustrated me enough to put together this PR.

Issue #2225, opened 2023-02-27
Incorrect return value for getOrderbook method


develoQ:
The return value of the getOrderbook method has a buy field and a sell field, but in the following, TakerGets = XRP and TakerGets = USD are mixed in the buy.

```js
const { Client } = require('xrpl');
const client = new Client('wss://xrpl.ws');

const main = async () => {
  await client.connect();
  const orderbook = await client.getOrderbook(
    { issuer: 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B', currency: 'USD' },
    { currency: 'XRP' },
    { limit: 10, ledger_index: 78070258 }
  );
  console.log(orderbook);
};

main();
```

```js
{
  buy: [
    ...
    {
      Account: 'rsFaWztpAqZKoBSLTuhrhYfBV4nwYg7TFP',
      BookDirectory: 'DFA3B6DDAB58C7E8E5D944E736DA4B7046C30E4F460FD9DE4E0DCE23664B3487',
      BookNode: '0',
      Flags: 0,
      LedgerEntryType: 'Offer',
      OwnerNode: '0',
      PreviousTxnID: 'F11D96C1BFF8DE0E6A731460747B97E77C453C1D874FA4ECC0186FFC3F0A6D7E',
      PreviousTxnLgrSeq: 78017755,
      Sequence: 75621130,
      TakerGets: '5236750',
      TakerPays: [Object],
      index: '5BF1EBD7C7DF6729C4EB6F3E8A12EE4721563F67E6DB29ACD1E81A38B9569111',
      quality: '0.0000003885826132620423'
    },
    {
      Account: 'rsgi7ENscrVaXC44JE2m94XzKQrmAVX2gV',
      BookDirectory: '4627DFFCFF8B5A265EDBD8AE8C14A52325DBFEDAF4F5C32E5B096C6EFAA91865',
      BookNode: '0',
      Expiration: 762310783,
      Flags: 0,
      LedgerEntryType: 'Offer',
      OwnerNode: '0',
      PreviousTxnID: '2B90439DA0A732E3CB1EF6E0D6DD8E39D378E17B750C7ED595B98798EACE097A',
      PreviousTxnLgrSeq: 78069539,
      Sequence: 72701643,
      TakerGets: [Object],
      TakerPays: '121568490',
      index: 'EC40BD275BF22AA9C40F024EDF6A642ABD3D73A4AA8CD7313330C45C6C4DA6EA',
[...]

PR #2294, opened 2023-04-28
Update orderbook docs and params

JST5000:

Make the docs & parameters for `getOrderbook` match it's functionality.

### Context of Change

Follow up to https://github.com/XRPLF/xrpl.js/issues/2225 - The function docs are confusing for what the function actually does. 

I think the actual function itself is confusing because `lsfSell` doesn't seem like a great way to organize the information, but rewriting it to normalize and only look up one way seems like it would create a totally different function. Might be worth doing as a separate task to create a separate helper.

### Type of Change

- [X] Documentation Updates

## Before / After

* Renamed parameters from `takerGets` and `takerPays` to `currency1` and `currency2` to avoid connotations of directionality which aren't backed up in the functionality
* Updated the description for the function

## Test Plan

CI Passes

### DIFF
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -100,6 +100,7 @@ Wallet.fromMmnemonic()
 ## 2.2.1 (2022-04-21)
 ### Fixed
 * Fix return field of NFT offer
+* Updated `getOrderbook` docs and param names to reflect actual behavior of checking both sides of order book.
 
 ## 2.2.0 (2022-04-19)
 ### Added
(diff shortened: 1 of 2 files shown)


JST5000:
@ckniffen I'm not sure if changing the parameter names constitutes a breaking change in JavaScript - thoughts?

ckniffen:
> @ckniffen I'm not sure if changing the parameter names constitutes a breaking change in JavaScript - thoughts?

I am in favor of changing the param names.  You are right it is not a breaking change.

ckniffen:
You should also make a note about how limit is what is passed to both calls but could result in more entries than the limit in the returned array.

ckniffen (approved):

pdp2121 (approved):

PR #2273, opened 2023-04-13
feat: use noble libs for cryptography

sublimator:
### Authors 

This PR is a collab with @ckniffen / @sublimator  

### Use @noble/{curbes,hashes}

Replace elliptic/hash.js/bn.js with noble suite (incl native bigint implementations of ecc)

See: [repo](https://github.com/paulmillr/noble-curves/)
See: [most recent audit](https://github.com/trailofbits/publications/blob/master/reviews/2023-01-ryanshea-noblecurveslibrary-securityreview.pdf
)
[image](https://github.com/paulmillr/noble-curves)

### TODO
- [x] Check: https://github.com/paulmillr/noble-curves/commit/42de6200102ce226e6e6466d16eb5c64a4578b32 (ed25519 pertinent?)
 - [x] For wider compatibility set ZIP 215 setting to false to override zcash friendly defaults (stricter than NIST)
- [x] @noble/hashes iso wrapper https://github.com/paulmillr/noble-hashes/issues/56
  - [x] @noble/xxx not interested 
  - [x] @xrpl/crypto (was ripple-iso-crypto)
    - [x] isomorphic support potentially overkill given @noble speed (see @ckniffen' bench [results](https://gist.github.com/sublimator/3558238b8aabc1cb7bb5a5a9563e6617))
      - [x] noble sha512 support isn't as good (relative to native) as its sha256/ripemd
    - [x] seems like a library that could be useful in many contexts (i.e. @xrpl/xxx potentially overly XRPL specific ? on the other hand XRPLF can "own" it with no other pesky use cases/users popping up)
- [x] Some issues with brorand replacement with @noble/hashes/utils#randomBytes
- [x] Look at bundle size
  - [x] Take [first look](https://github.com/XRPLF/xrpl.js/pull/2273#issuecomment-1507892188)
  - [x] Remove crypto-browserify which brings in an unused elliptic due to no tree shaking
    - [x] Update xrpl sha512Half util to use @noble/hashes
    - [x] Browser tests are failing for some reason (see #2275 )
  - [x] Investigate possibility of reducing use of crypto-browserify super package
    - ES6 is more amenable to tree shaking/static analysis {optimizations.usedExports}  
- [x] ripple-binary-codec uses createHash, could replace with tiny noble/hashes/sha512 ?
- [x] ripple-keypairs is using only @noble/hashes (and before only hash.js), never node native
[...]

PR #2303, opened 2023-05-03
feat: add support for node v20

mvadari:

Title says it all. I also updated our linter and browser tests to use v16 instead of now-deprecated v14.

### Context of Change

node v20 was just released

### Type of Change

## Test Plan

CI passes with node 20.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -57,7 +57,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [14.x, 16.x, 18.x, 20.x]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


.github/workflows/nodejs.yml
```diff
@@ -57,7 +57,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [14.x, 16.x, 18.x, 20.x]
```
sublimator, line 60: What are the plans re: node 14 which is EOL now ? 
Depending upon those plans, could potentially remove the need for another lib: 
https://github.com/XRPLF/xrpl.js/pull/2273#discussion_r1177267048

Thanks!

mvadari: You'll have to ask some of the other maintainers about that - it's not my responsibility

justinr1234 (approved):

.github/workflows/nodejs.yml
ckniffen: We should drop in 3.0.

ckniffen (approved):

.github/workflows/nodejs.yml
sublimator: > You'll have to ask some of the other maintainers 

I see, you just had that air of authority :)

PR #2309, opened 2023-05-11
add userAgent and usageContext to faucet transactions

jonathanlei:

add userAgent and useCase information to the json body of a faucet transaction request for analytics purposes. 
### Context of Change

This is part of the testnet analytics effort. 

### Type of Change
- [ X] New feature (non-breaking change which adds functionality)


justinr1234 (approved):

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -112,6 +115,8 @@ async function fundWallet(
       JSON.stringify({
         destination: walletToFund.classicAddress,
         xrpAmount: options?.amount,
+        userAgent: 'xrpl.js',
+        useCase: options?.useCase,
```
mvadari, line 119: I'm still not a big fan of the name `useCase` - it's not very descriptive for someone who's looking at this library and trying to figure out how they are supposed to use it.

jonathanlei: what would you name it instead? 
also I did add an explanation of what useCase is in the function params

justinr1234: I think useCase is the best I can think of

ckniffen (approved):

PR #2332, opened 2023-06-06
Add generics of TransactionType for TxResponse

develoQ:

Make Tx command response possible by specifying a transaction.

### Type of Change

- [ x] Documentation Updates

### DIFF
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -13,6 +13,7 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 * Fixed the location of `signer_lists` in the `account_info` response so that it matches rippled
 * Guard check for signing algorithm used in `Wallet.generate()`
 * Null and undefined values in transactions are now treated as though the field was not passed in.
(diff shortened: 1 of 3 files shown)


justinr1234:
@develoQ can you show the intended use case that isn’t possible today and how this helps?

develoQ:
Additional commits have been made.

For example, it would be possible to provide a more detailed type for the response value of the submitAndWait method.

``
const paymentTx: Payment = {
...
}

const response = client.submitAndWait(paymentTx)
```

In this case, response will be resolved as TxResponse<Payment>, not TxResponse<Transaction>.

mvadari:
I was a bit worried that this would make DX more complicated, since users would need to do `TxResponse<txtype>` instead of `TxResponse` when constructing a new `TxResponse`, but users never need to construct a `TxResponse` themselves so it's fine.

Does the `request` definition need to be adjusted with this to return `TxResponse<Transaction>` instead of just `TxResponse`? Seems like we don't actually run a TypeScript check in our CI to confirm.

mvadari (approved):

PR #2337, opened 2023-06-09
fix: make docs not output confusing information in xrpl client

justinr1234:

Docs are super confusing today. They don't output the correct information due to how functions are assigned to the class. We're also missing docs for several function. Additionally, there's overuse over overloaded functions which get output as separate functions in the docs page making it cluttered and difficult to follow.

### TODO

- [x] add a before and after image of the documentation

### Context of Change

- Refactor class methods to live on the class instead of imported and assigned to member variables
- Replace function overloads with typescript conditional types
- Improve jsdoc strings for multiple functions
- Improve missing test for `requestNextPage`

### Type of Change

- [X] Documentation Updates
- [ ] Release

## Test Plan

- CI passes
- Docs page looks right and works well

## Before

image

## After (properly marked as member functions and not variable constants)

image

## Before (submit bare documentation)

image

## After (documentation and examples are shown)

image

image

## Before (request has a bajillion overloads and is missing an example)

image

## After (request is clean and has example)

image

## Before (multiple client methods missing examples)

image

## After (all client methods have examples)

image


justinr1234:
Closes #2337

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -20,128 +20,57 @@ const INTERVAL_SECONDS = 1
...
- * The fundWallet() method is used to send an amount of XRP (usually 1000) to a new (randomly generated)
- * or existing XRP Ledger wallet.
+ * Generate a new wallet to fund if no existing wallet is provided or its address is invalid.
  *
- * @example
- *
- * Example 1: Fund a randomly generated wallet
- * const { Client, Wallet } = require('xrpl')
- *
- * const client = new Client('wss://s.altnet.rippletest.net:51233')
```
ckniffen, line 31: I am not a fan of getting rid of these examples.  Are these being moved somewhere else?

justinr1234: These examples live in the client folder the function `fundWallet` was moved on the class instead of in a separate file to allow the docs to follow the class method. The examples were kept.

justinr1234:
I will add a separate PR to generate the docs

pdp2121:
Hmm I'm not sure if this is the only way to generate better docs but this solution will make a long index files (most of functions in this file are abstraction, with exception for making a request and connect/disconnect) and also it would be a bit confusing imo to have the fundWallet or autofill files in sugar folder without the fundWallet and autofill function (seems more like utils files for me)

justinr1234:
> Hmm I'm not sure if this is the only way to generate better docs but this solution will make a long index files (most of functions in this file are abstraction, with exception for making a request and connect/disconnect) and also it would be a bit confusing imo to have the fundWallet or autofill files in sugar folder without the fundWallet and autofill function (seems more like utils files for me)

I’m fairly sure this is the only way, but open to another solution if you have one.

Aside from that, I tried to balance making it smaller by refactoring certain internal details to a separate file.

pdp2121:
> > Hmm I'm not sure if this is the only way to generate better docs but this solution will make a long index files (most of functions in this file are abstraction, with exception for making a request and connect/disconnect) and also it would be a bit confusing imo to have the fundWallet or autofill files in sugar folder without the fundWallet and autofill function (seems more like utils files for me)
> 
> I’m fairly sure this is the only way, but open to another solution if you have one.
> 
> Aside from that, I tried to balance making it smaller by refactoring certain internal details to a separate file.

I'm not familiar with docs generation so if that's the only way to output desired output, please go ahead with it.

ckniffen:
Can you add a before and after image of the documentation?

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -20,133 +20,58 @@ const INTERVAL_SECONDS = 1
...
- * @param options.usageContext - An optional field to indicate the use case context of the faucet transaction
- * Ex: integration test, code snippets.
- * @returns A Wallet on the Testnet or Devnet that contains some amount of XRP,
- * and that wallet's balance in XRP.
- * @throws When either Client isn't connected or unable to fund wallet address.
+ * @param wallet - Optional existing wallet.
+ * @returns The wallet to fund.
  */
-// eslint-disable-next-line max-lines-per-function -- All lines necessary
-async function fundWallet(
```
JST5000, line 89: Does this change fully remove the `fundWallet` function? I can't see where else it was moved to if it was moved.

packages/xrpl/test/client/requestNextPage.test.ts
```diff
@@ -57,4 +57,31 @@ describe('client.requestNextPage', function () {
       'response does not have a next page',
     )
   })
+
+  // TODO: Write this test to verify multiple types of commands can be run
```
JST5000, line 61: Can you create a follow up task to do this?

packages/xrpl/src/sugar/submit.ts
```diff
@@ -248,8 +182,42 @@ function isSigned(transaction: Transaction | string): boolean {
   return tx.SigningPubKey != null && tx.TxnSignature != null
 }
 
-// initializes a transaction for a submit request
-async function getSignedTx(
+/**
+ * Retrieves the signed transaction from the client.
```
JST5000, line 186: We don't want to give the impression that we're doing something which could reveal secrets online. The signing happens offline, but the client is used to update the transaction before that.
```suggestion
 * Updates a transaction with `autofill` then signs it if it is unsigned.
```

packages/xrpl/snippets/src/paths.ts
```diff
@@ -15,7 +15,7 @@ async function createTxWithPaths(): Promise<void> {
     issuer: 'rVnYNK9yuxBz4uP8zC8LEFokM2nqH3poc',
   }
 
-  const request = {
+  const resp = await client.request({
     command: 'ripple_path_find',
```
mvadari, line 19: It's actually bad practice to use this command, we should be recommending that people use the `path_find` subscription instead

packages/xrpl/src/sugar/submit.ts
```diff
@@ -248,8 +182,42 @@ function isSigned(transaction: Transaction | string): boolean {
...
+ *
+ * @param client - The client from which to retrieve the signed transaction.
+ * @param transaction - The transaction to retrieve. It can be either a Transaction object or
+ * a string (encode from ripple-binary-codec) representation of the transaction.
+ * @param [options={}] - Optional. Additional options for retrieving the signed transaction.
+ * @param [options.autofill=true] - Optional. Determines whether the transaction should be autofilled (true)
+ * or not (false). Default is true.
+ * @param [options.failHard] - Optional. If true, and the transaction fails locally, do not retry or relay the
```
JST5000, line 194: Potentially tangential to this PR, but it seems that `options.failHard` is never used and doesn't really make sense in this function (other than allowing opts to be passed in directly from the higher level function) - This function doesn't submit to ledger, so it can't fail like that.

packages/xrpl/src/sugar/submit.ts
```diff
@@ -187,29 +122,28 @@ async function waitForFinalTransactionOutcome<
...
+  //       lastLedger,
+  //       submissionResult,
+  //     )
+  //   }
+  //   throw new Error(
+  //     `${message} \n Preliminary result: ${submissionResult}.\nFull error details: ${String(
+  //       error,
+  //     )}`,
+  //   )
+  // })
```
JST5000, line 146: Why was this error handling commented out? (I believe this was originally added due to a race condition which could occur where the transaction wasn't quite validated yet which was a very intermittent error)

packages/xrpl/src/client/index.ts
```diff
@@ -661,48 +523,530 @@ class Client extends EventEmitter {
...
+   * in an unsigned transaction along with your wallet to be submitted.
+   *
+   * @template T
+   * @param transaction - A {@link Transaction} in JSON format
+   * @param signersCount - The expected number of signers for this transaction.
+   * Only used for multisigned transactions.
+   * @returns The autofilled transaction.
    */
-  public autofill = autofill
+  public async autofill<T extends Transaction>(
```
mvadari, line 570: I really don't like the idea of moving all this code into the `Client` file - it makes it way more difficult to find the code that you need and it makes this file monstrously large

packages/xrpl/src/Wallet/fundWallet.ts
justinr1234: It is moved to the client index file on the class

packages/xrpl/src/sugar/submit.ts
```diff
@@ -161,14 +61,49 @@ async function submitRequest(
...
 
-/*
- * The core logic of reliable submission.  This polls the ledger until the result of the
- * transaction can be considered final, meaning it has either been included in a
- * validated ledger, or the transaction's lastLedgerSequence has been surpassed by the
- * latest ledger sequence (meaning it will never be included in a validated ledger).
+/**
+ * Waits for the final outcome of a transaction by polling the ledger until the result can be considered final,
+ * meaning it has either been included in a validated ledger, or the transaction's lastLedgerSequence has been
+ * surpassed by the latest ledger sequence (meaning it will never be included in a validated ledger).
```
JST5000, line 67: If we're exporting this, I think it's worth including in this description that likely the function people will want to use is `submit` or `submitAndWait` instead of this one. The use for this is very niche in my head, and I'm worried someone will accidentally use it if VSCode gives them code autocomplete and then get stuck if they don't know about the more generally applicable functions. 

I may suggest keeping it private to avoid the confusion altogether since I don't see when someone would want to use this function instead of `submitAndWait`.

packages/xrpl/test/client/requestNextPage.test.ts
justinr1234: https://github.com/XRPLF/xrpl.js/issues/2384

packages/xrpl/src/Wallet/fundWallet.ts
JST5000: Nevermind, found it in client.

packages/xrpl/snippets/src/paths.ts
justinr1234: https://github.com/XRPLF/xrpl.js/issues/2385

packages/xrpl/src/client/index.ts
```diff
@@ -126,6 +95,28 @@ export interface ClientOptions extends ConnectionUserOptions {
   timeout?: number
 }
 
+type RequestNextPageType =
```
JST5000, line 98: I think it's worth adding a reminder to keep these two types in sync
```suggestion
// Make sure to update both this and `RequestNextPageReturnMap` at the same time
type RequestNextPageType =
```

packages/xrpl/src/sugar/submit.ts
justinr1234: Added back

packages/xrpl/src/client/index.ts
justinr1234: Yeah Phu felt the same, but I don't think there's another solution. We could remove some of the body of the functions, but honestly that feels unnecessary to me and more confusing in my eyes. Open to a vote by reviewers for moving the body of the functions out to separate files as I did with `doFundWalletRequest`

JST5000: When you say it makes it more difficult to find the code that you need, what specifically are you referring to? (What was the easier path people had before which is gone?)

packages/xrpl/src/sugar/submit.ts
justinr1234: It doesn't show up publicly. This file isn't exported for people to render in the docs.

packages/xrpl/src/client/index.ts
JST5000: Unsure if it's be much better - but if filesize is an issue, instead of `public autofill = autofill` we could just do a one-line wrapper here with the same docstring to avoid the "this" issue:

e.g. 
```
public async autofill<T extends Transaction> ( ... ) {
    return doAutofill(this, ...)
}
```

I think that'd keep the proper doc generation, although I could be wrong.

mvadari: You could go to the file in `src/sugar` that had the code you wanted to look at, instead of scrolling through an absolutely massive file (in general, large files are harder to work with and should be split up where possible)

mvadari: So does doc gen not work if you do something like this?
```js
/*
* doc stuff here
*/
public autofill = autofill
```

justinr1234: > So does doc gen not work if you do something like this?
> 
> ```js
> /*
> * doc stuff here
> */
> public autofill = autofill
> ```

Correct

justinr1234: > Unsure if it's be much better - but if filesize is an issue, instead of `public autofill = autofill` we could just do a one-line wrapper here with the same docstring to avoid the "this" issue:
> 
> e.g.
> 
> ```
> public async autofill<T extends Transaction> ( ... ) {
>     return doAutofill(this, ...)
> }
> ```
> 
> I think that'd keep the proper doc generation, although I could be wrong.

That's what I meant yup!

justinr1234: Also you'd have to have a double docstring if you move it into a different file, since we require docstrings on file exports.

mvadari: > > Unsure if it's be much better - but if filesize is an issue, instead of `public autofill = autofill` we could just do a one-line wrapper here with the same docstring to avoid the "this" issue:
> > e.g.
> > ```
> > public async autofill<T extends Transaction> ( ... ) {
> >     return doAutofill(this, ...)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think that'd keep the proper doc generation, although I could be wrong.
> 
> That's what I meant yup!

Yeah, this would be better. I think it's fine to disable the docstrings (or have a dummy docstring)

packages/xrpl/src/sugar/submit.ts
JST5000: Ah I didn't realize that :)

packages/xrpl/src/client/index.ts
JST5000: @justinr1234 Sorry for asking about what you already proposed - your replies didn't show up when I was writing the message.

packages/xrpl/src/models/methods/ledger.ts
```diff
@@ -96,30 +202,57 @@ export interface LedgerBinary
...
+
+interface LedgerResponseExpandedResult extends LedgerResponseBase {
+  /** The complete header data of this {@link Ledger}. */
+  ledger: Ledger
+}
+
+// expand = true
+// accounts OR transactions OR both
+// binary missing altogether
+// matches LedgerRequestExpanded
```
JST5000, line 250: These comments seem like they were notes during development, but I could be wrong

packages/xrpl/src/models/methods/ledger.ts
```diff
@@ -96,30 +202,57 @@ export interface LedgerBinary
...
+
+interface LedgerResponseResult extends LedgerResponseBase {
+  /** The complete header data of this {@link Ledger}. */
+  ledger: LedgerBinary
+}
+
+// This is the default
+// expand = false
+// binary = false
+// Matches basic LedgerRequest
```
JST5000, line 232: Ditto

packages/xrpl/src/models/methods/ledger.ts
```diff
@@ -67,6 +67,112 @@ export interface LedgerRequest extends BaseRequest, LookupByLedgerRequest {
   queue?: boolean
 }
 
+/**
+ * Retrieve information about the public ledger. Expects a response in the form
+ * of a {@link LedgerResponseExpanded}. Will return full JSON-formatted transaction data instead of string hashes.
+ *
```
JST5000, line 73: Probably should be a follow up issue, but I think it'd be worth linking to the xrpl.org docs pages for these requests since not only do they go into more depth on how the requests work, they also have a "Try It" button that let's you see the request in action using the websocket tool :)

Ex. https://xrpl.org/ledger.html#ledger

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -286,7 +255,15 @@ async function fetchAccountDeleteFee(client: Client): Promise<BigNumber> {
...
 }
 
-async function calculateFeePerTransactionType(
+/**
+ * Calculates the fee per transaction type.
+ *
+ * @param client - The client object.
+ * @param tx - The transaction object.
+ * @param [signersCount=0] - The number of signers (default is 0).
+ * @returns A promise that resolves with void.
```
JST5000, line 264:
```suggestion
 * @returns A promise that resolves with void. Modifies the `tx` parameter to give it the calculated fee.
```

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -331,11 +308,25 @@ async function calculateFeePerTransactionType(
...
   return new BigNumber(value).times(multiplier).toString()
 }
 
-async function setLatestValidatedLedgerSequence(
+/**
+ * Sets the latest validated ledger sequence for the transaction.
+ *
+ * @param client - The client object.
+ * @param tx - The transaction object.
+ * @returns A promise that resolves with void.
```
JST5000, line 327:
```suggestion
 * @returns A promise that resolves with void. Modifies the `tx` parameter setting `LastLedgerSequence`.
```

packages/xrpl/src/models/methods/ledger.ts
justinr1234: I left them there because it seemed helpful for devs, but I can remove if you think it looks bad

JST5000: I think I was mostly confused by the styling. It has a different comment style than the rest. 

I'd recommend moving those into the docstring instead of having them be separate. 

For example: 

"This is the default request response, triggered when `expand` and `binary` are both false."

packages/xrpl/src/sugar/getOrderbook.ts
```diff
@@ -101,7 +82,30 @@ async function getOrderbook(
...
+/**
+ * Creates a request object for retrieving book offers.
+ *
+ * @param currency1 - The first currency in the pair.
+ * @param currency2 - The second currency in the pair.
+ * @param options - Additional options for the request.
+ * @param [options.limit] - The maximum number of offers to retrieve.
+ * @param [options.ledger_index] - The ledger index to use for retrieval.
+ * @param [options.ledger_hash] - The ledger hash to use for retrieval.
+ * @param [options.taker] - The taker address for retrieval.
```
JST5000, line 96: Does adding the `[...]` brackets impact the generated docs? (It's not a convention I've been following, but can start following if it's helpful)

packages/xrpl/src/sugar/getOrderbook.ts
```diff
@@ -30,44 +29,26 @@ const getOrderbookOptionsSet = new Set([
   'taker',
 ])
 
+export interface GetOrderBookOptions {
```
JST5000, line 32: I think `Options` should have docstrings for each field since this is exported and used elsewhere.

packages/xrpl/src/models/methods/ledger.ts
justinr1234: Fixed

packages/xrpl/src/sugar/getOrderbook.ts
justinr1234: That's how you denote optional in jsdoc

packages/xrpl/src/sugar/balances.ts
```diff
@@ -1,123 +1,21 @@
...
-  AccountLinesTrustline,
-  LedgerIndex,
-  AccountInfoRequest,
-} from '../models'
-import { dropsToXrp } from '../utils'
+export interface Balance {
+  value: string
+  currency: string
+  issuer?: string
+}
```
JST5000, line 7: Does the order of fields here impact anything? 

(Previously I noticed `Balance` had currency -> issuer -> value)
https://github.com/XRPLF/xrpl.js/blob/d47a277e25ae1a1ff95c7bd5a35dd133286c4330/packages/xrpl/src/models/common/index.ts#L20

JST5000: Also, should the `Commmon/index.ts` implementation of `Balance` be removed? (I don't see changes to that file here)

packages/xrpl/src/sugar/getOrderbook.ts
justinr1234: Fixed

packages/xrpl/src/sugar/balances.ts
justinr1234: No

packages/xrpl/src/models/methods/index.ts
```diff
@@ -248,6 +254,120 @@ type Response =
...
+  ? AccountOffersResponse
+  : T extends AccountTxRequest
+  ? AccountTxResponse
+  : T extends GatewayBalancesRequest
+  ? GatewayBalancesResponse
+  : T extends NoRippleCheckRequest
+  ? NoRippleCheckResponse
+  : // NOTE: The order of these LedgerRequest types is important
+  // to get the proper type matching overrides based on parameters set
+  // in the request.
```
JST5000, line 279: Is there documentation this comment could link to which helps inform what the proper order should be? (Could be useful if we have to add another one "in the right spot" later on.

packages/xrpl/package.json
```diff
@@ -30,6 +30,7 @@
     "ripple-address-codec": "^4.3.0",
     "ripple-binary-codec": "^1.7.0",
     "ripple-keypairs": "^1.3.0",
+    "run-s": "0.0.0",
```
JST5000, line 33: Should this be a dev dependency instead of a normal depenency?

packages/xrpl/src/sugar/balances.ts
justinr1234: I think it autofixed it because putting optional params at the end is easier to read

packages/xrpl/src/models/methods/index.ts
justinr1234: Added docs inline, but they're a bit hefty

packages/xrpl/package.json
justinr1234: yes

justinr1234: Fixed

packages/xrpl/src/models/methods/ledger.ts
justinr1234: Fxied

ckniffen:
We usually dont commit the docs folder until the release.

justinr1234:
> We usually dont commit the docs folder until the release.

I knew it was going to sneak in there if I didn't delete it

packages/xrpl/src/sugar/balances.ts
```diff
@@ -1,123 +1,21 @@
...
 
-import type { Balance, Client } from '..'
-import {
-  AccountLinesRequest,
-  AccountLinesTrustline,
-  LedgerIndex,
-  AccountInfoRequest,
-} from '../models'
-import { dropsToXrp } from '../utils'
+export interface Balance {
```
ckniffen, line 3: You should use https://github.com/XRPLF/xrpl.js/blob/d47a277e25ae1a1ff95c7bd5a35dd133286c4330/packages/xrpl/src/models/common/index.ts#L20

justinr1234: Fixed

JST5000 (approved):
LGTM

ckniffen:
Created #2403 to handle deleting extraneous package-lock.json files that can cause confusion.

JST5000:
Notes on what I did during the merge to resolve the trickier conflicts:
1. I replaced the `fundWallet` function that was moved to the `Client` with the updated version from previous 3.0 changes to remove the HTTP polyfill. 
2. I removed a couple lodash imports which spawned from the merge (both from `flatMap` so switched to the es2019 syntax of `.flatMap` instead)
3. I made sure to double check that the removed imports in the package.json from other 3.0 changes stayed out

package-lock.json
ckniffen, line 1: What caused this to remove many entries?  The only dependency change was  `"run-s": "^0.0.0",` which I would think might add entries and not remove.

packages/xrpl/package.json
```diff
@@ -44,6 +44,7 @@
     "lodash": "^4.17.4",
     "node-polyfill-webpack-plugin": "^2.0.1",
     "react": "^18.2.0",
+    "run-s": "^0.0.0",
```
ckniffen, line 47: I found no new usages of run-s.  I think this can be removed along with package-lock.json changes.

package-lock.json
JST5000: Ah, I may have chosen the wrong version to merge then - there was a package-lock conflict and I assumed the 3.0 version would be shorter.

packages/xrpl/package.json
JST5000: run-s is used for `npm run build` - does it need to be a dev dependency for that?

package-lock.json
JST5000: I'm not sure why it still changed - but I used the 3.0 version and ran `npm i` - does that new result look sensible, or is there something I accidentally changed?

packages/xrpl/package.json
ckniffen: Good call.  Wonder why it wasn't there before.

package-lock.json
ckniffen: Try copying the 3.0 version of the file and run n

packages/xrpl/package.json
JST5000: TODO: Re-add this since I removed it in another commit

JST5000: Re-added.

ckniffen (approved):

jonathanlei (approved):

PR #2338, opened 2023-06-09
test: remove use of `fs` in tests

mvadari:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/packages/xrpl/test/webpack.config.js
+++ b/packages/xrpl/test/webpack.config.js
@@ -102,7 +102,6 @@ function webpackForTest(testFileName) {
         module: false,
         assert: require.resolve('assert-browserify'),
         constants: require.resolve('constants-browserify'),
-        fs: require.resolve('browserify-fs'),
         buffer: require.resolve('buffer/'),
         assert: require.resolve('assert/'),
         url: require.resolve('url/'),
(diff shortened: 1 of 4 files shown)


packages/xrpl/test/webpack.config.js
```diff
@@ -100,15 +100,11 @@ function webpackForTest(testFileName) {
       extensions: ['.ts', '.js', '.json'],
       fallback: {
         module: false,
-        assert: require.resolve('assert-browserify'),
```
ckniffen, line 103: I think assert should still be required.  Maybe we dont have a test that hits any assets.

ckniffen (approved):

packages/xrpl/test/webpack.config.js
mvadari: There's another assert webpack config a couple of lines below - we don't need two

mvadari:
Karma depends on `fs`, so this won't work.

mvadari:
Karma depends on `fs`, so this won't work.

ckniffen:
> Karma depends on `fs`, so this won't work.

Does it still need the browserify version?

mvadari:
> > Karma depends on `fs`, so this won't work.
> 
> Does it still need the browserify version?

Yes, apparently (tests were still failing on my last commit). Feel free to reopen this and take it over if you find a solution.

PR #2349, opened 2023-06-21
feat: add ledger entry types for NFTokens

ckniffen:

Add `NFToken`, `NFTokenPage`, and `NFTokenOffer` types.

### Type of Change

### DIFF
--- a/packages/xrpl/src/models/common/index.ts
+++ b/packages/xrpl/src/models/common/index.ts
@@ -119,7 +119,7 @@ export interface NFTOffer {
 }
 
 /**
- * One NFToken that might be returned from either an {@link NFTInfoResponse}
+ * One NFToken that might be returned from a {@link NFTInfoResponse}
  *
  * @category Responses
  */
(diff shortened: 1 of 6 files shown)


justinr1234 (approved):

packages/xrpl/src/models/common/index.ts
```diff
@@ -119,7 +119,7 @@ export interface NFTOffer {
 }
 
 /**
- * One NFToken that might be returned from either an {@link NFTInfoResponse}
+ * One NFToken that might be returned from a {@link NFTInfoResponse}
```
JST5000, line 122: nit: (I think "N" is pronounced as "en" so has an implicit vowel in front so uses "an" in this case)
```suggestion
 * One NFToken that might be returned from an {@link NFTInfoResponse}
```

packages/xrpl/src/models/methods/nftInfo.ts
```diff
@@ -3,8 +3,7 @@ import { LedgerIndex, NFToken } from '../common'
 import { BaseRequest, BaseResponse } from './baseMethod'
 
 /**
- * The `nft_info` method retrieves information about NFToken
- * NFToken.
+ * The `nft_info` method retrieves information about a NFToken.
```
JST5000, line 6: ditto
```suggestion
 * The `nft_info` method retrieves information about an NFToken.
```

JST5000 (approved):
LGTM

PR #2350, opened 2023-06-21
fix: escrow tests

justinr1234:

Escrow tests fail at different times. We theorize it is due to being later in the testing cycle and is affected by calling `ledger_accept` multiple times.

### Context of Change

Change the order of tests

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Tests pass


packages/xrpl/integrationTestSequencer.js
```diff
@@ -0,0 +1,33 @@
+const Sequencer = require('@jest/test-sequencer').default
+
+class CustomSequencer extends Sequencer {
+  sort(tests) {
+    // Test structure information
+    // https://github.com/facebook/jest/blob/6b8b1404a1d9254e7d5d90a8934087a9c9899dab/packages/jest-runner/src/types.ts#L17-L21
+    const copyTests = Array.from(tests)
+    const sorted = copyTests.sort((testA, testB) => {
+      // We run escrow tests first because they rely on the ledger number via ledger_accept
+      if (testA.path.includes('escrow') && testA.path.includes('escrow')) {
```
JST5000, line 10: These are checking the same thing twice
```suggestion
      if (testA.path.includes('escrow') && testB.path.includes('escrow')) {
```

ckniffen:
@justinr1234 After looking at it with @JST5000  the issue  might now be with `calculateWaitTimeForTransaction`.  It can be lower than 10 seconds.  If you call ledger_accept before 10 seconds pass it will only increment the close_time 1 second.

`calculateWaitTimeForTransaction` takes into account times that are in the future but not stunted time progression.

packages/xrpl/integrationTestSequencer.js
```diff
@@ -0,0 +1,33 @@
...
+        return 1
+      }
+
+      if (testB.path.includes('finalTest')) {
+        return -1
+      }
+
+      return testA.path > testB.path ? 1 : -1
+    })
+    console.log('sorted', sorted)
```
ckniffen, line 28: You should remove this console.log.  It makes the test output very noisy.

justinr1234: Yes, it was there only for debugging

justinr1234:
> @justinr1234 After looking at it with @JST5000 the issue might now be with `calculateWaitTimeForTransaction`. It can be lower than 10 seconds. If you call ledger_accept before 10 seconds pass it will only increment the close_time 1 second.
> 
> `calculateWaitTimeForTransaction` takes into account times that are in the future but not stunted time progression.

Where did you find that info? Through testing?

JST5000:
> Where did you find that info? Through testing?

This is the docs on the ledger_close time behavior - I didn't run the code myself to verify whether this was the root cause. 
https://xrpl.org/ledgers.html#ledger-close-times

Issue #2219, opened 2023-02-17
Support `featureXRPFees` amendment


ckniffen:
This amendment is part of rippled@1.10.0 and updates the fees to be formatted as Amounts.

XRPLF/rippled#4247

- ~~For server_info favor the following in order~~
  - ~~`reserve_base_drops`, `reserve_base`, `reserve_base_xrp`~~
  - ~~`base_fee_drops`, `base_fee`, `base_fee_xrp`~~
  - ~~`reserve_inc_drops`, `reserve_inc_fee`, `reserve_inc_fee_xrp`~~
- ~~If specifically looking for fees such as in the client libraries when autofilling use the command `fee` which is not changed by this PR and is already in drops.~~
- ~~Validations processing will need to be handled similar to `server_info`.~~
- `Fee` property on all transactions remains the same
- `SetFee` transactions
  - `BaseFeeDrops`, `ReserveBaseDrops`, and `ReserveIncrementDrops` will be present when the amendment is passed and on new transactions.  These will be of the type Amount 
  - `BaseFee`, `ReserveBase`, `ReferenceFeeUnits`, `ReserveIncrement` are valid only historically after the amendment passes
  - No need for validation updates in client libraries as this is a pseudo transaction
  - Types should reflect there will be only one of the two sets of fields
- FeeSettings
  - New fields need to be added to the [type](https://github.com/XRPLF/xrpl.js/blob/6fbfab4fa49f9f82e07716609de0d8e30bc5d33d/packages/xrpl/src/models/ledger/FeeSettings.ts#LL9C26-L9C37).

sappenin:
Thanks for reporting this @ckniffen -- @nkramer44 for visibility.

PR #2351, opened 2023-06-26
feat(types): Add pseudo transactions

ckniffen:

Add types for `EnableAmendment`, `SetFee`, and `UNLModify` transactions.

Related to #2219

### Type of Change

- [x] Documentation Updates


packages/xrpl/src/models/transactions/UNLModify.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @category Pseudo Transaction Models
+ */
+export interface UNLModify {
+  TransactionType: 'UNLModify'
+  /**
+   * The ledger index where this pseudo-transaction appears. This distinguishes the pseudo-transaction from other occurrences of the same change.
+   */
+  LedgerSequence: number
+  /**
+   * If 1, this change represents adding a validator to the Negative UNL. If 0, this change represents removing a validator from the Negative UNL. (No other values are allowed.)
```
JST5000, line 13: I think the "No other types are allowed" is implied by the type definition, and if we flip the order of explanations then it'll match the type definition (which I think is especially helpful since this is a double-negative situation so it's easy to get tripped up trying to map it all out in your head) 
```suggestion
   * If 0, this change represents removing a validator from the Negative UNL. If 1, this change represents adding a validator to the Negative UNL.
```

ckniffen: I copied these from xrpl.org.  I will clean up for this medium.

packages/xrpl/src/models/transactions/setFee.ts
```diff
@@ -0,0 +1,57 @@
...
+  TransactionType: 'SetFee'
+}
+
+export interface SetFeePreAmendment extends SetFeeBase {
+  /**
+   * The charge, in drops of XRP, for the reference transaction, as hex. (This is the transaction cost before scaling for load.)
+   */
+  BaseFee: string
+  /**
+   * The cost, in fee units, of the reference transaction
```
JST5000, line 13:
```suggestion
   * The cost, in fee units, of the [reference transaction](https://xrpl.org/transaction-cost.html#reference-transaction-cost)
```

packages/xrpl/src/models/transactions/setFee.ts
```diff
@@ -0,0 +1,57 @@
...
+   */
+  ReserveBase: number
+  /**
+   * The incremental reserve, in drops
+   */
+  ReserveIncrement: number
+
+  BaseFeeDrops?: never
+  ReserveBaseDrops?: never
+  ReserveIncrementDrops?: never
```
JST5000, line 27: Can you add comments to these 3? - https://xrpl.org/feesettings.html (same below)

JST5000: Or a comment explaining why their type is set to `never`? - I thought these could be assigned values still?

ckniffen: I do define them below on the type where they are set.  I can add a comment explaining that those fields will never be set before the amendment.

ckniffen: I ended up not treating this like the same object and went with this style of documentation below.  The reason was that the previous way typedoc would interpret fields that are actual required in pre or post amendment as optional when in fact they are optional collectively.

Monosnap SetFee | xrpl 2023-06-26 13-55-38

JST5000 (approved):
LGTM

pdp2121 (approved):

PR #2352, opened 2023-06-27
feat(types): define new `account_flags` field

ckniffen:

The `account_flags` field was added to the `account_info` response for `rippled@1.11.0`

### Type of Change

- [x] Documentation Updates

### DIFF
--- a/packages/xrpl/src/models/methods/accountInfo.ts
+++ b/packages/xrpl/src/models/methods/accountInfo.ts
@@ -73,6 +73,62 @@ interface QueueData {
   transactions?: QueueTransaction[]
 }
 
+export interface AccountInfoAccountFlags {
+  /**
+   * Enable rippling on this address's trust lines by default. Required for issuing addresses; discouraged for others.
+   */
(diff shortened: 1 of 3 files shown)


mvadari (approved):

packages/xrpl/src/models/methods/accountInfo.ts
```diff
@@ -89,6 +145,11 @@ export interface AccountInfoResponse extends BaseResponse {
      * present.
      */
     account_data: AccountRoot & { signer_lists?: SignerList[] }
+
+    /**
+     * A map of account flags parsed out.  This will only be available for rippled nodes 1.11.0 and higher.
+     */
+    account_flags?: AccountInfoAccountFlags
```
JST5000, line 152: Is this actually a map of flags, or is the data type an integer? (Do we have the integer mapping?)

ckniffen: It is a map of booleans.

JST5000 (approved):
LGTM

PR #2353, opened 2023-06-27
Clawback

shawnxie999:
- `Clawback` transaction
- Account flag `asfAllowTrustLineClawback `

### Context of Change
Spec https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-39d-clawback

### Type of Change

- [ ] Documentation Updates
- [ ] Release


jonathanlei (approved):

ckniffen:
`AccountRoot` object needs to be updated as well.

packages/xrpl/src/models/transactions/clawback.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Verify the form and type of an Clawback at runtime.
+ *
+ * @param tx - An Clawback Transaction.
+ * @throws When the Clawback is Malformed.
+ */
+export function validateClawback(tx: Record<string, unknown>): void {
+  validateBaseTransaction(tx)
+
+  if (tx.Amount == null) {
```
ckniffen, line 34: There should be a check that the issuer field is not the Account field.  I bet many people will incorrectly do that.

ckniffen (approved):

packages/xrpl/src/models/transactions/clawback.ts
shawnxie999: will fix it later

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -36,8 +36,12 @@ export function parseAccountRootFlags(
 ): AccountRootFlagsInterface {
   const flagsInterface: AccountRootFlagsInterface = {}
 
-  Object.keys(AccountRootFlags).forEach((flag) => {
-    if (isFlagEnabled(flags, AccountRootFlags[flag])) {
+  // If we use keys all will be strings and enums are reversed during transpilation
+  Object.values(AccountRootFlags).forEach((flag) => {
```
mvadari, line 40: What does this change do?

packages/xrpl/test/integration/transactions/clawback.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      const tx: Clawback = {
+        TransactionType: 'Clawback',
+        Account: testContext.wallet.classicAddress,
+        Amount: {
+          currency: 'USD',
+          issuer: wallet2.classicAddress,
+          value: '1000',
+        },
+      }
+      await testTransaction(testContext.client, tx, testContext.wallet)
```
mvadari, line 94: This test should also check the state of the trustline after, to make sure the money was actually clawed back

packages/xrpl/src/models/utils/flags.ts
ckniffen: This was to fix an issue with flags.  enums get turned into an object with double the values.  Key value like you define and then the reverse of value key.  This meant that isFlagEnabled was being passed strings for the flag value which was blowing up isFlagEnabled after making them be `bigint`

mvadari (approved):

PR #2355, opened 2023-06-28
feat(types): Make DepositPreauth easier to guard

ckniffen:

`DepositPreauth` transaction is now a union type to correctly model that the `Authorize` and `Unauthorize` fields cannot appear at the same time.

### Type of Change

- [x] Documentation Updates


packages/xrpl/HISTORY.md
```diff
@@ -13,6 +13,9 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 * Add type for NFToken object that is stored on a `NFTokenPage`.
 * Add type for `account_info`'s `account_flags` property.
 
+### Changed
+* `DeposityPreauth` has a new type definition that will not allow `Authorize` and `Unauthorize` on the same instance. This provides better guard support.
```
mvadari, line 17:
```suggestion
* `DepositPreauth` has a new type definition that will not allow `Authorize` and `Unauthorize` on the same instance. This provides better guard support.
```

packages/xrpl/src/models/transactions/index.ts
```diff
@@ -10,7 +10,11 @@ export { AccountDelete } from './accountDelete'
 export { CheckCancel } from './checkCancel'
 export { CheckCash } from './checkCash'
 export { CheckCreate } from './checkCreate'
-export { DepositPreauth } from './depositPreauth'
+export {
+  DepositPreauth,
+  DepositPreauthWithAuthorize,
+  DepositPreauthWithUnauthorize,
```
mvadari, line 16: Do these two need to be exported?

ckniffen: Yes if people want to type guard.

ckniffen:
This makes the documentation output look goofy.  Closing for now.

PR #2356, opened 2023-06-28
feat: enable 1.10.0 amendments for tests

ckniffen:

- `NFTokenMint` needs `NonFungibleTokensV1_1` enabled instead of `NonFungibleTokensV1` .  
- `fee_ref` is no longer returned after `XRPFees` amendment is enabled.

### Type of Change


pdp2121 (approved):

packages/xrpl/test/integration/requests/subscribe.test.ts
```diff
@@ -156,16 +156,7 @@ describe('subscribe', function () {
...
-          fee_base: ledgerResponse.fee_base,
-          fee_ref: ledgerResponse.fee_ref,
-          ledger_hash: ledgerResponse.ledger_hash,
-          ledger_index: ledgerResponse.ledger_index,
-          ledger_time: ledgerResponse.ledger_time,
-          reserve_base: ledgerResponse.reserve_base,
-          reserve_inc: ledgerResponse.reserve_inc,
-          validated_ledgers: ledgerResponse.validated_ledgers,
-        }
+        const expectedResult = { ...ledgerResponse }
```
JST5000, line 159: I think this defeats the point of this test - the reason those fields were explicitly defined was that we were trying to ensure that if a new field was added, or one was removed, our test would fail. (As it is now, the assertion is irrelevant because we're dynamically updating our expected value to have the same shape no matter what)

ckniffen: I see.  In this case the `fee_ref` field is removed if the SetFee amendment is enabled.  Do you suggest I remove that one field?

JST5000: I think removing that one field makes sense to me :)

ckniffen: Done. PR description updated.

PR #2362, opened 2023-07-01
Fix #2361 - Custom definitions in STArray»STObject

### LINKED ISSUE #2361: Nested STObject » STArray » STObject custom definitions error
Consider TX Json like this:

```javascript
{ 
  Hooks: [
    {
      Hook: {
        CreateCode: "...hex...",
        Flags: 1,
        HookApiVersion: 0,
        HookNamespace: "F...",
        HookOn: "F...",
      }
    }
  ]
}
```

And imagine custom definitions providing a specification for e.g. the `HookOn` field, this will not work (be ignored).

The reason is the `definitions` are passed on to the STObject:
- https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/types/st-object.ts#L97

But now if it contains e.g. an STArray, when sent through here:
- https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/types/st-object.ts#L135

... The `from()` doesn't pass the definitions.

Now imagine (as in the example above) this lands you in STArray, yielding yet another STObject (like in the example above, STObject contains STArray contains STObject):
- https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/types/st-array.ts#L64

... This is where the `from` lost the custom definitions.

Same happens with `decode`, where binary is read to JSON, and `fromParser` etc. don't pass definitions.

Add the following to `definitions.test.js` (missing test case) to see the problem:

```javascript
  test('can encode and decode a new Field nested in STObject in STArray in STObject', function () {
    const tx = {
      ...txJson,
      NewFieldArray: [
        {
          NewField: {
            NewFieldValue: 10,
          },
        },
      ],
    }

    // Before updating the types, undefined fields will be ignored on encode
    expect(decode(encode(tx))).not.toStrictEqual(tx)

    // Normally this would be generated directly from rippled with something like `server_definitions`.
    // Added here to make it easier to see what is actually changing in the definitions.json file.
    const definitions = JSON.parse(JSON.stringify(normalDefinitionsJson))

    definitions.FIELDS.push([
      'NewFieldArray',
      {
        nth: 100,
        isVLEncoded: false,
        isSerialized: true,
        isSigningField: true,
        type: 'STArray',
      },
    ])

    definitions.FIELDS.push([
      'NewField',
      {
        nth: 101,
        isVLEncoded: false,
        isSerialized: true,
        isSigningField: true,
        type: 'STObject',
      },
    ])

    definitions.FIELDS.push([
      'NewFieldValue',
      {
        nth: 102,
        isVLEncoded: false,
        isSerialized: true,
        isSigningField: true,
        type: 'UInt32',
      },
    ])

    const newDefs = new XrplDefinitions(definitions)

    const encoded = encode(tx, newDefs)
    expect(() => decode(encoded)).toThrow()
    const decoded = decode(encoded, newDefs)
    expect(decoded).toStrictEqual(tx)
  })
```

WietseWind:
See: https://github.com/XRPLF/xrpl.js/issues/2361

This PR fixes passing Definitions from STObject to STArray back to STObject, both for signing & decoding (to JSON).


ckniffen (approved):

JST5000:
Looks good to me - can you update the HISTORY.md before merging with a note on the change?

JST5000 (approved):

PR #2369, opened 2023-07-05
fix docs broken link

jonathanlei:
The docs site js.xrpl.org has broken links for [Case by Case Setup Steps](https://js.xrpl.org/#md:case-by-case-setup-steps), this PR fixes it

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -53,11 +53,11 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
 </ol>
 <a id="md:case-by-case-setup-steps" class="tsd-anchor"></a><h3><a href="#md:case-by-case-setup-steps">Case by Case Setup Steps</a></h3><p>If you&#39;re using xrpl.js with React or Deno, you&#39;ll need to do a couple extra steps to set it up:</p>
 <ul>
-<li><a href="./UNIQUE_SETUPS.md#using-xrpljs-from-a-cdn">Using xrpl.js with a CDN</a></li>
-<li><a href="./UNIQUE_SETUPS.md#using-xrpljs-with-create-react-app">Using xrpl.js with <code>create-react-app</code></a></li>
-<li><a href="./UNIQUE_SETUPS.md#using-xrpljs-with-react-native">Using xrpl.js with <code>React Native</code></a></li>
-<li><a href="./UNIQUE_SETUPS.md#using-xrpljs-with-vite-react">Using xrpl.js with <code>Vite React</code></a></li>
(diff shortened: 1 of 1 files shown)


docs/index.html
```diff
@@ -53,11 +53,11 @@ <h2>xrpl</h2></div>
 </ol>
 <a id="md:case-by-case-setup-steps" class="tsd-anchor"></a><h3><a href="#md:case-by-case-setup-steps">Case by Case Setup Steps</a></h3><p>If you&#39;re using xrpl.js with React or Deno, you&#39;ll need to do a couple extra steps to set it up:</p>
 <ul>
-<li><a href="./UNIQUE_SETUPS.md#using-xrpljs-from-a-cdn">Using xrpl.js with a CDN</a></li>
-<li><a href="./UNIQUE_SETUPS.md#using-xrpljs-with-create-react-app">Using xrpl.js with <code>create-react-app</code></a></li>
-<li><a href="./UNIQUE_SETUPS.md#using-xrpljs-with-react-native">Using xrpl.js with <code>React Native</code></a></li>
-<li><a href="./UNIQUE_SETUPS.md#using-xrpljs-with-vite-react">Using xrpl.js with <code>Vite React</code></a></li>
-<li><a href="./UNIQUE_SETUPS.md#using-xrpljs-with-deno">Using xrpl.js with <code>Deno</code></a></li>
+<li><a href="https://github.com/XRPLF/xrpl.js/blob/main/UNIQUE_SETUPS.md#using-xrpljs-from-a-cdn">Using xrpl.js with a CDN</a></li>
```
ckniffen, line 56: Is there a way in typedoc to populate urls?  This approach will not link properly for branches.

jonathanlei: the problem is that the docs don't generate separate pages for github pages, I'll look into the typedoc again and find a solution

jonathanlei:
> I'm not sure this is the right place to fix the links. Isn't `docs/index.html` generated by TypeDoc?
> 
> https://github.com/XRPLF/xrpl.js/blob/74de24cf75ff5a3d42ebe2fa17c9733812103881/CONTRIBUTING.md#generate-reference-docs

the problem is that the docs don't generate separate pages for github pages (so the relative url before would go to 404), I'll look into the typedoc again and find a solution

ckniffen:
> I'm not sure this is the right place to fix the links. Isn't `docs/index.html` generated by TypeDoc?
> 
> https://github.com/XRPLF/xrpl.js/blob/74de24cf75ff5a3d42ebe2fa17c9733812103881/CONTRIBUTING.md#generate-reference-docs

It is created from typedoc from the README.  There are several issues opened for this with typedoc.  What it comes down to is typedoc cannot generate markdown except for the index.  One solution is to use docusaurus and its typedoc plugin that will generate the api pages in typedoc and all other documention is generated from markdown files.

PR #2375, opened 2023-07-12
feat: remove 3 http related polyfills

ckniffen:

Switch to using fetch for browser and `node-fetch` for node for the faucet calls.  This reduces the webpack bundle by 3.2% or 16.5kb gzipped.

The fundWallet code has been refactored to be much more straight forward due to not having to do such low level operations.

This improves the frontend setup process by no longer requiring several polyfills such as `url`, `stream-http`, and `https-browserify`.

### Type of Change

## Test Plan

Some new tests might need to be added to validate error checking.


packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -83,205 +91,134 @@ const MAX_ATTEMPTS = 20
...
-    )
-  } catch {
-    /* startingBalance remains '0' */
+  if (!createWallet) {
+    try {
+      startingBalance = Number(
+        await this.getXrpBalance(walletToFund.classicAddress),
+      )
+    } catch {
+      /* startingBalance remains '0' */
```
jonathanlei, line 126: is there supposed to be code here for catch?

ckniffen: No there was before.

ckniffen: Relooking at this i bet it always fails because there is no `this`

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -83,205 +91,134 @@ const MAX_ATTEMPTS = 20
...
-            INTERVAL_SECONDS * MAX_ATTEMPTS
-          } seconds`,
-        ),
-      )
+    return {
+      wallet: walletToFund,
+      balance,
     }
+
+    throw new XRPLFaucetError(
```
jonathanlei, line 212: did you get rid of comparing existing and updated balance logic?

packages/xrpl/src/Wallet/fundWallet.ts
jonathanlei: you meant if there wasn't a catch logic before?

packages/xrpl/src/Wallet/fundWallet.ts
ckniffen: There was a catch

ckniffen: It will always be greater than starting amount and then it would retrieve the balance yet again.

packages/xrpl/src/Wallet/fundWallet.ts
ckniffen: There was a catch before

packages/xrpl/src/Wallet/fundWallet.ts
jonathanlei: oh yeah the logic should be if it isn't greater then retrieve again, instead of it is greater. There could be an edge case where the balance remained unchanged but the error in the transaction wasn't caught. if that never happens, `startingBalance` shouldn't be a param since it's not used?

ckniffen: I just realized that I still needed the check just not the additional balance retrieval.

JST5000:
FYI I added branch protections for 3.0 (same ones that apply to main)

packages/xrpl/src/Wallet/fundWallet.ts
JST5000: nit: Separate issue on this line - it could fail due to network connectivity instead of ledger reasons - in which case a previously funded account would have the same funds as before which can be greater than zero.
```suggestion
      /* startingBalance remains what it was previously */
```

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -110,151 +123,83 @@ async function fundWallet(
...
-  startingBalance: number,
-  walletToFund: Wallet,
-  resolve: (response: { wallet: Wallet; balance: number }) => void,
-  reject: (err: ErrorConstructor | Error | unknown) => void,
-): Promise<void> {
-  const body = Buffer.concat(chunks).toString()
+    body: JSON.stringify(postBody),
+  })
 
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Its a FaucetWallet
```
JST5000, line 171: nit:
```suggestion
  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- It's a FaucetWallet
```

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -110,151 +123,83 @@ async function fundWallet(
...
+  const body = (await response.json()) as FaucetWallet
   // "application/json; charset=utf-8"
```
JST5000, line 173: Is this a useful comment?

JST5000 (approved):
LGTM

packages/xrpl/src/Wallet/fundWallet.ts
ckniffen: I dont think so but it was left over.

jonathanlei (approved):

PR #2378, opened 2023-07-12
feat: remove lodash as a dependency

### LINKED ISSUE #2118: Remove `lodash` from the xrpl package
When connecting to XRPL with Metamask Snap, due to a Hardend Javascript (SES) limitation, xrpl.js cannot be used (resulting in an error).

This is due to the `lodash` package.

After replacing the `lodash` package with `just`, the error no longer occurs, but I am not sure if this package can be included in xrpl.js.

https://github.com/angus-c/just

ckniffen:

This will reduce the bundle size by ~23%(117kb).

Only 4 methods were used `flatten`, `flatMap`, `omitBy`, and `groupBy`.

`omitBy` and `groupBy` were recreated while the es2019 implementations of `flatten` and `flatMap` are used.

`lodash` is still used in the tests which is fine because it makes the tests cleaner.

Closes #2118

### Type of Change


jonathanlei (approved):

packages/xrpl/src/utils/collections.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @returns a map of arrays
+ */
+export function groupBy<T>(
+  array: T[],
+  iteratee: (value: T, index: number, array: T[]) => string,
+): { [p: string]: T[] } {
+  return array.reduce<{ [key: string]: T[] }>(
+    (acc, value, index, arrayReference) => {
+      ;(acc[iteratee(value, index, arrayReference)] ||= []).push(value)
```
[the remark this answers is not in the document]
ckniffen: yeah it is added by eslint.  It is needed when starting a line normally with a parethesis and not allowing lines to end with ;.

packages/xrpl/src/utils/collections.ts
```diff
@@ -0,0 +1,52 @@
...
+  predicate: (objElement: ValueOf<T>, k: string | number | symbol) => boolean,
+): Partial<T> {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- We know the keys are properties of T
+  const keys: Array<keyof T> = Object.keys(obj) as Array<keyof T>
+  const keysToKeep = keys.filter((kb) => !predicate(obj[kb], kb))
+  return keysToKeep.reduce((acc: Partial<T>, key: keyof T) => {
+    acc[key] = obj[key]
+    return acc
+  }, {})
+}
```
xrpl-grants, line 52: Can you add unit tests for these helper functions?

JST5000, line 52: Can you add unit tests for these helper functions?

justinr1234 (approved):

tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "composite": true,
     "module": "commonjs",
+    "lib": ["es2019", "dom"],
```
JST5000, line 5: This is something that comes up every once in a while - is this a breaking change?

PR #2379, opened 2023-07-12
feat: remove `decimal.js` and `big-integer`

ckniffen:
### `decimal.js`

`BigNumber` was already being used by xrpl and they had almost identical
interface to decimal.js as they are both based on Java.

Reduces filesize by 6% (31kb)

### `big-integer`

`BigInt` is now well supported so the library is no longer needed.

Reduces the bundle size by 2% (10kb)

BREAKING CHANGE:
- `bigInt` from `big-integer` is no longer accepted by `BinarySerializer.writeType` or type serializer's `from` methods . Use the native `BigInt` instead.
- `quality.decode` returns a `BigNumber` instead of `Decimal`.

### Context of Change

This is an interesting read: https://docs.ethers.org/v5/api/utils/bignumber/.  Might be something to consider as well.

### Type of Change


packages/xrpl/package.json
```diff
@@ -56,7 +56,7 @@
     "build:lib": "tsc --build tsconfig.build.json",
     "build:web": "webpack",
     "build:browserTests": "webpack --config ./test/webpack.config.js",
-    "analyze": "run-s build:web --analyze",
+    "analyze": "webpack --analyze",
```
jonathanlei, line 59: why is this in the PR?

ckniffen: It fixes the webpack bundle size tool which i used to weigh whether or not this change was worth it.

jonathanlei (approved):

packages/ripple-binary-codec/src/ledger-hashes.ts
```diff
@@ -121,7 +120,7 @@ function accountStateHash(param: Array<JsonObject>): Hash256 {
  */
 interface ledgerObject {
   ledger_index: number
-  total_coins: string | number | bigInt.BigInteger
```
mvadari, line 124: Is this a breaking change?

packages/ripple-binary-codec/src/quality.ts
```diff
@@ -28,10 +27,10 @@ class quality {
    * @param arg hex-string denoting serialized quality
    * @returns deserialized quality
    */
-  static decode(quality: string): Decimal {
```
mvadari, line 31: Is this a breaking change?

packages/ripple-binary-codec/src/types/uint-64.ts
```diff
@@ -1,11 +1,9 @@
 import { UInt } from './uint'
 import { BinaryParser } from '../serdes/binary-parser'
-import bigInt = require('big-integer')
-import { isInstance } from 'big-integer'
 import { Buffer } from 'buffer/'
 
 const HEX_REGEX = /^[a-fA-F0-9]{1,16}$/
-const mask = bigInt(0x00000000ffffffff)
+const mask = BigInt(0x00000000ffffffff)
```
khancode, line 6: There's actually a second way of assigning a BigInt literal by appending `n` to the end of the literal. For example, `0x00000000ffffffffn` instead of using the constructor `BigInt(0x00000000ffffffff)`. I do like the constructor way because it's more apparent it's a BigInt versus appending an `n` at the end of the literal where it's more subtle. Not sure if there are type conversion implications with using one over the other.

khancode (approved):
LGTM, just added a comment.

packages/ripple-binary-codec/src/types/uint-64.ts
ckniffen: With hex it for sure could be confusing to add the n.

PR #2386, opened 2023-07-17
feat: remove node 14 support

ckniffen:

This will allow for other libraries such as noble libs.

BREAKING CHANGE: Node 14 is no longer supported

### Type of Change


package.json
```diff
@@ -73,7 +73,7 @@
     "./packages/*"
   ],
   "engines": {
-    "node": ">=12.0.0",
-    "npm": ">=7.0.0 < 10.0.0"
+    "node": ">=16.0.0",
+    "npm": ">=8.0.0 < 10.0.0"
```
mvadari, line 77: Why are we changing the npm version?

packages/xrpl/package.json
```diff
@@ -34,7 +34,7 @@
   },
   "devDependencies": {
     "@geut/browser-node-core": "^2.0.13",
-    "@types/node": "^14.18.36",
+    "@types/node": "16.11.7",
```
mvadari, line 37:
```suggestion
    "@types/node": "^16.11.7",
```

package.json
ckniffen: good call `Now using node v16.0.0 (npm v7.10.0)`.  I was thinking of the latest 16 which defaults to 8

mvadari: So are we going to leave the npm versioning the same or....

mvadari (commented):
Note: you'll also have to turn off the 14.x tests in the branch settings

package.json
ckniffen: sorry I thought it updated.  On it.

mvadari (approved):

package.json
```diff
@@ -73,7 +73,7 @@
     "./packages/*"
   ],
   "engines": {
-    "node": ">=12.0.0",
-    "npm": ">=7.0.0 < 10.0.0"
+    "node": ">=16.0.0",
```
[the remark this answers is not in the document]
ckniffen: I think we dont have an explicit top range in hopes that it "just works" in future releases.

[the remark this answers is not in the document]
khancode: Oh also in the README `**[Node.js v16](https://nodejs.org/)** is recommended. We also support v18 and v20. Other versions may work but are not frequently tested.`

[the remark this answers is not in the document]
ckniffen: Yeah that is what we test and support but people can try higher.  Lower will be an issue because there will be changes made to dependencies that wont work less than 16.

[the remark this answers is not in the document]
khancode: Since newer versions of Node are mostly backwards compatible, then I would say it's fine to exclude. Just curious :)

khancode (approved):

PR #2410, opened 2023-07-26
feat: use `isomorphic-ws` to clean up bundlers

ckniffen:

This removes the need to specify a mapping in bundler configs for `ws`. Only downside is that it is 47bytes gzipped larger but with how much less config is needed I think is worth it.

- [x] Documentation Updates
- [ ] Release


ckniffen:
Turns out there is a heavy reliance on the event emitter version of the websocket wrapper.  This will require quite the rewrite and will be deprioritized.

PR #2428, opened 2023-08-08
test: remove extra polyfills only for testing

ckniffen:

Removes the testing polyfills by using jasmine and not jest in the browser. We were already using jasmine but where overriding those methods with jest versions which was not needed.

The previous karma setup also used a single entrypoint which meant that not all integration tests were running because all files were not imported in integrations/index.ts.

This also eliminates 79 dev dependencies.

### Type of Change

## Future Work

- Run all `xrpl` unit tests in browser 
- Run other package's unit tests in the browser.


packages/xrpl/package.json
```diff
@@ -33,9 +33,6 @@
   "devDependencies": {
     "@geut/browser-node-core": "^2.0.13",
     "@types/node": "^16.18.38",
-    "browserify-fs": "^1.0.0",
```
ckniffen, line 36: Hopefully soon we can eliminate `node-polyfill-webpack-plugin` which will remove even more dependencies from `package-lock.json`

packages/xrpl/test/integration/index.ts
```diff
@@ -1,62 +1 @@
-/* eslint-disable import/export -- Tells webpack which files exist. */
-
-// These go first because they're affected by the `ledger_accept`s
-export * from './transactions/escrowFinish.test'
-export * from './transactions/escrowCancel.test'
```
JST5000, line 5: Does this new setup mess up the escrow fix here? (We originally needed them in the beginning to avoid ledger accept's from other tests creating a race condition where the escrow was unable to be finished I believe)

justinr1234: I think my last fix avoids the problem

ckniffen: Agreed.  Also this previous fix in the index.js file was only being used by the browser tests.

packages/xrpl/test/integration/index.ts
```diff
@@ -1,62 +1 @@
...
-export * from './onConnect.test'
-export * from './regularKey.test'
-export * from './submitAndWait.test'
-
-// Because this does 256 ledger accepts, we do it last
-export * from './transactions/accountDelete.test'
-
-// Ensure you export all added tests above "export * from './finalTest'", otherwise they will not be run.
-export * from './finalTest.test'
+/* eslint-disable import/unambiguous -- silence is golden */
```
JST5000, line 1: If the file is empty, why does it need to exist?

ckniffen: If it wasnt empty it errored and the webpack needed an entry point.  Might be worth investigating why that causes problems

JST5000 (approved):
LGTM

packages/xrpl/package.json
ckniffen: actually we can get rid of `node-polyfill-webpack-plugin` now.  Will eliminate 22 more deps.

jonathanlei (approved):

PR #2429, opened 2023-08-09
fix: Fix invariant assertion in binary-codec

JST5000:

Found a tiny assertion typo while reviewing the merge in #2337

### Context of Change

This assertion was previously 
```
assert.ok(0 < n && n <= 4, 'invalid n')
```

And since the new version needs to invert the condition (since it throws when true instead of passing when true) the `&&` should be a `||`. 

Link to the line in the 2.x branch: https://github.com/XRPLF/xrpl.js/blob/bd46f1960499974d245e51ba45490cccba9e4517/packages/ripple-binary-codec/src/serdes/binary-parser.ts#L73C1-L74C1

### Type of Change

## Test Plan

CI Passes

### DIFF
--- a/packages/ripple-binary-codec/src/serdes/binary-parser.ts
+++ b/packages/ripple-binary-codec/src/serdes/binary-parser.ts
@@ -75,7 +75,7 @@ class BinaryParser {
    * @return The number represented by those bytes
    */
   readUIntN(n: number): number {
-    if (0 >= n && n > 4) {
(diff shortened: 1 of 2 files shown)


ckniffen:
Can you add a test for this?  I don't want it slipping by again.

JST5000:
@ckniffen Added :)

ckniffen (approved):

ckniffen:
Thank you for you eagle eye.

PR #2430, opened 2023-08-09
Fix typo in documentation: on('transactions') -> on('transaction')

movitto:

Fixes typo in typedoc example comment.

### Context of Change

The documentation describing the client.on method provides an example where the 'transactions' event stream is listened to. This is a mistake as it should not be plural. This patch fixes it to demonstrate listening on the 'transaction' event stream

### Type of Change

- [X] Documentation Updates

## Test Plan

Copy the example and make sure it works! :smile:


JST5000 (approved):
LGTM

PR #2431, opened 2023-08-09
ci: Add a script to generate changes to ci-config

JST5000:

Automate figuring out which amendments to enable in our CI docker container.

### Context of Change

Periodically we need to update the `rippled.cfg` we use in the standalone node we use during CI so that the library is being tested against the latest amendments. This adds a script to get the hashes of any new amendments so it's easy to update our CI without a bunch of manual lookups / comparisons.

### Type of Change

- [X] Documentation Updates

## Test Plan

Manually tested the script against various networks / removing entries from the config file to see if it noticed.

## Future Tasks
Automatically generate a PR to update the CI file directly with hashes + names instead of requiring a manual lookup on the known amendments page.


.ci-config/rippled.cfg
```diff
@@ -167,3 +163,7 @@ F1ED6B4A411D8B872E65B9DCB4C8B100375B0DD3D62D07192E011D6D7F339013 fixTrustLinesTo
 # 1.12.0-b1 Amendments
 56B241D7A43D40354D02A9DC4C8DF5C7A1F930D92A9035C4E12291B3CA3E1C2B featureClawback
 27CD95EE8E1E5A537FF2F89B6CEB7C622E78E9374EBD7DCBEDFAE21CD6F16E0A fixReducedOffersV1
+# Added August 9th, 2023
+86E83A7D2ECE3AD5FA87AB2195AE015C950469ABF0B72EAACED318F74886AE90 CryptoConditionsSuite
```
ckniffen, line 167: CryptoConditionsSuite is super old deprecated. https://xrpscan.com/amendment/86E83A7D2ECE3AD5FA87AB2195AE015C950469ABF0B72EAACED318F74886AE90

.ci-config/rippled.cfg
```diff
@@ -167,3 +163,7 @@ F1ED6B4A411D8B872E65B9DCB4C8B100375B0DD3D62D07192E011D6D7F339013 fixTrustLinesTo
...
+AE35ABDEFBDE520372B31C957020B34A7A4A9DC3115A69803A44016477C84D6E fixNFTokenRemint
+B2A4DB846F0891BF2C76AB2F2ACC8F5B4EC64437135C6E56F3F859DE5FFD5856 ExpandedSignerList
```
ckniffen, line 169: Should probably be added above since it was much older. Added in 1.9.1.

.ci-config/rippled.cfg
ckniffen, line 167: Move between 1.10 and 1.12.  It was added in 1.11.0.

.ci-config/rippled.cfg
ckniffen: Might need to add an integration test for `WalletLocator` and the size of SignerLists

.ci-config/rippled.cfg
JST5000: It's still enabled on mainnet though

.ci-config/getNewAmendments.js
```diff
@@ -0,0 +1,56 @@
...
+];
+
+async function main() {
+  const client = new xrpl.Client(networkToEmulate);
+  await client.connect();
+
+  // Looks up what amendments have been enabled via their hash
+  const request = {
+    command: "ledger_entry",
+    index: "7DB0788C020F02780A673DC74757F23823FA3014C1866E72CC4CD8B226CD6EF4",
```
ckniffen, line 21: Nit: you can import the constant from `xrpl`.

packages/ripple-binary-codec/HISTORY.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Added
+* Updated to include latest updates to `definitions.json` including changes for the `fixNFTokenRemint` amendment.
```
ckniffen, line 6: nit: note that this includes AMM definitions updates too.

.ci-config/getNewAmendments.js
JST5000: I don't think it's properly exported outside the repo. Importing it directly would require compiling the typescript, which would make it more annoying to run the script, so I don't think it's worth doing that for this.

packages/ripple-binary-codec/HISTORY.md
mvadari: Shouldn't the AMM definitions wait for the AMM PR instead of being done here?

mvadari (commented):
Can this automatically be run in a Github Action, say once a week?

packages/ripple-binary-codec/HISTORY.md
JST5000: I'd prefer to just match rippled's `develop` as best as possible instead of manually editing the `definitions.json` file. Plus, I don't think there's any harm in having more `definitions` available.

mvadari: Can we hold off on this PR until that's merged in, then? Since that one's going to be merged soon anyways.

JST5000: This was generated from `develop` already - not sure why the definitions were there before the big AMM PR, but it seems they're already defined.

.ci-config/rippled.cfg
JST5000: Would it be alright to leave that for a separate ticket? (I'm not exactly sure what we'd be testing for that - maybe you can flesh out an Issue to add it w/ an idea for what we'd want to test on the client library side?)

.ci-config/getNewAmendments.js
ckniffen: Will need to look into how why it is not on the exported object.

ckniffen (approved):

jonathanlei (approved):

mvadari:
super nit: could the PR title avoid the word "automate"? That's part of the reason why I asked about a GitHub Action - I was expecting something that automatically ran the script.

JST5000:
> super nit: could the PR title avoid the word "automate"? That's part of the reason why I asked about a GitHub Action - I was expecting something that automatically ran the script.

I think I wanted to automate it originally, but didn't see an easy way - I think your question was fair, I just didn't find an answer just yet and might look into it more as a follow up task. Updated the title for now since I think the changes are helpful as is.

PR #2436, opened 2023-08-15
fix docs broken links 

jonathanlei:
 Unique setup links on docs are broken because they are relative links which aren't generated in js.xrpl.org, changed the links to guide back to github pages. 

### Type of Change

- [X ] Bug fix (non-breaking change which fixes an issue)


JST5000 (approved):
LGTM

justinr1234 (approved):
Nice temp fix until we can improve the docs site as a whole!

PR #2445, opened 2023-08-24
feat: add `xrpl-secret-numbers` to the monorepo

ckniffen:

- Add `xrpl-secret-numbers` by @WietseWind  to the mono repo.
- `unpkg` and `jsdelivr` support was simplified by adding fields to package.json.
- Unit tests run in a browser and node.

BREAKING CHANGES:
- `xrpl-secret-numbers` is now `@xrplf/secret-numbers`.
- Changed the bundled file produced changed from  `dist/browerified.js` to `build/xrpl-secret-numbers-latest.js`.
- Bundle variable is `xrplf_secret_numbers` instead of using browserify's loader.
- If using CDN instructions in README will need to update to `https://cdn.jsdelivr.net/npm/@xrpl/secret-numbers`

### Context of Change

Making this library part of the monorepo was discussed in #1788.  This solves several problems with the upcoming `xrpl.js@3.0`.

#### Release Process
`xrpl-secret-numbers` has a dependency of `ripple-keypairs` and `xrpl` has a depenendecy on `xrpl-secret-numbers`.  This means that any change would require a separate release of `ripple-keypairs`, then a release of `xrpl-secret-numbers`, followed by `xrpl`.  Now they can be released all at once

#### Fewer Frontend Runtime Dependencies
In 3.0 we eliminated the need for many polyfills like `util`, `assert`, and soon to be `buffer` (after noble libs PR is merged).  `xrpl-secret-numbers` still needs those.  This will also eliminate them and anytime similar changes in the future will be much easier.  This further reduces the bundle size of 3.0 branch by over 10% as well as reducing bundler setup.

### Type of Change

- [x] Documentation Updates

## Test Plan

Unit tests running in node and chrome.


ckniffen:
Looks like this still requires updates to linting.  Will try to get those addressed.

packages/secret-numbers/.eslintrc.js
```diff
@@ -0,0 +1,47 @@
+module.exports = {
```
ckniffen, line 1: I looked into normalizing this across all the packages.  That will end up being a project all on its own.  I choose to base it off the simplest one which was `ripple-address-codec`.

packages/secret-numbers/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Wietse Wind
```
ckniffen, line 3: @WietseWind Can the license be updated to maybe ISC like the other packages?

webpack.config.js
```diff
@@ -0,0 +1,97 @@
+"use strict";
+const webpack = require("webpack");
+const { BundleAnalyzerPlugin } = require("webpack-bundle-analyzer");
+const { merge } = require("webpack-merge");
+
+const bnJsReplaces = [
```
ckniffen, line 6: I created this base webpack config so that all the crazyness with polyfills can be in one place.  Most of this will be able to be deleted with #2273 when `stream-browserify` and `crypto-browserify` is eliminated.

webpack.config.js
```diff
@@ -0,0 +1,97 @@
...
+  "public-encrypt",
+  "elliptic",
+];
+
+function getDefaultConfiguration() {
+  return {
+    cache: true,
+    performance: {
+      hints: false,
+      maxEntrypointSize: 512000,
```
ckniffen, line 20: In future commits I will be tweaking this value so that we have a way to catch future inflations of bundle size

packages/secret-numbers/karma.config.js
```diff
@@ -0,0 +1,39 @@
+const webpackConfig = require("./test/webpack.config");
+delete webpackConfig.entry;
+
+module.exports = function (config) {
```
ckniffen, line 4: When I update the other packages to run unit tests in the browsers most of this will live in a shared config file.

mvadari:
Is there any way to keep the git history, like we did when originally creating the monorepo?

intelliot:
It is possible to keep the history, but there is no need to do so in this case. While we would be happy to move over the history if @WietseWind really wants it, I think it's short enough that anyone who wants to see the evolution of the code can just go to the old repo to find it.

intelliot:
note: this PR is ready for review/approval - we have asked @WietseWind to have a look, at his convenience.

packages/secret-numbers/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Development
+
+Run `npm run prepublish` to clean, lint, test and build. Or just run `npm run build`, `npm run test` or `npm run lint`. 
+
+Tests are in `./test/`
+
+## Credits
+
+This concept is based on an idea by [@nbougalis](https://github.com/nbougalis).
```
JST5000, line 97: This README should probably link to the `samples/es5.js` file :)

packages/secret-numbers/test/utils.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  it("calculateChecksum: 8th position", () => {
+    expect(utils.calculateChecksum(7, 49962)).toEqual(0);
+  });
+
+  it("checkChecksum: 2nd position, split numbers", () => {
+    expect(utils.checkChecksum(1, 55450, 3)).toBeTruthy();
+  });
+
+  it("checkChecksum: 7nd position, split numbers", () => {
```
JST5000, line 25: nit:
```suggestion
  it("checkChecksum: 7th position, split numbers", () => {
```

webpack.config.js
```diff
@@ -0,0 +1,97 @@
...
+    },
+    stats: "errors-only",
+    devtool: "source-map",
+    plugins: [
+      new webpack.ProvidePlugin({ process: "process/browser" }),
+      new webpack.ProvidePlugin({ Buffer: ["buffer", "Buffer"] }),
+      // this is a bit of a hack to prevent 'bn.js' from being installed 6 times
+      // TODO: any package that is updated to use bn.js 5.x needs to be removed from `bnJsReplaces` above
+      // https://github.com/webpack/webpack/issues/5593#issuecomment-390356276
+      new webpack.NormalModuleReplacementPlugin(/^bn.js$/, (resource) => {
```
JST5000, line 31: Maybe not for this PR - but can we resolve this TODO now?

JST5000 (approved):
LGTM

packages/secret-numbers/LICENSE
WietseWind: That's fine 👍

packages/secret-numbers/.eslintrc.js
WietseWind: 👍

WietseWind:
Wonderful @ckniffen - thanks for the effort!

webpack.config.js
ckniffen: This logic will be removed by #2273 but I added it to this PR so that neither is holding the other back.

packages/xrpl/test/webpack.config.js
```diff
@@ -38,37 +25,13 @@ function webpackForTest(testFileName) {
...
-        }
-      }),
     ],
     module: {
       rules: [
-        {
-          test: /jayson/,
-          use: 'null',
-        },
+        // Compile compile the tests
```
jonathanlei, line 34: nit: compile compile?

jonathanlei (approved):

Issue #2316, opened 2023-05-19
Add `nftoken_id`, `nftoken_ids` and `offer_id` meta fields into NFT Tx responses


ckniffen:
NFToken transactions have these meta fields being added to a transactions response in rippled@1.11.0.  Typing information needs to be added for them.

See: https://github.com/XRPLF/rippled/pull/4447

JST5000:
Closed by #2450

PR #2450, opened 2023-08-29
Add `nftoken_id`, `nftoken_ids`, `offer_id` to meta fields

develoQ:
- https://github.com/XRPLF/xrpl.js/issues/2316

### Type of Change

- [x ] New feature (non-breaking change which adds functionality)


packages/xrpl/src/models/transactions/metadata.ts
```diff
@@ -59,11 +67,35 @@ export function isDeletedNode(node: Node): node is DeletedNode {
...
   TransactionIndex: number
   TransactionResult: string
-}
+} & (T extends Payment
+  ? {
+      delivered_amount?: Amount | 'unavailable'
+    }
+  : T extends NFTokenMint
+  ? {
+      // rippled 1.11.0 or later
```
ckniffen, line 83: super nit: double space between "or" and "later"

packages/xrpl/src/models/transactions/metadata.ts
```diff
@@ -59,11 +67,35 @@ export function isDeletedNode(node: Node): node is DeletedNode {
...
-export interface TransactionMetadata {
+export type TransactionMetadata<T extends BaseTransaction = Transaction> = {
   AffectedNodes: Node[]
   DeliveredAmount?: Amount
   // "unavailable" possible for transactions before 2014-01-20
   delivered_amount?: Amount | 'unavailable'
   TransactionIndex: number
   TransactionResult: string
-}
+} & (T extends Payment
```
ckniffen, line 77: My only hesitation about this approach is it separates transaction specific meta from the rest of the transaction information.

Can we have each transaction file define its extended meta definition and then reference them here?  This will make it cleaner if we add more in the future.

packages/xrpl/src/models/transactions/transaction.ts
```diff
@@ -97,9 +97,11 @@ export type Transaction =
 /**
  * @category Transaction Models
  */
-export interface TransactionAndMetadata {
-  transaction: Transaction
-  metadata: TransactionMetadata
+export interface TransactionAndMetadata<
+  T extends BaseTransaction = Transaction,
+> {
+  transaction: T
```
ckniffen, line 103: nit: The tricky part of modifying this type is that `ledger` command and streams will not return these nice new properties.  Might be worth adding a comment about that.

ckniffen (commented):
Thank you for picking this up.

packages/xrpl/src/models/transactions/metadata.ts
develoQ: What names would you like to define within each transaction file?

for example, 
1.`PaymentMetadata`,`NFTokenMintMetadata`?
2.`MetadataForPayment`,`MetadataForNFTokenMint`?

mvadari: definitely the former over the latter

ckniffen: Agreed.  I like `PaymentMetadata` and `NFTokenMintMetadata`.

ckniffen (approved):

pdp2121 (approved):

PR #2451, opened 2023-08-29
fix: Avoid casting when using `Currency` type for AMM objects

JST5000:

Makes the interface of `XRP` and `IssuedCurrency` compatible to avoid casting when referencing `Currency.issuer`

### Context of Change

Found when trying to access `Currency.issuer` in the explorer while trying to add the support for `AMMDelete`

### Type of Change

## Test Plan

Manually via type checking when experimenting with types in the explorer.


ckniffen (approved):

mvadari (approved):
Nice!

mvadari (approved):

PR #2452, opened 2023-08-31
fix: faucet errors not showing returned body

ckniffen:

`fetch` only lets you read the body once so we need to save the variable

Before it would throw `TypeError: body used already for: https://faucet.devnet.rippletest.net/accounts`

### Type of Change

## Test Plan

Needs tests written for mocking failed requests.


packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -158,22 +158,22 @@ export async function requestFunding(
     body: JSON.stringify(postBody),
   })
 
-  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- It's a FaucetWallet
-  const body = (await response.json()) as FaucetWallet
-  // "application/json; charset=utf-8"
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- it can be anything
```
justinr1234, line 161: Confused what you mean

justinr1234: It seems fetch and fetch.json don't have template types :(

Would it be easier to just assert the type?

ckniffen: The fetch api only lets you retrieve the body once.  At this point it could be an error and not guaranteed to be a FaucetWallet.  That is why the casting was moved lower.

justinr1234 (approved):

packages/xrpl/test/integration/fundWallet.test.ts
```diff
@@ -150,4 +151,26 @@ describe('fundWallet', function () {
...
+        amount: '-1000',
+        usageContext: 'integration-test',
+      })
+
+      throw new Error('Error not thrown')
+    } catch (error) {
+      await api.disconnect()
+      expect(error).toEqual(
+        new XRPLFaucetError(
+          'Request failed: {"body":{"error":"Invalid amount","detail":"Must be an integer"},"contentType":"application/json; charset=utf-8","statusCode":400}',
```
jonathanlei, line 171: nit: can't be fixed here but shouldn't the message say "Must be an positive integer"?

jonathanlei (approved):
LGTM

packages/xrpl/test/integration/fundWallet.test.ts
ckniffen: This is the response that comes back from the faucet server.  This test gets the live response from the testnet faucet.

PR #2471, opened 2023-09-11
test: add AMM integration tests

khancode:

Adds AMM integration tests.

### Context of Change

Adds AMM integration tests.

Transactions:
- `AMMCreate`
- `AMMDeposit`
- `AMMWithdraw`
- `AMMVote`
- `AMMBid`

Requests:
- `amm_info`

`AMMDelete` is omitted since it requires for at least 512 trustlines to be created which seems overkill for an integ test. There's already a unit test for the model so that should be sufficient since it only requires 2 params (asset pair).

### Type of Change

- Not applicable for integ tests.

## Test Plan

Run this against the CI once the amendment is added to the Docker container.

## Future Tasks

- [x] Add AMM amendment to Docker container.


.github/workflows/nodejs.yml
```diff
@@ -105,7 +105,7 @@ jobs:
 
       - name: Run docker in background
         run: |
-          docker run --detach --rm --name rippled-service -p 6006:6006 --volume "${{ github.workspace }}/.ci-config/":"/config/" --health-cmd="wget localhost:6006 || exit 1" --health-interval=5s --health-retries=10 --health-timeout=2s --env "ENV_ARGS=-a --start" --env GITHUB_ACTIONS=true --env CI=true xrpllabsofficial/xrpld:1.12.0-b1
+          docker run --detach --rm --name rippled-service -p 6006:6006 --volume "${{ github.workspace }}/.ci-config/":"/config/" --health-cmd="wget localhost:6006 || exit 1" --health-interval=5s --health-retries=10 --health-timeout=2s --env "ENV_ARGS=-a --start" --env GITHUB_ACTIONS=true --env CI=true xrpllabsofficial/xrpld:1.12.0
```
mvadari, line 108: Can you make the version a YAML variable so it's easier to edit instead of being buried in the command?

packages/xrpl/HISTORY.md
```diff
@@ -3,6 +3,7 @@
 Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xrpl-announce) for release announcements. We recommend that xrpl.js (ripple-lib) users stay up-to-date with the latest stable release.
 
 ## Unreleased
+* Add AMM integration tests
```
mvadari, line 6: This doesn't need to be in the history since it doesn't affect how people use it.

packages/xrpl/test/integration/transactions/ammBid.test.ts
```diff
@@ -0,0 +1,191 @@
...
+  let lptoken: IssuedCurrencyAmount
+
+  beforeAll(async () => {
+    testContext = await setupClient(serverUrl)
+    wallet = testContext.wallet
+    wallet2 = await generateFundedWallet(testContext.client)
+    wallet3 = await generateFundedWallet(testContext.client)
+    currencyCode = 'USD'
+
+    const ammInfoRes = await setupAMMPool(
```
mvadari, line 28: Should there be some sort of "reusable value" generation like we have in xrpl-py for this, so you don't need to generate an AMM for every single test? It would be useful for XLS-38d integration tests as well.

khancode: Yeah that would definitely improve efficiency. But I'm trying to determine where the generated AMM would be saved to be queried by all tests. I'm thinking it can could be added to the test context`XrplIntegrationTestContext`.

packages/xrpl/test/integration/transactions/ammDeposit.test.ts
```diff
@@ -0,0 +1,312 @@
...
+      asset2: {
+        currency: currencyCode,
+        issuer: wallet2.classicAddress,
+      },
+    })
+
+    const { amm } = ammInfoRes.result
+    const { amount, amount2, lp_token } = amm
+
+    if (typeof amount !== 'string') {
```
ckniffen, line 216: For these types of tests you can use `expect(amount).toBeInstanceOf(String)`  That will turn these assertions into one line.

mvadari:
Can you test the `AMMDelete` model by hand once just to make sure there isn't a serious issue?

packages/xrpl/test/integration/transactions/ammBid.test.ts
```diff
@@ -0,0 +1,191 @@
...
+        currency: currencyCode,
+        issuer: wallet2.classicAddress,
+      },
+    })
+
+    const { amm } = ammInfoRes.result
+    const { auction_slot, lp_token } = amm
+
+    if (auction_slot === undefined) {
+      throw new Error('auction_slot should not be undefined')
```
mvadari, line 103:
```suggestion
      throw new Error('auction_slot should be defined')
```

mvadari: I believe we prefer false asserts over throwing an error for integration tests.

packages/xrpl/test/integration/transactions/ammBid.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
```
mvadari, line 104:
```suggestion
    assert.true(auction_slot === undefined, 'auction_slot should be defined')
```
(I don't remember the exact syntax for this)

ckniffen: `assert.ok` works as it checks if it is truthy.

packages/xrpl/test/integration/utils.ts
```diff
@@ -10,13 +10,21 @@ import {
...
   NotConnectedError,
+  Currency,
 } from '../../src'
 import { Payment, Transaction } from '../../src/models/transactions'
 import { hashSignedTx } from '../../src/utils/hashes'
 
 const masterAccount = 'rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh'
 const masterSecret = 'snoPBrXtMeMyMHUVTgbuqAfg1SUTb'
 
+export interface SetupAMMPool {
```
mvadari, line 21: IMO this should just be `AMMPool` - I saw the import in `setup.ts` and thought you had mis-capitalized a function.

packages/xrpl/test/integration/utils.ts
```diff
@@ -337,3 +346,13 @@ export async function waitForAndForceProgressLedgerTime(
...
+
+/**
+ * Sets up an AMM pool for integration testing.
+ *
+ * @param client - The client object.
+ * @param lpWallet - The liquidity provider wallet that creates the AMM pool.
+ * @param issuerWallet - The issuer wallet of the new token (using currencyCode) to be as asset in the AMM pool.
+ * @param currencyCode - The currency code of the new token to be an asset in the AMM pool.
+ * @returns - A promise that resolves to the amm_info response of the AMM pool.
+ */
```
mvadari, line 358: ????

packages/xrpl/test/integration/setup.ts
```diff
@@ -32,15 +50,100 @@ async function connectWithRetry(client: Client, tries = 0): Promise<void> {
 export async function setupClient(
```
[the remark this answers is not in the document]
khancode: Ah you're right. Let me see if I can add cache logic within here for `setupAMMPool()`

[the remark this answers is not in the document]
mvadari: The way I did it is with a global variable that is only instantiated if it's `undefined`

ckniffen (approved):

packages/xrpl/test/integration/transactions/ammCreate.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+describe('AMMCreate', function () {
+  let testContext: XrplIntegrationTestContext
+
+  beforeAll(async () => {
+    testContext = await setupClient(serverUrl)
+  })
+  afterAll(async () => teardownClient(testContext))
+
+  it('base', async function () {
```
justinr1234, line 21: Could you add more of what it "should" do? Across all the files

mvadari: That's what the model docs are for.

justinr1234: When the test runs it will show `ammCreate base` which doesn't seem useful when you're trying to follow it. It's a nit though, so up to omar

mvadari: That's what all the basic tests show for the models.

khancode: I'd like to keep it consistent with all test models. I think base is self-explanatory.

packages/xrpl/test/integration/setup.ts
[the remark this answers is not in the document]
khancode: Jest doesn't support global test data to be used among tests, so the only solution I came up with is by saving to an external json file. I was going to use this method for AMM integ tests to use one amm pool for all tests but since the test run order is indeterministic, the assertion values would fluctuate a lot.

However, I think this will still be beneficial for sidechains integ tests so here's my POC - https://github.com/XRPLF/xrpl.js/pull/2557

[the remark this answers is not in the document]
mvadari: It was established offline that browser tests fail because they use jasmine, not jest.

justinr1234 (approved):
Nice work on this!

PR #2474, opened 2023-09-12
fix: fix escrow `OfferSequence` validation

mvadari:

This PR fixes a couple of validation issues with `EscrowCancel` and `EscrowFinish`. It also fixes the issue a lot of VS Code devs are having where the `HISTORY.md` file automatically reformats via `prettier` every time it's edited and saved, and adjusts the linter rules for `validate` functions.

### Context of Change

This was an issue discovered during the Delft hackathon.

### Type of Change

- [x] Yes

## Test Plan

Fixed + added tests.

## Future Tasks
These validations likely need to be modified for all `UInt32` types in the repo.


packages/xrpl/test/models/escrowFinish.test.ts
```diff
@@ -51,7 +58,7 @@ describe('EscrowFinish', function () {
   })
 
   it(`throws w/ invalid OfferSequence`, function () {
-    escrow.OfferSequence = '10'
+    escrow.OfferSequence = 'ten'
```
JST5000, line 61: nit: I'd use a different string like `abcd` instead of `ten` because I can totally see myself reading this and accidentally internally thinking "10" while I look for the problem elsewhere hahaha

JST5000 (approved):
LGTM

ckniffen (approved):

PR #2491, opened 2023-09-28
feat: add support for XLS-40d + add script to auto-generate models from rippled code

mvadari:

This PR adds support for XLS-40 and adds a script to automatically generate transaction models from rippled source code.

### Context of Change

https://github.com/XRPLF/XRPL-Standards/pull/136
https://github.com/XRPLF/rippled/pull/4636

### Type of Change

### Did you update CHANGELOG.md?

- [x] Yes

## Test Plan

Added model tests (integration tests will be added later). CI passes.


packages/xrpl/tools/createValidateTests.js
```diff
@@ -0,0 +1,197 @@
+const fs = require('fs')
```
mvadari, line 1: This is mostly the same as the old `createValidateTests` file (I didn't even change very much), but I guess git doesn't think they're the same file anymore.

packages/xrpl/src/models/transactions/DIDSet.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * @category Transaction Models
+ */
+export interface DIDSet extends BaseTransaction {
+  TransactionType: 'DIDSet'
+
+  Data?: string
+
+  DIDDocument?: string
```
ckniffen, line 16: Should these be documented?

packages/xrpl/tools/generateModels.js
```diff
@@ -0,0 +1,265 @@
...
+    fs.writeFileSync(
+      path.join(path.dirname(__filename), `../test/models/${tx}.test.ts`),
+      validateTests,
+    )
+
+    updateTransactionFile(transactionMatch, tx)
+
+    updateIndexFile(tx)
+
+    console.log('Added ' + tx)
```
ckniffen, line 255: Use template string

packages/xrpl/tools/generateModels.js
```diff
@@ -0,0 +1,265 @@
...
+  allTransactions.sort()
+  const newTransactionMatch =
+    firstLine + '\n  | ' + allTransactions.join('\n  | ')
+  let newJsTxFile = jsTransactionFile.replace(
+    transactionMatch,
+    newTransactionMatch,
+  )
+
+  newJsTxFile = newJsTxFile.replace(
+    `import {
```
ckniffen, line 111: This feels oddly specific to one type.

packages/ripple-binary-codec/src/enums/definitions.json
```diff
@@ -140,40 +141,40 @@
     [
       "LedgerEntry",
       {
-        "nth": 1,
+        "nth": 257,
```
ckniffen, line 144: Why did these fields change?

mvadari: Fixing discrepancies between the `server_definitions` RPC and the `xrpl-codec-gen` script. Tests all pass because these fields aren't used for signing, they only exist for rippled technical reasons.

packages/xrpl/tools/generateModels.js
mvadari: It's the last type in the list, it was easier to do it this way and have the linter fix the sorting than do some complex regex that sorts alphabetically.

packages/xrpl/src/models/transactions/DIDSet.ts
mvadari: There's no documentation yet, I will add it when it's available. But that won't be for another couple of weeks.

packages/xrpl/tools/generateModels.js
mvadari: I don't think it matters much here

ckniffen (approved):

packages/xrpl/src/models/transactions/transaction.ts
```diff
@@ -91,18 +93,20 @@ import {
...
-  | AccountSet
   | AMMBid
+  | AMMCreate
   | AMMDelete
   | AMMDeposit
-  | AMMCreate
   | AMMVote
   | AMMWithdraw
+  | AccountDelete
+  | AccountSet
```
JST5000, line 103: nit: Alphabetically `AccountDelete` and `AccountSet` should be before `AMM...`

JST5000: (Same nit applies below with the case statement if we want to maintain alphabetical order)

packages/xrpl/test/integration/transactions/didDelete.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      // confirm that the DID no longer exists
+      const accountOffersResponse = await testContext.client.request({
+        command: 'account_objects',
+        account: testContext.wallet.classicAddress,
+        type: 'did',
+      })
+      assert.lengthOf(
+        accountOffersResponse.result.account_objects,
+        0,
+        'Should be no DID on the ledger',
```
JST5000, line 65: nit: More verbose error message to help debug
```suggestion
        'Should be no DID on the ledger after submitting a DIDDelete transaction',
```

packages/xrpl/test/integration/transactions/didDelete.test.ts
```diff
@@ -0,0 +1,70 @@
...
+        TransactionType: 'DIDDelete',
+        Account: testContext.wallet.classicAddress,
+      }
+
+      await testTransaction(testContext.client, tx, testContext.wallet)
+
+      // confirm that the DID no longer exists
+      const accountOffersResponse = await testContext.client.request({
+        command: 'account_objects',
+        account: testContext.wallet.classicAddress,
```
JST5000, line 59: nit: `address` is more straightforward (I'd recommend using it by default throughout this PR)
```suggestion
        account: testContext.wallet.address,
```

packages/xrpl/test/integration/transactions/didDelete.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      // double check the DID was properly created
+      const initialAccountOffersResponse = await testContext.client.request({
+        command: 'account_objects',
+        account: testContext.wallet.classicAddress,
+        type: 'did',
+      })
+      assert.lengthOf(
+        initialAccountOffersResponse.result.account_objects,
+        1,
+        'Should be exactly one DID on the ledger',
```
JST5000, line 45: nit: More verbose error message
```suggestion
        'Should be exactly one DID on the ledger after submitting a DIDSet transaction',
```

packages/xrpl/test/integration/transactions/didSet.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      // confirm that the DID was actually created
+      const accountOffersResponse = await testContext.client.request({
+        command: 'account_objects',
+        account: testContext.wallet.classicAddress,
+        type: 'did',
+      })
+      assert.lengthOf(
+        accountOffersResponse.result.account_objects,
+        1,
+        'Should be exactly one DID on the ledger',
```
JST5000, line 45: Same nit as above :)
```suggestion
        'Should be exactly one DID on the ledger after a DIDSet transaction',
```

packages/xrpl/test/models/DIDDelete.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  let tx
+
+  beforeEach(function () {
+    tx = {
+      Account: 'rfmDuhDyLGgx94qiwf3YF8BUV5j6KSvE8',
+      Fee: '10',
+      Flags: 2147483648,
+      Sequence: 4,
+      TransactionType: 'DIDDelete',
+    } as any
```
JST5000, line 21: Why is typed as `any` instead of `DIDDelete`?

packages/xrpl/test/models/DIDDelete.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  })
+
+  it('verifies valid DIDDelete', function () {
+    assert.doesNotThrow(() => validateDIDDelete(tx))
+    assert.doesNotThrow(() => validate(tx))
+  })
```
JST5000, line 27: I think there should be a test case verifying that the validate functions can throw (doesn't have to be comprehensive, just one with an error)

packages/xrpl/tools/generateModels.js
JST5000, line 255: Might be worth adding a note on what the next steps that people have to do by hand would be :)
```suggestion
    const pathname = path/to/new/model
    ...
    console.log(`Added  ${tx} to ${pathname}`)
    console.log(`Next you should: \n  1. Add docstrings to the model fields.\n  2. Add an integration test for the model like ${pathToExample}`)  
```

packages/xrpl/tools/generateModels.js
JST5000: Can you add a comment with that context?

packages/xrpl/tools/generateModels.js
```diff
@@ -0,0 +1,265 @@
...
+
+function readFile(filename) {
+  return fs.readFileSync(filename, 'utf-8')
+}
+
+let jsTransactionFile
+
+function processRippledSource(folder) {
+  const sfieldCpp = readFile(
+    path.join(folder, 'src/ripple/protocol/impl/SField.cpp'),
```
JST5000, line 14: Can you add a comment explaining the setup you need to do before running this tool? 
(Ex. Do you need to build rippled, or just clone it?)

JST5000 (approved):
All my changes are docs updates, logic looks good :)

packages/xrpl/src/models/transactions/transaction.ts
mvadari: I was sorting with capitals before lowercase, which makes more sense to me, but if you disagree I'm happy to change that.

packages/xrpl/test/models/DIDDelete.test.ts
mvadari: So that you can remove required fields without TS complaining (see other model tests)

packages/xrpl/tools/generateModels.js
mvadari: Just clone it - it just needs access to the source code.

packages/xrpl/src/models/transactions/transaction.ts
JST5000: I personally don't consider capitalization when scanning lists, so I'd prefer to ignore it when sorting alphabetically

PR #2493, opened 2023-09-28
Update the dependabot interval to weekly

### LINKED ISSUE #2394: Update `dependabot` interval to twice a month
`dependabot` is checking for packages updates and security vulnerabilities monthly now. We should update the interval to bimonthly to ensure the security of our libraries.

JST5000:

Changes the dependabot interval to weekly

### Context of Change

Closes #2394 - There's unfortunately no option for twice a month, but weekly should be about as good. Anything which keeps getting re-opened we can use the dependabot commands to indicate our preferences for when to be notified about changes. (E.g. on the next major version)

### Type of Change

- [X] Automations

## Test Plan

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,6 +3,6 @@ updates:
 - package-ecosystem: npm
   directory: "/"
   schedule:
-    interval: monthly
+    interval: weekly
     time: "15:00"
   open-pull-requests-limit: 10


mvadari:
There's also a limit on how many PRs dependabot can have open, so that should prevent it from being too overwhelming.

mvadari (approved):

ckniffen (approved):

PR #2502, opened 2023-10-03
docs: Update xrpl.js keywords to make it easier to find

JST5000:

Since we've deprecated ripple-lib, we should make xrpl.js show up when they search for that instead. 

### Context of Change

Helping to make xrpl.js more discoverable, especially for people who knew older names for the library.

### Type of Change

- [X] Documentation Updates

## Test Plan

Inspection + Search npm for ripple-lib after the next release and see if this shows up.

### DIFF
--- a/packages/xrpl/package.json
+++ b/packages/xrpl/package.json
@@ -80,6 +80,10 @@
     "url": "git@github.com:XRPLF/xrpl.js.git"
   },
   "readmeFilename": "README.md",
+  "keywords": [
+    "ripple-lib",
+    "ripple"
+  ],
   "engines": {
     "node": ">=10.13.0"
   }
--- a/packages/xrpl/package.json
+++ b/packages/xrpl/package.json
@@ -82,7 +82,10 @@
   "readmeFilename": "README.md",
   "keywords": [
     "ripple-lib",
-    "ripple"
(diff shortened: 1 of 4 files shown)


jonathanlei (approved):

packages/xrpl/package.json
```diff
@@ -80,6 +80,10 @@
     "url": "git@github.com:XRPLF/xrpl.js.git"
   },
   "readmeFilename": "README.md",
+  "keywords": [
+    "ripple-lib",
+    "ripple"
+  ],
```
intelliot, line 86:
```suggestion
  "keywords": [
    "ripple-lib",
    "ripple",
    "xrp",
    "xrp ledger"
  ],
```

intelliot (approved):
any harm in adding a couple more while we're introducing the `keywords` property?

packages/xrpl/package.json
ckniffen: I recommend adding these to all of the packages.

JST5000: I added all of them except ripple-lib because I think if people are searching for ripple-lib, they really are looking just for the `xrpl` package.

ckniffen (approved):

PR #2503, opened 2023-10-04
fix: add `type` param to `ledger` & `ledger_data`

mvadari:

- Add missing `type` param
- Clean up typing of `type` param and the response property `account_objects` of the `account_objects` request.
- Create interfaces `HasPreviousTxnID` and `MissingPreviousTxnID`. This helps clean up type checking with regard to the presence of `PreviousTxnID` and `MissingPreviousTxnID`

### Context of Change

This field was missing.

https://xrpl.org/ledger_data.html

### Type of Change

- [x] Yes

## Test Plan

N/A

### DIFF
--- a/packages/xrpl/src/models/ledger/AMM.ts
+++ b/packages/xrpl/src/models/ledger/AMM.ts
@@ -1,6 +1,6 @@
 import { AuthAccount, Currency, IssuedCurrencyAmount } from '../common'
 
-import BaseLedgerEntry from './BaseLedgerEntry'
+import { BaseLedgerEntry, MissingPreviousTxnID } from './BaseLedgerEntry'
 
 export interface VoteSlot {
   VoteEntry: {
@@ -15,7 +15,7 @@ export interface VoteSlot {
  *
  * @category Ledger Entries
  */
-export default interface AMM extends BaseLedgerEntry {
+export default interface AMM extends BaseLedgerEntry, MissingPreviousTxnID {
   LedgerEntryType: 'AMM'
   /**
    * The address of the special account that holds this AMM's assets.
@@ -75,12 +75,4 @@ export default interface AMM extends BaseLedgerEntry {
    * type, so this value is always 0.
    */
   Flags: 0
-  /**
-   * This field is missing on AMM but is present on all other account_object returned objects.
-   */
-  PreviousTxnID: never
-  /**
-   * This field is missing on AMM but is present on all other account_object returned objects.
-   */
-  PreviousTxnLgrSeq: never
 }
--- a/packages/xrpl/src/models/ledger/Amendments.ts
+++ b/packages/xrpl/src/models/ledger/Amendments.ts
@@ -1,4 +1,4 @@
-import BaseLedgerEntry from './BaseLedgerEntry'
+import { BaseLedgerEntry, MissingPreviousTxnID } from './BaseLedgerEntry'
 
 /**
  * The unique id for the Amendments object https://xrpl.org/amendments-object.html#amendments-id-format
@@ -24,7 +24,9 @@ export interface Majority {
  *
  * @category Ledger Entries
  */
-export default interface Amendments extends BaseLedgerEntry {
+export default interface Amendments
+  extends BaseLedgerEntry,
+    MissingPreviousTxnID {
   LedgerEntryType: 'Amendments'
   /**
    * Array of 256-bit amendment IDs for all currently-enabled amendments. If
--- a/packages/xrpl/src/models/methods/accountObjects.ts
+++ b/packages/xrpl/src/models/methods/accountObjects.ts
@@ -1,35 +1,11 @@
-import {
-  AMM,
-  Bridge,
-  Check,
-  DepositPreauth,
-  Escrow,
-  Offer,
-  PayChannel,
-  RippleState,
-  SignerList,
-  Ticket,
-  XChainOwnedClaimID,
-  XChainOwnedCreateAccountClaimID,
-} from '../ledger'
-
+import { Amendments, FeeSettings, LedgerHashes } from '../ledger'
+import { LedgerEntry, LedgerEntryFilter } from '../ledger/LedgerEntry'
 import { BaseRequest, BaseResponse, LookupByLedgerRequest } from './baseMethod'
 
-export type AccountObjectType =
-  | 'amm'
-  | 'bridge'
-  | 'check'
-  | 'deposit_preauth'
-  | 'escrow'
-  | 'nft_offer'
-  | 'offer'
-  | 'payment_channel'
-  | 'signer_list'
-  | 'state'
-  | 'ticket'
-  | 'xchain_owned_create_account_claim_id'
-  | 'xchain_owned_claim_id'
-
+export type AccountObjectType = Exclude<
+  LedgerEntryFilter,
+  'amendments' | 'fee' | 'hashes'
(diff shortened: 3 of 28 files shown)


packages/xrpl/src/models/methods/accountObjects.ts
```diff
@@ -46,8 +46,6 @@ export interface AccountObjectsRequest
   account: string
   /**
    * If included, filter results to include only this type of ledger object.
-   * The valid types are: Check , DepositPreauth, Escrow, Offer, PayChannel,
```
mvadari, line 49: I removed this part of the docs because it's hard to keep updated (it's already missing a bunch of object types), and IMO it's worse to have this be outdated than to not have it (since the docs have the types).
https://js.xrpl.org/types/AccountObjectType.html

ckniffen (dismissed):

JST5000 (dismissed):
LGTM

JST5000 (approved):
LGTM

packages/xrpl/src/models/methods/accountObjects.ts
```diff
@@ -59,20 +37,6 @@ export interface AccountObjectsRequest
   marker?: unknown
 }
 
-/**
- * Account Objects can be a Check, a DepositPreauth, an Escrow, an Offer, a
- * PayChannel, a SignerList, a Ticket, or a RippleState.
- */
-export type AccountObject =
```
ckniffen, line 66: you could keep this the same and have it be as subset of LedgerEntry

mvadari: Looks like that might be necessary:
```
test/integration/transactions/escrowCancel.test.ts:72:42 - error TS2339: Property 'PreviousTxnID' does not exist on type 'LedgerEntry'.
xrpl:       Property 'PreviousTxnID' does not exist on type 'Amendments'.
xrpl:     72           transaction: accountObjects[0].PreviousTxnID,
xrpl:                                                 ~~~~~~~~~~~~~
```

ckniffen: Also see how we solved this with the AMM Object.  Part of this issue stems for the fact that most ledger objects have PreviousTxnID.  Actually all of them that an account owns does except for the newly added AMM one.  So we had to solve this same issue with its type by using never.

intelliot:
Regarding the PR title: the way to indicate a breaking change with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) is to use an exclamation point, e.g.
```
fix!: switch out `type` object filters to all be the same
```

and/or put "BREAKING CHANGE: <description>" in the footer (i.e. the bottom of the commit message body).

mvadari:
> Regarding the PR title: the way to indicate a breaking change with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) is to use an exclamation point, e.g.
> 
> ```
> fix!: switch out `type` object filters to all be the same
> ```
> 
> and/or put "BREAKING CHANGE: <description>" in the footer (i.e. the bottom of the commit message body).

Given that we haven't fully switched over to conventional commits (and aren't documenting our use), I'd rather keep it like this so that it's obvious. The `!` isn't yet because it's not entirely the norm. If you think I should still stick with the conventional commit formatting I'm happy to change it.

ckniffen:
> Given that we haven't fully switched over to conventional commits (and aren't documenting our use), I'd rather keep it like this so that it's obvious. The `!` isn't yet because it's not entirely the norm. If you think I should still stick with the conventional commit formatting I'm happy to change it.

I have been putting `BREAKING CHANGES` in the footer of the commit for 3.0

packages/xrpl/src/models/methods/accountObjects.ts
```diff
@@ -1,35 +1,7 @@
...
-  SignerList,
-  Ticket,
-  XChainOwnedClaimID,
-  XChainOwnedCreateAccountClaimID,
-} from '../ledger'
+import { LedgerEntryFilter, LedgerEntry } from '../ledger'
 
 import { BaseRequest, BaseResponse, LookupByLedgerRequest } from './baseMethod'
 
-export type AccountObjectType =
```
ckniffen, line 18: I find having a subset type helpful for autocomplete and type checking for example if I try `account_objects` with `fee` I would like typescript to fail.

ckniffen: This shows how to do this with fun fruits. https://fjolt.com/article/typescript-exclude-type

mvadari: rippled actually is fine with using `fee` in `account_objects` (Clio complains though)

intelliot:
I prefer conventional commits, but it's not a big deal. Go with what works best for you.

packages/xrpl/src/models/ledger/Amendments.ts
```diff
@@ -42,4 +42,13 @@ export default interface Amendments extends BaseLedgerEntry {
    * type, so this value is always 0.
    */
   Flags: 0
+
+  /**
+   * This field is missing on Amendments but is present on all other returned objects.
```
ckniffen, line 47:
```suggestion
   * This field is missing on Amendments but is present on most other returned objects.
```
Other places could be edited too.

packages/xrpl/src/models/ledger/Amendments.ts
```diff
@@ -42,4 +42,13 @@ export default interface Amendments extends BaseLedgerEntry {
...
+   */
+  PreviousTxnID: never
```
ckniffen, line 49: Would prefer for this to be abstracted into an interface for objects without these two properties.

packages/xrpl/src/models/ledger/AMM.ts
```diff
@@ -76,11 +76,11 @@ export default interface AMM extends BaseLedgerEntry {
    */
   Flags: 0
   /**
-   * This field is missing on AMM but is present on all other account_object returned objects.
+   * This field is missing on AMM but is present on all other returned objects.
    */
   PreviousTxnID: never
   /**
-   * This field is missing on AMM but is present on all other account_object returned objects.
+   * This field is missing on AMM but is present on all other returned objects.
```
ckniffen, line 83: The previous comment was more factual.

jonathanlei (approved):
LGTM other than Caleb's suggestions on docstrings

packages/xrpl/src/models/ledger/AMM.ts
mvadari: This type can be returned with other methods as well, like `ledger_entry`. It's confusing in those contexts.

packages/xrpl/src/models/methods/accountObjects.ts
```diff
@@ -98,7 +55,7 @@ export interface AccountObjectsResponse extends BaseResponse {
      * Array of objects owned by this account. Each object is in its raw
      * ledger format.
      */
-    account_objects: AccountObject[]
+    account_objects: LedgerEntry[]
```
ckniffen, line 58: Can you do something similar to AccountObjectsType and then add back AccountObject.  Small tweak will help intellisense.

ckniffen:
Updated to no longer be breaking change and targeted 2.x

khancode (approved):

jonathanlei (approved):

pdp2121 (approved):

PR #2514, opened 2023-10-09
feat: isomorphic sockets and use eventemitter3

ckniffen:

Reduces filesize by 3kb minified and gzipped or 2.8% and reduces bundler config steps.

- Move `WSWrapper` to `@xrpl/isomorphic` to remove the need to add mapping of `ws` to `WSWrapper` file in bundler configs
- Switch to `eventemitter3` which is smaller than `events` by almost 1kb and will not require a mapping to node's `events` in `vite` bundling. `webpack` always automatically maps it.
    - max listeners is not a thing for `eventemitter3` so we do not need to set it to `Infinity`. `ws` uses the native event emitter which does still need that to be set.
- Remove `eventemitter2` which was only used in tests and was replaced with `eventemitter3`

BREAKING CHANGE: Configuration for frontend bundlers has changed for `ws`

### Type of Change

- [x] Documentation Updates

- [x] Yes

### DIFF
--- a/UNIQUE_SETUPS.md
+++ b/UNIQUE_SETUPS.md
@@ -37,11 +37,6 @@ To use `xrpl.js` with React, you need to install shims for core NodeJS modules. 
       const webpack = require("webpack");
 
       module.exports = function override(config) {
-        const fallback = config.resolve.fallback || {};
-        Object.assign(fallback, {
-          ws: require.resolve("xrpl/dist/npm/client/WSWrapper"),
-        });
-        config.resolve.fallback = fallback;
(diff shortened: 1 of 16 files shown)


justinr1234:
Does this affect minimum browser versions?

ckniffen:
> Does this affect minimum browser versions?

No because we were already using native WebSocket in the browser.  This just moves it around to avoid consumers to have to specify our custom `ws` polyfill which implement an `EventEmitter`.

justinr1234 (approved):

pdp2121 (approved):

PR #2515, opened 2023-10-09
fix: pseudo-transactions for `ledger` and `tx`

pdp2121:

Currently the explicit responses for `tx` and `ledger` methods do not include psedo-transactions (EnableAmendment, UNLModify, SetFee) even though these transactions should still be found with the above methods. This PR will add these types to the responses.

The response type for `ledger` will allow specifying a custom ledger entry type.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

### DIFF
--- a/packages/xrpl/src/models/ledger/Ledger.ts
+++ b/packages/xrpl/src/models/ledger/Ledger.ts
@@ -1,4 +1,5 @@
 import { Transaction, TransactionMetadata } from '../transactions'
+import { PseudoTransaction } from '../transactions/transaction'
 
 import LedgerEntry from './LedgerEntry'
 
@@ -61,5 +62,7 @@ export default interface Ledger {
    * either JSON or binary depending on whether the request specified binary
    * as true.
    */
-  transactions?: Array<Transaction & { metaData?: TransactionMetadata }>
+  transactions?: Array<
+    (Transaction | PseudoTransaction) & { metaData?: TransactionMetadata }
+  >
 }
--- a/packages/xrpl/src/models/ledger/Ledger.ts
+++ b/packages/xrpl/src/models/ledger/Ledger.ts
@@ -1,4 +1,5 @@
-import { AnyTransaction, TransactionMetadata } from '../transactions'
+import { Transaction, TransactionMetadata } from '../transactions'
+import { PseudoTransaction } from '../transactions/transaction'
 
 import LedgerEntry from './LedgerEntry'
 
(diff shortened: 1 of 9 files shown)


packages/xrpl/src/models/methods/tx.ts
```diff
@@ -39,8 +40,9 @@ export interface TxRequest extends BaseRequest {
  *
  * @category Responses
  */
-export interface TxResponse<T extends BaseTransaction = Transaction>
-  extends BaseResponse {
+export interface TxResponse<
+  T extends BaseTransaction | PseudoTransaction = Transaction,
```
ckniffen, line 44: Dont all `PseudoTransaction` options extend BaseTransactions.

Would this work? 

```
T extends BaseTransaction = Transaction | PseudoTransaction,
```

pdp2121: I tried that earlier but `UNLModify` does not extend `BaseTransaction`

ckniffen: It should be able to.

ckniffen:
You should also update `LedgerEntryResponse` to be a generic that defaults to `LedgerEntry` this will allow people using the library in side chains/test networks to specify ledger object types that are not officially supported yet.

https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/models/methods/ledgerEntry.ts#L187C18-L202

packages/xrpl/src/models/ledger/Ledger.ts
```diff
@@ -61,5 +62,7 @@ export default interface Ledger {
    * either JSON or binary depending on whether the request specified binary
    * as true.
    */
-  transactions?: Array<Transaction & { metaData?: TransactionMetadata }>
+  transactions?: Array<
+    (Transaction | PseudoTransaction) & { metaData?: TransactionMetadata }
```
khancode, line 66:
```suggestion
    (Transaction | PseudoTransaction) & { metaData?: TransactionMetadata }
```
Is it possible to combine `Transaction | PseudoTransaction` into one type? I see typing also occurs `hashTxTree`

intelliot (approved):
I don't have a way to test this, but LGTM

khancode (approved):
LGTM. Would like to have this refactored in a separate PR with transaction type `Transaction = SubmittableTransaction | PseudoTransaction`

Issue #2517, opened 2023-10-10
Sending a payment without destination address does not raise an exception anymore


SaxenaKaustubh:
When the user sends a payment transaction using xrpl.js without specifying the destination address (set to empty string), the destination address gets auto-populated to `rrrrrrrrrrrrrrrrrrrrrhoLvTp`, no exception gets thrown and the response is null.

**XRPL.js version** 2.12.0

Issue #2475, opened 2023-09-12
Add a validation check to see if an `Address` accidentally has a `secret` instead


JST5000:
We may want to add some "common error" checks to the validation function. 

This one comes from https://github.com/XRPLF/xrpl.js/issues/2472#issuecomment-1715249081 where the error message for using a secret instead of an address was very unclear. 

These validation checks could be on the Transaction objects, the validate functions before submitting, or somewhere else. But ideally these would give clearer errors for typos / mismatched datatypes which are technically right from a type perspective, but not XRPL perspective (ex. address vs secret)

mvadari:
The easier way would be to use the `ripple-address-codec` function `isValidAddress`, which would check all of this

ckniffen:
Fixed by #2525

PR #2525, opened 2023-10-16
fix: tx fields for accounts allowing empty string

ckniffen:

Fields that were encoded as STAccounts were turning `''` into `rrrrrrrrrrrrrrrrrrrrrhoLvTp` the ACCOUNT_ZERO account. They will now throw a `ValidationError`.

Invalid addresses on a transaction now throws a `ValidationError` when submitting a transaction instead of `Error('checksum_invalid')`.

This change updates some error messages for fields a few fields that aren't accounts, `DestinationTag` or `NFTokenID`.  They will still be of type `ValidationError`.

Fixes: #2517 and #2475

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

### DIFF
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -7,9 +7,11 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 ### Fixed
 - Allow flag maps when submitting `NFTokenMint` and `NFTokenCreateOffer` transactions like others with flags
 - Add pseudo transaction types to `tx` and `ledger` methods responses.
+- Transaction fields that equate to an address no longer allow an empty string `''`. If you want to specify the genesis account you must manually specify `rrrrrrrrrrrrrrrrrrrrrhoLvTp`
 
 ### Updated
 - Make `LedgerEntryResponse` a generic so it can be used like `LedgerEntryResponse<Escrow>`
+- Error messages for fields that equate to an address, `DestinationTag`, or `NFTokenID`.  They will still be of type `ValidationError`.
 
 ## 2.12.0 (2023-09-27)
 ### Added
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -10,11 +10,13 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 * Add missing `type` param to `ledger_data` and `ledger` requests
 * Type assertions around `PreviousTxnID` and `PreviousTxnLgrSeq` missing on some ledger objects
 * Transaction fields that represent an address no longer allow an empty string (`''`). If you want to specify [ACCOUNT_ZERO](https://xrpl.org/addresses.html#special-addresses), you can specify `rrrrrrrrrrrrrrrrrrrrrhoLvTp`. ⚠️ **WARNING:** `rrrrrrrrrrrrrrrrrrrrrhoLvTp` is a black hole address, with no corresponding private key. Accounts/funds controlled by this address are not accessible.
+* Invalid addresses on a transaction now throws a `ValidationError` when submitting a transaction instead of `Error('checksum_invalid')`
 
 ### Updated
 * Make `LedgerEntryResponse` a generic so it can be used like `LedgerEntryResponse<Escrow>`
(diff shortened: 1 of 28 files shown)


packages/xrpl/test/models/NFTokenCreateOffer.test.ts
```diff
@@ -16,7 +16,7 @@ describe('NFTokenCreateOffer', function () {
       TransactionType: 'NFTokenCreateOffer',
       NFTokenID: NFTOKEN_ID,
       Amount: '1',
-      Owner: 'r9LqNeG6qHxjeUocjvVki2XR35weJ9mZgQ',
+      Owner: 'rcXY84C4g14iFp6taFXjjQGVeHqSCh9RX',
```
mvadari, line 19: Why were all of these changed?

packages/xrpl/test/models/payment.test.ts
```diff
@@ -116,12 +116,26 @@ describe('Payment', function () {
...
     )
     assert.throws(
       () => validate(paymentTransaction),
       ValidationError,
-      'PaymentTransaction: invalid Destination',
+      'Payment: invalid field Destination',
+    )
+  })
+
+  it(`throws when Destination is invalid classic address`, function () {
```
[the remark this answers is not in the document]
mvadari: Can there also be a test for an empty string?

packages/xrpl/test/models/NFTokenCreateOffer.test.ts
ckniffen: The previous value was an invalid classic address.  Looking at it now I could have just deleted a character.

ckniffen: That didnt work because it is still not valid.  The examples lower were too short.

packages/xrpl/test/models/NFTokenCreateOffer.test.ts
```diff
@@ -36,7 +36,7 @@ describe('NFTokenCreateOffer', function () {
         tfSellNFToken: true,
       },
       Expiration: 1000,
-      Destination: 'r9LqNeG6qHxjeUocjvVki2XR35weJ9mZgQ',
+      Destination: 'rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn',
```
ckniffen: Reverting this one because it did not need to change.

packages/xrpl/HISTORY.md
```diff
@@ -9,10 +9,12 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 * Add pseudo transaction types to `tx` and `ledger` method responses.
 * Add missing `type` param to `ledger_data` and `ledger` requests
 * Type assertions around `PreviousTxnID` and `PreviousTxnLgrSeq` missing on some ledger objects
+* Transaction fields that equate to an address no longer allow an empty string `''`. If you want to specify the genesis account you must manually specify `rrrrrrrrrrrrrrrrrrrrrhoLvTp`
```
intelliot, line 12:
```suggestion
* Transaction fields that represent an address no longer allow an empty string (`''`). If you want to specify [ACCOUNT_ZERO](https://xrpl.org/addresses.html#special-addresses), you can specify `rrrrrrrrrrrrrrrrrrrrrhoLvTp`. ⚠️ **WARNING:** `rrrrrrrrrrrrrrrrrrrrrhoLvTp` is a black hole address, with no corresponding private key. Accounts/funds controlled by this address are not accessible.
```

intelliot (changes requested):
The genesis account is `rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh`, not `rrrrrrrrrrrrrrrrrrrrrhoLvTp`.

`rrrrrrrrrrrrrrrrrrrrrhoLvTp` is `0` (or the empty string) encoded with the XRPL's base58 address format. Named ACCOUNT_ZERO, it is a known black hole address, meaning that it was not derived from a secret key.

packages/xrpl/HISTORY.md
mvadari: `rrrrrrrrrrrrrrrrrrrrrhoLvTp` is not the genesis account, it's the zero account

intelliot (approved):

PR #2526, opened 2023-10-16
feat: remove `Buffer` support and bundle polyfill

ckniffen:

- Removes need for bundlers to polyfill the `Buffer` class. `UInt8Array` are used instead which are native to the browser and node.
- Reduces bundle size 7.1kb gzipped and eliminates 4 runtime dependencies: `base-x`, `base64-js`, `buffer`, and `ieee754`.

Breaking Change: All methods that previously took a `Buffer` now accept a `UInt8Array`.

### Type of Change

- [x] Documentation Updates
- [ ] Release

- [x] Yes


packages/isomorphic/src/utils/shared.ts
```diff
@@ -0,0 +1,41 @@
+import { concatBytes } from '@noble/hashes/utils'
+
+export function concat(views: Uint8Array[]): Uint8Array {
+  return concatBytes(...views)
+}
+
+//
+// export function concat(views: ArrayBufferView[]) {
```
ckniffen: Haven't decided if this util should be used always or if i needs to have separate node and browser implementations.

packages/isomorphic/src/utils/index.ts
```diff
@@ -70,3 +70,6 @@ export const randomBytes: typeof RandomBytesFn = (size) => {
   return toUint8Array(cryptoRandomBytes(size))
 }
 /* eslint-enable func-style */
+export * from './shared'
+// eslint-disable-next-line node/prefer-global/text-encoder, node/prefer-global/text-decoder
+export { TextDecoder, TextEncoder } from 'util'
```
ckniffen, line 75: Needed because I was getting errors in several places when using them directly.

packages/ripple-address-codec/src/xrp-codec.ts
```diff
@@ -258,3 +251,9 @@ export function isValidClassicAddress(address: string): boolean {
   }
   return true
 }
+
+function checkByteLength(bytes: ByteArray, expectedLength: number): boolean {
```
ckniffen, line 255: Had to add this checker because arrays and Uint8Array length's are different.  Looking back encodeChecked was typed to just `Buffer` but was expanded in the noble libs PR.  If keep it more strictly typed we can eliminate this checking logic.

packages/ripple-address-codec/test/xrp-codec.test.ts
```diff
@@ -11,12 +13,12 @@ import {
   isValidClassicAddress,
 } from '../src'
 
-function toHex(bytes: Buffer): string {
-  return Buffer.from(bytes).toString('hex').toUpperCase()
+function toBytes(hex: string): Uint8Array {
```
ckniffen, line 16: TODO: use `hexToBytes` directly in the tests

packages/ripple-binary-codec/src/utils.ts
```diff
@@ -0,0 +1,86 @@
+import TypedArray = NodeJS.TypedArray
+
+export function writeUInt8(
```
ckniffen, line 3: TODO: Clean up and document.

These were helpful methods that lived directly on the `Buffer` object.

packages/ripple-binary-codec/src/shamap.ts
```diff
@@ -3,13 +3,12 @@ import { HashPrefix } from './hash-prefixes'
...
 
 /**
  * Abstract class describing a SHAMapNode
  */
 abstract class ShaMapNode {
-  abstract hashPrefix(): Buffer
+  abstract hashPrefix(): Uint8Array
   abstract isLeaf(): boolean
   abstract isInner(): boolean
   abstract toBytesSink(list: BytesList): void
```
sublimator, line 14: toBytesSink was never meant to have a dependency on a particular instance of a BytesSink ... BytesList should "satisifies"/"implements" BytesSink, simply `{ put(b: Uint8Array) }`

so, while you're at it :)

packages/ripple-binary-codec/src/utils.ts
sublimator: DataView ?

packages/ripple-binary-codec/test/binary-serializer.test.js
```diff
@@ -151,10 +150,18 @@ function check(type, n, expected) {
...
     }
     serializer.writeType(type, n)
-    expect(bl.toBytes()).toEqual(Buffer.from(expected))
+    expect(bl.toBytes()).toEqual(Uint8Array.from(expected))
   })
 }
 
+test(`Uint16 serializes 5 as 0,5`, function () {
+  const bl = new BytesList()
+  const serializer = new BinarySerializer(bl)
```
sublimator, line 159: BinarySerializer should just take a sink, no hard list dep

packages/xrpl/src/client/connection.ts
```diff
@@ -68,10 +68,10 @@ function createWebSocket(
     options.headers = config.headers
   }
   if (config.authorization != null) {
-    const base64 = Buffer.from(config.authorization).toString('base64')
+    // const base64 = Buffer.from(config.authorization).toString('base64')
```
sublimator, line 71: TODO

packages/ripple-binary-codec/src/utils.ts
ckniffen: It is the inner type of a UInt array.  Usefully for working with the different implementations of a TypedArray

UNIQUE_SETUPS.md
```diff
@@ -18,9 +18,7 @@ To use `xrpl.js` with React, you need to install shims for core NodeJS modules.
 1. Install shims (you can use `yarn` as well):
 
    ```shell
-   npm install --save-dev \
-       buffer \
-       process
+   npm install --save-dev process
```
ckniffen, line 21: The integration tests need process to be polyfilled but I dont think running it regularly does.  I need to come up with a good way to make it clear that you shouldnt need any polyfills with typical operation.

JST5000: If it is only necessary for testing, I wouldn't include it in these steps. Instead I'd just make a note in the `CONTRIBUTING.md`. 

(Is there a reason we don't include `process` as a dev dependency to automatically install it for people testing, or would that conflict with another package we use?

packages/isomorphic/src/utils/browser.ts
```diff
@@ -13,5 +12,24 @@ export const bytesToHex: typeof BytesToHexFn = (bytes) => {
   return hex.toUpperCase()
 }
 
-export const hexToBytes: typeof HexToBytesFn = nobleHexToBytes
+// A clone of hexToBytes from @noble/hashes without the length checks. This allows us to do our own checks.
+/* eslint-disable-next-line func-style -- Typed to ensure uniformity between node and browser implementations and docs */
```
JST5000, line 16: nit: This comment doesn't quite make sense to me considering the es-lint error is for using the `const hexToBytes = (hex) => {...}` syntax for the function instead of `function hexToBytes(hex) {...}`, but the comment is talking about typing.

packages/isomorphic/src/utils/index.ts
JST5000: Could you add that as a comment? (Along with the specific errors you would get if you remember?)

packages/ripple-address-codec/README.md
```diff
@@ -67,7 +67,7 @@ Check whether a classic address (starting with `r`...) is valid.
 
 Returns `false` for X-addresses (extended addresses). To validate an X-address, use `isValidXAddress`.
 
-### encodeSeed(entropy: Buffer, type: 'ed25519' | 'secp256k1'): string
+### encodeSeed(entropy: UInt8Array, type: 'ed25519' | 'secp256k1'): string
```
JST5000, line 70: Make sure to list all these changed functions in the migration guide if you haven't already :)

packages/ripple-address-codec/src/index.ts
```diff
@@ -90,38 +92,38 @@ function xAddressToClassicAddress(xAddress: string): {
...
 function decodeXAddress(xAddress: string): {
-  accountId: Buffer
+  accountId: Uint8Array
   tag: number | false
   test: boolean
 } {
   const decoded = codec.decodeChecked(xAddress)
   /* eslint-disable @typescript-eslint/naming-convention --
    * TODO 'test' should be something like 'isTest', do this later
    */
```
JST5000, line 102: nit: Can you resolve this TODO while you're here?

packages/ripple-address-codec/src/xrp-codec.ts
```diff
@@ -2,30 +2,24 @@
...
+  private readonly _codec: BytesCoder
 
-  public constructor(options: {
-    sha256: (bytes: ByteArray) => Uint8Array
-    alphabet: string
-  }) {
+  public constructor(options: { sha256: (bytes: ByteArray) => Uint8Array }) {
     this._sha256 = options.sha256
-    this._alphabet = options.alphabet
-    this._codec = baseCodec(this._alphabet)
```
JST5000, line 22: Note to make sure this gets a special note in the migration guide since this is a reduction in functionality for this class.

packages/isomorphic/src/utils/browser.ts
ckniffen: This is a complaint about `typeof HexToBytesFn` which is needed to ensure uniformity.

packages/ripple-address-codec/src/index.ts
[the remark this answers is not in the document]
ckniffen: That change doesn't feel too important to me.

packages/ripple-address-codec/src/xrp-codec.ts
```diff
@@ -97,42 +91,42 @@ class Codec {
...
+    return this._encodeRaw(Uint8Array.from(concatArgs(bytes, check)))
   }
 
-  public decodeChecked(base58string: string): Buffer {
-    const buffer = this._decodeRaw(base58string)
-    if (buffer.length < 5) {
+  public decodeChecked(base58string: string): Uint8Array {
+    const intArray = this._decodeRaw(base58string)
+    if (intArray.byteLength < 5) {
       throw new Error('invalid_input_size: decoded data must have length >= 5')
```
JST5000, line 102: This error message seems incorrect for what's being checked here

packages/ripple-address-codec/test/utils.test.ts
```diff
@@ -9,20 +9,20 @@ it('elements must be in the same order', () => {
 })
 
 it('sequences do not need to be the same type', () => {
```
JST5000, line 11: This test name should change to something like `Uint8Array is equivalent to other representations`

JST5000: Or `sequences with multiple values` to match the naming convention of the other `it(...)` cases

packages/ripple-address-codec/src/xrp-codec.ts
JST5000: Is this function used in multiple files? (If so, can we pull it into a shared spot?)

packages/ripple-binary-codec/src/shamap.ts
JST5000: So instead of having an abstract function here, just including a concrete one like:

```
toBytesSink(list: BytesList) {
  ...
}
```

(I don't really follow what you mean by `{ put(b: Uint8Array }` or that BytesList satisfies BytesSink unfortunately)

packages/ripple-binary-codec/src/types/account-id.ts
```diff
@@ -76,9 +79,7 @@ class AccountID extends Hash160 {
    * @returns the base58 string defined by this.bytes
    */
   toBase58(): string {
-    /* eslint-disable @typescript-eslint/no-explicit-any */
-    return encodeAccountID(this.bytes as any)
-    /* eslint-enable @typescript-eslint/no-explicit-any */
+    return encodeAccountID(this.bytes)
```
JST5000, line 82: Nice :)

packages/ripple-binary-codec/src/types/uint-16.ts
```diff
@@ -41,7 +46,7 @@ class UInt16 extends UInt {
    * @returns the number represented by this.bytes
    */
   valueOf(): number {
-    return this.bytes.readUInt16BE(0)
+    return parseInt(readUInt16BE(this.bytes, 0))
```
JST5000, line 49: Why is this wrapped in a parseInt? (It seems that previously this worked with just the `readUInt16BE` function, which makes me wonder if the implementation in `utils` has the wrong signature?

packages/ripple-binary-codec/src/types/uint-32.ts
```diff
@@ -48,7 +52,7 @@ class UInt32 extends UInt {
    * @returns the number represented by this.bytes
    */
   valueOf(): number {
-    return this.bytes.readUInt32BE(0)
+    return parseInt(readUInt32BE(this.bytes, 0), 10)
```
JST5000, line 55: Ditto

packages/ripple-binary-codec/src/types/uint.ts
```diff
@@ -47,6 +47,14 @@ abstract class UInt extends Comparable<UInt | number> {
    * @returns the value
    */
   abstract valueOf(): number | bigint
+
+  static checkUintRange(val: number, min: number, max: number): void {
```
JST5000, line 51: Should this be re-used in `uint8`? (Logic seems to be rewritten there)

packages/ripple-binary-codec/src/utils.ts
```diff
@@ -0,0 +1,149 @@
+// Don't know why this comes from NodeJS as it is valid in the browser
```
JST5000, line 1: nit: I prefer writing in an instructional tone rather than a 1st person one when committing. Conveys the same information without conveying the feeling of confusion since the confusion can be a bit distracting.
```suggestion
// Even though this comes from NodeJS it is valid in the browser
```

packages/ripple-binary-codec/src/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+export function writeUInt16BE(
+  array: Uint8Array,
+  value: number,
+  offset: number,
+): void {
+  value = Number(value)
+
+  array[offset++] = value >>> 8
+  array[offset++] = value
```
JST5000, line 34: This modifies the offset parameter - I'd prefer to see `(offset)` and `(offset + 1)` instead to avoid any confusion (although I know that in JavaScript it's pass-by-value instead of pass-by-reference, it took me a sec to verify whether this had side-effects or not)

I also think it's easy to misread this as `offset + 1` and `offset + 2` since the prepended ++offset is easily confused with appended offset++.

JST5000, line 34: This modifies the offset parameter - I'd prefer to see `(offset)` and `(offset + 1)` instead to avoid any confusion (although I know that in JavaScript it's pass-by-value instead of pass-by-reference, it took me a sec to verify whether this had side-effects or not)

I also think it's easy to misread this as `offset + 1` and `offset + 2` since the prepended ++offset is easily confused with appended offset++.

```suggestion
  array[offset] = value >>> 8
  array[offset + 1] = value
```

JST5000: (Unless I'm wrong and there is a side effect from modifying the parameter which ends up being used - in which case that's very surprising, and out of line with what the similar functions here do)

packages/ripple-binary-codec/src/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+  offset: number,
+): void {
+  array[offset] = (value >>> 24) & 0xff
+  array[offset + 1] = (value >>> 16) & 0xff
+  array[offset + 2] = (value >>> 8) & 0xff
+  array[offset + 3] = value & 0xff
+}
+
+/**
+ * Reads an unsigned, big-endian 16-bit integer from buf at the specified offset.
```
JST5000, line 55: `buf` was the old name for the parameter I think
```suggestion
 * Reads an unsigned, big-endian 16-bit integer from array at the specified offset.
```

JST5000: Please apply this to all the docstrings in this file

JST5000: (Maybe better to say `from the array` instead of `from array` since array can be confused for a word instead of the name of a parameter as it's being used in this case

packages/ripple-binary-codec/src/utils.ts
```diff
@@ -0,0 +1,149 @@
...
+  const ub = new Uint32Array(b.buffer, b.byteOffset, b.byteLength / 4)
+  return compare(ua, ub)
+}
+
+/**
+ * Compare two TypedArrays
+ * @param a first array to compare
+ * @param b second array to compare
+ */
+export function compare(a: TypedArray, b: TypedArray): 1 | -1 | 0 {
```
JST5000, line 127: Please also check the lengths of these functions. 

As written I believe this would incorrectly return 0 for `a = [1]` and `b = [1, 2]` when it should return `-1` or maybe throw an error?

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -176,15 +178,34 @@ function bufferToArray(buf: Buffer): number[] {
...
-    16,
-  )
+function swap128(buf: Uint8Array): Uint8Array {
+  // // Interprets buffer as an array of (two, in this case) 64-bit numbers and swaps byte order in-place.
+  const reversedBytes = swap64(buf)
+  // // Swap the two 64-bit numbers since our buffer is 128 bits.
+  return concat([reversedBytes.slice(8, 16), reversedBytes.slice(0, 8)])
+}
+
+function swap(b: Uint8Array, n: number, m: number): void {
```
JST5000, line 188: nit: This function order is a bit confusing. I'd recommend ordering them in `swap`, `swap64`, `swap128` so that by the time you read the last one, you know all the details it depends on.

JST5000: (The opposite order is also fine - it's just odd to have a function in between two functions which both depend on it)

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -176,15 +178,34 @@ function bufferToArray(buf: Buffer): number[] {
...
+  const i = b[n]
+  // eslint-disable-next-line no-param-reassign -- we have to swap
+  b[n] = b[m]
+  // eslint-disable-next-line no-param-reassign -- see above
+  b[m] = i
+}
+
+function swap64(arr: Uint8Array): Uint8Array {
```
JST5000, line 196: This seems like it should error if `len >= 192` instead of just returning an unswapped array.

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -176,15 +178,34 @@ function bufferToArray(buf: Buffer): number[] {
...
-function swap128(buf: Buffer): Buffer {
-  // Interprets buffer as an array of (two, in this case) 64-bit numbers and swaps byte order in-place.
-  const reversedBytes = buf.swap64()
-
-  // Swap the two 64-bit numbers since our buffer is 128 bits.
-  return Buffer.concat(
-    [reversedBytes.slice(8, 16), reversedBytes.slice(0, 8)],
-    16,
-  )
+function swap128(buf: Uint8Array): Uint8Array {
```
JST5000, line 181: nit: `buf should maybe be renamed to `arr` or `array` here and throughout functions in this file to match similar naming convention changes

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -176,15 +178,34 @@ function bufferToArray(buf: Buffer): number[] {
...
+  // eslint-disable-next-line no-param-reassign -- we have to swap
+  b[n] = b[m]
+  // eslint-disable-next-line no-param-reassign -- see above
+  b[m] = i
+}
+
+function swap64(arr: Uint8Array): Uint8Array {
+  const len = arr.length
+
+  if (len < 192) {
```
JST5000, line 199: I'm confused where this 192 number comes from, it doesn't seem to be from the 64, and doesn't match the name of the function.

packages/xrpl/src/Wallet/rfc1751.ts
```diff
@@ -176,15 +178,34 @@ function bufferToArray(buf: Buffer): number[] {
...
-  const reversedBytes = buf.swap64()
-
-  // Swap the two 64-bit numbers since our buffer is 128 bits.
-  return Buffer.concat(
-    [reversedBytes.slice(8, 16), reversedBytes.slice(0, 8)],
-    16,
-  )
+function swap128(buf: Uint8Array): Uint8Array {
+  // // Interprets buffer as an array of (two, in this case) 64-bit numbers and swaps byte order in-place.
+  const reversedBytes = swap64(buf)
```
JST5000, line 183: This puts 128 bytes into a 64 byte swap, I wouldn't expect this to work.

packages/xrpl/src/utils/stringConversion.ts
```diff
@@ -6,7 +13,7 @@
  * @category Utilities
  */
 function convertStringToHex(string: string): string {
```
JST5000, line 15: Should this be used in place of `bytesToHex(new TextEncoder().encode(string))` in more places in the code? (I think I saw instances of that in files I reviewed, but I am not sure if they were from within the xrpl package or other ones which wouldn't have access to this helper)

webpack.config.js
```diff
@@ -13,10 +13,7 @@ function getDefaultConfiguration() {
     },
     stats: "errors-only",
     devtool: "source-map",
-    plugins: [
-      new webpack.ProvidePlugin({ process: "process/browser" }),
-      new webpack.ProvidePlugin({ Buffer: ["buffer", "Buffer"] }),
-    ],
+    plugins: [new webpack.ProvidePlugin({ process: "process/browser" })],
```
JST5000, line 16: Can we only include `process/browser` for tests? (Or does it matter?)

packages/ripple-binary-codec/src/utils.ts
```diff
@@ -0,0 +1,149 @@
+// Don't know why this comes from NodeJS as it is valid in the browser
+import TypedArray = NodeJS.TypedArray
+
+//
```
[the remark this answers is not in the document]
ckniffen: Fixed in aa343f183a934ea60f37fec6f00d16a7b20ce89b

packages/ripple-binary-codec/src/utils.ts
ckniffen: Fixed in aa343f183a934ea60f37fec6f00d16a7b20ce89b

packages/xrpl/src/Wallet/rfc1751.ts
ckniffen: I derived the code from this.  https://github.com/nodejs/node/blob/95b198911382b0aa0a0448e2a86214558665e02b/lib/buffer.js#L1198-L1215. I am now seeing the issue though.  I just return it non swapped if it is greater than 192.

packages/ripple-binary-codec/src/types/uint-16.ts
ckniffen: Used to make `BigInt`s else where

packages/xrpl/src/Wallet/rfc1751.ts
ckniffen: Fixed in 2d66c38a06eb47f13102054958b599bace783b9d

packages/xrpl/src/Wallet/rfc1751.ts
ckniffen: Resolved by 2d66c38a06eb47f13102054958b599bace783b9d

UNIQUE_SETUPS.md
ckniffen: Resolved by 01dce8581da25469b7a436d0c2af5e14559ff8b3

packages/xrpl/src/Wallet/rfc1751.ts
ckniffen: Resolved by 2d66c38a06eb47f13102054958b599bace783b9d

packages/xrpl/src/utils/stringConversion.ts
ckniffen: Resolved in 11dd21c101084f6c00ee9d606bbdb0ad1dc048e7

packages/isomorphic/src/utils/index.ts
ckniffen: Addressed in 11dd21c101084f6c00ee9d606bbdb0ad1dc048e7

packages/isomorphic/src/utils/browser.ts
ckniffen: Moved to a wrapping enable/disable since the two additional methods I added need it too.

Addressed in 11dd21c101084f6c00ee9d606bbdb0ad1dc048e7

packages/ripple-address-codec/README.md
JST5000: You can, but I think the migration guide is far more useful if the specific function names are available - that way when someone sees "oh shoot, `encodeSeed` is broken, 'how do I migrate encodeSeed?'" they find the migration guide, and specific instructions for it. 

The fix should be outlined in one spot (how to replace `Buffer` with `UInt8Array` in the common functions like hexToBytes, Buffer.alloc(...), etc., but then having a list at the bottom of that section with "affected functions" can help.

packages/ripple-address-codec/test/utils.test.ts
JST5000: Bump :)

JST5000 (approved):
LGTM

packages/ripple-address-codec/test/utils.test.ts
ckniffen: Addressed in 3568f04a61d7f3ea188e12d30b9c3e0497dbd0ed

pdp2121 (approved):
LGTM

khancode (approved):
LGTM, great work!

PR #2534, opened 2023-10-19
chore: switch to rippleci Docker container

mvadari:

Title says it all.

### Context of Change

This Docker container is more frequently updated as a part of rippled CI/CD. For example, it has support for XLS-38 already.

### Type of Change

### Did you update CHANGELOG.md?

- [x] Yes

## Test Plan

CI passes.


ckniffen (approved):

CONTRIBUTING.md
```diff
@@ -64,7 +64,7 @@ From the top-level xrpl.js folder (one level above `packages`), run the followin
 ```bash
 npm install
 # sets up the rippled standalone Docker container - you can skip this step if you already have it set up
-docker run -p 6006:6006 --interactive -t --volume $PWD/.ci-config:/config/ xrpllabsofficial/xrpld:1.12.0-b1 -a --start
+docker run -p 6006:6006 --interactive -t --volume $PWD/.ci-config:/opt/ripple/etc/ --platform linux/amd64 rippleci/rippled:2.0.0-b3 /opt/ripple/bin/rippled -a --conf /opt/ripple/etc/rippled.cfg
```
JST5000, line 67: Do we need to specify 2.0.0-b3 or can we specify an image that we consistently update with the latest version? (like `:latest` with Weitse's repo?)

afaik we only specified a version specifically because we needed a custom release that wasn't an official release, but if it's our docker container we can have a :develop image potentially that doesn't have the same "release only" requirement

.ci-config/rippled.cfg
```diff
@@ -105,65 +105,66 @@ validators.txt
 # Note: The version of rippled you use this config with must have an implementation for the amendments you attempt to enable or it will crash.
 # If you need the version of rippled to be more up to date, you may need to make a comment on this repo: https://github.com/WietseWind/docker-rippled
 
-[amendments]
+[features]
```
JST5000, line 108: I thought the `features` stanza didn't work on startup which is why we needed the `amendments` stanza. When you run this locally are all these amendments enabled?

.github/workflows/nodejs.yml
```diff
@@ -162,7 +165,7 @@ jobs:
 
       - name: Run docker in background
         run: |
-          docker run --detach --rm --name rippled-service -p 6006:6006 --volume "${{ github.workspace }}/.ci-config/":"/config/" --health-cmd="wget localhost:6006 || exit 1" --health-interval=5s --health-retries=10 --health-timeout=2s --env "ENV_ARGS=-a --start" --env GITHUB_ACTIONS=true --env CI=true xrpllabsofficial/xrpld:1.12.0-b1
+          docker run --detach --rm --name rippled-service -p 6006:6006 --volume "${{ github.workspace }}/.ci-config/":"/opt/ripple/etc/" --health-cmd="wget localhost:6006 || exit 1" --health-interval=5s --health-retries=10 --health-timeout=2s --env GITHUB_ACTIONS=true --env CI=true ${{ env.RIPPLED_DOCKER_IMAGE }} /opt/ripple/bin/rippled -a --conf /opt/ripple/etc/rippled.cfg
```
JST5000, line 168: Are there docs on how to update the rippled container? (May be worth linking here or noting down somewhere?)

.ci-config/rippled.cfg
mvadari: Tests pass, so yes.

.github/workflows/nodejs.yml
mvadari: I believe these are going to be automatically updated as a part of rippled CI (cc @manojsdoshi).

CONTRIBUTING.md
mvadari: As of now there is no `latest` tag, so the version does need to be specified: https://hub.docker.com/r/rippleci/rippled/tags

JST5000 (approved):
LGTM

PR #2563, opened 2023-11-06
feat: remove `assert` library as a testing dep

ckniffen:

Right now we have 5 testing library dependencies. This eliminates `assert`. `jest`, `jasmine`, `chai` and `karma` are the remaining ones.

### Type of Change


mvadari (commented):
Why keep `chai` instead of `assert`?

ckniffen:
> Why keep `chai` instead of `assert`?

Chai is used in way more places.  The goal should be to remove chai too.  This was just a step toward that.

JST5000 (approved):
LGTM

jonathanlei (approved):

khancode (approved):

PR #2566, opened 2023-11-09
test: run binary-codec tests in the browser

ckniffen:

- Convert tests to typescript
- Update type definitions causing errors in tests
  - `makeParser` to accept a `Buffer` in addition to `string`
  - `SerializedType` constructor allows not passing in a byte array
  - `Comparable` is now a generic type so that it allows `compareTo` methods to take more that the type itself. Example: `Uint64.compareTo` can accept `number`
- Update tests to use jasmine compatible functions
  - Switching from `test` to `it`.
  - Updated test checking if coretypes all implement SerializedType
  - Import fixtures directly instead of using `loadFixture` utility
- Remove importing of `buffer/` explicitly.  It was throwing off type checking in tests.  Buffer is going away in a future PR anyway.
- Fixed `npm run clean` not clearing `.tsbuildinfo` files for keypairs
- Remove unused `account-tx-transactions.db`.  It was likely used in the past to test historical ledgers.

### Context of Change

Unit tests should be run in the browser and needed to be in Typescript

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue))

- [ ] Yes


packages/ripple-binary-codec/src/types/serialized-type.ts
```diff
@@ -80,26 +80,30 @@ class SerializedType {
 }
 
 /**
- * Base class for SerializedTypes that are comparable
+ * Base class for SerializedTypes that are comparable.
+ *
+ * @template T - What types you want to allow comparisons between. You must specify all types.
```
JST5000, line 85: I think this extra sentence helps explain why we're using type generics more explicitly 
```suggestion
 * @template T - What types you want to allow comparisons between. You must specify all types. Primarily used to allow comparisons between built-in types (like `string`) and SerializedType subclasses (like `Hash`). 
```

packages/ripple-binary-codec/test/shamap.test.ts
```diff
@@ -72,18 +80,18 @@ describe('ShaMap', () => {
...
       expect(map.hash().toHex()).toBe(ledger.account_hash)
       // console.log('took seconds: ', (now() - t));
     })
   }
-  factory('ledger-full-38129.json')
-  factory('ledger-full-40000.json')
+  factory(ledgerFull38129)
+  factory(ledgerFull40000)
   // factory('ledger-4320277.json');
   // factory('14280680.json');
```
JST5000, line 96: Should we get rid of these comments or add these tests? (Do these files still exist?)

JST5000 (approved):
LGTM

packages/ripple-binary-codec/test/shamap.test.ts
ckniffen: Good call.  These are gone now.

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -13,20 +13,23 @@ import {
 } from './enums'
 import { STObject } from './types/st-object'
 import { JsonObject } from './types/serialized-type'
-import { Buffer } from 'buffer/'
 
 /**
  * Construct a BinaryParser
  *
  * @param bytes hex-string to construct BinaryParser from
```
khancode, line 20:
```suggestion
 * @param bytes hex-string or Buffer to construct BinaryParser from
```

khancode (approved):
LGTM, left a comment.

PR #2574, opened 2023-11-13
feat: getXrpBalance and dropsToXRP return number

ckniffen:

BREAKING CHANGE: `dropsToXRP` and `Client.getXrpBalance` now return a `number` instead of a `string`

### Context of Change

@mvadari requested this be updated in a breaking release.

### Type of Change

- [x] Documentation Updates

- [x] Yes

### DIFF
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -12,6 +12,7 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
   * `Client.prepareTransaction`
   * `getSignedTx`
   * `isAccountDelete`
+* BREAKING CHANGE: `dropsToXRP` and `Client.getXrpBalance` now return a `number` instead of a `string`
 
 ## 3.0.0 Beta 1 (2023-10-19)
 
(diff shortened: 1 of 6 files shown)


packages/xrpl/HISTORY.md
```diff
@@ -12,6 +12,7 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
   * `Client.prepareTransaction`
   * `getSignedTx`
   * `isAccountDelete`
+* BREAKING CHANGE: `dropsToXRP` and `Client.getXrpBalance` now return a `number` instead of a `string`
```
mvadari, line 15: It's already in the `Breaking Changes` section
```suggestion
* `dropsToXRP` and `Client.getXrpBalance` now return a `number` instead of a `string`
```

packages/xrpl/src/client/index.ts
```diff
@@ -938,8 +938,8 @@ class Client extends EventEmitter<EventTypes> {
         const accountLinesBalance = linesResponses.flatMap((response) =>
           formatBalances(response.result.lines),
         )
-        if (xrpBalance !== '') {
-          balances.push({ currency: 'XRP', value: xrpBalance })
+        if (xrpBalance !== 0) {
+          balances.push({ currency: 'XRP', value: xrpBalance.toString() })
```
mvadari, line 942: This might be a broader question than this PR, but perhaps there should be an error thrown if the XRP balance is somehow 0, since that shouldn't be possible?

mvadari (approved):

justinr1234 (approved):

packages/xrpl/HISTORY.md
ckniffen: good catch.  I copied it from the commit subject.

PR #2596, opened 2023-12-05
Support alt. native assets

WietseWind:

Support for setting alternative native asset for currency encoding using:

```
nativeAsset.set('XRP')
```

Required when dealing with issued currencies on a network where XRP isn't the native asset, e.g. `XAH` or `USD`.

### DIFF
--- a/packages/ripple-binary-codec/src/index.ts
+++ b/packages/ripple-binary-codec/src/index.ts
@@ -10,6 +10,7 @@ import {
 } from './enums'
 import { XrplDefinitions } from './enums/xrpl-definitions'
 import { coreTypes } from './types'
+import { nativeAsset } from './nativeasset'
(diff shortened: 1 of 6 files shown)


ckniffen:
I would like to see these two methods on XrplDefinitions.  I believe that would allow it to pass through better and setup for it to be a field on server_definitions.

WietseWind:
> I would like to see these two methods on XrplDefinitions. I believe that would allow it to pass through better and setup for it to be a field on server_definitions.

See #2597 for that.

ckniffen:
> Member

Never mind just now seeing the other PR #2597

justinr1234:
@coderabbit full review

justinr1234:
@coderabbitai full review

PR #2597, opened 2023-12-05
Support for native asset & hash in custom Definitions

WietseWind:

Add native asset & hash as returned by the `server_definitions` method to `XrplDefinitions` objects, and allow getting them through:

- `getNativeAsset()`
- `getHash()`

This allows to use these params as present in constructed `XrplDefinitions` by other libs building on top, potentially wanting to pass `getNativeAsset` to #2596 and using `getHash` to save bandwidth when checking if new definitions are to be loaded.


ckniffen (dismissed):

packages/ripple-binary-codec/src/enums/definitions-xahau.json
```diff
@@ -0,0 +1,2740 @@
+{
```
mvadari, line 1: This should go in the `test` folder, not the `src` folder. And perhaps should be renamed to something like `definitions-sample.json` so that people don't think it's the current production file.

This is because this file would increase the size of the package by about 6%, and it's therefore not sustainable to do for every XRPL-based chain. I'd recommend maintaining this file in a separate repo and pointing people towards that.

WietseWind: Will do in separate repo 👍

mvadari: The rest of the PR was fine, just this one file was what I suggested you move to a new repo 🙂

WietseWind: @mvadari Done, loading JSON from external dev-dependency.

packages/ripple-binary-codec/test/definitions.test.js
```diff
@@ -1,5 +1,6 @@
 const { encode, decode, XrplDefinitions } = require('../src')
 const normalDefinitionsJson = require('../src/enums/definitions.json')
+const altDefinitions = require('xrpl-binary-codec-prerelease/dist/enums/definitions-xahau.json')
```
mvadari, line 3: I'd recommend having the definitions file be just a copy-pasted test file in the `fixtures` folder, since it doesn't need to be kept up-to-date for the tests to still work.

WietseWind: You know what: I propose that you decide internally how you want to resolve this :) Lmk if there's anything I can do at this point.

ckniffen: I agree with @mvadari approach.

justinr1234:
@coderabbit full review

justinr1234:
@coderabbitai full review

PR #2617, opened 2024-01-17
Add a CONTRIBUTING.md to `ripple-binary-codec`

JST5000:

It's not obvious how to add a new type to the library, run tests, and how to write tests. This documents that process better.

### Context of Change

This should help any potential contributors understand the binary-codec library faster.

### Type of Change

- [X] Documentation Updates

## Test Plan

N/A

### DIFF
--- a/packages/ripple-binary-codec/CONTRIBUTING.md
+++ b/packages/ripple-binary-codec/CONTRIBUTING.md
@@ -1,0 +1,29 @@
+# Code structure
+
+ - `enums` contains the logic for mapping types to their numerical id in rippled, which is used to serialize / deserialize later on. For more details, see its specific [README.md](src/enums/README.md)
+ - `serdes` contains the two main classes of this repo `BinarySerializer` and `BinaryParser` which actually do the serializing / deserializing.
+ - `types` defines how the library should convert between objects and binary data for each data type. These are then used by the `BinarySerializer` and `BinaryParser` to serialize / deserialize rippled data.
+ - At the top-level, we have helper functions for specific situations (like `encodeForMultisigning` in [index.ts](src/index.ts)). These functions are generally what are used by consumers of the library (with `BinarySerializer` and `BinaryParser` being used as the implementation under the hood)
+
+# Running tests
+You can run tests for this repo with:
+`npm i` to install necessary dependencies followed by either:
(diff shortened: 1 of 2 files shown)


packages/ripple-binary-codec/CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
...
+For a good example unit test file, look at [test/hash.test.ts](test/hash.test.ts).
+
+If you add a new serializable type, please add a new file with tests that ensure it can be encoded / decoded, and that it throws any relevant errors.
+
+# Adding new serializable types
+To add a new serializable type, first read through `enum`'s [README.md](src/enums/README.md) as it explains how to update `definitions.json` which ties `TransactionType`s and `Field`s to specific ids rippled understands.
+
+After that, if you need to add a new type of data to be serialized / deserialized (for example adding a bigger int than [uint-64.ts](src/types/uint-64.ts)) you can follow these steps:
+1. Create a new class that extends `SerializedType`
+  - If your type is intended to be comparable to native values like `number`, you should extend `Comparable<YourObjectType | nativeValue>`. See [uint-64.ts](src/types/uint-64.ts) for an example of this.
```
sublimator, line 26: Is Comparable even used any more? I was the original author of the repo way back in the day, when it was raw JS (see https://github.com/sublimator/ripple-core/tree/master/src). I was using it for tests and actually used the internal objects. One of my colleagues at the time took it for use as ripple-binary-codec, adding the facade we all know and love (or not, you might like base64 instead of hex) 

In any case, the internals aren't really exported, and last I looked, there wasn't actual uses of Comparable outside of the tests? Or maybe I'm misremembering, or confused myself, or was led astray by intellij?

ckeshava (approved):

packages/ripple-binary-codec/CONTRIBUTING.md
ckniffen: They are still exported. Other libraries may use them. https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/index.ts#L140

ckniffen (approved):

sublimator:
Oh, interesting! Back in the day they weren't, but I may be misremembering.

Thanks

PR #2626, opened 2024-01-27
Update migration guide

JST5000:

Couple points of confusion / clarification added from migrating xrpl.org

### Context of Change

- Adds a note that `xrpToDrops` is unchanged (which can be easily assumed since the "reverse" is modified)
- Adds a list of polyfills that are removed to be very explicit about how to migrate if you already have polyfills (it's backwards compatible, but this is where the size reduction happens!)

### Type of Change

- [X] Documentation Updates

- [ ] Yes

## Test Plan

Inspection (these additional steps / explanations point to things I had to figure out when migrating xrpl.org)

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -49,10 +49,11 @@ The easiest to replace were `assert` (which was replaced by simple conditions & 
 
 Lastly, the `buffer` polyfill turned out to be the trickiest to remove, resulting in the largest number of breaking changes. Since the `Buffer` object is not native to the browser all apis were migrated to the superclass of `Buffer` → `Uint8Array`s. For a detailed write up of why we and many libraries are choosing to make this transition, check out this [blog post](https://sindresorhus.com/blog/goodbye-nodejs-buffer) by Sindre Sorhus.
 
-List of all replaced polyfills that can be removed from your webpack.config.js / vite.config.js / other bundling config files:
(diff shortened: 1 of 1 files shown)


justinr1234 (approved):

MIGRATION.md
```diff
@@ -49,6 +49,17 @@ The easiest to replace were `assert` (which was replaced by simple conditions &
 
 Lastly, the `buffer` polyfill turned out to be the trickiest to remove, resulting in the largest number of breaking changes. Since the `Buffer` object is not native to the browser all apis were migrated to the superclass of `Buffer` → `Uint8Array`s. For a detailed write up of why we and many libraries are choosing to make this transition, check out this [blog post](https://sindresorhus.com/blog/goodbye-nodejs-buffer) by Sindre Sorhus.
 
+List of all replaced polyfills that can be removed from your webpack.config.js / vite.config.js / other bundling config files:
```
ckniffen, line 52: Having helped others or looked at their code these polyfills cannot always be safely removed espefially if they are using other libraries that require them.  You might also mention `events` which used to have to be specified in the vite config

ckniffen (approved):

ckeshava (approved):

MIGRATION.md
JST5000: Added along with a note that these may still be needed for other libraries / code written.

PR #2630, opened 2024-02-01
Update docs for Channel, ClientOptions, and ConnectionUserOptions

JST5000:

Three small objects were missing docs

### Context of Change

Found while investigating writing a schema for types.

### Type of Change

- [X] Documentation Updates


packages/xrpl/src/client/index.ts
```diff
@@ -125,9 +125,24 @@ import {
 } from './partialPayment'
 
 export interface ClientOptions extends ConnectionUserOptions {
+  /**
+   * Factor to multiply estimated fee by to provide a cushion in case the
```
ckeshava, line 129:
```suggestion
   * Multiplication factor of the estimated fee to provide a buffer, in case the
```

packages/xrpl/src/client/index.ts
```diff
@@ -125,9 +125,24 @@ import {
...
+   * number. Defaults to '2'.
+   *
+   * @category Fee
+   */
   maxFeeXRP?: string
   proxy?: string
+  /**
+   * Duration to wait for a request to timeout.
+   */
   timeout?: number
```
ckeshava, line 146: This `timeout` is not used in the codebase. Removal of it didn't cause any failures in the unit or integration tests.

Why is it present?

packages/xrpl/src/models/methods/accountChannels.ts
```diff
@@ -1,18 +1,111 @@
...
-  expiration?: number
-  cancel_after?: number
-  source_tab?: number
-  destination_tag?: number
+  // Required Fields
+  /**
+   * The source address that owns this payment channel.
+   * This comes from the sending address of the transaction that created the channel.
+   */
+  Account: string
```
ckeshava, line 12: The other files in the `methods/` folder use `snake_case` for the interface attributes of ledger objects. Do we have a naming convention?

This is a minor nit, not a binding suggestion

packages/xrpl/src/models/methods/accountChannels.ts
```diff
@@ -1,18 +1,111 @@
...
+
+  /**
+   * The mutable expiration time for this payment channel, in seconds since the Ripple Epoch.
+   * The channel is expired if this value is present and smaller than the previous ledger's close_time field.
+   * See Channel Expiration for more details.
+   */
+  Expiration?: number
+
+  /**
+   * An arbitrary tag to further specify the source for this payment channel,
```
ckeshava, line 105: thanks for adding the docstrings, it makes it very readable

packages/xrpl/src/client/index.ts
JST5000: Used 1st half of change but kept the `cushion` since this parameter is called `feeCushion` and `buffer` is a meaningful data type name we don't want to accidentally reference :)

packages/xrpl/src/client/index.ts
JST5000: `timeout` is used in `Client.request` on lines 371-374 as a max timeout for a request to be sent, it would only matter if you were in a poor internet connection situation so wouldn't show up easily in the tests. 

Our test coverage isn't good enough to trust passing tests as a sign that a field isn't being used since we're not exercising all code / branches.

packages/xrpl/src/models/methods/accountChannels.ts
JST5000: This seems to be a bug - nice catch :) 

Judging from the response type on xrpl.org the fields are lowercase + snake_case in the response type.

JST5000: I was using the `PaymentChannel` transaction type, but should've been using a similar (but not identical) type that's embedded within the `account_channels` request.

packages/xrpl/src/models/methods/accountChannels.ts
```diff
@@ -1,17 +1,65 @@
...
+   * This expiration date is mutable. If this is before the close time of the most recent validated ledger, the channel is expired.
+   */
   expiration?: number
+
+  /**
+   * Time, in seconds since the Ripple Epoch, of this channel's immutable expiration, if one was specified at channel creation.
+   * If this is before the close time of the most recent validated ledger, the channel is expired.
+   */
   cancel_after?: number
-  source_tab?: number
```
JST5000, line 14: Hahaha, I guess there was a typo... oops!

jonathanlei (approved):

khancode (approved):

PR #2649, opened 2024-02-15
fix: get `client.requestAll` to handle filters better

mvadari:

The `client.requestAll` method was failing when it had an initial empty list of objects. This sometimes happens when there is an object filter, such as on `account_objects. This PR fixes that issue.

### Context of Change

This issue caused problems in some production code.

### Type of Change

- [x] Yes

## Test Plan

Added a unit test.

### DIFF
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -4,6 +4,9 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 
 ## Unreleased
 
+### Fixed
+* Fix `client.requestAll` to handle filters better
+
 ## 3.0.0 (2024-02-01)
 
 ### BREAKING CHANGES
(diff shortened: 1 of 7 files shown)


packages/xrpl/HISTORY.md
```diff
@@ -52,80 +55,6 @@ Bundler configurations are much more simplified. See [../UNIQUE_STEPS](Unique St
 * Deprecated:
   * `convertHexToString` in favor of `@xrplf/isomorphic/utils`'s `hexToString`
   * `convertStringToHex` in favor of `@xrplf/isomorphic/utils`'s `stringToHex`
-
```
JST5000, line 55: I think you accidentally deleted some HISTORY.md here (Specifically the beta notes)

mvadari: That was on purpose. You don't need the beta logs in the HISTORY after the new version is released, since it's not an official release.

JST5000 (approved):
LGTM (other than the history note)

packages/xrpl/HISTORY.md
JST5000: I see, ok.

intelliot (approved):

PR #2650, opened 2024-02-18
Set default api_version in requests

khancode:

Set default `api_version` in requests if not supplied. Default api_version will use `buildVersion` of what server the client is connected to.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

### DIFF
--- a/packages/xrpl/src/client/index.ts
+++ b/packages/xrpl/src/client/index.ts
@@ -199,6 +199,12 @@ class Client extends EventEmitter<EventTypes> {
   public buildVersion: string | undefined
 
   /**
+   * API Version used by the server this client is connected to
+   *
+   */
+  public apiVersion: number | undefined
+
+  /**
    * Creates a new Client with a websocket connection to a rippled server.
    *
    * @param server - URL of the server to connect to.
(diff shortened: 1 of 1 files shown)


packages/xrpl/src/client/index.ts
```diff
@@ -489,6 +496,7 @@ class Client extends EventEmitter<EventTypes> {
       })
       this.networkID = response.result.info.network_id ?? undefined
       this.buildVersion = response.result.info.build_version
+      this.apiVersion = parseInt(this.buildVersion.split('.')[0], 10)
```
ckniffen, line 499: I am not a fan of this.  This will be a major breaking change.  Anyone using public infrastructure will be one api 2.0.  also without complete 2.0 type support this should not be done.

khancode: @ckniffen That's fair. What do you recommend doing? Should we just leave it as it (defaults to rippled v1 API) and the user supplies what API version to use?

ckniffen: Yes default to not sending one and if they set version to 2 then we derive different types

PR #2656, opened 2024-03-04
feat: add rippled API v2 support and use as default

khancode:

Add rippled `api_version` support for v2 and use as default while maintaining support for v1.

### Context of Change

- rippled `api_version` 2 is supported by 100% of rippled servers now that numerous v2 amendments have activated. (as of April 8, 2024.). Therefore, we should default to using v2 and maintain support of v1 response types.
- This feature will be released as a major bump version since it's a significant breaking change.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Updated existing unit/integ tests and added new ones.


packages/xrpl/test/integration/transactions/signerListSet.test.ts
```diff
@@ -45,7 +45,10 @@ describe('SignerListSet', function () {
       }
       await testTransaction(testContext.client, tx, testContext.wallet)
 
-      const accountInfoResponse = await testContext.client.request({
+      const accountInfoResponse = await testContext.client.request<
+        AccountInfoRequest,
+        1
```
ckniffen, line 50: Does this not default to 1?

khancode: For TypeScript, it's not able to infer it's api_version 1. The type that gets assigned is the parent type `AccountInfoResponse` when type assertion isn't specified; it should be `AccountInfoV1Response` type.

justinr1234: I put in a non-complete change to show:

https://github.com/XRPLF/xrpl.js/pull/2677/files#diff-a17f03a61f343bf490e5f4379ef5650ac8fcdc8c9dba35c8fd914c5502f83a7bR553-R563

- See if it's API version 2
  - If it is, go to V2 types
    - Check if account info
      - If it is, return account info v2
      - If it isn't, return the base types (which excludes account info)
  - If it isn't, go to V1 types
    - Check if account info
      - If it is, return account info v1
      - If it isn't, return the base types (which excludes account info)

khancode:
api_version 2 is supported by 100% of rippled servers now that numerous v2 amendments have activated. (as of April 8, 2024.) Therefore, we decided to drop type support for API v2 and just default to use v2 -  https://github.com/XRPLF/xrpl.js/pull/2687

khancode:
Reopening since we decided it's best to continue TS type support of v1 and only deprecate features/versions when rippled does.

packages/xrpl/test/integration/transactions/signerListSet.test.ts
khancode: @justinr1234 Thanks for looking into it but unfortunately it didn't make a difference since type inference goes undetectable with TS Union and other complex types. And our codebase is filled with these types.

packages/xrpl/test/integration/README.md
```diff
@@ -1,7 +1,7 @@
 To run integration tests:
 1. Run rippled in standalone node, either in a docker container (preferred) or by installing rippled.
   * Go to the top-level of the `xrpl.js` repo, just above the `packages` folder.
-  * With docker, run `docker run -p 6006:6006 --interactive -t --volume $PWD/.ci-config:/config/ xrpllabsofficial/1.12.0-b1 -a --start`
+  * With docker, run `docker run -p 6006:6006 --interactive -t --volume $PWD/.ci-config:/opt/ripple/etc/ --platform linux/amd64 rippleci/rippled:2.2.0-b3 /opt/ripple/bin/rippled -a --conf /opt/ripple/etc/rippled.cfg`
```
ckeshava, line 4: this rippled version appears to be different from the CONTRIBUTING.md file

packages/xrpl/test/integration/fundWallet.test.ts
```diff
@@ -83,40 +83,6 @@ describe('fundWallet', function () {
   //   )
   // })
 
-  it(
```
ckeshava, line 86: is this removed with an intent to remove support for the hooks testnet faucet? If so, can we include other files under this purview and also make a mention of it in the `HISTORY.md` file?

I took a stab at it here: https://github.com/ckeshava/xrpl.js/tree/removeHooksSupport

packages/xrpl/test/integration/requests/accountInfo.test.ts
```diff
@@ -80,4 +80,105 @@ describe('account_info', function () {
     },
     TIMEOUT,
   )
+
+  it(
+    'uses api_version 1',
```
ckeshava, line 85: regarding the integration tests that mention `api_version 1`, how are we enforcing that? -- the present CI/CD defaults to v2.

henceforth, are we executing integration tests against two versions of rippled? v1 and v2? I suspect we'll need to make changes to the Github Actions workflow right?

packages/xrpl/src/client/index.ts
```diff
@@ -213,6 +219,12 @@ class Client extends EventEmitter<EventTypes> {
    */
   public buildVersion: string | undefined
 
+  /**
+   * API Version used by the server this client is connected to
+   *
+   */
+  public apiVersion: APIVersion = DEFAULT_API_VERSION
```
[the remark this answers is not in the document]
khancode: They can either set `apiVersion` variable or if a request object contains `api_version` it will override it.

packages/xrpl/src/models/methods/accountInfo.ts
```diff
@@ -148,8 +144,7 @@ export interface AccountInfoResponse extends BaseResponse {
      * at most one SignerList, this array must have exactly one member if it is
```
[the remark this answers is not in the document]
khancode: Why should it be removed? It was already present here.

packages/xrpl/src/models/methods/accountTx.ts
```diff
@@ -58,7 +63,15 @@ export interface AccountTxTransaction {
    */
   meta: string | TransactionMetadata
   /** JSON object defining the transaction. */
```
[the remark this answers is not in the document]
khancode: My opinion is for default versions, we don't specify API v2 since it's used by most validators now.

packages/xrpl/src/models/methods/accountTx.ts
```diff
@@ -58,7 +63,15 @@ export interface AccountTxTransaction {
...
-  tx?: Transaction & ResponseOnlyTxInfo
+  tx_json?: Version extends typeof RIPPLED_API_V2
+    ? Transaction & ResponseOnlyTxInfo
+    : never
+  /** JSON object defining the transaction. */
+  tx?: Version extends typeof RIPPLED_API_V1
+    ? Transaction & ResponseOnlyTxInfo
+    : never
+  /** The hash of the transaction. */
```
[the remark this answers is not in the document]
khancode: ditto explanation

packages/xrpl/test/integration/fundWallet.test.ts
khancode: Yes, we decided to no longer support Hooks testnet faucet. Can you create a PR with this included in your branch?

packages/xrpl/test/integration/requests/accountInfo.test.ts
khancode: It doesn't default to API v2. The client library integ tests decide which api_version to use. I don't think it's necessary to have separate GitHub Action workflows since not all requests were affected by API v2. For the ones that were, I kept their API v1 tests and added v2.

packages/xrpl/src/models/methods/accountTx.ts
```diff
@@ -69,41 +82,41 @@ export interface AccountTxTransaction {
...
- * @category Responses
+ * Base interface for account transaction responses.
  */
-export interface AccountTxResponse extends BaseResponse {
+interface AccountTxResponseBase<Version extends APIVersion>
+  extends BaseResponse {
   result: {
-    /** Unique Address identifying the related account. */
     account: string
-    /**
```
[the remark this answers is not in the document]
khancode: Oops, I'll add them back in. Not sure how that happened.

packages/xrpl/src/models/methods/accountInfo.ts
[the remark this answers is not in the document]
ckeshava: The `SignerLists` portion of the interface has been moved into the new V2 interface. The comment is relevant wrt to the child-interface

packages/xrpl/test/integration/fundWallet.test.ts
ckeshava: ok. https://github.com/XRPLF/xrpl.js/pull/2711

packages/xrpl/test/integration/requests/accountInfo.test.ts
ckeshava: ah okay, I didn't realise that.

The xrpl docs do not mention the `api_version` parameter, so I was slightly confused -- https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_info/

packages/xrpl/test/integration/requests/accountTx.test.ts
```diff
@@ -32,6 +32,102 @@ describe('account_tx', function () {
...
+                date: 685747005,
+                hash: '2E68BC15813B4A836FAC4D80E42E6FDA6410E99AB973937DEA5E6C2E9A116BAB',
+                ledger_index: 1734,
+              },
+            },
+          ],
+        },
+        type: 'response',
+      }
+      assert.equal(response.type, expected.type)
```
ckeshava, line 63: could you include an `assert` for `tx_blob` and `meta_blob` fields? -- these would be visible only if "binary" format was requested. Nevertheless, these fields were newly introduced in V2.

I'm looking at the table here for reference: https://github.com/XRPLF/clio/issues/722

packages/xrpl/src/models/methods/tx.ts
```diff
@@ -41,35 +42,67 @@ export interface TxRequest extends BaseRequest {
...
+  /** The SHA-512 hash of the transaction. */
+  hash: string
+  /**
+   * The Concise Transaction Identifier of the transaction (16-byte hex string)
+   */
+  ctid?: string
+  /** The ledger index of the ledger that includes this transaction. */
+  ledger_index?: number
+  /** Unique hashed string Transaction metadata blob, which describes the results of the transaction.
+   *  Can be undefined if a transaction has not been validated yet. */
```
ckeshava, line 63: `meta_blob` field is absent if the user has not requested for the `binary` format too

packages/xrpl/test/integration/requests/bookOffers.test.ts
```diff
@@ -35,6 +35,7 @@ describe('book_offers', function () {
       const response = await testContext.client.request(bookOffer)
 
       const expectedResponse: BookOffersResponse = {
+        api_version: 2,
```
ckeshava, line 38: can we remove these `api_version: 2` lines in the integration tests? In the future when we upgrade to V3, we'll need to update all these lines again.

I understand that it defaults to V2, if no version is specified

packages/xrpl/test/integration/requests/accountInfo.test.ts
intelliot: api_version can be included with any request, so it's mentioned here: https://xrpl.org/docs/references/http-websocket-apis/api-conventions/request-formatting/

ckeshava: thanks @intelliot 

there's an interesting discrepancy here -- rippled defaults to API v1 whereas the xrpl.js client library defaults to using V2. I hope developers don't get confused by that. It shouldn't be much of a problem because there are no V1 machines (afaik)

khancode: @ckeshava that's a good callout about being a discrepancy between rippled and client libraries but I still agree to default to API v2. And I would assume eventually rippled will default to v2 too.

packages/xrpl/src/models/methods/accountInfo.ts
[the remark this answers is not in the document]
khancode: Ah good catch

packages/xrpl/test/integration/requests/bookOffers.test.ts
khancode: It's better to keep this because it will serve as a safeguard/reminder that whenever we change the default `api_version` then we should also update the integ tests.

packages/xrpl/test/integration/requests/accountInfo.test.ts
```diff
@@ -80,4 +80,105 @@ describe('account_info', function () {
     },
     TIMEOUT,
   )
+
```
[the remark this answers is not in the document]
khancode: It's already covered in `tx.test.ts`

packages/xrpl/src/models/methods/tx.ts
khancode: How would a user request for `binary` format? I'm having trouble finding it.

packages/xrpl/test/integration/requests/accountTx.test.ts
khancode: How would a user request for `binary` format? I'm having trouble finding it.

ckeshava: https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_tx/
`"binary": true`

packages/xrpl/test/integration/requests/accountInfo.test.ts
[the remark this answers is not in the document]
ckeshava: I observed only `AccountSet` transactions in that file, not `AccountInfo`

packages/xrpl/src/models/methods/accountTx.ts
```diff
@@ -49,7 +54,7 @@ export interface AccountTxRequest extends BaseRequest, LookupByLedgerRequest {
   marker?: unknown
 }
 
-export interface AccountTxTransaction {
+export interface AccountTxTransaction<Version extends APIVersion> {
   /** The ledger index of the ledger version that included this transaction. */
   ledger_index: number
   /**
```
ckeshava, line 60: `ledger_hash` should be present in both the request and response of `AccountTx`.

Here's an example usage in the response (from cpp unit tests) - https://github.com/XRPLF/rippled/blob/c706926ee3c7d0f3abbbd3545cc85d6886643444/src/test/rpc/AccountTx_test.cpp#L162

here's is the usage inside a request: https://github.com/XRPLF/rippled/blob/c706926ee3c7d0f3abbbd3545cc85d6886643444/src/test/rpc/AccountTx_test.cpp#L316

I think the JS client library didn't have `ledger_hash` field since before this PR

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -351,6 +351,7 @@ export async function checkAccountDeleteBlockers(
     account: tx.Account,
     ledger_index: 'validated',
     deletion_blockers_only: true,
+    api_version: client.apiVersion,
```
ckeshava, line 354: there is one more invocation of `account_info` in this file (line 229). For the sake of consistency, I think it's better to specify `api_version: client.apiVersion` at that line too.

ckeshava: and also an invocation to `server_state` request -- but I don't think that request is impacted by API V2 changes

ckeshava (changes requested):

packages/xrpl/src/sugar/autofill.ts
khancode: Good catch. This line should actually be deleted because `client.apiVersion` is eventually used in `request()`.

packages/xrpl/src/models/methods/accountTx.ts
khancode: I just added it in `ResponseOnlyTxInfo`.

packages/xrpl/src/models/methods/accountInfo.ts
```diff
@@ -180,3 +175,59 @@ export interface AccountInfoResponse extends BaseResponse {
...
+    }
+  }
+}
+
+/**
+ * Type to map between the API version and the response type.
+ *
+ * @category Responses
+ */
+export type AccountInfoVersionResponseMap<Version extends APIVersion> =
```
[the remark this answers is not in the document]
khancode: I tried `api_version: 3` and rippled did throw an error `RippledError: invalid_API_version`. I don't think it's really worth adding a check since this is unlikely to happen and when it does it's a clear error from rippled.

packages/xrpl/test/integration/requests/accountInfo.test.ts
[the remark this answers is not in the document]
khancode: I don't think integ tests should be this exhaustive for checking existence of all possible fields. They should only be done for testing transaction functionality which `account_info` test already does.

packages/xrpl/src/models/methods/accountInfo.ts
```diff
@@ -180,3 +175,59 @@ export interface AccountInfoResponse extends BaseResponse {
...
+ *
+ * @category Responses
+ */
+export interface AccountInfoResponse extends BaseAccountInfoResponse {
+  result: BaseAccountInfoResponse['result'] & {
+    /**
+     * Array of SignerList ledger objects associated with this account for Multi-Signing.
+     * Since an account can own at most one SignerList, this array must have exactly one
+     * member if it is present.
+     * Quirk: In API version 1, this field is nested under account_data. For this method,
```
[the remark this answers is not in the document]
khancode: I agree, just pushed the change.

packages/xrpl/src/models/methods/accountTx.ts
```diff
@@ -58,7 +63,15 @@ export interface AccountTxTransaction {
    */
   meta: string | TransactionMetadata
```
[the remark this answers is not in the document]
khancode: It's already added in `BaseTxResult`

packages/xrpl/test/integration/requests/accountTx.test.ts
khancode: Like I mentioned in https://github.com/XRPLF/xrpl.js/pull/2656#discussion_r1649162438 , I don't believe we should test for this specifically.

packages/xrpl/src/models/methods/submitMultisigned.ts
```diff
@@ -24,28 +25,58 @@ export interface SubmitMultisignedRequest extends BaseRequest {
...
+  }
+}
+
+/**
+ * Response expected from a {@link SubmitMultisignedRequest} using api_version 1.
+ *
+ * @category Responses
+ */
+export interface SubmitMultisignedV1Response extends BaseResponse {
+  result: BaseSubmitMultisignedResult & {
```
ckeshava, line 69: Hey Omar, is my understanding correct? 
`BaseSubmitMultisignedResult` already contains a `tx_json` field, but we are inserting another `tx_json` field as a child property in this line.

I suspect that the javascript runtime flexibly manages the layout of the object, depending on whether its a V1-response or a V2-response, right?

packages/xrpl/src/models/ledger/Ledger.ts
```diff
@@ -63,3 +58,39 @@ export default interface Ledger {
...
+ * A ledger is a block of transactions and shared state data. It has a unique
+ * header that describes its contents using cryptographic hashes.
+ *
+ * @category Ledger Entries
+ */
+export interface Ledger extends BaseLedger {
+  /**
+   * The ledger index of the ledger. Represented as a native JSON number.
+   */
+  ledger_index: number
```
ckeshava, line 72: In hindsight, I like your other design better. Since this is the only aspect of difference between the two versions, couldn't we template based on the `APIVersion` ? This could reduce a lot of boiler-plate code downstream.

```suggestion
export interface Ledger<APIVersion> extends BaseLedger {

ledger_index : (APIVersion extends RIPPLED_API_V2 ? number : string)
}

// now all the downstream classes can just inherit from "Ledger<APIVersion>", instead of having to make two copies
// Ex
export interface LedgerBinary<APIVersion>
  extends Omit<Omit<Ledger<APIVersion>, 'transactions'>, 'accountState'> {
  accountState?: string[]
  transactions?: string[]
}
```

I'm not asking you to change the code, I was wondering if this approach has any demerits?

packages/xrpl/test/integration/requests/accountTx.test.ts
ckeshava: okay yeah. The APIs have a wide surface area, its hard to cover every possibility

packages/xrpl/src/models/methods/tx.ts
```diff
@@ -41,35 +42,68 @@ export interface TxRequest extends BaseRequest {
...
-     * If true, this data comes from a validated ledger version; if omitted or.
-     * Set to false, this data is not final.
-     */
-    validated?: boolean
-    /**
-     * This number measures the number of seconds since the "Ripple Epoch" of January 1, 2000 (00:00 UTC)
-     */
-    date?: number
-  } & T
+  result: BaseTxResult<typeof RIPPLED_API_V2, T> & { tx_json: T }
```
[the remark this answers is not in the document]
ckeshava: Hmm, it's present in `SubmitResponse`. I suspect that backstops all the responses

packages/xrpl/test/integration/requests/ledger.test.ts
```diff
@@ -55,7 +117,7 @@ describe('ledger', function () {
       assert.typeOf(ledgerResponse.result.ledger_hash, 'string')
       assert.typeOf(ledgerResponse.result.ledger_index, 'number')
```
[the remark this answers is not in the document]
khancode: This is from v1 JSDoc:`The ledger index of the ledger. Some API methods display this as a quoted integer; some display it as a native JSON number.` So in this case, it's being returned as a JSON `number`.

packages/xrpl/src/models/methods/index.ts
```diff
@@ -442,14 +458,14 @@ export type RequestAllResponseMap<T> = T extends AccountChannelsRequest
   : T extends AccountOffersRequest
   ? AccountOffersResponse
   : T extends AccountTxRequest
-  ? AccountTxResponse
+  ? AccountTxVersionResponseMap<Version>
   : T extends LedgerDataRequest
   ? LedgerDataResponse
   : T extends AccountTxRequest
```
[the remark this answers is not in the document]
khancode: Good catch, the second one should be deleted.

packages/xrpl/src/client/index.ts
```diff
@@ -319,16 +330,20 @@ class Client extends EventEmitter<EventTypes> {
    * console.log(response)
    * ```
    */
-  public async request<R extends Request, T = RequestResponseMap<R>>(
-    req: R,
-  ): Promise<T> {
-    const response = await this.connection.request<R, T>({
+  public async request<
+    R extends Request,
+    V extends APIVersion = typeof DEFAULT_API_VERSION,
```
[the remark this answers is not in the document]
khancode: `APIVersion` is type that can be either `1` or `2` number. We're setting a default value to it using `DEFAULT_API_VERSION` which is `2` number.

packages/xrpl/src/models/methods/tx.ts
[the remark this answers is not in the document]
khancode: Yes, it's present in other models too.

packages/xrpl/src/models/methods/submitMultisigned.ts
khancode: It's the property location of `hash` that makes it different between v1 and v2. In `SubmitMultisignedV1Response`, we're overwriting `tx_json` to include `hash` as a child prop.

packages/xrpl/src/models/methods/accountTx.ts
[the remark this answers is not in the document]
khancode: Ah my mistake, you're right it's not included. Can `meta_blob` be included in any transaction response? If so, then it should be added in `BaseResponse`

packages/xrpl/src/models/methods/tx.ts
```diff
@@ -41,35 +42,68 @@ export interface TxRequest extends BaseRequest {
...
+ * Common properties of transaction responses.
+ *
+ * @category Responses
+ */
+interface BaseTxResult<
+  Version extends APIVersion,
+  T extends BaseTransaction = Transaction,
+> {
+  /** The SHA-512 hash of the transaction. */
+  hash: string
```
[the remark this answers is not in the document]
khancode: I don't see `close_time_iso` returned in the integration test response. Is it meant to be an optional field?

packages/xrpl/src/models/ledger/Ledger.ts
khancode: I'd rather keep the design consistent of having separate interfaces for each api version: `Ledger` for v2, `LedgerV1` for v1. Thus, I think it's better to keep it as is.

packages/xrpl/test/integration/requests/ledger.test.ts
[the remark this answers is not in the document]
ckeshava: hmm. I'm glad this has a uniform, consistent type now.

packages/xrpl/src/models/methods/submitMultisigned.ts
ckeshava: hmm right, makes sense 👍

packages/xrpl/src/models/methods/tx.ts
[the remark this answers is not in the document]
Bronek: It is only populated in validated transactions, and I think I was not able to pupulate it in every API operations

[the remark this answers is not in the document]
ckeshava: the `close_time_iso` is an optional field that's set on certain validated transactions only. It's appropriate to model it as such, in the `Tx` request's responses.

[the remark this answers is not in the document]
khancode: I pushed the change.

packages/xrpl/src/models/methods/accountTx.ts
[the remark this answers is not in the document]
khancode: I just added it to `BaseResponse`

packages/xrpl/src/models/methods/baseMethod.ts
```diff
@@ -38,6 +38,7 @@ export interface BaseResponse {
   warnings?: ResponseWarning[]
   forwarded?: boolean
   api_version?: number
+  meta_blob?: string
```
ckeshava, line 41: `meta_blob` needs to be at the same depth/level as `meta` field i.e. within `response.result`. At the moment, I can access both `response.result.meta_blob` and `response.meta_blob`.

Docs for tx request: https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/transaction-methods/tx/

This inclusion of `meta_blob` is not needed. 

I didn't mean to confuse you, I originally wanted a clarification on how `meta_blob` was being inherited across the classes.

ckeshava: `meta_blob` is already included in `TxResponse<Transaction>.result : BaseTxResult` -- you were initially correct, there is no need for this line

khancode: I reverted the change.

packages/xrpl/src/models/methods/index.ts
```diff
@@ -276,12 +286,15 @@ type Response =
   // Price Oracle methods
   | GetAggregatePriceResponse
 
-export type RequestResponseMap<T> = T extends AccountChannelsRequest
+export type RequestResponseMap<
+  T,
+  Version extends APIVersion,
```
justinr1234, line 291: @khancode is there a benefit to defaulting this value?

khancode: `Version` is already being defaulted from its upper level in `Client.request()` so it's defaulted indirectly. The benefit of doing this is so TS can default to using API v2 response model. Otherwise the response model returned ends up being a union type `v1 | v2`.

justinr1234: @khancode why not default it all the way down the stack?

khancode: Sounds like a good design and brings added benefit. I just pushed the change.

packages/xrpl/src/client/RequestManager.ts
```diff
@@ -206,7 +209,7 @@ export default class RequestManager {
     // status no longer needed because error is thrown if status is not "success"
     delete response.status
     // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Must be a valid Response here
-    this.resolve(response.id, response as unknown as Response)
+    this.resolve(response.id, response as unknown as Response<APIVersion>)
```
justinr1234, line 212: You shouldn't need to use type assertions here if you change `resolve` to be `Partial<Response<`

packages/xrpl/src/client/index.ts
```diff
@@ -319,16 +330,20 @@ class Client extends EventEmitter<EventTypes> {
...
+    T = RequestResponseMap<R, V>,
+  >(req: R): Promise<T> {
+    const request = {
       ...req,
       account: req.account
```
justinr1234, line 340: Change this to `typeof req.account === 'string'` instead of using `as`

packages/xrpl/src/client/index.ts
```diff
@@ -1037,9 +1053,13 @@ class Client extends EventEmitter<EventTypes> {
    * ```
    */
   public async getLedgerIndex(): Promise<number> {
-    const ledgerResponse = await this.request({
+    const ledgerResponse = await this.request<
+      LedgerRequest,
+      typeof this.apiVersion
```
justinr1234, line 1058: This can be removed

packages/xrpl/src/client/partialPayment.ts
```diff
@@ -63,7 +63,10 @@ function isPartialPayment(
   }
 
   const delivered = meta.delivered_amount
-  const amount = tx.Amount
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- DeliverMax is a valid field on Payment response
```
justinr1234, line 66: This doesn't seem like the right way to handle this. We should fix the types.

packages/xrpl/src/client/partialPayment.ts
```diff
@@ -82,14 +85,17 @@ function txEntryHasPartialPayment(response: TransactionEntryResponse): boolean {
 
 function accountTxHasPartialPayment(response: AccountTxResponse): boolean {
   const { transactions } = response.result
-  const foo = transactions.some((tx) => isPartialPayment(tx.tx, tx.meta))
+  const foo = transactions.some((tx) =>
+    // API v2 returns tx_json, API v1 returns tx
+    isPartialPayment(tx.tx_json ?? tx.tx, tx.meta),
```
justinr1234, line 90: This line isn't accurate. If you mouseover `tx.tx` it's showing as always `undefined` because `AccountTxresponse` defaults to `AccountResponseBase<2>`. Your code "works" but the type checking is wrong.

You need to make `accountTxHasPartialPayment` a template function that properly passes in the type of the API version and then `AccountTxResponse` also needs to be made generic in order to pass the version to that which is eventually passed to `AccountResponseBase` which gives you the proper typing (1 or 2)

packages/xrpl/src/models/ledger/Ledger.ts
```diff
@@ -63,3 +58,40 @@ export default interface Ledger {
...
+
+/**
+ * A ledger is a block of transactions and shared state data. It has a unique
+ * header that describes its contents using cryptographic hashes.
+ *
+ * @category Ledger Entries
+ */
+export interface Ledger extends BaseLedger {
+  /**
+   * The ledger index of the ledger. Represented as a native JSON number.
```
justinr1234, line 70: nit: weirdly worded to say "native json number"

packages/xrpl/src/models/methods/accountInfo.ts
```diff
@@ -180,3 +171,58 @@ export interface AccountInfoResponse extends BaseResponse {
...
+     * member if it is present.
+     */
+    signer_lists?: SignerList[]
+  }
+}
+
+/**
+ * Response expected from a {@link AccountInfoRequest} using API version 1.
+ *
+ * @category Responses
```
justinr1234, line 194: Maybe we should also add an @category for `v1` or otherwise make it so our output docs render better to tag things as `v1`

I would utilize both `@since` for api v2 functions and `@version` for API v1 stuff.

see: https://jsdoc.app/

packages/xrpl/src/sugar/submit.ts
```diff
@@ -153,7 +153,8 @@ export async function waitForFinalTransactionOutcome<
     })
 
   if (txResponse.result.validated) {
-    return txResponse
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- we know that txResponse is of type TxResponse
+    return txResponse as TxResponse<T>
```
justinr1234, line 157: Can you leave a comment to fix this later --- the typing is really messy

justinr1234 (commented):
- [ ] For all your tests, shouldn't we be keeping v1 tests? I know you added some v1 tests, but you changed a lot of tests to use `tx_json` but I didn't see tests for making sure v1 still works
- [ ] I think we should add a warning to people when they're trying to access v1 data but the library is defaulting to v2 -- Likely we can use a class getter than throws an error when they try to access the wrong `tx` or `tx_json`

packages/xrpl/src/models/ledger/Ledger.ts
khancode: Yeah, I can replace it with just "number".

packages/xrpl/src/client/partialPayment.ts
khancode: Technically, this is correct because `DeliverMax` is an RPC field and not a protocol field. So that's why it's not added to `Payment` transaction. Refer to this PR for more details - https://github.com/XRPLF/xrpl.js/pull/2689

packages/xrpl/src/sugar/submit.ts
khancode: Done

packages/xrpl/src/models/methods/accountInfo.ts
khancode: Sure, I can change it to `@category ResponsesV1` for v1 types and leave `@category Responses` for v2.

khancode: However, `@since` and `@version` require version values such as `2.0.0` so `v2` wouldn't think. I think this could confuse users thinking `2.0.0` means client library version and not rippled.

packages/xrpl/src/client/partialPayment.ts
justinr1234: @khancode right, but our library should work around they without type assertions

ckeshava: @justinr1234 how do you propose that we use model the `DeliverMax` field?

`DeliverMax` is not a property on the `Payment` transaction type. But V2 rippled responses allude to `DeliverMax` only, not to the erstwhile `Amount` field.

We could use the `any` type for these objects, but that removes all type-information for future engineers. Annotating these objects with `Payment` type (and some ts-lint exceptions) is more informative.

justinr1234 (approved):

PR #2657, opened 2024-03-07
fix: throw error if `hexToBytes` or `hexToString` is provided a string that is not in hex

mvadari:

Title says it all.

### Context of Change

The current methods return an empty bytes object/string if they are not provided a string in hex, which caused an issue.

### Type of Change

### Did you update `HISTORY.md`?

- [x] Yes

I also cleaned up the other `HISTORY.md` files.

## Test Plan

Added tests, they pass.

### DIFF
--- a/packages/isomorphic/test/utils.test.ts
+++ b/packages/isomorphic/test/utils.test.ts
@@ -23,8 +23,12 @@ describe('utils', function () {
     expect(hexToBytes('DEADBEEF')).toEqual(new Uint8Array([222, 173, 190, 239]))
   })
 
-  it('bytesToHex - DEADBEEF', () => {
-    expect(bytesToHex([222, 173, 190, 239])).toEqual('DEADBEEF')
+  it('hexToBytes - DEADBEEF', () => {
+    expect(hexToBytes('DEADBEEF')).toEqual(new Uint8Array([222, 173, 190, 239]))
+  })
+
+  it('bytesToHex - bad hex', () => {
+    expect(() => hexToBytes('hello')).toThrow(new Error('Invalid hex string'))
   })
 
   it('bytesToHex - 010203', () => {
@@ -41,6 +45,10 @@ describe('utils', function () {
 
   it('hexToString - deadbeef+infinity symbol (HEX)', () => {
     expect(hexToString('6465616462656566D68D')).toEqual('deadbeef֍')
+  })
+
+  it('hexToString - bad hex', () => {
+    expect(() => hexToString('hello')).toThrow(new Error('Invalid hex string'))
   })
 
   it('stringToHex - deadbeef+infinity symbol (utf8)', () => {
--- a/packages/ripple-address-codec/HISTORY.md
+++ b/packages/ripple-address-codec/HISTORY.md
@@ -15,25 +15,6 @@
 * Eliminates 4 runtime dependencies: `base-x`, `base64-js`, `buffer`, and `ieee754`.
 * Execute test in a browser in addition to node
 
-## 5.0.0 Beta 1 (2023-11-30)
-
-### Breaking Changes
-* `Buffer` has been replaced with `UInt8Array` for both params and return values. `Buffer` may continue to work with params since they extend `UInt8Arrays`.
-
-### Changes
-* Eliminates 4 runtime dependencies: `base-x`, `base64-js`, `buffer`, and `ieee754`.
-
-## 5.0.0 Beta 0 (2023-10-19)
-
-### Breaking Changes
-* Bump typescript to 5.x
-* Remove Node 14 support
-* Remove `assert` dependency. If you were catching `AssertionError` you need to change to `Error`.
-* Remove `create-hash` in favor of `@noble/hashes`
-
-### Changes
-* Execute test in a browser in addition to node
-
 ## 4.3.1 (2023-09-27)
 ### Fixed
 * Fix source-maps not finding their designated source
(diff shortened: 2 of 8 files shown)


packages/ripple-address-codec/HISTORY.md
```diff
@@ -15,25 +15,6 @@
 * Eliminates 4 runtime dependencies: `base-x`, `base64-js`, `buffer`, and `ieee754`.
 * Execute test in a browser in addition to node
 
-## 5.0.0 Beta 1 (2023-11-30)
```
ckniffen, line 18: This still feels relevant.  There were releases.  It also shows we released betas.

packages/isomorphic/test/utils.test.ts
```diff
@@ -23,8 +23,12 @@ describe('utils', function () {
     expect(hexToBytes('DEADBEEF')).toEqual(new Uint8Array([222, 173, 190, 239]))
   })
 
-  it('bytesToHex - DEADBEEF', () => {
-    expect(bytesToHex([222, 173, 190, 239])).toEqual('DEADBEEF')
```
ckniffen, line 27: Why was this one removed?

packages/isomorphic/test/utils.test.ts
```diff
@@ -23,8 +23,12 @@ describe('utils', function () {
...
+  it('hexToBytes - DEADBEEF', () => {
+    expect(hexToBytes('DEADBEEF')).toEqual(new Uint8Array([222, 173, 190, 239]))
+  })
+
+  it('bytesToHex - bad hex', () => {
+    expect(() => hexToBytes('hello')).toThrow(new Error('Invalid hex string'))
```
ckniffen, line 31: Should this be a special Error that can be more easily detected?

packages/ripple-address-codec/HISTORY.md
mvadari: It's not useful in the changelog IMO - it's already in the release list.

packages/isomorphic/test/utils.test.ts
ckeshava: this is in the unit tests, so I don't expect a special error code to be useful here. On the other hand, I feel the main code base could make use of a special error code / error status message

packages/isomorphic/src/utils/browser.ts
```diff
@@ -22,6 +22,9 @@ export const bytesToHex: typeof BytesToHexFn = (bytes) => {
 export const hexToBytes: typeof HexToBytesFn = (hex): Uint8Array => {
   const len = hex.length
   const array = new Uint8Array(len / 2)
+  if (!/^[A-F0-9]*$/iu.test(hex)) {
```
ckeshava, line 25: can this regex be placed in a global variable and re-used?

packages/isomorphic/test/utils.test.ts
ckniffen: Yeah that is what I mean.

packages/isomorphic/src/utils/browser.ts
ckniffen: Yes, it could be.

packages/isomorphic/test/utils.test.ts
mvadari: I checked what errors were used in the rest of the package and it was just `Error`, so I just copied the norm

ckeshava: hmm. Assigning special error codes (for all errors) might be a significant deviation from the intent of the PR. I'm okay if it's not included as a part of this change list

packages/ripple-address-codec/HISTORY.md
mvadari: I personally find negative value in having beta info in changelogs, even when looking at other packages. I think it just clutters up the file. But if you disagree I'm happy to roll back this part of the change.

packages/isomorphic/test/utils.test.ts
ckeshava: @mvadari  reminder : did you mean the retain the existing test case and add more?

ckeshava (changes requested):
Apart from the missing test case (I've tagged you on the comment), I'm okay with the rest of the PR.

I don't have a strong opinion on removing the beta changelist information. I'm weakly in favor of removing the information

packages/isomorphic/test/utils.test.ts
mvadari: Oops, typo

mvadari: Hm I'm not sure actually - either way I can add it back

ckeshava (approved):

packages/isomorphic/src/utils/shared.ts
```diff
@@ -1,5 +1,7 @@
 import { concatBytes } from '@noble/hashes/utils'
 
+export const HEX_REGEX = /^[A-F0-9]*$/iu
```
justinr1234, line 3: Do we want to change from `*` to `+` to prevent empty strings? @mvadari

mvadari: I purposefully kept support for empty strings - I don't see any reason why these functions shouldn't support empty strings, which would convert to an empty bytestring or empty ASCII string.

ckeshava: Yes, but in what cases would an empty string be a legitimate input?

pdp2121 (approved):

packages/isomorphic/src/utils/shared.ts
mvadari: One example: your app has a field that accepts memos and runs the result through `convertHexToString`

Issue #2666, opened 2024-03-19
tfTwoAssetIfEmpty flag is missing


zgrguric:
https://github.com/XRPLF/xrpl.js/blob/b9af7bdb6c6d601356da73f4531a035d56d73767/packages/xrpl/src/models/transactions/AMMDeposit.ts#L18

ckeshava:
hello @zgrguric ,
thank you for bringing this issue to our attention. we are working on fixing it.

PR #2667, opened 2024-03-20
additional amm flags are introduced

ckeshava:

Fix https://github.com/XRPLF/xrpl.js/issues/2666

This PR introduces additional flags used in the AMM-related transactions. I'm using these lines of code as the source of truth: https://github.com/XRPLF/rippled/blob/69143d71f8973e33b701d7becc19da2ad6a68b68/src/ripple/protocol/TxFlags.h#L166

### Context of Change
This change brings forth certain AMMFlags into the xrpl-js client library.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes
Do I need to update HISTORY file? this change implements what was promised in the AMM-transaction documentation.

## Test Plan

I'm not familiar with the AMM-transactions to write unit tests for all these flags. Please let me know if you have any ideas.

### DIFF
--- a/packages/xrpl/src/models/transactions/AMMDeposit.ts
+++ b/packages/xrpl/src/models/transactions/AMMDeposit.ts
@@ -21,6 +21,7 @@ export enum AMMDepositFlags {
   tfTwoAsset = 0x00100000,
   tfOneAssetLPToken = 0x00200000,
   tfLimitLPToken = 0x00400000,
+  tfTwoAssetIfEmpty = 0x00800000,
 }
 
 export interface AMMDepositFlagsInterface extends GlobalFlags {
@@ -29,6 +30,7 @@ export interface AMMDepositFlagsInterface extends GlobalFlags {
   tfTwoAsset?: boolean
   tfOneAssetLPToken?: boolean
   tfLimitLPToken?: boolean
+  tfTwoAssetIfEmpty?: boolean
 }
 
 /**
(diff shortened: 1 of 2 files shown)


packages/xrpl/src/models/transactions/AMMDeposit.ts
```diff
@@ -21,6 +21,25 @@ export enum AMMDepositFlags {
   tfTwoAsset = 0x00100000,
   tfOneAssetLPToken = 0x00200000,
   tfLimitLPToken = 0x00400000,
+  tfWithdrawAll = 0x00020000,
+  tfOneAssetWithdrawAll = 0x00040000,
+  tfTwoAssetIfEmpty = 0x00800000,
+  tfWithdrawSubTx = tfLPToken |
+    tfSingleAsset |
```
ckeshava, line 28: Can I not use existing flags to designate new flags? I get the following lint error ://

`error  Explicit enum value must only be a literal value (string, number, boolean, etc)  `

mvadari: There shouldn't be any masks here - the enum should just be the flags.
I would use https://xrpl.org/docs/references/protocol/transactions/types/ammdeposit/#ammdeposit-flags as your source of truth instead of rippled.

mvadari: It's the dev's job to `|` the flags together, not ours. `rippled` reuses a lot of these combinations, which is why that code has some combinations. They aren't independent flags, though.

ckeshava: ok thanks, fixed it

mvadari:
This should probably be released before AMM goes live tomorrow

mvadari (commented):
Please update the changelog

khancode (approved):

mvadari (approved):

PR #2679, opened 2024-04-05
length check warning on arg publicKeyBytes for method deriveAddressFromBytes

mattlockyer:
Currently addresses are derived from any length of bytes, but documentation states that addresses are only derived from compressed public key points:

https://xrpl.org/docs/concepts/accounts/cryptographic-keys/

Not sure where this should go. Perhaps in the codec package.

### Context of Change

I faced an error that was hard to debug since I was passing in an uncompressed public key point to deriveAddress and signing. Everthing was working locally with the xrpl libraries but the nodes were rejecting the transactions.

TLDR; I needed to call deriveAddress with the compressed public key point, as per documentation.

Nothing in the code prevents me from calling deriveAddress with any length of bytes, uncompressed or arbitrary length.

There should be a check.

### Type of Change

## Test Plan

Please add commit to put this check in the right place and add warning to the codebase. It will help other developers in the future.

### DIFF
--- a/packages/ripple-keypairs/src/index.ts
+++ b/packages/ripple-keypairs/src/index.ts
@@ -88,6 +88,9 @@ function computePublicKeyHash(publicKeyBytes: Uint8Array): Uint8Array {
 }
 
 function deriveAddressFromBytes(publicKeyBytes: Uint8Array): string {
+  if (publicKeyBytes.byteLength > 33) {
+    console.warn('publicKeyBytes should be compressed ed25519 or secp256k1 compressed point bytes, 32 bvytes prefixed with 02 or 03');
+  }
   return encodeAccountID(computePublicKeyHash(publicKeyBytes))
 }
 


packages/ripple-keypairs/src/index.ts
```diff
@@ -88,6 +88,9 @@ function computePublicKeyHash(publicKeyBytes: Uint8Array): Uint8Array {
 }
 
 function deriveAddressFromBytes(publicKeyBytes: Uint8Array): string {
+  if (publicKeyBytes.byteLength > 33) {
+    console.warn('publicKeyBytes should be compressed ed25519 or secp256k1 compressed point bytes, 32 bvytes prefixed with 02 or 03');
```
intelliot, line 92:
```suggestion
    console.warn('publicKeyBytes should be compressed ed25519 or secp256k1 compressed point bytes, 32 bytes prefixed with 02 or 03');
```

mvadari (commented):
Please update the `HISTORY.md` file.

justinr1234:
@coderabbit full review

justinr1234:
@coderabbitai full review

packages/ripple-keypairs/src/index.ts
```diff
@@ -88,6 +88,9 @@ function computePublicKeyHash(publicKeyBytes: Uint8Array): Uint8Array {
 }
```
ckeshava, line 88: Is it possible to make a similar update to `computePublicKeyHash` method on line 86 ? That is another function making use of `publicKeyBytes`

packages/ripple-keypairs/src/index.ts
```diff
@@ -88,6 +88,9 @@ function computePublicKeyHash(publicKeyBytes: Uint8Array): Uint8Array {
...
 
 function deriveAddressFromBytes(publicKeyBytes: Uint8Array): string {
+  if (publicKeyBytes.byteLength > 33) {
```
ckeshava, line 91: Are there any disadvantages to having a stricter check?
```suggestion
  if (publicKeyBytes.byteLength != 33) {
```

packages/ripple-keypairs/src/index.ts
```diff
@@ -88,6 +88,9 @@ function computePublicKeyHash(publicKeyBytes: Uint8Array): Uint8Array {
 }
 
 function deriveAddressFromBytes(publicKeyBytes: Uint8Array): string {
```
ckeshava, line 90: It occurs to me that these types of checks could be avoided by having an explicit user-defined class for `PublicKey` and `PrivateKey`. The constructor of the class can enforce these checks. But that change requires a lot more effort. 

Thanks for pointing this out!

PR #2684, opened 2024-04-23
introduce DeliverMax alias in Payment transactions

ckeshava:

This is a sister PR of the following xrpl-py work: https://github.com/XRPLF/xrpl-py/pull/684

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Unit tests have been added, but I am unable to add integration tests. I couldn't find any examples of specifying a JSON representation (not `Payment, AccountSet` or any `SubmittableTransaction` type) as an input to the `submitTransaction` function. This is necessary because `DeliverMax` is an alias at the RPC level, it is not recognised at the protocol level. Let me know if I'm missing something.

### DIFF
--- a/packages/xrpl/src/models/transactions/payment.ts
+++ b/packages/xrpl/src/models/transactions/payment.ts
@@ -167,6 +167,24 @@ export function validatePayment(tx: Record<string, unknown>): void {
   validateBaseTransaction(tx)
 
   if (tx.Amount === undefined) {
+    // If only DeliverMax is provided, use it to populate the Amount field
+    if (tx.DeliverMax !== undefined) {
+      tx.Amount = tx.DeliverMax
+    }
+  }
+
+  // If Amount is not identical to DeliverMax, throw an error
+  if (
+    tx.DeliverMax !== undefined &&
+    tx.Amount !== undefined &&
+    tx.Amount !== tx.DeliverMax
+  ) {
+    throw new ValidationError(
+      'PaymentTransaction: Amount and DeliverMax fields must be identical',
+    )
+  }
+
+  if (tx.Amount === undefined) {
     throw new ValidationError('PaymentTransaction: missing field Amount')
   }
 
--- a/packages/xrpl/src/models/transactions/payment.ts
+++ b/packages/xrpl/src/models/transactions/payment.ts
@@ -166,17 +166,17 @@ export interface PaymentMetadata extends TransactionMetadataBase {
 export function validatePayment(tx: Record<string, unknown>): void {
   validateBaseTransaction(tx)
 
-  if (tx.Amount === undefined) {
+  if (tx.Amount == null) {
     // If only DeliverMax is provided, use it to populate the Amount field
-    if (tx.DeliverMax !== undefined) {
+    if (tx.DeliverMax != null) {
       tx.Amount = tx.DeliverMax
     }
   }
 
   // If Amount is not identical to DeliverMax, throw an error
   if (
-    tx.DeliverMax !== undefined &&
-    tx.Amount !== undefined &&
+    tx.DeliverMax != null &&
+    tx.Amount != null &&
     tx.Amount !== tx.DeliverMax
   ) {
     throw new ValidationError(
--- a/packages/xrpl/test/models/payment.test.ts
+++ b/packages/xrpl/test/models/payment.test.ts
@@ -10,129 +10,13 @@ import { validatePayment } from '../../src/models/transactions/payment'
  */
 describe('Payment', function () {
   let paymentTransaction
-  let paytxn1, paytxn2, paytxn3, paytxn4, paytxn5
+  // let paytxn1, paytxn2, paytxn3, paytxn4, paytxn5
 
   beforeEach(function () {
     paymentTransaction = {
       TransactionType: 'Payment',
       Account: 'rUn84CUYbNjRoTQ6mSW7BVJPSVJNLb1QLo',
       Amount: '1234',
-      Destination: 'rfkE1aSy9G8Upk4JssnwBxhEv5p4mn2KTy',
-      DestinationTag: 1,
-      Fee: '12',
-      Flags: 2147483648,
-      LastLedgerSequence: 65953073,
-      Sequence: 65923914,
-      SigningPubKey:
-        '02F9E33F16DF9507705EC954E3F94EB5F10D1FC4A354606DBE6297DBB1096FE654',
-      TxnSignature:
-        '3045022100E3FAE0EDEC3D6A8FF6D81BC9CF8288A61B7EEDE8071E90FF9314CB4621058D10022043545CF631706D700CEE65A1DB83EFDD185413808292D9D90F14D87D3DC2D8CB',
-      InvoiceID:
-        '6F1DFD1D0FE8A32E40E1F2C05CF1C15545BAB56B617F9C6C2D63A6B704BEF59B',
-      Paths: [
-        [{ currency: 'BTC', issuer: 'r9vbV3EHvXWjSkeQ6CAcYVPGeq7TuiXY2X' }],
-      ],
-      SendMax: '100000000',
-    } as any
-
-    // JSON contains only Amount field
-    paytxn1 = {
-      TransactionType: 'Payment',
-      Account: 'rUn84CUYbNjRoTQ6mSW7BVJPSVJNLb1QLo',
-      Amount: '1234',
-      Destination: 'rfkE1aSy9G8Upk4JssnwBxhEv5p4mn2KTy',
-      DestinationTag: 1,
-      Fee: '12',
-      Flags: 2147483648,
-      LastLedgerSequence: 65953073,
-      Sequence: 65923914,
-      SigningPubKey:
-        '02F9E33F16DF9507705EC954E3F94EB5F10D1FC4A354606DBE6297DBB1096FE654',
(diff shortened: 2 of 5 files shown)


packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -166,6 +166,24 @@ export interface PaymentMetadata extends TransactionMetadataBase {
 export function validatePayment(tx: Record<string, unknown>): void {
   validateBaseTransaction(tx)
 
+  if (tx.Amount === undefined) {
```
khancode, line 169:
```suggestion
  if (tx.Amount == null) {
```
It's better to use value comparison with null `== null` since it will cover cases for both `undefined` and `null`.

packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -166,6 +166,24 @@ export interface PaymentMetadata extends TransactionMetadataBase {
...
+    // If only DeliverMax is provided, use it to populate the Amount field
+    if (tx.DeliverMax !== undefined) {
```
khancode, line 171:
```suggestion
    if (tx.DeliverMax != null) {
```

packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -166,6 +166,24 @@ export interface PaymentMetadata extends TransactionMetadataBase {
...
+      tx.Amount = tx.DeliverMax
+    }
+  }
+
+  // If Amount is not identical to DeliverMax, throw an error
+  if (
+    tx.DeliverMax !== undefined &&
+    tx.Amount !== undefined &&
+    tx.Amount !== tx.DeliverMax
```
khancode, line 180:
```suggestion
    tx.DeliverMax != null &&
    tx.Amount != null &&
    tx.Amount !== tx.DeliverMax
```

packages/xrpl/test/models/payment.test.ts
```diff
@@ -258,4 +375,21 @@ describe('Payment', function () {
...
+
+    assert.doesNotThrow(() => validatePayment(paytxn3))
+    assert.doesNotThrow(() => validate(paytxn3))
+
+    assert.throws(() => validatePayment(paytxn4))
+    assert.throws(() => validate(paytxn4))
+
+    assert.throws(() => validatePayment(paytxn5))
+    assert.throws(() => validate(paytxn5))
+  })
```
khancode, line 394: Each different test should be separated into its own `it()` function.

khancode (changes requested):

packages/xrpl/test/models/payment.test.ts
ckeshava: ok, I've addressed this comment in a8036ae7e54219003e03ed41ffb5bb347335eb30

packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -166,6 +166,24 @@ export interface PaymentMetadata extends TransactionMetadataBase {
 export function validatePayment(tx: Record<string, unknown>): void {
   validateBaseTransaction(tx)
 
+  if (tx.Amount == null) {
```
mvadari, line 169: This feels like it's changing the remit of the `validate` functions, which should only be throwing errors and not mutating the objects.

ckeshava: Hmm, I agree. But I couldn't find another appropriate place to sneak in this mutation.
Perhaps I could include this change inside `ripple-binary-codec`. But since this is an RPC-related change and has nothing to do with protocol level serialization/de-serialization, it doesn't feel appropriate.

I'm also questioning the necessity of this change. rippled performs validation checks on the `Amount` and `DeliverMax` fields in the `Payment` transactions. So even if the client libraries send incorrect input, the user should get a `tem...` error code from rippled.

mvadari: I think having such a check in `autofill` would be nice.

Yes you get a `tem` code, but it's better to fail fast. Otherwise, why even bother to have the `validate` functions in the first place?

packages/xrpl/test/models/payment.test.ts
```diff
@@ -33,6 +34,122 @@ describe('Payment', function () {
...
+        '02F9E33F16DF9507705EC954E3F94EB5F10D1FC4A354606DBE6297DBB1096FE654',
+      TxnSignature:
+        '3045022100E3FAE0EDEC3D6A8FF6D81BC9CF8288A61B7EEDE8071E90FF9314CB4621058D10022043545CF631706D700CEE65A1DB83EFDD185413808292D9D90F14D87D3DC2D8CB',
+      InvoiceID:
+        '6F1DFD1D0FE8A32E40E1F2C05CF1C15545BAB56B617F9C6C2D63A6B704BEF59B',
+      Paths: [
+        [{ currency: 'BTC', issuer: 'r9vbV3EHvXWjSkeQ6CAcYVPGeq7TuiXY2X' }],
+      ],
+      SendMax: '100000000',
+    } as any
```
khancode, line 152: It's more appropriate to put these objects in their corresponding tests.

packages/xrpl/src/models/transactions/payment.ts
ckeshava: I believe you are referring to this `autofill` : https://github.com/XRPLF/xrpl.js/blob/8b596a6687ff89578973429f2b05d622a77b4249/packages/xrpl/src/client/index.ts#L642

It accepts only `SubmittableTransaction`'s. JSON-string representation of certain `Payment` transactions, which include `DeliverMax` field are not serialization friendly with `encode` and `decode` JS functions.


Please refer to this branch, it has two unit tests demonstrating this issue: https://github.com/XRPLF/xrpl.js/commit/d262da9dc4fbd7bc6928b5a7a952ea674429b0f9

If we want to insert this change in `autofill`, we will need to remove Type restrictions of `SubmittableTransaction` and perform cosmetic changes prior to the call to `decode` here: https://github.com/XRPLF/xrpl.js/blob/8b596a6687ff89578973429f2b05d622a77b4249/packages/xrpl/src/sugar/submit.ts#L248

packages/xrpl/test/models/payment.test.ts
ckeshava: fixed in [109a88d](https://github.com/XRPLF/xrpl.js/pull/2684/commits/109a88d4057e8f3b4f1b31b8642254e59b6e7262)

packages/xrpl/src/models/transactions/payment.ts
mvadari: If you're using TS, it'll get caught anyways, so that's not who we're worried about.

ckeshava: what do you mean? I don't understand.

Irrespective of TS or JS, since the input is a JSON string representation of tx_json, we will need to (1) either update the `decode` function, or (2) modify `tx_json` before and after the decode/encode function calls.

mvadari: `encode`/`decode` are never called in `autofill`, I'm not sure what your concern is.

ckeshava: I can't invoke `autofill` directly on a tx_json, because autofill accepts only a `SubmittableTransaction`. A JSON input with `DeliverMax` does not fit the requirements of a `SubmittableTransaction`. So `autofill` cannot be used for our purposes.

ckeshava: I need to throw an error if `DeliverMax` != `Amount`. `autofill`'s API does not allow for this possibility

ckeshava: I'm exploring the alternative `submit` function: it could be used for populating the `DeliverMax` field. Here is a typical function call hierarchy:
`submit` -> `getSignedTx` -> `autofill`. This is a viable approach. But throwing the above-mentioned error is still not possible with this approach.

what do you think?

mvadari: That's what I'm referring to on TS vs JS. TS will error if you have a `DeliverMax` field, JS will let you continue because it has no type checking. I think an `as any` is fine for this check.

mvadari: A lot of folks call `autofill`/`sign`/`submit` on their own, so updating `submit` wouldn't do anything.

ckeshava: Okay, I have implemented this idea here: https://github.com/ckeshava/xrpl.js/pull/new/autofill-temp

But we are losing transaction model checks, as shown by the last unit test -- https://github.com/XRPLF/xrpl.js/blob/13125564baafcc013c9f911bd647ab577c80608f/packages/xrpl/test/client/autofill.test.ts#L165

Is this what you had in mind?

mvadari: What I'm suggesting is that only lines 672-696 use an `as any` version of the object. The rest of the checks can use the normal transaction.

ckeshava: Okay, here is another PR: https://github.com/XRPLF/xrpl.js/pull/2689

khancode:
Closing this since the other solution was merged in - https://github.com/XRPLF/xrpl.js/pull/2689

PR #2688, opened 2024-05-03
feat: add Price Oracles support

khancode:
## High Level Overview of Change

Add Price Oracles support - https://github.com/XRPLF/XRPL-Standards/discussions/129

### Context of Change

Adds transactions: `OracleSet`, `OracleDelete`

Adds request: `get_aggregate_price`

Adds ledger entry: `Oracle`

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### Did you update HISTORY.md?

- [x] Yes
- [ ] No, this change does not impact library users

## Test Plan

Adds unit/integration tests for Price Oracle transactions and request.

PR #2693, opened 2024-05-07
fix: better error handling for the binary codec

mvadari:

Title says it all.

### Context of Change

#2688 issues.

### Type of Change

- [x] Yes

## Test Plan

Tested by hand. TODO: add error test cases


packages/ripple-binary-codec/src/types/blob.ts
```diff
@@ -33,6 +33,9 @@ class Blob extends SerializedType {
     }
 
     if (typeof value === 'string') {
+      if (!/^[A-F0-9]*$/iu.test(value)) {
+        throw new Error('Cannot construct Bob from a non-hex string')
```
khancode, line 37: I don't think we'd want to construct Bob 👷‍♂️ from a non-hex string
```suggestion
        throw new Error('Cannot construct Blob from a non-hex string')
```

packages/ripple-binary-codec/src/types/st-array.ts
```diff
@@ -13,8 +13,14 @@ const OBJECT_END_MARKER = Uint8Array.from([0xe1])
...
-    Array.isArray(args) && (args.length === 0 || typeof args[0] === 'object')
+  if (!Array.isArray(args)) {
+    return false
+  }
+  return args.every(
+    (arg) =>
+      typeof arg === 'object' &&
+      Object.keys(arg).length === 1 &&
+      typeof Object.values(arg)[0] === 'object',
   )
```
khancode, line 24:
```suggestion
  return (
    Array.isArray(args) || args.every(
      (arg) =>
        typeof arg === 'object' &&
        Object.keys(arg).length === 1 &&
        typeof Object.values(arg)[0] === 'object',
    )
)
```

packages/ripple-binary-codec/test/binary-parser.test.ts
```diff
@@ -238,19 +238,19 @@ function fieldParsingTests() {
...
     expect(() => parser.readFieldOrdinal()).toThrow(
-      new Error('Cannot read FieldOrdinal, field_code out of range'),
+      new Error('Cannot read FieldOrdinal, field_code 1 out of range'),
     )
   })
   it('Field throws when both type and field code out of range', () => {
     const parser = makeParser('000101')
     expect(() => parser.readFieldOrdinal()).toThrow(
-      new Error('Cannot read FieldOrdinal, type_code out of range'),
+      new Error('Cannot read FieldOrdinal, type_code 1 out of range'),
```
khancode, line 253: `1 out of range` doesn't look correct.. I also have no idea how this works.

packages/ripple-binary-codec/src/types/blob.ts
mvadari: 🤣

khancode (approved):
LGTM 👍

packages/ripple-binary-codec/test/binary-parser.test.ts
ckeshava: in my local runs of this test, I observe the value `1`, as reported in these print statements. But I don't understand the parsing logic.

mvadari: It doesn't matter, the only reason these test changes are needed are due to the better error reporting in `binary-parser.ts`. This test is written to cause those errors.

packages/ripple-binary-codec/src/types/st-array.ts
```diff
@@ -14,7 +14,13 @@ const OBJECT_END_MARKER = Uint8Array.from([0xe1])
  */
 function isObjects(args): args is Array<JsonObject> {
   return (
-    Array.isArray(args) && (args.length === 0 || typeof args[0] === 'object')
+    Array.isArray(args) &&
+    args.every(
+      (arg) =>
+        typeof arg === 'object' &&
+        Object.keys(arg).length === 1 &&
```
ckeshava, line 21: could you explain why we need lines 21 and 22?

```
{
"key1": "value1",
"key2": "value2"
}
```

An input like this is rejected because there are 2 keys, and the values are "string", not "object". Why do we need this condition?

mvadari: That's not what valid inner objects look like.
This is what they must look like:
```
{
  "PriceData": {
    "BaseAsset": "XRP",
    "QuoteAsset": "USD",
    "AssetPrice": 740,
    "Scale": 3
  }
}
```

ckeshava: hmm okay. this format slipped my mind 👍

ckeshava: thanks

ckeshava (approved):

PR #2719, opened 2024-07-01
feat: add `feature` RPC

mvadari:

This PR adds the `feature` RPC to the models.

### Context of Change

There is now a non-admin user version of this RPC: https://github.com/XRPLF/rippled/pull/4781

### Type of Change

- [x] Yes

## Test Plan

Added integration tests.


packages/xrpl/test/integration/requests/feature.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      assert.equal(amendmentData.name, 'AMM')
+      // TODO: rippled says "false" for standalone nodes for some reason
+      assert.typeOf(amendmentData.enabled, 'boolean')
+      assert.equal(amendmentData.supported, true)
+    },
+    TIMEOUT,
+  )
+
+  it(
+    'all',
```
ckeshava, line 46: could you provide a different description for this integration test?

packages/xrpl/test/integration/requests/feature.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      const featureRequest: FeatureRequest = {
+        command: 'feature',
+        feature: AMENDMENT,
+      }
+      const featureResponse = await testContext.client.request(featureRequest)
+
+      assert.equal(featureResponse.type, 'response')
+      assert.typeOf(featureResponse.result, 'object')
+      assert.isTrue(AMENDMENT in featureResponse.result)
+
```
ckeshava, line 57:
```suggestion
      assert.equal(Object.keys(featureResponse.result).length, 1) // only one relevant feature must be returned
```

ckeshava (changes requested):

packages/xrpl/src/models/methods/feature.ts
```diff
@@ -0,0 +1,68 @@
+import { BaseRequest, BaseResponse } from './baseMethod'
+
+export interface FeatureAllRequest extends BaseRequest {
+  command: 'feature'
+
+  feature?: never
```
ckeshava, line 6:
```suggestion
```

Can we remove this line?
I've seen `never` used in the context of infeasible control flows. Why do we need to state the absence of a field in an interface?

mvadari: It's necessary to distinguish the types. One weirdness about TS types is that you can add on extra fields and it'll still be that type. So this is still a valid `FeatureAllRequest` object according to TS without this field:
```typescript
{
  command: 'feature'
  feature: 'something'
}
```
Adding the `never` is needed to ensure that this isn't actually a valid `FeatureAllRequest` type.

The `?` is needed because for some reason you get errors for `'feature' field not included` in the `FeatureAllRequest` type if you don't have it.

mvadari: cc @justinr1234, our local expert on TS types

justinr1234: Will look

mvadari: An example in our codebase: https://github.com/XRPLF/xrpl.js/blob/7e733c4446b9f2bc80b144bff03d46bdd5c9df93/packages/xrpl/src/models/ledger/BaseLedgerEntry.ts#L18-L26

ckeshava: yes, I did come across other examples in the codebase

ckeshava: I only saw control-flow related examples in the Typescript docs website: https://www.typescriptlang.org/docs/handbook/2/functions.html#never

It just feels strange that we are listing non-existent variables in the interface. But I get the rationale as a work-around to typescript's type inference.

This is not a strong objection, but it's nice to not have to do this

ckeshava (approved):

packages/xrpl/src/models/methods/feature.ts
mvadari: I agree that it's kind of ugly, but I think it's necessary in TypeScript's typing system.

justinr1234 (approved):

Issue #1903, opened 2022-01-21
Parse transaction flags into map of names:booleans


mDuo13:
xrpl.js has [`setTransactionFlagsToNumber()`](https://js.xrpl.org/modules.html#setTransactionFlagsToNumber) to convert flags _from_ a map of names to an integer, and [`parseAccountRootFlags()`](https://js.xrpl.org/modules.html#parseAccountRootFlags) for going from an AccountRoot object's flags integer to a map of booleans, but there's no helper to parse transaction flags from an integer into a map of booleans.

This is a common point of confusion for new XRPL devs, who don't know for example why, for example, [this transaction](https://livenet.xrpl.org/transactions/C1356ED3033210B0D153D7CDFBD4BF1076858E430E8ABA6D34C942C17051A5E0/raw) has `"Flags":2148007936` when there is no flag with the value 2148007936. (It's actually the combination of two flags, tfSell and tfFullyCanonicalSig.)

I suggest adding a `parseTransactionFlags(transaction)` method that returns a map of flag names with booleans, e.g.:

```json
{
  "tfSell": true,
  "tfFullyCanonicalSig": true
}
```

I'm ambivalent about whether it should include flags with the value `false` (meaning, ones that are valid for this transaction type but not enabled).

achowdhry-ripple:
Will pick this up

slurpyone:
## Issue Already Resolved ✅

This issue has been **resolved** in the current codebase! The `parseTransactionFlags(transaction)` method has been implemented and is available.

### Implementation Details

The function is located in `packages/xrpl/src/models/utils/flags.ts` and is exported in `packages/xrpl/src/models/index.ts`.

**Function signature:**
```typescript
export function parseTransactionFlags(tx: Transaction): object
```

### Example Usage

For the transaction mentioned in this issue with `"Flags": 2148007936`:

```typescript
import { parseTransactionFlags } from 'xrpl'

const tx = {
  TransactionType: 'OfferCreate',
  Flags: 2148007936,
  // ... other transaction fields
}

const flagsMap = parseTransactionFlags(tx)
console.log(flagsMap)
// Output: { "tfSell": true, "tfFullyCanonicalSig": true }
```

### Verification

The flags value `2148007936` correctly decomposes to:
- `tfFullyCanonicalSig` (0x80000000 = 2147483648) 
- `tfSell` (0x00080000 = 524288)
- Combined: 2148007936 ✅

### Recommendation

[...]

PR #2734, opened 2024-07-11
Parse transaction flags into map of names:booleans

achowdhry-ripple:
Add a new utility function that converts a transaction's flag field into a map response object, returning the flags that are enabled.

### Context of Change
https://github.com/XRPLF/xrpl.js/issues/1903

### Type of Change
Improvement

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes


pdp2121:
Please update CHANGELOG.md

packages/xrpl/test/models/utils.test.ts
```diff
@@ -207,5 +208,43 @@ describe('Models Utils', function () {
...
       assert.isUndefined(parsed.lsfAllowTrustLineClawback)
     })
+
+    it('parseTransactionFlags all enabled', function () {
+      const tx: PaymentChannelClaim = {
+        Account: 'r...',
+        TransactionType: 'PaymentChannelClaim',
+        Channel:
+          'C1AE6DDDEEC05CF2978C0BAD6FE302948E9533691DC749DCDD3B9E5992CA6198',
+        Flags: {
```
pdp2121, line 218: We should add a test case for when Flags is a number. (e.g. [this transaction](https://livenet.xrpl.org/transactions/C1356ED3033210B0D153D7CDFBD4BF1076858E430E8ABA6D34C942C17051A5E0/raw)

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -90,3 +89,30 @@ function convertFlagsToNumber(flags: GlobalFlags, flagEnum: any): number {
     return flags[flag] ? resultFlags | flagEnum[flag] : resultFlags
   }, 0)
 }
+
+/**
+ * Convert a Transaction flags property into an interface for easy interpretation.
+ *
+ * @param tx - A transaction to parse flags for.
+ * @returns An interface with all flags as booleans.
```
khancode, line 97:
```suggestion
 * @returns A map with all flags as booleans.
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -90,3 +89,30 @@ function convertFlagsToNumber(flags: GlobalFlags, flagEnum: any): number {
     return flags[flag] ? resultFlags | flagEnum[flag] : resultFlags
   }, 0)
 }
+
+/**
+ * Convert a Transaction flags property into an interface for easy interpretation.
```
khancode, line 94:
```suggestion
 * Convert a Transaction flags property into a map interface for easy interpretation.
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -90,3 +89,30 @@ function convertFlagsToNumber(flags: GlobalFlags, flagEnum: any): number {
...
+ * @returns An interface with all flags as booleans.
+ */
+export function parseTransactionFlags(tx: Transaction): object {
+  setTransactionFlagsToNumber(tx)
+  if (typeof tx.Flags !== 'number' || !tx.Flags || tx.Flags === 0) {
+    return {}
+  }
+
+  const flags = tx.Flags
+  const flagsMap = {}
```
khancode, line 106: Why not use JavaScripts native `Map` class instead of an object?

packages/xrpl/test/models/utils.test.ts
achowdhry-ripple: added

packages/xrpl/src/models/utils/flags.ts
achowdhry-ripple: My thinking here was to keep it consistent with the parseAccountRootFlags logic in the utils, and couldn't see how much added efficiency it would add in this case vs. the extra line for exporting as an object to return. I don't feel too strongly either way, but is there a significant benefit you see for this?

achowdhry-ripple: Is it possible to have duplication in flags for this case? I assumed not but that would be a hole in this logic if its the case

khancode: Ah I see. I agree to rather not break consistency and go with using an Object then. I don't think there can be duplication using an Object, it will just overwrite the existing key entry with the new value.

khancode (approved):

packages/xrpl/HISTORY.md
```diff
@@ -2,6 +2,9 @@
 
 Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xrpl-announce) for release announcements. We recommend that xrpl.js (ripple-lib) users stay up-to-date with the latest stable release.
 
+## Unreleased Changes
```
pdp2121, line 5: nit:

## Unreleased
### Added

pdp2121 (approved):
LGTM. Just a small nit change in HISTORY

PR #2795, opened 2024-10-02
Integration test for OracleSet txn

ckeshava:
Validate highest admissible `AssetPrice` value in the OracleSet transaction.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes


pdp2121 (approved):
LGTM
I think it's worth to mention this fix in HISTORY.md so people can update accordingly and avoid this issue

ckeshava:
Thanks for the quick approval @pdp2121 
The PriceOracle amendment hasn't been enabled on the mainnet yet, so I don't believe it needs a mention in the HISTORY file.

If you feel strongly about it, I can include that change too

mvadari:
> Thanks for the quick approval @pdp2121 The PriceOracle amendment hasn't been enabled on the mainnet yet, so I don't believe it needs a mention in the HISTORY file.
> 
> If you feel strongly about it, I can include that change too

It's a change in the library, it should be mentioned. It matters for people testing on Devnet.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -81,7 +81,7 @@ export function isString(str: unknown): str is string {
  * @returns Whether the number is properly formed.
  */
 export function isNumber(num: unknown): num is number {
-  return typeof num === 'number'
+  return typeof num === 'number' || typeof num === 'bigint'
```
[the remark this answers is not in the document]
ckeshava: - I'm not aware of cases where `OracleSet` transaction's `AssetPrice` field is used as an `Amount` field.
- Even if that is the case, type-coercion of `bigint` into `number` type causes loss of precision. I'd prefer to do it the other way around.
- The precision supported by the `Amount` field is not as large as that of `bigint`. `IssuedCurrencies` use `54` bits whereas `XRP` Amounts need `57` bits, to represent their highest values. The former case needs further investigation, appropos the `parseAmountValue` function.

mvadari (commented):
Why are there now keypairs changes in this PR?

packages/xrpl/src/models/transactions/common.ts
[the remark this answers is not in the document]
ckeshava: Continuing on the last bullet point, I incorporated the suggestions in this commit d53dd5eb6726177d218f3da11bc9eb266b4c0085

ckeshava:
> Why are there now keypairs changes in this PR?

Sorry, my bad. I messed up the git history. I rectified it with the relevant cherry-picked commits.

packages/xrpl/src/models/transactions/common.ts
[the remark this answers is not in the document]
ckeshava: thanks rabbit :)

ckeshava:
@achowdhry-ripple  can you please review this PR at your convenience?

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -400,15 +400,19 @@ export function validateBaseTransaction(common: Record<string, unknown>): void {
  *
  * @param amount - An Amount to parse for its value.
  * @returns The parsed amount value, or NaN if the amount count not be parsed.
+ * @throws ValidationError, if the input Amount is invalid
+ * @throws SyntaxError, if Amount cannot be parsed by BigInt constructor
  */
-export function parseAmountValue(amount: unknown): number {
+export function parseAmountValue(amount: unknown): bigint {
```
khancode, line 406: Couldn't the amount be a floating point number? In that case, it would be `number` instead of `bigint`. `bigint` doesn't support floating point numbers.

ckeshava: Thanks for pointing this out. I missed the fact that `bigint` cannot handle floats.

I wrote this PR with the perspective of `AssetPrice` field in OracleSet transaction which only accepts `std::uint_64` values.

However, you are correct. We cannot make this assumption for the rest of `Amount` fields. Do you have any suggestion for javascript libraries which accept floating-point numbers and also are compatible with rippled's `AMOUNT` internal-type field?

packages/xrpl/test/integration/transactions/oracleSet.test.ts
```diff
@@ -39,6 +43,14 @@ describe('OracleSet', function () {
               Scale: 3,
             },
           },
+          {
+            PriceData: {
+              BaseAsset: 'XRP',
+              QuoteAsset: 'INR',
+              AssetPrice: BigInt(MAX_64_BIT_UNSIGNED_INT),
```
mvadari, line 50: Does this not work out of the box if you specify the number as a string?

packages/xrpl/src/models/transactions/common.ts
khancode: You can just make the return type a union: `number | bigint`

packages/xrpl/test/integration/transactions/oracleSet.test.ts
ckeshava: @mvadari  you are correct, thanks for pointing it out. I didn't realize that `str` type preserves the range of the UINT64 type.

@khancode I'd like to close this PR -- Modifying the parseFloat method might not be needed and I don't have the time to look into it.

However, this [if-condition](https://github.com/XRPLF/xrpl.js/blob/0a45c58934620d8cad127a8d5652e7bc9f377836/packages/xrpl/src/models/transactions/oracleSet.ts#L149) needs to be removed for this integration test to work. Presently, developers cannot use `str` type inputs for the `AssetPrice` field in `OracleSet` transaction.

khancode: Then create another PR to remove the if-condition?

ckeshava: I can't take it up for the next two weeks. I want to complete other tasks, I can take it up after that.

khancode: Then follow the process of creating a ticket and moving it to the appropriate sprint to work on later.

khancode: Let's discuss this offline but close this PR and create a ticket to track the fix.

PR #2801, opened 2024-10-07
feat: add support for Batch amendment

mvadari:

This PR:
* Adds models for the `Batch` transactions
* Adds helper functions for signing and combining multi-account `Batch` transactions
* Adds support to `autofill` for filling in the `BatchTxn` and `TxIDs` fields

### Context of Change

https://github.com/XRPLF/rippled/pull/5060

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Added tests for the new features.


packages/xrpl/src/client/index.ts
```diff
@@ -683,33 +684,11 @@ class Client extends EventEmitter<EventTypes> {
     if (tx.TransactionType === 'AccountDelete') {
       promises.push(checkAccountDeleteBlockers(this, tx))
     }
-
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- ignore type-assertions on the DeliverMax property
-    // @ts-expect-error -- DeliverMax property exists only at the RPC level, not at the protocol level
-    if (tx.TransactionType === 'Payment' && tx.DeliverMax != null) {
```
mvadari, line 689: This is moved to `autofill.ts`

mvadari:
cc @dangell7 @tequdev

mvadari:
@coderabbitai full review

dangell7:
We should think about adding the ability to generate the index/hash for all objects especially NFTokenID.

mvadari:
> We should think about adding the ability to generate the index/hash for all objects especially NFTokenID.

Agreed, but I think that can be a separate PR.

mvadari:
Integration tests are expected to fail, since there is an integration test for `Batch` (which is not in the Docker container yet).

packages/xrpl/test/integration/transactions/batch.test.ts
```diff
@@ -0,0 +1,113 @@
...
+            RawTransaction: {
+              TransactionType: 'Payment',
+              Account: wallet2.classicAddress,
+              Destination: destination.classicAddress,
+              Amount: '10000000',
+            },
+          },
+        ],
+      }
+      const autofilled = await testContext.client.autofill(tx, 2)
```
ckeshava, line 107:
```suggestion
      const autofilled = await testContext.client.autofill(tx)
```

Do Batch transactions need to manually specify the number of signers? `calculateFeePerTransactionType` does not consider `signersCount` whilst computing the transactions fees for `Batch` transaction.

packages/xrpl/src/Wallet/defaultFaucets.ts
```diff
@@ -14,11 +14,13 @@ export interface FaucetWallet {
 export enum FaucetNetwork {
   Testnet = 'faucet.altnet.rippletest.net',
   Devnet = 'faucet.devnet.rippletest.net',
+  Batchnet = 'batch.faucet.nerdnest.xyz',
```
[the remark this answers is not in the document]
mvadari: > Can we run the integration tests against this devnet?

That would require a lot of code changes to the integration tests, since they're designed to be run on a standalone node.

> With this devnet, we don't have to wait for final approval of cpp `rippled` code

We still do, in case there are last-minute changes to rippled code - we don't want to have to introduce breaking changes.

> I'm wondering if this devnet has the latest version of rippled with Batch amendment code.

I'm not sure.

packages/ripple-binary-codec/src/index.ts
```diff
@@ -110,6 +111,13 @@ function encodeForMultisigning(
   )
 }
 
+function encodeForSigningBatch(json: object): string {
+  if (typeof json !== 'object') {
+    throw new Error()
+  }
```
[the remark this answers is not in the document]
mvadari: The other fields are checked in `signingBatchData`

packages/xrpl/src/client/RequestManager.ts
```diff
@@ -195,7 +196,9 @@ export default class RequestManager {
       // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- We know this must be true
       const errorResponse = response as Partial<ErrorResponse>
       const error = new RippledError(
-        errorResponse.error_message ?? errorResponse.error,
+        errorResponse.error_message ??
+          errorResponse.error_exception ??
+          errorResponse.error,
```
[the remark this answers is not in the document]
mvadari: Not that I know of. Different error fields are used in different cases - `error_code` will not always show up (and will also be more opaque for users).

packages/xrpl/src/utils/hashes/hashLedger.ts
```diff
@@ -66,7 +68,7 @@ function addLengthPrefix(hex: string): string {
  *
  * @param tx - A transaction to hash. Tx may be in binary blob form. Tx must be signed.
  * @returns A hash of tx.
- * @throws ValidationError if the Transaction is unsigned.\
+ * @throws ValidationError if the Transaction is unsigned.
```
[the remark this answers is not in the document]
mvadari: That's a given.

packages/xrpl/test/integration/transactions/batch.test.ts
mvadari: I'm using `signersCount` as the total number of multisign signers _and_ the total number of Batch signers.

packages/xrpl/src/Wallet/defaultFaucets.ts
```diff
@@ -36,6 +38,10 @@ export function getFaucetHost(client: Client): FaucetNetwork | undefined {
     return FaucetNetwork.Testnet
   }
 
+  if (connectionUrl.includes('batchnet')) {
+    return FaucetNetwork.Batchnet
+  }
```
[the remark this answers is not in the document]
mvadari: Depends on #2950

packages/xrpl/test/models/Batch.test.ts
```diff
@@ -0,0 +1,136 @@
...
+          },
+        },
+      ],
+      TransactionType: 'Batch',
+      TxIDs: [
+        'ABE4871E9083DF66727045D49DEEDD3A6F166EB7F8D1E92FE868F02E76B2C5CA',
+        '795AAC88B59E95C3497609749127E69F12958BC016C600C770AEEB1474C840B4',
+      ],
+    } as any
+  })
```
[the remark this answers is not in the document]
mvadari: `as any` is fine for tests

packages/xrpl/test/models/Batch.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      () => validateBatch(tx),
+      ValidationError,
+      'Batch: missing field RawTransactions',
+    )
+    assert.throws(
+      () => validate(tx),
+      ValidationError,
+      'Batch: missing field RawTransactions',
+    )
+  })
```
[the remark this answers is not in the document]
mvadari: `delete` is fine for tests

packages/xrpl/src/models/transactions/batch.ts
```diff
@@ -0,0 +1,149 @@
...
+// eslint-disable-next-line max-lines-per-function -- needed here due to the complexity
+export function validateBatch(tx: Record<string, unknown>): void {
+  validateBaseTransaction(tx)
+
+  validateRequiredField(tx, 'RawTransactions', isArray)
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- checked above
+  const rawTransactions = tx.RawTransactions as unknown[]
+  rawTransactions.forEach((field, index) => {
+    if (!isObject(field)) {
+      throw new ValidationError(`Batch: RawTransactions[${index} is not object`)
```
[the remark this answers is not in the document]
mvadari: Fixed

packages/xrpl/src/models/transactions/batch.ts
```diff
@@ -0,0 +1,149 @@
...
+  })
+  // Full validation of each `RawTransaction` object is done in `validate` to avoid dependency cycles
+
+  validateOptionalField(tx, 'BatchSigners', isArray)
+
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- checked above
+  const batchSigners = tx.BatchSigners as unknown[] | undefined
+  batchSigners?.forEach((field, index) => {
+    if (!isObject(field)) {
+      throw new ValidationError(`Batch: BatchSigners[${index} is not object`)
```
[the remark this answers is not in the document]
mvadari: Fixed

packages/xrpl/src/models/transactions/batch.ts
```diff
@@ -0,0 +1,149 @@
...
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- needed here
+    // @ts-expect-error -- checked above
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- checked above
+    const rawTx = field.RawTransaction as Record<string, unknown>
+    if (!isObject(rawTx)) {
+      throw new ValidationError(`Batch: RawTransactions[${index} is not object`)
+    }
+    if (rawTx.TransactionType === 'Batch') {
+      throw new ValidationError(
+        `Batch: RawTransactions[${index} is a Batch transaction. Cannot nest Batch transactions.`,
```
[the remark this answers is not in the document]
mvadari: Fixed

packages/xrpl/test/wallet/batchSigner.test.ts
```diff
@@ -0,0 +1,248 @@
...
+  algorithm: ECDSA.secp256k1,
+})
+const edWallet = Wallet.fromSeed('spkcsko6Ag3RbCSVXV2FJ8Pd4Zac1', {
+  algorithm: ECDSA.ed25519,
+})
+const submitWallet = Wallet.fromSeed('sEd7HmQFsoyj5TAm6d98gytM9LJA1MF', {
+  algorithm: ECDSA.ed25519,
+})
+const otherWallet = Wallet.generate()
+
```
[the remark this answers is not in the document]
mvadari: These seeds don't matter and aren't protected data at all.

packages/xrpl/test/models/Batch.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      () => validateBatch(tx),
+      ValidationError,
+      'Batch: invalid field TxIDs',
+    )
+    assert.throws(
+      () => validate(tx),
+      ValidationError,
+      'Batch: invalid field TxIDs',
+    )
+  })
```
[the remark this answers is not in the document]
mvadari: Not an issue anymore

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -177,11 +177,50 @@
...
+  const txIDsLength = coreTypes.UInt32.from(batch.txIDs.length).toBytes()
+
+  const bytesList = new BytesList()
+
+  bytesList.put(prefix)
+  bytesList.put(flags)
+  bytesList.put(txIDsLength)
+  batch.txIDs.forEach((txID: string) => {
+    bytesList.put(coreTypes.Hash256.from(txID).toBytes())
+  })
```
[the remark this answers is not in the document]
mvadari: fixed

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -164,30 +164,54 @@
...
+
+/**
+ * Verify the form and type of an Array at runtime.
+ *
+ * @param input - The object to check the form and type of.
+ * @returns Whether the Array is properly formed.
+ */
+export function isArray(input: unknown): boolean {
+  return Array.isArray(input)
+}
```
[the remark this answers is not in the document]
mvadari: fixed, thanks

packages/xrpl/src/models/transactions/batch.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @param tx - A Batch Transaction.
+ * @throws When the Batch is malformed.
+ */
+// eslint-disable-next-line max-lines-per-function -- needed here due to the complexity
+export function validateBatch(tx: Record<string, unknown>): void {
+  validateBaseTransaction(tx)
+
+  validateRequiredField(tx, 'RawTransactions', isArray)
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- checked above
+  const rawTransactions = tx.RawTransactions as unknown[]
```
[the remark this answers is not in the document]
mvadari: Fixed

packages/xrpl/src/models/transactions/batch.ts
```diff
@@ -0,0 +1,149 @@
...
+      fieldObject,
+      'RawTransaction',
+      isObject,
+      `RawTransactions[${index}].RawTransaction`,
+    )
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- needed here
+    // @ts-expect-error -- checked above
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- checked above
+    const rawTx = field.RawTransaction as Record<string, unknown>
+    if (!isObject(rawTx)) {
```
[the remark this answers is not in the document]
mvadari: Fixed

packages/xrpl/src/Wallet/index.ts
```diff
@@ -407,6 +407,10 @@ export class Wallet {
      */
     // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- validate does not accept Transaction type
     validate(tx as unknown as Record<string, unknown>)
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- needed here for JS validation
+    if ((tx as unknown as Record<string, unknown>).BatchTxn != null) {
```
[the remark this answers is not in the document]
mvadari: No longer relevant

packages/xrpl/src/Wallet/batchSigner.ts
```diff
@@ -0,0 +1,214 @@
...
+  }
+  /*
+   * This will throw a more clear error for JS users if the supplied transaction has incorrect formatting
+   */
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- validate does not accept Transaction type
+  validate(transaction as unknown as Record<string, unknown>)
+  const fieldsToSign = {
+    flags: transaction.Flags,
+    txIDs: transaction.TxIDs,
+  }
```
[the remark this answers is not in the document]
mvadari: That's not the norm in this package

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -359,3 +370,82 @@
...
+        'PaymentTransaction: Amount and DeliverMax fields must be identical when both are provided',
+      )
+    }
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- ignore type-assertions on the DeliverMax property
+    // @ts-expect-error -- DeliverMax property exists only at the RPC level, not at the protocol level
+    // eslint-disable-next-line no-param-reassign -- needed here
+    delete tx.DeliverMax
+  }
+}
```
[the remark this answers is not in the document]
mvadari: This code was just moved in this PR and changing it is out of scope

packages/xrpl/src/Wallet/defaultFaucets.ts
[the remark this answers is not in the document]
ckeshava: okay 👍

packages/xrpl/src/utils/hashes/hashLedger.ts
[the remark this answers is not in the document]
ckeshava: alright

packages/xrpl/test/integration/transactions/batch.test.ts
ckeshava: I don't understand three aspects of the fee calculation:
1. The [specification (Section 2.1)](https://github.com/XRPLF/XRPL-Standards/discussions/162) states that the fees of the inner-Batch transactions are incorporated. How does the implementation ensure that? Suppose the inner batch transaction contains AMMCreate, AccountDelete (or) EscrowFinish transactions, should't their fees be computed in a recursive/iterative manner by going through all the inner transactions?

2. Based on my understanding of the spec, the appropriate fees would be:
```
// The definition of baseFee follows from the existing code.
const n = (tx.BatchSigners?.length ?? 0)
fee_accumulator = (n + 2) * baseFee

for await (inner_txn in inner_transactions) {
    fee_accumulator += calculateFeePerTransactionType(client, inner_txn, <I don't know the signerCount value>)
}
```

My understanding is in-line with [this comment of yours](https://github.com/XRPLF/XRPL-Standards/discussions/162#discussioncomment-7800870). Has the spec changed?

3. If there are multiple signers, then the trasaction pays for it in two factors: the length of BatchSigners array and the signersCount. Is that accurate? Is this explained in the spec?

The UX is not good. The `signersCount` could have been auto-filled by the library. There is no need for users to specify that.

image
image

packages/xrpl/src/client/RequestManager.ts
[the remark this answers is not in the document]
mvadari: Not all errors returned from rippled contain an error code, so that doesn't work. Also, why add that in rippled if it's always returned alongside the error code?

packages/xrpl/test/integration/transactions/batch.test.ts
mvadari: > 1. The [specification (Section 2.1)](https://github.com/XRPLF/XRPL-Standards/discussions/162) states that the fees of the inner-Batch transactions are incorporated. How does the implementation ensure that? Suppose the inner batch transaction contains AMMCreate, AccountDelete (or) EscrowFinish transactions, should't their fees be computed in a recursive/iterative manner by going through all the inner transactions?

Good catch - fixed.

> 3. If there are multiple signers, then the trasaction pays for it in two factors: the length of BatchSigners array and the signersCount. Is that accurate? Is this explained in the spec?
> 
> The UX is not good. The `signersCount` could have been auto-filled by the library. There is no need for users to specify that.

`signersCount` cannot be autofilled, because if those are filled in, the transaction has already been signed. `autofill` has to run before the transaction is signed.

packages/xrpl/src/models/transactions/transaction.ts
```diff
@@ -307,6 +309,19 @@ export function validate(transaction: Record<string, unknown>): void {
       validateAccountSet(tx)
       break
 
+    case 'Batch':
+      validateBatch(tx)
+      // This is done here to avoid issues with dependency cycles
```
ckeshava, line 314: what is done here? Which issues of the dependency cycle are you referring to?

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -117,7 +120,15 @@ export function convertTxFlagsToNumber(tx: Transaction): number {
     }, 0)
   }
 
-  return 0
+  return Object.keys(tx.Flags).reduce((resultFlags, flag) => {
```
ckeshava, line 123: If the previous `if` condition is True (`isTxToFlagKey(tx.TransactionType)`), then the Global flags are not parsed. i.e. In cases where transaction-specific flags are specified, this return statement is not executed.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -140,41 +141,46 @@ export function setValidAddresses(tx: Transaction): void {
...
-    if (tx[tagField] && tx[tagField] !== tag) {
+    if (tagField in tx && tx[tagField] !== tag) {
       throw new ValidationError(
         `The ${tagField}, if present, must match the tag of the ${accountField} X-address`,
       )
     }
-    // eslint-disable-next-line no-param-reassign -- param reassign is safe
-    tx[tagField] = tag
+    // eslint-disable-next-line no-param-reassign, @typescript-eslint/consistent-type-assertions -- param reassign is safe
+    ;(tx as Record<K2, number | false>)[tagField] = tag
```
ckeshava, line 166: I'm not familiar with this syntax. Why do you have a statement terminator `;` at the beginning of this line?

packages/xrpl/test/integration/transactions/payment.test.ts
```diff
@@ -74,7 +74,6 @@ describe('Payment', function () {
   it(
     'Validate Payment transaction API v2: Payment Transaction: Specify Only DeliverMax field',
     async () => {
-      // @ts-expect-error -- DeliverMax is a non-protocol, RPC level field in Payment transactions
```
ckeshava, line 77: Thanks for removing these declarations. When I wrote the DeliverMax changes, Typescript transpiler complained about them.

packages/xrpl/test/models/batch.test.ts
```diff
@@ -0,0 +1,163 @@
...
+            TransactionType: 'Payment',
+          },
+        },
+      ],
+      TransactionType: 'Batch',
+    }
+  })
+
+  it('verifies valid Batch', function () {
+    assert.doesNotThrow(() => validateBatch(tx))
```
ckeshava, line 61: the next statement encompasses this one, this is not useful.
```suggestion
```

packages/xrpl/test/models/batch.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        },
+      ],
+      TransactionType: 'Batch',
+    }
+    assert.doesNotThrow(() => validateBatch(tx))
+    assert.doesNotThrow(() => validate(tx))
+  })
+
+  it('throws w/ invalid BatchSigners', function () {
+    tx.BatchSigners = 0
```
ckeshava, line 102: How does the typescript transpiler allow this statement? This statement should fail static type checks.

packages/xrpl/test/testUtils.ts
```diff
@@ -52,6 +57,22 @@ export function assertResultMatch(
...
 
+/**
+ * Check that a transaction error validation fails properly.
+ *
+ * @param tx The transaction that should fail validation.
+ * @param validateTx The transaction-specific validation function (e.g. `validatePayment`).
+ * @param errorMessage The error message that should be included in the error.
+ */
+export function assertTxValidationError(
+  tx: any,
```
ckeshava, line 68: Can you nail down a type for this parameter? I think `Transaction` (or) `SubmitableTransaction` are suitable candidates.

packages/xrpl/test/wallet/batchSigner.test.ts
```diff
@@ -0,0 +1,370 @@
...
+  BatchFlags,
+  BatchInnerTransaction,
+  BatchSigner,
+} from '../../src/models/transactions/batch'
+import {
+  combineBatchSigners,
+  signMultiBatch,
+} from '../../src/Wallet/batchSigner'
+
+const secpWallet = Wallet.fromSeed('spkcsko6Ag3RbCSVXV2FJ8Pd4Zac1', {
```
ckeshava, line 21: Can you use these variables when you construct transaction payloads? Visually, I can't understand the goal of these tests.

packages/xrpl/src/Wallet/batchSigner.ts
```diff
@@ -0,0 +1,222 @@
...
+          flags: tx.Flags,
+          transactionIDs: tx.RawTransactions.map((rawTx) =>
+            hashSignedTx(rawTx.RawTransaction),
+          ),
+        }) !== exampleTransaction,
+    )
+  ) {
+    throw new ValidationError(
+      'Flags and transaction hashes are not the same for all provided transactions.',
+    )
```
ckeshava, line 174: I prefer that you compare the entire transaction payload. I understand that `flags` and `txIDs` must be identical for all equivalent Batch transactions. However, this implementation completely ignores all the other fields.

Do we take a large performance hit if we compare all the fields of the Transaction object? (except `BatchSigner`, of course)

packages/xrpl/src/Wallet/batchSigner.ts
```diff
@@ -0,0 +1,222 @@
...
+    .sort(compareBatchSigners)
+
+  return { ...transactions[0], BatchSigners: sortedSigners }
+}
+
+/**
+ * If presented in binary form, the BatchSigners array must be sorted based on
+ * the numeric value of the signer addresses, with the lowest value first.
+ * (If submitted as JSON, the submit_multisigned method handles this automatically.)
+ * https://xrpl.org/multi-signing.html.
```
ckeshava, line 192: Is this comment relevant for Batch transactions? The listed docs-page does not make a mention of signing Batch transaction inputs. I suspect it only works for usual multi-sign purposes.

packages/xrpl/src/Wallet/batchSigner.ts
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * @param left - A BatchSigner to compare with.
+ * @param right - A second BatchSigner to compare with.
+ * @returns 1 if left \> right, 0 if left = right, -1 if left \< right, and null if left or right are NaN.
+ */
+function compareBatchSigners(left: BatchSigner, right: BatchSigner): number {
+  return addressToBigNumber(left.BatchSigner.Account).comparedTo(
```
ckeshava, line 199: Is this implementation mandated by the spec?

Can we define an `operator<()` in the Account class and use that for comparison?

 `addressToBigNumber` is an expensive operation. I'm especially concerned about explicit typecasts into large `BigNumber` objects.

packages/xrpl/test/integration/transactions/batch.test.ts
ckeshava: > signersCount cannot be autofilled, because if those are filled in, the transaction has already been signed. autofill has to run before the transaction is signed.

Okay, that makes sense.

packages/xrpl/src/client/RequestManager.ts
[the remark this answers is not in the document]
ckeshava: > Not all errors returned from rippled contain an error code, so that doesn't work.

I'm proposing a more comprehensive API error management system inside rippled itself. It is better to standardize the responses of all APIs using numeric error-codes. This results in compact, clear communication to the end user. Presently, it is annoying to go through error_message, error_exception, etc and these fields are haphazardly populated by rippled.

I'm basing my understanding off of HTTP error codes, as implemented by web-server protocols.

ckeshava:
I will go through the rest of the files/comments once the cpp code is finalized and this PR is ready for review.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -210,30 +216,54 @@ export function isXChainBridge(input: unknown): input is XChainBridge {
   )
 }
 
+/**
+ * Verify the form and type of an Array at runtime.
+ *
+ * @param input - The object to check the form and type of.
+ * @returns Whether the Array is properly formed.
+ */
+export function isArray<T = unknown>(input: unknown): input is T[] {
```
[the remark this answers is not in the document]
mvadari: It's a type check function - we don't know the type of the input, this function is what's checking it.

packages/xrpl/src/models/transactions/XChainAddAccountCreateAttestation.ts
```diff
@@ -110,13 +110,13 @@ export function validateXChainAddAccountCreateAttestation(
...
     'WasLockingChainSend',
-    (inp) => inp === 0 || inp === 1,
+    (inp: unknown): inp is 0 | 1 => inp === 0 || inp === 1,
   )
 
   validateRequiredField(
     tx,
     'XChainAccountCreateCount',
-    (inp) => isNumber(inp) || isString(inp),
+    (inp: unknown): inp is number | string => isNumber(inp) || isString(inp),
```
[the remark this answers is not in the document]
mvadari: It's due to the changes to `validateRequiredField` and `validateOptionalField`. I can move those to a separate PR if you'd prefer, but it does really simplify the typing in this PR.

packages/xrpl/src/models/transactions/XChainAddClaimAttestation.ts
```diff
@@ -104,14 +104,14 @@ export function validateXChainAddClaimAttestation(
   validateRequiredField(
     tx,
     'WasLockingChainSend',
-    (inp) => inp === 0 || inp === 1,
+    (inp: unknown): inp is 0 | 1 => inp === 0 || inp === 1,
```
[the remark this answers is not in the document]
mvadari: They're included in the other file - you even have a comment on it.

packages/xrpl/src/models/transactions/batch.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  tx.BatchSigners?.forEach((signerObj, index) => {
+    if (!isRecord(signerObj)) {
+      throw new ValidationError(`Batch: BatchSigners[${index}] is not object.`)
+    }
+
+    const signerRecord = signerObj
+    validateRequiredField(signerRecord, 'BatchSigner', isRecord, {
+      paramName: `BatchSigners[${index}].BatchSigner`,
+      txType: 'Batch',
```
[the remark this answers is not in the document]
mvadari: Why?

packages/xrpl/src/models/transactions/transaction.ts
mvadari: You can't call `validate` from within `validateBatch` on the inner transactions because that would result in a dependency cycle. So this was my way around that.

packages/xrpl/test/integration/transactions/payment.test.ts
mvadari: The changes to `validateRequiredField` and `validateOptionalField` is what resolved them.

packages/xrpl/src/sugar/autofill.ts
mvadari: I don't really understand it either, but it's what's expected.

packages/xrpl/test/testUtils.ts
mvadari: No, because this is purposefully checking error cases - so they're often not valid `Transaction`s.

packages/xrpl/test/wallet/batchSigner.test.ts
mvadari: What do you mean?

packages/xrpl/src/Wallet/batchSigner.ts
mvadari: You're only signing the flags and transaction IDs, so that's all that matters.

packages/xrpl/src/Wallet/batchSigner.ts
mvadari: There is no Account class.

packages/xrpl/test/models/batch.test.ts
mvadari: Note the `let tx: any` at the top

packages/xrpl/src/client/RequestManager.ts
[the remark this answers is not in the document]
mvadari: That isn't something you or I can make happen, and it is most certainly beyond the scope of this PR.

packages/xrpl/src/models/transactions/AMMWithdraw.ts
```diff
@@ -25,7 +25,7 @@ export enum AMMWithdrawFlags {
   tfLimitLPToken = 0x00400000,
 }
```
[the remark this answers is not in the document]
mvadari: enums can't inherit in TypeScript

packages/xrpl/src/models/transactions/batch.ts
```diff
@@ -0,0 +1,161 @@
...
+  validateRequiredField,
+} from './common'
+import type { SubmittableTransaction } from './transaction'
+
+/**
+ * Enum representing values of {@link Batch} transaction flags.
+ *
+ * @category Transaction Flags
+ */
+export enum BatchFlags {
```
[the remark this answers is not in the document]
mvadari: enums can't inherit in TypeScript

packages/xrpl/test/models/batch.test.ts
mvadari: This one checks one with `BatchSigners`

packages/xrpl/src/Wallet/batchSigner.ts
mvadari: Removed

packages/ripple-binary-codec/src/binary.ts
```diff
@@ -177,11 +177,54 @@ function multiSigningData(
...
+
+  const bytesList = new BytesList()
+
+  bytesList.put(prefix)
+  bytesList.put(flags)
+  bytesList.put(txIDsLength)
+  batch.txIDs.forEach((txID: string) => {
+    if (!/^[A-F0-9]{64}$/i.test(txID)) {
+      throw new Error(`Invalid transaction ID format: ${txID}`)
+    }
```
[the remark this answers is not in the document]
mvadari: Fixed

packages/xrpl/src/models/utils/flags.ts
mvadari: Good catch - fixed

packages/xrpl/src/models/transactions/XChainAddAccountCreateAttestation.ts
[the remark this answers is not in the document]
mvadari: #2966

mvadari:
The `develop` Docker container needs to be updated before integration tests will pass. They do pass locally.

packages/ripple-binary-codec/src/types/hash.ts
```diff
@@ -27,6 +27,9 @@ class Hash extends Comparable<Hash | string> {
     }
 
     if (typeof value === 'string') {
+      if (!/^[A-F0-9]*$/i.test(value)) {
```
khancode: Can you define this as a variable and then use it?

packages/ripple-binary-codec/test/signing-data-encoding.test.ts
```diff
@@ -244,4 +245,27 @@ describe('Signing data', function () {
       ].join(''),
     )
   })
+
+  it('can create batch blob', function () {
```
khancode, line 249: Can you create a negative test case for testing `'Need an object to encode a Batch transaction'` error?

packages/xrpl/package.json
```diff
@@ -61,7 +61,7 @@
     "test": "jest --config=jest.config.unit.js --verbose false --silent=false",
     "test:integration": "TS_NODE_PROJECT=tsconfig.build.json jest --config=jest.config.integration.js --verbose false --silent=false --runInBand",
     "test:browser": "npm run build && npm run build:browserTests && karma start ./karma.config.js",
-    "test:watch": "jest --watch --verbose false --silent=false --runInBand ./test/**/*.test.ts --testPathIgnorePatterns=./test/integration --testPathIgnorePatterns=./test/fixtures",
+    "test:watch": "jest --watch --config=jest.config.unit.js --verbose false --silent=false",
```
khancode, line 64: Why did you modify this command?

packages/xrpl/src/Wallet/batchSigner.ts
```diff
@@ -0,0 +1,198 @@
...
+ *                       The actual address is only needed in the case of regular key usage.
+ * @throws ValidationError if the transaction is malformed.
+ */
+export function signMultiBatch(
+  wallet: Wallet,
+  transaction: Batch,
+  opts: { batchAccount?: string; multisign?: boolean | string } = {},
+): void {
+  const batchAccount = opts.batchAccount ?? wallet.classicAddress
+  let multisignAddress: boolean | string = false
```
khancode, line 65: How can this be a `boolean` value?

packages/xrpl/src/Wallet/batchSigner.ts
```diff
@@ -0,0 +1,198 @@
...
+  /*
+   * This will throw a more clear error for JS users if the supplied transaction has incorrect formatting
+   */
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- validate does not accept Transaction type
+  validate(transaction as unknown as Record<string, unknown>)
+
+  const involvedAccounts = transaction.RawTransactions.map(
+    (raw) => raw.RawTransaction.Account,
+  )
+  if (!involvedAccounts.includes(batchAccount)) {
```
khancode: Would be better to initialize a Set and use `involvedAccounts.has(batchAccount)`

packages/xrpl/src/models/transactions/batch.ts
```diff
@@ -0,0 +1,161 @@
...
+        rawTx.Flags |= GlobalFlags.tfInnerBatchTxn
+      }
+    } else if (typeof rawTx.Flags === 'object') {
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- checked above
+      const flags = rawTx.Flags as Record<string, boolean>
+      if (!flags.tfInnerBatchTxn) {
+        // txInnerBatchTxn is either false or null
+        flags.tfInnerBatchTxn = true
+      }
+    }
```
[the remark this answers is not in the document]
khancode: Should this modification logic be inside `autofill` instead? I don't think we should do modifications in `validate` functions.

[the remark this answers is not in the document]
mvadari: IMO `autofill` is a bit too late tbh - either we should enforce those flags in `validate` or add them if they're missing.

packages/xrpl/package.json
mvadari: I think it didn't work when I tried to use it - and either way, it could be drastically simplified given the separate configs for unit and integration tests (I don't think this command was modified when that change was made)

packages/xrpl/src/Wallet/batchSigner.ts
mvadari: It's the same signature as with the multisign helper functions.

packages/xrpl/src/Wallet/batchSigner.ts
mvadari: It's got a max size of 8 so I don't think it matters too much, but I can make the change if you want.

khancode: Oh I see. Since it's a tiny change, might as well do it.

packages/xrpl/src/models/transactions/batch.ts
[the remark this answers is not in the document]
khancode: In that case, I would just throw an error here rather than setting it to 'true'

mvadari:
@khancode I had to make some changes to resolve a dependency cycle, you might want to do a quick re-review

packages/xrpl/src/Wallet/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @returns A Transaction object.
+ * @throws If the input is not a valid Transaction or transaction blob.
+ */
+export function getDecodedTransaction(
+  txOrBlob: Transaction | string,
+): Transaction {
+  if (typeof txOrBlob === 'object') {
+    // We need this to handle X-addresses in multisigning
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- We are casting here to get strong typing
+    return decode(encode(txOrBlob)) as unknown as Transaction
```
ckeshava, line 53: I don't understand why you are doing `decode(encode(txOrBlob))` here. If we receive a `Transaction` object as input, can't you return it directly ? How does this implementation ensure strong typing?

packages/xrpl/test/wallet/batchSigner.test.ts
```diff
@@ -0,0 +1,372 @@
...
+      tx1 = { ...newTx }
+      tx2 = { ...newTx }
+      const tx3 = { ...newTx }
+      signMultiBatch(edWallet, tx1)
+      signMultiBatch(secpWallet, tx2)
+      signMultiBatch(submitWallet, tx3)
+
+      // run test
+      const result = combineBatchSigners([tx1, tx2, tx3])
+      const expected = (tx1.BatchSigners ?? []).concat(tx2.BatchSigners ?? [])
```
ckeshava, line 322: For the purposes of readability, you should have used `edWallet.address`, `secpWallet.address` and `submitWallet.address` in all these tests. In the current state of the implementation, I do not know which addresses (for instance `rJCxK2hX9tDMzbnn3cg1GU2g19Kfmhzxkp`) correspond to which wallets.

Consequently, It is not obvious to me why the `expected` solution is missing the `tx3.BatchSigners` value.

packages/xrpl/test/wallet/batchSigner.test.ts
```diff
@@ -0,0 +1,372 @@
...
+        () => combineBatchSigners([secpWallet.sign(tx1).tx_blob, tx2]),
+        ValidationError,
+        'Batch transaction must be unsigned.',
+      )
+    })
+
+    it('fails with different inner transactions', function () {
+      const badTx2 = { ...tx2 }
+      badTx2.Flags = BatchFlags.tfIndependent
+      signMultiBatch(secpWallet, tx2)
```
ckeshava, line 371: This test case will fail irrespective of whether or not you sign `tx2`. Perhaps, you meant to demonstrate some behavior with `signMultiBatch(secpWallet, badTx2)` ?

```suggestion
```

packages/xrpl/test/wallet/batchSigner.test.ts
```diff
@@ -0,0 +1,372 @@
...
+
+    it('fails with signed inner transaction', function () {
+      assert.throws(
+        () => combineBatchSigners([secpWallet.sign(tx1).tx_blob, tx2]),
+        ValidationError,
+        'Batch transaction must be unsigned.',
+      )
+    })
+
+    it('fails with different inner transactions', function () {
```
ckeshava:
```suggestion
    it('fails with different flags amongst the inner transactions', function () {
```

packages/xrpl/test/wallet/batchSigner.test.ts
```diff
@@ -0,0 +1,372 @@
...
+        },
+      ]
+      assert.property(transaction, 'BatchSigners')
+      assert.strictEqual(
+        JSON.stringify(transaction.BatchSigners),
+        JSON.stringify(expected),
+      )
+    })
+
+    it('succeeds with a different account', function () {
```
ckeshava, line 130: Can you help me understand the below three test cases. This is a regular key for which wallet? How do `regkeyWallet` and `batchAccount` play a role in the `expected` value? I'm not understanding the relationship between these accounts.

This design appears to be slightly different from that of xrpl-py Batch implementation.

packages/xrpl/src/Wallet/utils.ts
mvadari: See the comment above - this is to handle all the possible X-addresses. This code is taken directly from the existing multisigning code and is not new, just moved.

packages/xrpl/test/wallet/batchSigner.test.ts
```diff
@@ -0,0 +1,248 @@
+import { assert } from 'chai'
+
+import {
+  Batch,
+  decode,
+  ECDSA,
+  encode,
```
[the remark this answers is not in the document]
mvadari: Out of scope of this PR

packages/xrpl/test/wallet/batchSigner.test.ts
mvadari: That test is less secure then, because it should also be testing that the account added is correct. This is especially important with the `regkeyWallet`. I added comments with the addresses, though.

The comments in the test address why the `tx3.BatchSigners` doesn't exist - the Batch outer transaction submitter's signature should not be included in `BatchSigners` since it's already essentially included in the normal transaction signature.

mvadari: See https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0056d-batch#213-batchsigners

packages/xrpl/test/wallet/batchSigner.test.ts
mvadari: That's just because the error is called first. It's still testing the proper flow.

packages/xrpl/test/wallet/batchSigner.test.ts
mvadari: Same design, just a different test. This is ensuring that the wallets can properly handle a regular key wallet, where the public key differs from the account specified in the signing info.

packages/xrpl/test/wallet/batchSigner.test.ts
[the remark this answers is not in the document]
mvadari: No

packages/xrpl/test/wallet/batchSigner.test.ts
ckeshava: > That test is less secure then, because it should also be testing that the account added is correct.

What do you mean? How does using variable names like `edWallet.address` instead of the `account_id` like `rJCxK2hX9tDMzbnn3cg1GU2g19Kfmhzxkp ` decrease the security of the tests?

mvadari: An important part of this test is ensuring that the right account address is used. That's much more certain with a string rather than a variable.

mvadari: An important part of this test is ensuring that the right account address is used. That's much more certain with a string rather than a variable.

ckeshava: That goal is easily verifiable with a variable name, rather than a long string of alphanumeric characters.

mvadari: I added comments by the wallets to record their addresses.

mvadari: It's a signing function, I want to make sure all the signing is happening properly. The best way to do that is to hardcode _everything_ in the expected results.

packages/xrpl/src/Wallet/utils.ts
ckeshava: okay, I found the reason. [This line ](https://github.com/XRPLF/xrpl.js/blob/a85a4507aff8acb349397cf70ceb8c13d268d9f5/packages/ripple-binary-codec/src/types/account-id.ts#L42)automatically converts an X-address into a classic-address, if the input fails the regex.

packages/xrpl/src/models/transactions/batch.ts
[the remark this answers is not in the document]
ckeshava: I don't remember the context of this comment now, sorry

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -242,31 +272,50 @@ export function validateRequiredField(
...
+  V,
+>(
+  tx: T,
+  param: K,
+  checkValidity: (inp: unknown) => inp is V,
+  errorOpts: {
+    txType?: string
+    paramName?: string
+  } = {},
+): asserts tx is T & { [P in K]: V | undefined } {
```
ckeshava, line 295: Did you mean to use `T`? What is the `P` type variable?
```suggestion
): asserts tx is T & { [T in K]: V | undefined } {
```

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -210,30 +216,54 @@ export function isXChainBridge(input: unknown): input is XChainBridge {
...
+  V,
+>(
+  tx: T,
+  param: K,
+  checkValidity: (inp: unknown) => inp is V,
+  errorOpts: {
+    txType?: string
+    paramName?: string
+  } = {},
+): asserts tx is T & { [P in K]: V } {
```
ckeshava, line 255: Did you mean to use `T`? What is the `P` type variable?

```suggestion
): asserts tx is T & { [T in K]: V } {
```

packages/xrpl/test/integration/transactions/batch.test.ts
```diff
@@ -0,0 +1,117 @@
...
+              TransactionType: 'Payment',
+              Flags: 0x40000000,
+              Account: wallet2.classicAddress,
+              Destination: destination.classicAddress,
+              Amount: '10000000',
+            },
+          },
+        ],
+      }
+      const autofilled = await testContext.client.autofill(tx, 2)
```
ckeshava: Since `testContext.wallet` is the sender of the outer-batch-transaction, only `wallet2` needs to be included in the `BatchSigners` array.
```suggestion
      const autofilled = await testContext.client.autofill(tx, 1)
```

packages/xrpl/test/models/batch.test.ts
ckeshava: Yes, I understand that `validateBatch` also checks the `BatchSigners` field. 

However, `validate(tx)` invokes the respective `validateBatch(tx)` [here](https://github.com/XRPLF/xrpl.js/blob/a85a4507aff8acb349397cf70ceb8c13d268d9f5/packages/xrpl/src/models/transactions/transaction.ts#L277). So, I don't see the point of explicitly invoking the `validateBatch` method in the unit tests.

packages/xrpl/test/models/utils.test.ts
```diff
@@ -308,9 +344,77 @@ describe('Models Utils', function () {
...
+          tfInnerBatchTxn: true,
+        },
+      }
+
+      const expected = {
+        tfRenew: true,
+      }
+
+      const flagsMap = parseTransactionFlags(tx)
+      assert.notStrictEqual(flagsMap, expected)
```
ckeshava:
```suggestion
      const expected = {
        tfRenew: true,
        tfInnerBatchTxn: true
      }

      const flagsMap = parseTransactionFlags(tx)
      assert.strictEqual(flagsMap, expected)
```
`tfInnerBatchTxn: true` is missing in the `expected` value

packages/xrpl/src/models/transactions/common.ts
mvadari: From ChatGPT (who also helped me write this bit of syntax):

> P is a TypeScript mapped type key — specifically, it’s a type-level alias for the key K. It’s used inside the mapped type syntax:
> 
> ```{ [P in K]: V | undefined }```
> This means:
> 
> > For each key P in the union of keys K, define a property with the type V | undefined.
> 
> But since K is already a single key from T (see K extends keyof T), this mapping is just a long way of saying:
> 
> ```{ [K]: V | undefined }```
> 
> But TypeScript doesn't allow you to use K directly as a key name in an object type — so you use [P in K] instead, even if it's just one key.

packages/xrpl/src/models/transactions/common.ts
mvadari: Same as above

packages/xrpl/test/models/batch.test.ts
mvadari: It's the norm in tests. Better to keep it, to ensure you didn't forget to add that line of code.

packages/xrpl/test/models/utils.test.ts
mvadari: fixed

packages/xrpl/test/integration/transactions/batch.test.ts
mvadari: fixed

packages/xrpl/src/models/transactions/common.ts
ckeshava: okay interesting. I need to learn more about this, not very familiar with it.

mvadari: Shouldn't be a blocker for a merge, since it's an internal-only typing thing

packages/xrpl/test/wallet/batchSigner.test.ts
ckeshava: I forgot the context of this comment, sorry

packages/xrpl/test/wallet/batchSigner.test.ts
ckeshava: I believe the following summary captures the differences well:

```
The key differences are:
- The batchAccount determines which account's transactions we're signing for
- The multisign parameter determines how the signature will be structured (either as a multi-signature or with a specific signer)
- The regkeyWallet is the actual key pair being used to create the signature

This allows for flexible signing scenarios where you might want to:
- Sign transactions on behalf of another account using a regular key
- Create multi-signature transactions that require multiple approvals
- Specify exactly which account should be listed as the signer in the transaction
```

packages/xrpl/test/wallet/batchSigner.test.ts
```diff
@@ -0,0 +1,379 @@
...
+        () => signMultiBatch(otherWallet, transaction),
+        ValidationError,
+        'Must be signing for an address submitting a transaction in the Batch.',
+      )
+    })
+
+    it('fails with non-Batch transaction', function () {
+      assert.throws(
+        // @ts-expect-error - needed for JS/codecov
+        () => signMultiBatch(otherWallet, nonBatchTx),
```
ckeshava: This ensures that only the `transaction` (second function) parameter causes an error in the validation of the transaction.
```suggestion
        () => signMultiBatch(edWallet, nonBatchTx),
```

ckeshava (approved):
I have one last suggestion here: https://github.com/XRPLF/xrpl.js/pull/2801#discussion_r2132719629

PR #2812, opened 2024-10-25
Adjustments for fee voting

pdp2121:

Ensure the library works for changes FeeSettings values resulted by [Fee Voting](https://xrpl.org/docs/concepts/consensus-protocol/fee-voting):
- Remove hard-coded reference to 10 drops as the reference transaction cost.
- Ensure tests passed for all transaction fee scenarios.
- Modify `calculateFeePerTransaction` to get accurate tx fee for `AMMCreate` in case `owner_reserve` increases.

### Context of Change

Fee voting by mainnet UNL validators are approaching 200 drops for `base_fee`. Both `account_reserve` and `reserve_inc` are also getting proposed changes.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

All integration tests passed.


ckniffen (approved):

mvadari:
nit: this PR name should be changed - it has nothing to do with the `FeeSettings` object.

packages/xrpl/test/integration/transactions/payment.test.ts
```diff
@@ -25,14 +30,17 @@ describe('Payment', function () {
...
       Destination: 'rfkE1aSy9G8Upk4JssnwBxhEv5p4mn2KTy',
     }
   })
 
   beforeAll(async () => {
     testContext = await setupClient(serverUrl)
     senderWallet = await generateFundedWallet(testContext.client)
+    // Make sure the amount sent satisfies minimum reserve requirement to fund an account.
+    amount =
+      (await fetchAccountReserveFee(testContext.client)) ?? DEFAULT_AMOUNT
```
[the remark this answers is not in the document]
ckeshava: I agree with this suggestion. Moreover, how can you choose an appropriate `DEFAULT_AMOUNT` ?
Since the XRPL mainnet can vote for arbitrary fee values, I feel we should not advocate for any "Default Fees". If we are unable to retrieve the fees from the latest validated ledger, we should throw an error

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -230,13 +230,13 @@ export async function setNextValidSequenceNumber(
 }
 
 /**
- * Fetches the account deletion fee from the server state using the provided client.
+ * Fetches the owner reserve fee from the server state using the provided client.
  *
  * @param client - The client object used to make the request.
  * @returns A Promise that resolves to the account deletion fee as a BigNumber.
```
ckeshava, line 236: Please update the references to `Account Deletion` into Owner reserve in these comments

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -260,15 +260,14 @@ export async function calculateFeePerTransactionType(
...
   signersCount = 0,
 ): Promise<void> {
-  // netFee is usually 0.00001 XRP (10 drops)
   const netFeeXRP = await getFeeXrp(client)
   const netFeeDrops = xrpToDrops(netFeeXRP)
   let baseFee = new BigNumber(netFeeDrops)
 
   // EscrowFinish Transaction with Fulfillment
   if (tx.TransactionType === 'EscrowFinish' && tx.Fulfillment != null) {
     const fulfillmentBytesSize: number = Math.ceil(tx.Fulfillment.length / 2)
```
ckeshava, line 269: Do you know why we are dividing the `tx.Fulfillment.length` by `2` ? I know that you didn't write this code, but I didn't understand it.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -280,22 +279,21 @@ export async function calculateFeePerTransactionType(
...
   if (signersCount > 0) {
     baseFee = BigNumber.sum(baseFee, scaleValue(netFeeDrops, 1 + signersCount))
   }
 
   const maxFeeDrops = xrpToDrops(client.maxFeeXRP)
-  const totalFee =
-    tx.TransactionType === 'AccountDelete'
-      ? baseFee
-      : BigNumber.min(baseFee, maxFeeDrops)
+  const totalFee = ['AccountDelete', 'AMMCreate'].includes(tx.TransactionType)
```
ckeshava, line 294: What is the need for this comparison with `maxFeeDrops` ( == `2,000,000` drops) ?

the `baseFee` variable computes the final fees, accounting for the cases where multiple-signers are provided. 

Why are performing this arbitrary comparison with `2 XRP` ? Moreover, we are not using any `cushion` value in the `getFeeXrp()` function call, hence there is no possibility of the library over-estimating the fees.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -280,22 +279,21 @@ export async function calculateFeePerTransactionType(
     tx.TransactionType === 'AccountDelete' ||
```
ckeshava, line 279: The rounding operation a few lines above (on line 275), `baseFee = product.dp(0, BigNumber.ROUND_CEIL)` is not necessary.

We should perform rounding operation after completing all the calculations, otherwise we risk losing precision of values.

Since we are invoking `totalFee.dp(0)` eventually, this rounding operation can be removed.

Here is the docs: https://mikemcl.github.io/bignumber.js/#dp

packages/xrpl/test/integration/requests/fee.test.ts
```diff
@@ -26,17 +28,21 @@ describe('fee', function () {
...
         result: {
           current_ledger_size: '0',
           current_queue_size: '0',
           drops: {
-            base_fee: '10',
-            median_fee: '5000',
-            minimum_fee: '10',
-            open_ledger_fee: '10',
+            base_fee: referenceFee,
+            median_fee: new BigNumber(referenceFee)
```
ckeshava, line 41: We can omit the `median_fee` field from the deep-equality comparison. Quoting from the docs: `An approximation of the median transaction cost among transactions included in the previous validated ledger, represented in drops of XRP.`

Since this value depends on the transactions-mix of the previous validated ledger, it is not informative for integration test status. Docs: https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/server-info-methods/fee#response-format

packages/xrpl/test/integration/transactions/payment.test.ts
```diff
@@ -25,14 +30,17 @@ describe('Payment', function () {
...
+      Amount: amount,
       Destination: 'rfkE1aSy9G8Upk4JssnwBxhEv5p4mn2KTy',
     }
   })
 
   beforeAll(async () => {
     testContext = await setupClient(serverUrl)
     senderWallet = await generateFundedWallet(testContext.client)
+    // Make sure the amount sent satisfies minimum reserve requirement to fund an account.
+    amount =
```
ckeshava, line 42: Hello,
The intention in these tests is not to fund an account into existence. Hence, the Amount field could be any positive value, it does not have to be above the reserve requirement.

It looks like the `amount` value is always equal to `DEFAULT_AMOUNT`. So, I don't see a need for another `amount` variable.

Are changes in this file required? I don't this file is impacted by the fee-change-calculation updates

packages/xrpl/test/integration/utils.ts
```diff
@@ -444,3 +445,16 @@ export async function createAMMPool(client: Client): Promise<{
     asset2,
   }
 }
+
+export async function fetchAccountReserveFee(
```
ckeshava, line 449: Instead of creating a new function, can we re-use the one from `packages/xrpl/src/sugar/autofill.ts` ? Its titled `fetchOwnerReserve`

This will improve the test-coverage of the function which is already in the main codebase.

pdp2121: Hi Chenna, OwnerReserve and AccountReserve refers to 2 different fees. One would be `reserve_inc` and the other `reserve_base`.
Please refer to: https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/feesettings

packages/xrpl/src/sugar/autofill.ts
pdp2121: This part confuses me a bit as well since baseFee computed the final fee with multi-signers accounted so not sure why we need this comparison.
@khancode could you explain why since you wrote this code a while ago?
Btw, cushion is defaulted to 1.2 if not set.

packages/xrpl/src/sugar/autofill.ts
pdp2121: This is from the docs:

> The minimum [transaction cost](https://xrpl.org/docs/concepts/transactions/transaction-cost/) to submit an EscrowFinish transaction increases if it contains a fulfillment. If the transaction has no fulfillment, the transaction cost is the standard 10 drops. If the transaction contains a fulfillment, the transaction cost is 330 [drops of XRP](https://xrpl.org/docs/references/protocol/data-types/basic-data-types/#specifying-currency-amounts) plus another 10 drops for every 16 bytes in size of the preimage.

The division by 2 is due to the fact that each byte is represented by 2 characters in hex

packages/xrpl/test/integration/transactions/payment.test.ts
pdp2121: It is because the first payment transaction in this test `Validate Payment transaction API v2: Payment Transaction: Specify Only Amount field` will create the hardcoded account `rfkE1aSy9G8Upk4JssnwBxhEv5p4mn2KTy`, which does not exist beforehand. So if we increase `account_reserve = 20000000` instead of 10XRP then all payment transactions in these tests will fail:
`xrpl:     Message:
xrpl:       Destination does not exist. Too little XRP sent to create it.: expected 'tecNO_DST_INSUF_XRP' to equal 'tesSUCCESS'`

Hence we either have to make this first payment to be at least the account_reserve or setup a destination account before sending payments

packages/xrpl/src/sugar/autofill.ts
ckeshava: ok thanks

packages/xrpl/test/integration/transactions/payment.test.ts
ckeshava: ah okay, thanks for the clarification

packages/xrpl/test/integration/utils.ts
ckeshava: okay, thanks for the clarification. Personally, I'd have preferred the names `reserve_base` and `reserve_increment` because its easier to understand their intent. But its not a strong opinion.

ckeshava (approved):
Except for the removal of the `baseFee` comparison, this PR looks good

ckeshava (changes requested):
@pdp2121  I'm getting the following error message on many integration tests. I suspect the tests need to be updated to use a generic fee value. Did you observe this in your local run?

```
AMMWithdraw › withdraw with Amount and LPTokenIn
xrpl:     assert.strictEqual(received, expected)
xrpl:     Expected value to strictly be equal to:
xrpl:       "tesSUCCESS"
xrpl:     Received:
xrpl:       "telINSUF_FEE_P"
xrpl:     Message:
xrpl:       Fee insufficient.: expected 'telINSUF_FEE_P' to equal 'tesSUCCESS'
```

ckeshava:
For additional context, I'm using the fee-voting parameters as specified in this PR's `rippled.cfg` fee-voting section.

pdp2121:
@ckeshava all tests passed for me locally.
`xrpl: Test Suites: 84 passed, 84 total
xrpl: Tests:       136 passed, 136 total`

ckeshava (approved):
The CI is happy with all the tests, that should suffice 👍

Issue #2823, opened 2024-11-11
parseTransactionFlags unintentionally mutates the passed in Transaction object


achowdhry-ripple:
https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/models/utils/flags.ts#L100
This line of code mutates the provided Transaction object, which is an unexpected side effect from this function.

Use convertFlagsToNumber directly to eliminate this issue. 

Also remove all usages of setTransactionFlagsToNumber with convertFlagsToNumber and deprecate the method.

PR #2825, opened 2024-11-13
Fix: parseTransactionFlags unintentionally modifies transaction

achowdhry-ripple:

the parseTransactionFlags method unintentionally edited the Transaction object's flags when passed in as a parameter, because it used setTransactionFlagsToNumber inside. Instead, now it uses convertTxFlagsToNumber, which passes the numerical representation of flags back to the caller instead of automatically modifying the passed in object. This is clearer for a user, and avoids unexpected modification and reduces redundant need for both functions. 

As cleanup, all uses of the setter were removed and moving forward we will be using the converter instead. The setter remains as a deprecated function for compatibility support since it is an exported function.

convertFlagsToNumber flags has been renamed to convertTxFlagsToNumber for clarity

### Context of Change

https://github.com/XRPLF/xrpl.js/issues/2823

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes


packages/xrpl/src/models/utils/flags.ts
```diff
@@ -58,36 +56,33 @@ const txToFlag = {
...
 
-  tx.Flags = txToFlag[tx.TransactionType]
-    ? convertFlagsToNumber(tx.Flags, txToFlag[tx.TransactionType])
-    : 0
-}
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- safe member access
+  const flagEnum = txToFlag[tx.TransactionType]
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any -- added ValidationError check for flagEnum
-function convertFlagsToNumber(flags: GlobalFlags, flagEnum: any): number {
```
mvadari, line 80: Why did `convertFlagsToNumber` need to change?

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -60,34 +58,50 @@ const txToFlag = {
...
+}
+
+/**
+ * Returns a transaction's flags as its numeric representation.
+ *
+ * @param tx - A transaction to parse flags for
+ * @returns A numerical representation of a transaction's flags
+ */
+export function convertTxFlagsToNumber(tx: Transaction): number {
   if (typeof tx.Flags === 'number') {
```
mvadari, line 84: missing a `null` check

packages/xrpl/test/models/utils.test.ts
```diff
@@ -71,8 +71,8 @@ describe('Models Utils', function () {
       const { tfPassive, tfFillOrKill } = OfferCreateFlags
       const expected: number = tfPassive | tfFillOrKill
 
-      setTransactionFlagsToNumber(tx)
-      assert.strictEqual(tx.Flags, expected)
+      const flagNum = convertTxFlagsToNumber(tx)
```
mvadari, line 74: I would leave the `setTransactionFlagsToNumber` in as well, until it gets removed. Just to ensure there's no other breaking changes.

packages/xrpl/HISTORY.md
```diff
@@ -3,6 +3,9 @@
 Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xrpl-announce) for release announcements. We recommend that xrpl.js (ripple-lib) users stay up-to-date with the latest stable release.
 
 ## Unreleased Changes
+* Fix for parseTransactionFlags to no longer modify a passed in transaction when returning a parsed mapping
```
mvadari, line 6: This should be under a heading that matches the actions - see https://keepachangelog.com/en/1.1.0/

packages/xrpl/src/models/utils/flags.ts
achowdhry-ripple: There is a check for nulls on line 91 before the conversion logic, do you mean a null check for the tx param as a whole?

packages/xrpl/test/models/utils.test.ts
achowdhry-ripple: Issue here is that the tests now fail because `setTransactionFlagsToNumber` was marked as deprecated. Is there a way to override this?

ckeshava: @mvadari Why should we retain `setTransactionFlagsToNumber` inside a test file? How will that cause a breaking change?

packages/xrpl/src/models/index.ts
```diff
@@ -8,8 +8,9 @@
  */
 export * as LedgerEntry from './ledger'
 export {
-  setTransactionFlagsToNumber,
   parseAccountRootFlags,
+  setTransactionFlagsToNumber,
```
ckeshava, line 12: Is it possible to remove this import of `setTransactionFlagsToNumber` ? That would ensure that `packages/xrpl/src/models/*` will always use the newer pure function.

packages/xrpl/src/models/utils/flags.ts
ckeshava: I believe she's referring to the updated behavior of the `setTransactionFlagsToNumber` function.

In the existing code, if `tx.Flags != null`, then the `tx` object is not updated at all.
However, in your PR, the `tx.Flags` will be over-written with the output of the `convertTxFlagsToNumber`

packages/xrpl/src/models/index.ts
achowdhry-ripple: This was intentionally left in since removing it would create a breaking change. To prevent this, I've added deprecated annotations that cause linting errors and a console warning to the method, and maybe it would be good to remove two releases from now (one so the community can see the deprecation warning and react, another to actually remove)

packages/xrpl/src/models/utils/flags.ts
achowdhry-ripple: I was under the impression that the numerical flag value 0 is equivalent to tx.Flags = null, which is what would be the result of the current code changes. Is representing a null tx.Flags value as null numerically what we would want instead? In that case I'll add an error as well since I don't think we would ever want that

khancode:
Looks like you'll need to ignore deprecation warnings like this:
```
xrpl: /home/runner/work/xrpl.js/xrpl.js/packages/xrpl/test/models/utils.test.ts
xrpl:    18:3  warning  Deprecated: This utility function is deprecated.
xrpl: Use convertTxFlagsToNumber() instead and use the returned value to modify the tx.Flags from the caller  import/no-deprecated``
```

packages/xrpl/test/models/utils.test.ts
achowdhry-ripple: Figured out the test failure issue, I agree with @mvadari here. Since its still in use and is modified, we would want to make sure the expected behavior didn't changed (even though its deprecated now)

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -58,36 +56,33 @@ const txToFlag = {
...
-    return
+    return tx.Flags
   }
 
-  tx.Flags = txToFlag[tx.TransactionType]
-    ? convertFlagsToNumber(tx.Flags, txToFlag[tx.TransactionType])
-    : 0
-}
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment -- safe member access
+  const flagEnum = txToFlag[tx.TransactionType]
```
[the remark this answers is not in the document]
justinr1234: @achowdhry-ripple Agree with this

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -60,34 +58,52 @@ const txToFlag = {
...
+}
+
+/**
+ * Returns a transaction's flags as its numeric representation.
+ *
+ * @param tx - A transaction to parse flags for
+ * @returns A numerical representation of a transaction's flags
+ */
+export function convertTxFlagsToNumber(tx: Transaction): number {
   if (typeof tx.Flags === 'number') {
```
ckeshava, line 86:
```suggestion
  if (tx.Flags != null and typeof tx.Flags === 'number') {
```

packages/xrpl/src/models/utils/flags.ts
ckeshava: Shouldn't the last assert statement pass in this example?
```
tx.Flags = null
setTransactionFlagsToNumber(tx)
assert(tx.Flags === 0) // false
```

packages/xrpl/src/models/utils/flags.ts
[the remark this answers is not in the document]
ckeshava: @achowdhry-ripple  what do you think of these suggestions?

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -60,34 +58,52 @@ const txToFlag = {
...
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- safe member access
-    if (flagEnum[flag] == null) {
-      throw new ValidationError(
-        `flag ${flag} doesn't exist in flagEnum: ${JSON.stringify(flagEnum)}`,
-      )
-    }
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- safe member access
-    return flags[flag] ? resultFlags | flagEnum[flag] : resultFlags
-  }, 0)
+  if (flagEnum && tx.Flags) {
```
ckeshava, line 93: can you extract the case of `tx.Flags == null` as a pre-error-check in this method? I'm finding it difficult to parse the meaning of this code change.

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -97,22 +113,24 @@ function convertFlagsToNumber(flags: GlobalFlags, flagEnum: any): number {
  * @returns A map with all flags as booleans.
  */
 export function parseTransactionFlags(tx: Transaction): object {
-  setTransactionFlagsToNumber(tx)
-  if (typeof tx.Flags !== 'number' || !tx.Flags || tx.Flags === 0) {
+  const flags = convertTxFlagsToNumber(tx)
+  if (flags === 0) {
```
ckeshava, line 117: there is value in retaining the old checks. `convertTxFlagsToNumber` method does not retain the exact order of these checks in its current implementation.

packages/xrpl/test/models/utils.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-deprecated -- using deprecated setTransactionFlagsToNumbers to ensure no breaking changes */
 /* eslint-disable no-bitwise -- flags require bitwise operations */
```
ckeshava, line 2: can you add unit tests for `convertTxFlagsToNumber` method, at least along similar lines as line 51 in this file? Have you added the tests elsewhere?

packages/xrpl/HISTORY.md
```diff
@@ -2,16 +2,20 @@
 
 Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xrpl-announce) for release announcements. We recommend that xrpl.js (ripple-lib) users stay up-to-date with the latest stable release.
 
-## Unreleased Changes
+## Unreleased
```
ckeshava, line 5:
```suggestion
## Unreleased Changes
```

Can we retain the words here? I feel the existing wording is better.

ckeshava (changes requested):

packages/xrpl/HISTORY.md
achowdhry-ripple: Every other package uses "Unreleased", and thats whats consistent with https://keepachangelog.com/en/1.1.0/ so thats why I made the change. If you feel strongly, I can keep it but felt inconsistent

ckeshava: Okay, I wasn't aware of this industry practice.

packages/xrpl/src/models/utils/flags.ts
achowdhry-ripple: Not sure if this is necessary, wouldn't typeof null return 'object', which would fail this condition anyways?

ckeshava: ah, I wasn't aware of this JS behavior 👍

packages/xrpl/src/models/utils/flags.ts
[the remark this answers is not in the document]
achowdhry-ripple: this exact way doesn't work, but I added a type guard instead so should be cleaner with fewer lint overrides now

packages/xrpl/src/models/utils/flags.ts
achowdhry-ripple: Yes, and it still should since line 98 `if (tx.flags) ` wouldn't be true, so `convertTxFlagsToNumber` would return 0. In the old setter logic , if tx.Flags was null, it returned 0 as well so the behavior is the same, just without an explicit null check and early return

packages/xrpl/src/models/utils/flags.ts
achowdhry-ripple: All of the same checks still exist in the current logic, just in a different representation and in the helper method. I have extracted the null check to be earlier now so that it is clearer that it is the same

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -57,37 +55,61 @@ const txToFlag = {
...
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- safe member access
-    if (flagEnum[flag] == null) {
-      throw new ValidationError(
-        `flag ${flag} doesn't exist in flagEnum: ${JSON.stringify(flagEnum)}`,
-      )
-    }
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- safe member access
-    return flags[flag] ? resultFlags | flagEnum[flag] : resultFlags
-  }, 0)
+      return tx.Flags?.[flag] ? resultFlags | flagEnum[flag] : resultFlags
```
ckeshava, line 108: At this point, the existence of `tx.Flags` is guaranteed. The `return tx.Flags?` operator is not needed.

ckeshava (approved):

ckeshava:
@achowdhry-ripple  this looks good. I'd like to see the CI/CD test results. They will be executed once the merge conflicts are resolved.

packages/xrpl/HISTORY.md
```diff
@@ -2,16 +2,20 @@
...
 
 ### Added
-* parseTransactionFlags as a utility function in the xrpl package to streamline transactions flags-to-map conversion
+* `parseTransactionFlags` as a utility function in the xrpl package to streamline transactions flags-to-map conversion
```
ckeshava, line 8: What is the need to mention `parseTransactionFlags` method under the `Added` section ? This method was present in the codebase. Its internal implementation has been updated.

packages/xrpl/src/models/utils/flags.ts
achowdhry-ripple: removed the outdated logic

packages/xrpl/HISTORY.md
achowdhry-ripple: fair enough, removed that line

packages/xrpl/src/models/utils/flags.ts
achowdhry-ripple: Actually, this is how the typing works since we do an early return, instead of checking for it later inside of the loop. Will leave it like this to avoid restructuring the logic

packages/xrpl/HISTORY.md
```diff
@@ -2,7 +2,13 @@
 
 Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xrpl-announce) for release announcements. We recommend that xrpl.js (ripple-lib) users stay up-to-date with the latest stable release.
 
-## Unreleased Changes
+## Unreleased
+
+### Added
+* `convertTxFlagsToNumber` as a utility function in xrpl to return the number conversion of a transaction's flags
```
khancode, line 8: Not necessary to explain the the function since the name implies its functionality. One can look at the JSDoc for explanation.

```suggestion
* Adds utility function `convertTxFlagsToNumber`
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -63,37 +61,61 @@ const txToFlag = {
   XChainModifyBridge: XChainModifyBridgeFlags,
 }
 
+function istxToFlagKey(
```
khancode, line 64:
```suggestion
function isTxToFlagKey(
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -63,37 +61,61 @@ const txToFlag = {
...
+): transactionType is keyof typeof txToFlag {
+  return transactionType in txToFlag
+}
+
 /**
  * Sets a transaction's flags to its numeric representation.
  *
+ * @deprecated
+ * This utility function is deprecated.
+ * Use convertTxFlagsToNumber() instead and use the returned value to modify the tx.Flags from the caller.
```
khancode, line 75: Better to reference `Transaction` class name than variable name.

```suggestion
 * Use convertTxFlagsToNumber() instead and use the returned value to modify the Transaction.Flags from the caller.
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -63,37 +61,61 @@ const txToFlag = {
...
+ *
  * @param tx - A transaction to set its flags to its numeric representation.
  */
 export function setTransactionFlagsToNumber(tx: Transaction): void {
-  if (tx.Flags == null) {
-    tx.Flags = 0
-    return
+  // eslint-disable-next-line no-console -- intended deprecation warning
+  console.warn(
+    'This function is deprecated. Use convertTxFlagsToNumber() instead and use the returned value to modify the tx.Flags from the caller.',
```
khancode, line 82:
```suggestion
    'This function is deprecated. Use convertTxFlagsToNumber() instead and use the returned value to modify the Transaction.Flags from the caller.',
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -63,37 +61,61 @@ const txToFlag = {
...
+  )
+
+  if (tx.Flags) {
+    // eslint-disable-next-line no-param-reassign -- intended param reassign in setter, retain old functionality for compatibility
+    tx.Flags = convertTxFlagsToNumber(tx)
+  }
+}
+
+/**
+ * Returns a transaction's flags as its numeric representation.
```
khancode, line 92: I think it looks better to capitalize class and field names.

```suggestion
 * Returns a Transaction's Flags as its numeric representation.
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -63,37 +61,61 @@ const txToFlag = {
...
+ *
+ * @param tx - A transaction to parse flags for
```
khancode, line 94: ditto

```suggestion
 * @param tx - A Transaction to parse Flags for
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -63,37 +61,61 @@ const txToFlag = {
...
+ * @returns A numerical representation of a transaction's flags
```
khancode, line 95: ditto

```suggestion
 * @returns A numerical representation of a Transaction's Flags
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -63,37 +61,61 @@ const txToFlag = {
...
   if (typeof tx.Flags === 'number') {
-    return
+    return tx.Flags
   }
 
-  tx.Flags = txToFlag[tx.TransactionType]
-    ? convertFlagsToNumber(tx.Flags, txToFlag[tx.TransactionType])
-    : 0
-}
+  if (istxToFlagKey(tx.TransactionType)) {
```
khancode, line 105:
```suggestion
  if (isTxToFlagKey(tx.TransactionType)) {
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -63,37 +61,61 @@ const txToFlag = {
...
+    const flagEnum = txToFlag[tx.TransactionType]
+    return Object.keys(tx.Flags).reduce((resultFlags, flag) => {
+      if (flagEnum[flag] == null) {
+        throw new ValidationError(
+          `flag ${flag} doesn't exist in flagEnum: ${JSON.stringify(flagEnum)}`,
```
khancode, line 110: Better to not reference `flagEnum` variable within the function. Users who get this exception won't understand the variable unless they inspect the function.

```suggestion
          Invalid flag ${flag}. Valid flags are ${JSON.stringify(flagEnum)}`,
```

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -103,22 +125,24 @@ function convertFlagsToNumber(flags: GlobalFlags, flagEnum: any): number {
...
-    }
-  })
+  if (istxToFlagKey(tx.TransactionType)) {
+    const transactionTypeFlags = txToFlag[tx.TransactionType]
+    Object.values(transactionTypeFlags).forEach((flag) => {
+      if (
+        typeof flag === 'string' &&
+        isFlagEnabled(flags, transactionTypeFlags[flag])
+      ) {
+        booleanFlagMap[flag] = true
```
khancode, line 142:
```suggestion
        booleanFlagMap[flag] = typeof flag === 'string' && isFlagEnabled(flags, transactionTypeFlags[flag])
```

packages/xrpl/test/models/utils.test.ts
```diff
@@ -46,6 +48,65 @@ describe('Models Utils', function () {
...
+      assert.isUndefined(parsed.lsfGlobalFreeze)
+      assert.isUndefined(parsed.lsfNoFreeze)
+      assert.isUndefined(parsed.lsfPasswordSpent)
+      assert.isUndefined(parsed.lsfRequireAuth)
+      assert.isUndefined(parsed.lsfRequireDestTag)
+      assert.isUndefined(parsed.lsfDisallowIncomingNFTokenOffer)
+      assert.isUndefined(parsed.lsfDisallowIncomingCheck)
+      assert.isUndefined(parsed.lsfDisallowIncomingPayChan)
+      assert.isUndefined(parsed.lsfDisallowIncomingTrustline)
+      assert.isUndefined(parsed.lsfAllowTrustLineClawback)
```
khancode, line 106: This test is misleading because its title implies set to `false` but really they're `undefined`. `undefined != false` in JS.

khancode (changes requested):

packages/xrpl/src/models/utils/flags.ts
khancode: Oh actually, this will set all flags to either true/false. May not want to do this simplification since we only want to set enabled flags.

packages/xrpl/test/models/utils.test.ts
khancode: Also, it's worth creating another test when you explicitly set flags to `false` to cover all cases.

packages/xrpl/src/models/utils/flags.ts
achowdhry-ripple: Yeah this will change the expected outcome and include false values, so will omit this change

packages/xrpl/test/models/utils.test.ts
achowdhry-ripple: Since we are parsing "0" as the numerical form, we are checking an all-false flagset in this test. Also because the definition of the flag utility function returns only true flags, an empty object would by definition be equivalent to all false flags, without it being explicitly stated per each flag. So this logic is also equivalent to checking for an empty flag set object returned, but this seems clearer to show all the expected flags cannot be found

khancode: I see. I do agree in the sense being undefined is also considered to be false but in the JS context these are inequivalent. And I would also consider passing `0` as not setting any flags. Therefore, I still do think we should rename the test to `"all flags not set"`

achowdhry-ripple: sure can rename it

packages/xrpl/test/models/utils.test.ts
```diff
@@ -46,6 +48,65 @@ describe('Models Utils', function () {
...
+          parsed.lsfRequireDestTag &&
+          parsed.lsfDisallowIncomingNFTokenOffer &&
+          parsed.lsfDisallowIncomingCheck &&
+          parsed.lsfDisallowIncomingPayChan &&
+          parsed.lsfDisallowIncomingTrustline &&
+          parsed.lsfAllowTrustLineClawback,
+      )
+    })
+
+    it('all flags not set', function () {
```
khancode, line 90: My bad, this one is a better title :)

```suggestion
    it('no flags set', function () {
```

khancode (approved):
Left a small thing to change but LGTM. Great work!

PR #2856, opened 2024-12-18
add more mpt flag validations

shawnxie999:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

### DIFF
--- a/packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
+++ b/packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
@@ -1,5 +1,6 @@
 import { ValidationError } from '../../errors'
 import { isHex, INTEGER_SANITY_CHECK } from '../utils'
+import { isFlagEnabled } from '../utils'
 
 import {
   BaseTransaction,
@@ -157,5 +158,16 @@ export function validateMPTokenIssuanceCreate(
         'MPTokenIssuanceCreate: TransferFee out of range',
       )
     }
+  }
+
+  if (
+    typeof tx.Flags === 'number' &&
+    typeof tx.TransferFee === 'number' &&
+    !isFlagEnabled(tx.Flags, MPTokenIssuanceCreateFlags.tfMPTCanTransfer) &&
+    tx.TransferFee != 0
+  ) {
+    throw new ValidationError(
+      'MPTokenIssuanceCreate: TransferFee cannot be provided without enabling tfMPTCanTransfer flag',
+    )
   }
 }
--- a/packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
+++ b/packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
@@ -12,6 +12,7 @@ import type { TransactionMetadataBase } from './metadata'
 
 // 2^63 - 1
 const MAX_AMT = '9223372036854775807'
+const MAX_TRANSFER_FEE = 50000
 
 /**
  * Transaction Flags for an MPTokenIssuanceCreate Transaction.
@@ -159,11 +160,11 @@ export function validateMPTokenIssuanceCreate(
       ? isFlagEnabled(flags, MPTokenIssuanceCreateFlags.tfMPTCanTransfer)
       : flags.tfMPTCanTransfer ?? false
 
-  const MAX_TRANSFER_FEE = 50000
   if (typeof tx.TransferFee === 'number') {
     if (tx.TransferFee < 0 || tx.TransferFee > MAX_TRANSFER_FEE) {
       throw new ValidationError(
-        'MPTokenIssuanceCreate: TransferFee out of range',
+        'MPTokenIssuanceCreate: TransferFee must be between 0 and ' +
+          MAX_TRANSFER_FEE,
       )
     }
 
--- a/packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
+++ b/packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
@@ -115,7 +115,6 @@ export interface MPTokenIssuanceCreateMetadata extends TransactionMetadataBase {
 }
 
 /* eslint-disable max-lines-per-function -- Not needed to reduce function */
-
 /**
  * Verify the form and type of an MPTokenIssuanceCreate at runtime.
  *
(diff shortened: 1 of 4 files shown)


mvadari (commented):
Don't forget to update the changelog

shawnxie999:
@mvadari don't think that's necessary since the library hasn't released MPT yet and it is just an extra validation

khancode (approved):
LGTM

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -150,12 +152,26 @@ export function validateMPTokenIssuanceCreate(
...
   }
 
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Not necessary
+  const flags = tx.Flags as number | MPTokenIssuanceCreateFlagsInterface
+  const isTfMPTCanTransfer =
+    typeof flags === 'number'
+      ? isFlagEnabled(flags, MPTokenIssuanceCreateFlags.tfMPTCanTransfer)
+      : flags.tfMPTCanTransfer ?? false
+
   const MAX_TRANSFER_FEE = 50000
```
khancode, line 162: Can you move `MAX_TRANSFER_FEE` to top of this module?

khancode (changes requested):

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -150,12 +152,26 @@ export function validateMPTokenIssuanceCreate(
...
   if (typeof tx.TransferFee === 'number') {
```
khancode, line 163: Should add a ValidationError when `TransferFee` isn't a `number`.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -150,12 +152,26 @@ export function validateMPTokenIssuanceCreate(
...
     if (tx.TransferFee < 0 || tx.TransferFee > MAX_TRANSFER_FEE) {
       throw new ValidationError(
         'MPTokenIssuanceCreate: TransferFee out of range',
```
khancode, line 166: Specify the range in the error message.
```suggestion
        'MPTokenIssuanceCreate: TransferFee must be between 0 and ' + MAX_TRANSFER_FEE,
```

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
shawnxie999: don't think that's necessary since `TransferFee` only supports `number` type

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -112,6 +113,8 @@ export interface MPTokenIssuanceCreateMetadata extends TransactionMetadataBase {
   mpt_issuance_id?: string
 }
 
+/* eslint-disable max-lines-per-function -- Not needed to reduce function */
+
```
khancode, line 117: Is this applied to the entire module or function below?

shawnxie999: just the function below

khancode: Ok, move this below the docstring and above the function signature.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
shawnxie999: updated

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -150,12 +155,25 @@ export function validateMPTokenIssuanceCreate(
...
 
-  const MAX_TRANSFER_FEE = 50000
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Not necessary
+  const flags = tx.Flags as number | MPTokenIssuanceCreateFlagsInterface
+  const isTfMPTCanTransfer =
+    typeof flags === 'number'
+      ? isFlagEnabled(flags, MPTokenIssuanceCreateFlags.tfMPTCanTransfer)
+      : flags.tfMPTCanTransfer ?? false
+
   if (typeof tx.TransferFee === 'number') {
```
khancode, line 165: I would put the flags code inside the if block. And change the conditional check to 'tx.TransferFee != null' since the number check is done before it.

khancode (approved):
One more change and it LGTM.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
khancode: Ok, it looks confusing because ideally we're checking if the field is not null. Maybe add in != null with the number check?

khancode: Ok, it looks confusing because ideally we're checking if the field is not null. Maybe add in != null with the number check?

shawnxie999: adding the check would also give another lint error. The current condition `typeof tx.TransferFee === 'number' ` already implicitly checks that the field is not null

packages/xrpl/test/models/MPTokenIssuanceSet.test.ts
```diff
@@ -55,7 +55,7 @@ describe('MPTokenIssuanceSet', function () {
 
   it(`throws w/ conflicting flags`, function () {
     /* eslint-disable no-bitwise -- Bitwise operation needed for flag combination */
-    const invalid = {
+    let invalid = {
       TransactionType: 'MPTokenIssuanceSet',
       Account: 'rWYkbWkCeg8dP6rXALnjgZSjjLyih5NXm',
       MPTokenIssuanceID: TOKEN_ID,
```
achowdhry-ripple, line 61: Instead of redefining every test, might be easier to follow the pattern used in other test files (define the object in the beforeEach runner, and then selectively set missing fields to undefined or adding to the object over each test)

achowdhry-ripple (changes requested):
Test code cleanup but besides that LGTM

achowdhry-ripple (approved):
Approving preemptively so I don't block the release, but would recommend the small cleanup comment before merging

PR #2829, opened 2024-11-15
Credentials

achowdhry-ripple:
## High Level Overview of Change

JavaScript client implementation of https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0070d-credentials

Corresponding c++ changes: https://github.com/XRPLF/rippled/pull/5103/files

### Context of Change

Following changes are introduced in this PR:
(Quoting Verbatim from the XLS specification document)

Creating a Credential ledger object
Creating a CredentialCreate transaction type
Creating a CredentialAccept transaction type
Creating a CredentialDelete transaction type
Modifying the DepositPreauth ledger object
Modifying the DepositPreauth transaction
Modifying other transactions that are affected by Deposit Authorization -- Payment, PaymentChannelClaim, AccountDelete and EscrowFinish

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### Did you update HISTORY.md?

- [x] Yes
- [ ] No, this change does not impact library users

## Test Plan

unit and integration tests have been added, must use a version of rippled 2.3.0 or newer

PR #2874, opened 2025-01-07
PermissionedDomain XLS-80d

ckeshava:

This PR implements the XLS-80d specification titled "PermissionedDomains": https://github.com/XRPLF/XRPL-Standards/discussions/228

Here is the accompanying xrpl-py implementation: https://github.com/XRPLF/xrpl-py/pull/773

This PR has the following dependencies. The final merge needs to await the below two PRs:

Reference C++ Implementation: https://github.com/XRPLF/rippled/pull/5161
xrpl.js implementation of the Credentials feature: https://github.com/XRPLF/xrpl.js/pull/2829

### Context of Change

This PR implements the client library changes for two new transactions: `PermissionedDomainSet` and `PermissionedDomainDelete`

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes


packages/xrpl/src/models/ledger/PermissionedDomain.ts
```diff
@@ -0,0 +1,24 @@
+import { BaseLedgerEntry, HasPreviousTxnID } from './BaseLedgerEntry'
+
+export interface Credential {
+  Credential: {
+    Issuer: string
+    CredentialType: string
+  }
+}
```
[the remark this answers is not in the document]
ckeshava: no, the Credential structure needs to mirror the nested-object, as defined in the rippled codebase

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        // fetch the PD `index` from the previous account_objects RPC response
+        DomainID: pd.index
+      }
+
+      console.log(pd)
+
+      await testTransaction(testContext.client, tx_pd_delete, testContext.wallet)
+    },
+    TIMEOUT,
+  )
```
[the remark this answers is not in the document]
ckeshava: these cases are covered in the rippled cpp implementation

packages/xrpl/test/models/permissionedDomainDelete.test.ts
```diff
@@ -0,0 +1,40 @@
...
+describe('PermissionedDomainDelete', function () {
+  let tx
+
+  beforeEach(function () {
+    tx = {
+      TransactionType: 'PermissionedDomainDelete',
+      Account: 'rfmDuhDyLGgx94qiwf3YF8BUV5j6KSvE8',
+      DomainID: 'D88930B33C2B6831660BFD006D91FF100011AD4E67CBB78B460AF0A215103737',
+    } as any
+  })
```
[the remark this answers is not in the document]
ckeshava: specifying a generic JSON is not necessarily bad in these test cases. Users can submit JSON/map/dictionaries as input to the `validate` method.

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      // Step-3: Test the PDDelete transaction
+      const tx_pd_delete: PermissionedDomainDelete = {
+        TransactionType: 'PermissionedDomainDelete',
+        Account: testContext.wallet.classicAddress,
+        // fetch the PD `index` from the previous account_objects RPC response
+        DomainID: pd.index
+      }
+
+      console.log(pd)
```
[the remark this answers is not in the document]
ckeshava: thank you 0ad5503e

packages/xrpl/test/models/permissionedDomainSet.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    assert.throws(() => validatePermissionedDomainSet(tx), ValidationError, errorMessage)
+    assert.throws(() => validate(tx), ValidationError, errorMessage)
+  })
+
+  it(`throws w/ missing field AcceptedCredentials`, function () {
+    delete tx.AcceptedCredentials
+    const errorMessage = 'PermissionedDomainSet: missing field AcceptedCredentials'
+    assert.throws(() => validatePermissionedDomainSet(tx), ValidationError, errorMessage)
+    assert.throws(() => validate(tx), ValidationError, errorMessage)
+  })
```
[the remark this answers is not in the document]
ckeshava: thanks 0ad5503e

packages/xrpl/src/models/ledger/PermissionedDomain.ts
```diff
@@ -0,0 +1,24 @@
+import { BaseLedgerEntry, HasPreviousTxnID } from './BaseLedgerEntry'
+
+export interface Credential {
+  Credential: {
+    Issuer: string
+    CredentialType: string
+  }
```
achowdhry-ripple, line 7: I think this exists as AuthorizeCredential somewhere else from the Credentials feature. Can we either rename that or use that one?

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,47 @@
...
+  validateOptionalField,
+  validateRequiredField,
+} from './common'
+
+import { Credential } from '../ledger/PermissionedDomain'
+const ACCEPTED_CREDENTIALS_MAX_LENGTH = 10
+
+export interface PermissionedDomainSet extends BaseTransaction {
+  TransactionType: 'PermissionedDomainSet'
+
```
achowdhry-ripple, line 15: account field here too according to spec?

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,47 @@
...
+import {
+  BaseTransaction,
+  isString,
+  validateBaseTransaction,
+  validateOptionalField,
+  validateRequiredField,
+} from './common'
+
+import { Credential } from '../ledger/PermissionedDomain'
+const ACCEPTED_CREDENTIALS_MAX_LENGTH = 10
```
achowdhry-ripple, line 11: more of a spec design question but why is this 10 when the credentials array in the credeentials feature is always 8 max? Seems a little confusing to me

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    if (value.length > ACCEPTED_CREDENTIALS_MAX_LENGTH) {
+      throw new ValidationError(
+        `PermissionedDomainSet: AcceptedCredentials must have at most ${ACCEPTED_CREDENTIALS_MAX_LENGTH} Credential objects`,
+      )
+    }
+    else if (value.length === 0) {
+      throw new ValidationError(
+        `PermissionedDomainSet: AcceptedCredentials must have at least one Credential object`,
+      )
```
achowdhry-ripple, line 40: you'll also need to validate the credential shapes. There is a validateCredentialList helper function made as part of credentials that could be useful here

packages/xrpl/test/models/permissionedDomainDelete.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  it('verifies valid PermissionedDomainDelete', function () {
+    assert.doesNotThrow(() => validatePermissionedDomainDelete(tx))
+    assert.doesNotThrow(() => validate(tx))
+  })
+
+  it(`throws w/ missing field DomainID`, function () {
```
achowdhry-ripple, line 27: add tests for account parameter once added

packages/xrpl/test/models/permissionedDomainSet.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      TransactionType: 'PermissionedDomainSet',
+      Account: 'rfmDuhDyLGgx94qiwf3YF8BUV5j6KSvE8',
+      DomainID: 'D88930B33C2B6831660BFD006D91FF100011AD4E67CBB78B460AF0A215103737',
+      AcceptedCredentials: [sampleCredential]
+    } as any
+  })
+
+  it('verifies valid PermissionedDomainSet', function () {
+    assert.doesNotThrow(() => validatePermissionedDomainSet(tx))
+    assert.doesNotThrow(() => validate(tx))
```
achowdhry-ripple, line 33: same here, add account param test

packages/xrpl/test/models/permissionedDomainSet.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  it(`throws w/ missing field AcceptedCredentials`, function () {
+    delete tx.AcceptedCredentials
+    const errorMessage = 'PermissionedDomainSet: missing field AcceptedCredentials'
+    assert.throws(() => validatePermissionedDomainSet(tx), ValidationError, errorMessage)
+    assert.throws(() => validate(tx), ValidationError, errorMessage)
+  })
+
+  it('throws when AcceptedCredentials exceeds maximum length', function () {
+    tx.AcceptedCredentials = Array(11).fill(sampleCredential)
+    assert.throws(
```
achowdhry-ripple, line 52: will want invalid credential shapes tested here, see the tests added as a part of the credentials feature that test every case in the validateCredentialsList helper function

achowdhry-ripple (changes requested):
Few small questions and additions, overall looks solid though

packages/xrpl/src/models/transactions/permissionedDomainDelete.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  BaseTransaction,
+  isString,
+  validateBaseTransaction,
+  validateRequiredField
+} from './common'
+
+export interface PermissionedDomainDelete extends BaseTransaction {
+  TransactionType: 'PermissionedDomainDelete'
+
```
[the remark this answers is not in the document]
mvadari: That's in `BaseTransaction`.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
mvadari: The numbers were picked sort of arbitrarily tbh

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
mvadari: That's in `BaseTransaction`.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,47 @@
...
+// eslint-disable-next-line max-lines-per-function -- necessary to validate many fields
+export function validatePermissionedDomainSet(tx: Record<string, unknown>): void {
+  validateBaseTransaction(tx)
+
+  validateOptionalField(tx, 'DomainID', isString)
+
+  // eslint-disable-next-line max-lines-per-function -- necessary to validate many fields
+  validateRequiredField(tx, 'AcceptedCredentials', (value) => {
+    if (!Array.isArray(value)) {
+      throw new ValidationError('PermissionedDomainSet: AcceptedCredentials must be an array')
```
mvadari, line 29: Add a test covering this line

packages/xrpl/test/models/permissionedDomainSet.test.ts
ckeshava: do you remember which file contains your example references? I checked `packages/xrpl/test/models/CredentialCreate.test.ts`, `packages/xrpl/test/models/CredentialAccept.test.ts`, `packages/xrpl/test/models/CredentialDelete.test.ts` -- I didn't find any array-type fields in these models. Did I miss something?

packages/xrpl/test/models/permissionedDomainDelete.test.ts
ckeshava: I'm not convinced that `account`-related unit tests are required here. 
1. `account` is a part of the `BaseTransaction`, which gets tested with all the transaction-types.
2. Users cannot submit a transaction without an `account` field.
3. The `TxnSignature` field validates the correct-ness of the account, it's public and private keys. If there is any mismatch in these values, the digital-signature-scheme should fail. rippled's encryption libraries check for the correctness of signatures.
4. There are other required fields that we don't explicitly test. For instance, inserting an invalid `TransactionType` or a non-string `TransactionType` is an error that isn't covered by the unit tests.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: The `AcceptedCredentials` field is different from the `validateCredentialsList` method. The former accepts at most 10 members and Inner-Objects only (string IDs are rejected in `AcceptedCredentials`).

packages/xrpl/test/models/permissionedDomainDelete.test.ts
achowdhry-ripple: Ah yup forgot Account was in BaseTransaction, can ignore those comments from me

packages/xrpl/test/models/permissionedDomainSet.test.ts
achowdhry-ripple: escrowFinish unit tests is one example. Any of the other transactions that were modified in that spec to have a CredentialIDs[] parameter added also have those tests

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: 84061c07

packages/xrpl/test/models/permissionedDomainSet.test.ts
ckeshava: I have written some more tests here: https://github.com/XRPLF/xrpl.js/commit/84061c078b389b98b80e9a5741c096a3295faa82

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,63 @@
...
+        throw new ValidationError(
+          'PermissionedDomainSet: Invalid AcceptedCredentials format',
+        )
+      }
+    })
+
+    if (containsDuplicates(value)) {
+      throw new ValidationError(
+        `PermissionedDomainSet: AcceptedCredentials cannot contain duplicate elements`,
+      )
```
achowdhry-ripple, line 58: instead of duplicating the logic for validateCredentialsList could we reuse that method as a whole? It takes in transaction type as a parameter so the error messages are customized

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -500,7 +502,7 @@ export function validateCredentialsList(
   }
 }
 
-function containsDuplicates(objectList: object[]): boolean {
+export function containsDuplicates(objectList: object[]): boolean {
```
achowdhry-ripple, line 505: tied to other comment below, but if we use the whole validate function then these wouldn't need to be exported either

achowdhry-ripple:
A couple more cleanups, but LGTM after that. I really like the "lifecycle" test approach, it goes against the repo norm but it makes more sense for these kind of sequential transactions. In hindsight, this is what credentials tests in this repo should have done

mvadari:
A bunch of linter issues failing

.github/workflows/nodejs.yml
```diff
@@ -4,7 +4,7 @@
 name: Node.js CI
 
 env:
-  RIPPLED_DOCKER_IMAGE: rippleci/rippled:2.3.0-rc1
+  RIPPLED_DOCKER_IMAGE: rippleci/rippled:develop
```
mvadari, line 7: Hm I like this change - less updating needed in these files, and it serves as an additional form of testing for rippled

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,63 @@
...
+  BaseTransaction,
+  isString,
+  validateBaseTransaction,
+  validateOptionalField,
+  validateRequiredField,
+  isAuthorizeCredential,
+  containsDuplicates,
+} from './common'
+
+const ACCEPTED_CREDENTIALS_MAX_LENGTH = 10
```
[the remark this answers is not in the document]
mvadari: This is incorrect, since the rippled code has different limits for each.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: Yes, I wanted to re-use it, but there are two key differences as I explained here: https://github.com/XRPLF/xrpl.js/pull/2874#discussion_r1915394297

I don't want to over-complicate that method to accommodate both `Credential` and `PD`-related methods. 

These are the differing characteristics:
- The maximum length of the array of credentials
- The string format of credential IDs.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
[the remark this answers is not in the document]
ckeshava: this is out of scope for this PR. These values are determined as per user requirements.

packages/xrpl/src/models/ledger/PermissionedDomain.ts
ckeshava: [ce9193c](https://github.com/XRPLF/xrpl.js/pull/2874/commits/ce9193c434b85025ede03f14cd7b8d940b5c4501)

packages/xrpl/src/models/transactions/common.ts
ckeshava: I've responded to the earlier comment. It's difficult to port the existing validate method to validate the PDSet transaction. The parameters used are slightly different.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
achowdhry-ripple: The string format is already handled in the helper function, there is a boolean isStringID you can use to set that. That makes the only difference a list max of 10 vs 8, for which all the same checks are now duplicated again in a different part of the code. I'd strongly prefer if we do a small refactor that makes this helper function usable here instead of duplicating all of the checks

ckeshava: ok. fixed in cf772b93

packages/xrpl/src/models/transactions/common.ts
ckeshava: cf772b93

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: updated in cf772b93

mvadari: String IDs aren't accepted in either...?

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -455,13 +457,16 @@ export function validateCredentialType(tx: Record<string, unknown>): void {
...
+ * @param maxLengthCredentialsArray The maximum length of the credentials array.
+ *        PermissionedDomainSet transaction uses 10, other transactions use 8.
  * @throws Validation Error if the formatting is incorrect
  */
 // eslint-disable-next-line max-lines-per-function -- separating logic further will add unnecessary complexity
 export function validateCredentialsList(
   credentials: unknown,
   transactionType: string,
   isStringID: boolean,
+  maxLengthCredentialsArray: number = MAX_CREDENTIALS_LIST_LENGTH,
```
mvadari, line 469: This shouldn't have a default value

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -500,7 +505,13 @@ export function validateCredentialsList(
...
 
-function containsDuplicates(objectList: object[]): boolean {
+/**
+ * Check if an array of objects contains any duplicates.
+ *
+ * @param objectList - Array of objects to check for duplicates
+ * @returns True if duplicates exist, false otherwise
+ */
+export function containsDuplicates(objectList: object[]): boolean {
   const objSet = new Set(objectList.map((obj) => JSON.stringify(obj)))
```
mvadari, line 515: This will be incorrect if one has `{Issuer: ..., CredentialType: ...}` and the other has `{CredentialType: ... Issuer: ...}`

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,46 @@
+import { isArray } from 'lodash'
```
mvadari, line 1: We shouldn't use lodash

packages/xrpl/src/models/transactions/common.ts
ckeshava: Do you expect this method to be used in other contexts? It has been predominantly used in the CredentialCreate validation

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: is it because of low-frequency of recent commits? Do you have any alternative preferences?

mvadari: https://youmightnotneed.com/lodash

mvadari: lodash is a massive package and we don't want that size bloat in the xrpl.js package. https://bundlephobia.com/package/lodash@4.17.21

packages/xrpl/src/models/transactions/common.ts
mvadari: Neither option really feels like a "default", and I wouldn't want a library developer to make a mistake in the future (and use the wrong value) because of that.

packages/xrpl/src/models/transactions/common.ts
ckeshava: thanks for pointing it out. How does this look ? [da655b2](https://github.com/XRPLF/xrpl.js/pull/2874/commits/da655b20ac59172331999b5aa61ea22f62a60124)

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -15,6 +15,7 @@ import {
   MPTAmount,
 } from '../common'
 import { onlyHasFields } from '../utils'
+import { assert } from 'chai'
```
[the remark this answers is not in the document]
ckeshava: Does this work in non-node environments? For instance, a browser ?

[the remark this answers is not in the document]
ckeshava: This approach causes a repetition of logic. Other parts of the code already validate the structure of the nested object

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: okay. However, the size really matters only in memory-constrained environments right? Are any applications affected by the large size of the xrpl.js application ?

ckeshava: [2697a21](https://github.com/XRPLF/xrpl.js/pull/2874/commits/2697a218a1e8bb65e9099eaa6857e9f9b43a27dc)

mvadari: In JS, memory always matters, because it'll affect a lot of things when you're running the code in the browser.

packages/xrpl/src/models/transactions/common.ts
ckeshava: ok, fixed in c619920a

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @param tx - The transaction to validate.
+ * @throws {ValidationError} When the transaction is invalid.
+ */
+export function validatePermissionedDomainSet(
+  tx: Record<string, unknown>,
+): void {
+  validateBaseTransaction(tx)
+
+  validateOptionalField(tx, 'DomainID', isString)
+  validateRequiredField(tx, 'AcceptedCredentials', () => tx.AcceptedCredentials instanceof Array)
```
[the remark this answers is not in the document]
ckeshava: could you give concrete examples where the proposed code fails?

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -500,7 +506,36 @@ export function validateCredentialsList(
...
+
+    if (seen.has(key)) {
+      // Found a duplicate
+      return true
+    }
+
+    seen.add(key)
+  }
+
+  return false
```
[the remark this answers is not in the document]
ckeshava: fixed in [2316945](https://github.com/XRPLF/xrpl.js/pull/2874/commits/2316945923a7b16ecfaaedc27cc33ded096dc64b)

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -500,7 +505,42 @@ export function validateCredentialsList(
...
+
+  // Case-2: Process a list of nested objects
+  const seen = new Set<string>()
+
+  // Type guard to ensure we're working with AuthorizeCredential[]
+  // Note: This is not a rigorous type-guard. A more thorough solution would be to iterate over the array and check each item.
+  function isAuthorizeCredentialArray(
+    list: AuthorizeCredential[] | string[],
+  ): list is AuthorizeCredential[] {
+    return typeof list[0] !== 'string'
```
pdp2121, line 531: This is just the "else" case for the if statement on line 518. I think it's fine to just use the else statement with a comment to explain that this is the case for `AuthorizeCredential[]`

khancode:
It's helpful to include new transactions and requests that will be added in the PR description.

.github/workflows/nodejs.yml
khancode: I would say we point to a specific version because when `develop` updates, our integ tests may fail. Also, it's helpful to see what version of rippled the client libraries are testing against.

packages/xrpl/src/models/ledger/PermissionedDomain.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  Owner: string
+
+  Flags: 0
+
+  OwnerNode: string
+
+  Sequence: number
+
+  AcceptedCredentials: AuthorizeCredential[]
```
khancode, line 18: Easier to read to put transaction specific fields at top and matches the rest of the models in the codebase.

```suggestion
  LedgerEntryType: 'PermissionedDomain'

  Owner: string

  AcceptedCredentials: AuthorizeCredential[]

  Flags: 0

  OwnerNode: string

  Sequence: number
```

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -9,15 +9,15 @@ import {
...
   Memo,
   Signer,
   XChainBridge,
-  MPTAmount,
 } from '../common'
 import { onlyHasFields } from '../utils'
 
 const MEMO_SIZE = 3
-const MAX_CREDENTIALS_LIST_LENGTH = 8
+export const MAX_CREDENTIALS_LIST_LENGTH = 8
```
khancode, line 20: It's better to use JS terminology for array than using Python's list.
```suggestion
export const MAX_CREDENTIALS_ARRAY_LENGTH = 8
```

packages/xrpl/src/models/transactions/common.ts
khancode: It's unnecessary to have this inner function.

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  setupClient,
+  teardownClient,
+  type XrplIntegrationTestContext,
+} from '../setup'
+import { testTransaction } from '../utils'
+
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('PermissionedDomainSet', function () {
```
khancode, line 22: Same comment on your Python PR: Testing all feature transactions in one integ file doesn't match the the codebase's convention. For each new transaction, there should be a separate integ test file for it.

khancode (changes requested):

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,48 @@
...
+  validateCredentialsList,
+} from './common'
+
+const ACCEPTED_CREDENTIALS_MAX_LENGTH = 10
+
+export interface PermissionedDomainSet extends BaseTransaction {
+  TransactionType: 'PermissionedDomainSet'
+
+  DomainID?: string
+  AcceptedCredentials: AuthorizeCredential[]
```
mvadari, line 18:
```suggestion
  DomainID?: string
  
  AcceptedCredentials: AuthorizeCredential[]
```

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,48 @@
...
+  tx: Record<string, unknown>,
+): void {
+  validateBaseTransaction(tx)
+
+  validateOptionalField(tx, 'DomainID', isString)
+  validateRequiredField(
+    tx,
+    'AcceptedCredentials',
+    () => tx.AcceptedCredentials instanceof Array,
+  )
```
mvadari, line 37:
```suggestion
  validateRequiredField(
    tx,
    'AcceptedCredentials',
    () => Array.isArray(tx.AcceptedCredentials),
  )
```
You could also add `isArray` to the list of helpers in `common.ts`

packages/xrpl/test/models/permissionedDomainSet.test.ts
```diff
@@ -0,0 +1,104 @@
...
+      'PermissionedDomainSet: missing field AcceptedCredentials'
+    assert.throws(
+      () => validatePermissionedDomainSet(tx),
+      ValidationError,
+      errorMessage,
+    )
+    assert.throws(() => validate(tx), ValidationError, errorMessage)
+  })
+
+  it('throws when AcceptedCredentials exceeds maximum length', function () {
```
mvadari, line 60: All the tests should check `validatePermissionedDomainSet` and `validate`

packages/xrpl/test/models/utils.test.ts
```diff
@@ -28,6 +30,40 @@ import {
...
+      ]
+
+      // changing the order of the inner-objects in the list
+      const list_with_duplicates: AuthorizeCredential[] = [
+        { Credential: { Issuer: 'alice', CredentialType: 'Passport' } },
+        { Credential: { CredentialType: 'Passport', Issuer: 'alice' } },
+      ]
+
+      assert.isFalse(containsDuplicates(list_without_dups))
+      assert.isTrue(containsDuplicates(list_with_duplicates))
```
mvadari, line 47: These should be separate tests

packages/xrpl/test/models/utils.test.ts
```diff
@@ -28,6 +30,40 @@ import {
...
+        '5328F2D1D6EBBC6093DC10F1EA3DD630666F5B2491EB9BDD7DF9A6C45AC12C46',
+      ]
+
+      const list_with_duplicates: string[] = [
+        'EA85602C1B41F6F1F5E83C0E6B87142FB8957BD209469E4CC347BA2D0C26F66A',
+        'EA85602C1B41F6F1F5E83C0E6B87142FB8957BD209469E4CC347BA2D0C26F66A',
+      ]
+
+      assert.isFalse(containsDuplicates(list_without_dups))
+      assert.isTrue(containsDuplicates(list_with_duplicates))
```
mvadari, line 63: ditto

packages/xrpl/src/models/transactions/common.ts
ckeshava: @pdp2121  yes, you are correct. I have included a comment on line 526.

@khancode  I felt the type-guards offered by the inner function are more robust than a usual else-condition statement.

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
ckeshava: It feels more natural to validate the complete lifecycle of a ledger-object. If we split this into multiple files, there will be a lot of repetition of code. You cannot execute `PDDelete` without a preceding `PDSet` transaction.

There is no benefit to splitting this test into multiple files.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: I considered this suggestion, but I didn't fully understand the differences. Are our code-snippets very different?

packages/xrpl/test/models/permissionedDomainSet.test.ts
ckeshava: `validate` encompasses `validatePermissionedDomainSet`. Thanks, I've decided to only invoke the former method.

Should I make this change inside the PDSet model verification too? `validate` method multiplexes the validations for all transactions

packages/xrpl/test/models/utils.test.ts
ckeshava: [2080c21](https://github.com/XRPLF/xrpl.js/pull/2874/commits/2080c21bd425a596f844d2898ed65c465f19adfd)

packages/xrpl/test/models/utils.test.ts
ckeshava: [2080c21](https://github.com/XRPLF/xrpl.js/pull/2874/commits/2080c21bd425a596f844d2898ed65c465f19adfd)

packages/xrpl/test/models/permissionedDomainSet.test.ts
ckeshava: [2080c21](https://github.com/XRPLF/xrpl.js/pull/2874/commits/2080c21bd425a596f844d2898ed65c465f19adfd)

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: [2080c21](https://github.com/XRPLF/xrpl.js/pull/2874/commits/2080c21bd425a596f844d2898ed65c465f19adfd)

packages/xrpl/src/models/transactions/common.ts
ckeshava: [2080c21](https://github.com/XRPLF/xrpl.js/pull/2874/commits/2080c21bd425a596f844d2898ed65c465f19adfd)

packages/xrpl/src/models/ledger/PermissionedDomain.ts
ckeshava: [2080c21](https://github.com/XRPLF/xrpl.js/pull/2874/commits/2080c21bd425a596f844d2898ed65c465f19adfd)

ckeshava:
@khancode 
> I would say we point to a specific version because when develop updates, our integ tests may fail. Also, it's helpful to see what version of rippled the client libraries are testing against.

I'd prefer to run integration tests with the develop branch of rippled. This helps with faster detection of issues in both rippled and the client libraries. Furthermore, it absolves DGE developers from having to update the latest rippled versions.

khancode:
> @khancode
> 
> > I would say we point to a specific version because when develop updates, our integ tests may fail. Also, it's helpful to see what version of rippled the client libraries are testing against.
> 
> I'd prefer to run integration tests with the develop branch of rippled. This helps with faster detection of issues in both rippled and the client libraries. Furthermore, it absolves DGE developers from having to update the latest rippled versions.

That's fair. I support this change then :)

packages/xrpl/src/models/transactions/common.ts
khancode: It's best practice to avoid inner functions in JS because of how closure works. If you want to keep this helper function, then I would define it separately outside of this function.

pdp2121 (approved):

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
ckeshava: The benefits outweigh the conventional needs in this case

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,48 @@
...
+  validateRequiredField,
+  validateCredentialsList,
+} from './common'
+
+const ACCEPTED_CREDENTIALS_MAX_LENGTH = 10
+
+export interface PermissionedDomainSet extends BaseTransaction {
+  TransactionType: 'PermissionedDomainSet'
+
+  DomainID?: string
```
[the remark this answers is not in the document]
ckeshava: It was added in this commit: [7f0baad](https://github.com/XRPLF/xrpl.js/pull/2874/commits/7f0baad8d87fd8a4eeaf547cdb3f18fb7bf97a79)

Perhaps refreshing the website will show the updated version of the code

packages/xrpl/src/models/transactions/permissionedDomainDelete.ts
```diff
@@ -0,0 +1,26 @@
...
+  BaseTransaction,
+  isString,
+  validateBaseTransaction,
+  validateRequiredField,
+} from './common'
+
+export interface PermissionedDomainDelete extends BaseTransaction {
+  TransactionType: 'PermissionedDomainDelete'
+
+  DomainID: string
```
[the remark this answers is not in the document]
ckeshava: It was added in this commit: [7f0baad](https://github.com/XRPLF/xrpl.js/pull/2874/commits/7f0baad8d87fd8a4eeaf547cdb3f18fb7bf97a79)

Perhaps refreshing the website will show the updated version of the code

packages/xrpl/src/models/transactions/permissionedDomainDelete.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  BaseTransaction,
+  isString,
+  validateBaseTransaction,
+  validateRequiredField,
+} from './common'
+
+export interface PermissionedDomainDelete extends BaseTransaction {
```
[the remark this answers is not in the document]
ckeshava: It was added in this commit: [7f0baad](https://github.com/XRPLF/xrpl.js/pull/2874/commits/7f0baad8d87fd8a4eeaf547cdb3f18fb7bf97a79)

Perhaps refreshing github will show the updated version of the code

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,48 @@
...
+  isString,
+  validateBaseTransaction,
+  validateOptionalField,
+  validateRequiredField,
+  validateCredentialsList,
+} from './common'
+
+const ACCEPTED_CREDENTIALS_MAX_LENGTH = 10
+
+export interface PermissionedDomainSet extends BaseTransaction {
```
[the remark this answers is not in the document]
ckeshava: It was added in this commit: [7f0baad](https://github.com/XRPLF/xrpl.js/pull/2874/commits/7f0baad8d87fd8a4eeaf547cdb3f18fb7bf97a79)

Perhaps refreshing github will show the updated version of the code

packages/xrpl/src/models/transactions/common.ts
ckeshava: ok, updated in [2b3580a](https://github.com/XRPLF/xrpl.js/pull/2874/commits/2b3580a276dd73e2f6068397afc6a77ffde6a357)

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
khancode: I disagree. Breaking the convention for one feature while all others follow it isn't good practice.

ckeshava: I'd like to know the opinions of other reviewers to break this tie @pdp2121 @achowdhry-ripple  @mvadari

achowdhry-ripple: I personally didn't like this convention when implementing credentials, because separate files forces you to duplicate the create and accept transactions in each file, making the delete test cover the functionality of the first two anyways. I would prefer we switch to this convention going forward, and if the team agrees could you delete the create/accept tests for credentials and rename the delete file? Quick cleanup to create a more logical convention going forward

achowdhry-ripple (approved):
LGTM

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -9,15 +9,15 @@ import {
...
   Memo,
   Signer,
   XChainBridge,
-  MPTAmount,
 } from '../common'
 import { onlyHasFields } from '../utils'
 
 const MEMO_SIZE = 3
-const MAX_CREDENTIALS_LIST_LENGTH = 8
+export const MAX_CREDENTIALS_ARRAY_LENGTH = 8
```
mvadari, line 20:
```suggestion
export const MAX_AUTHORIZED_CREDENTIALS = 8
```

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -455,13 +457,16 @@ export function validateCredentialType(tx: Record<string, unknown>): void {
...
+ *        PermissionedDomainSet transaction uses 10, other transactions use 8.
  * @throws Validation Error if the formatting is incorrect
  */
-// eslint-disable-next-line max-lines-per-function -- separating logic further will add unnecessary complexity
+// eslint-disable-next-line max-lines-per-function, max-params -- separating logic further will add unnecessary complexity
 export function validateCredentialsList(
   credentials: unknown,
   transactionType: string,
   isStringID: boolean,
+  maxLengthCredentialsArray: number,
```
mvadari, line 469:
```suggestion
  maxCredentials: number,
```

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
mvadari: `Array.isArray` is better practice - https://stackoverflow.com/questions/28779255/is-instanceof-array-better-than-isarray-in-javascript

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
mvadari: I'm fine with tests like this.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -134,7 +134,9 @@ export function isIssuedCurrency(
  * @param input - The input to check the form and type of
  * @returns Whether the AuthorizeCredential is properly formed
  */
-function isAuthorizeCredential(input: unknown): input is AuthorizeCredential {
+export function isAuthorizeCredential(
```
mvadari, line 137: This function's name is too specific now that it is also being used for PDs

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
```diff
@@ -0,0 +1,54 @@
...
+import {
+  BaseTransaction,
+  isString,
+  validateBaseTransaction,
+  validateOptionalField,
+  validateRequiredField,
+  validateCredentialsList,
+} from './common'
+
+const ACCEPTED_CREDENTIALS_MAX_LENGTH = 10
```
mvadari, line 12:
```suggestion
const MAX_ACCEPTED_CREDENTIALS = 10
```

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    async () => {
+      const sampleCredential: AuthorizeCredential = {
+        Credential: {
+          CredentialType: stringToHex('Passport'),
+          Issuer: testContext.wallet.classicAddress,
+        },
+      }
+
+      // Step-1: Test the PermissionedDomainSet transaction
+      const tx_pd_set: PermissionedDomainSet = {
```
mvadari, line 41: Use `camelCase`, not `snakeCase` in JS

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      })
+
+      assert.equal(result.result.account_objects.length, 1)
+      const pd = result.result.account_objects[0] as PermissionedDomain
+
+      assert.equal(pd.Flags, 0)
+      expect(pd.AcceptedCredentials).toEqual([sampleCredential])
+
+      // validate the ledger_entry RPC
+      const ledger_entry_request: LedgerEntryRequest = {
```
mvadari, line 64: Use `camelCase`, not `snakeCase` in JS

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        command: 'ledger_entry',
+        // fetch the PD `index` from the previous account_objects RPC response
+        index: pd.index,
+      }
+      const ledger_entry_result = await testContext.client.request(
```
mvadari, line 69: Use `camelCase`, not `snakeCase` in JS

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        ledger_entry_request,
+      )
+      assert.deepEqual(pd, ledger_entry_result.result.node)
+
+      // Step-3: Test the PDDelete transaction
+      const tx_pd_delete: PermissionedDomainDelete = {
```
mvadari, line 75: Use `camelCase`, not `snakeCase` in JS

packages/xrpl/test/models/permissionedDomainSet.test.ts
mvadari: See other test files - it's better to check both, just in case.

packages/xrpl/test/models/permissionedDomainSet.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  it('verifies valid PermissionedDomainSet', function () {
+    assert.doesNotThrow(() => validate(tx))
+  })
+
+  it(`throws with invalid field DomainID`, function () {
+    // DomainID is expected to be a string
+    tx.DomainID = 1234
+    const errorMessage = 'PermissionedDomainSet: invalid field DomainID'
+    assert.throws(() => validate(tx), ValidationError, errorMessage)
+    assert.throws(() => validate(tx), ValidationError, errorMessage)
```
mvadari, line 39: This doesn't do anything.
```suggestion
    assert.throws(() => validatePermissionedDomainSet(tx), ValidationError, errorMessage)
    assert.throws(() => validate(tx), ValidationError, errorMessage)
```

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
ckeshava: @achowdhry-ripple  I'm happy to do that in a separate PR. I'd like to restrict the scope of this PR at the moment.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: Hmm, thanks for the above article.

I don't imagine `PDSet` transaction will be invoked across multiple windows in the browser. I expect transaction submission to complete in a single window. This is the edge case where `instanceof` throws false negatives.

Here is the original blog post: http://blog.niftysnippets.org/2010/09/say-what.html#instanceof

packages/xrpl/test/models/permissionedDomainSet.test.ts
ckeshava: I don't see the need to separately test `validatePermissionedDomainSet` over and above the `validate` method. What do you think?

Yes, other test files invoke both the methods, but that appears unnecessary.

packages/xrpl/src/models/transactions/common.ts
ckeshava: @mvadari  I don't have strong preference about your variable-name-change suggestions, but they are used in multiple places across the codebase. Furthermore, these names are suited to the subjective preferences of the author.

If you could provide me a commit/patch will all the changes, I'm happy to incorporate that. Sorry, I'm a little pressed for time.

packages/xrpl/test/models/permissionedDomainSet.test.ts
ckeshava: I have removed the second, duplicate function call. `validate` method invokes `validatePermissionedDomainSet` in its implementation. There is no need to invoke this method again.

[9154151](https://github.com/XRPLF/xrpl.js/pull/2874/commits/91541510cb43525583d0c32c3acec258831fd230)

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
ckeshava: [9154151](https://github.com/XRPLF/xrpl.js/pull/2874/commits/91541510cb43525583d0c32c3acec258831fd230)

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
ckeshava: [9154151](https://github.com/XRPLF/xrpl.js/pull/2874/commits/91541510cb43525583d0c32c3acec258831fd230)

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
ckeshava: [9154151](https://github.com/XRPLF/xrpl.js/pull/2874/commits/91541510cb43525583d0c32c3acec258831fd230)

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
ckeshava: [9154151](https://github.com/XRPLF/xrpl.js/pull/2874/commits/91541510cb43525583d0c32c3acec258831fd230)

packages/xrpl/src/models/methods/serverState.ts
```diff
@@ -51,6 +51,7 @@ export interface ServerStateResponse extends BaseResponse {
       load_factor_fee_queue?: number
       load_factor_fee_reference?: number
       load_factor_server?: number
+      network_id: number
```
mvadari, line 54: Please add this to the changelog as a fix

packages/xrpl/src/models/transactions/common.ts
mvadari: Most IDEs will allow you to update all at once with F2 (or you can do a ctrl-f-replace). IMO this name as-is is incorrect/misleading since DepositPreauth and Permissioned Domains use different lengths of arrays.

packages/xrpl/src/models/transactions/common.ts
mvadari: Yes, it's `AcceptedCredentials`, not `AuthorizeCredentials` there. That's my point.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
mvadari: Regardless, let's use best practices, because xrpl.js is used in the browser, and we should be avoiding things that may result in edge cases (especially when it's not very difficult to make that change).

packages/xrpl/test/integration/requests/serverState.test.ts
```diff
@@ -116,6 +117,8 @@ describe('server_state', function () {
         'node_size',
         'initial_sync_duration_us',
         'ports',
+        'git',
+        'network_id',
```
mvadari, line 121: Why does this need to be removed?

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    async () => {
+      const sampleCredential: AuthorizeCredential = {
+        Credential: {
+          CredentialType: stringToHex('Passport'),
+          Issuer: testContext.wallet.classicAddress,
+        },
+      }
+
+      // Step-1: Test the PermissionedDomainSet transaction
+      const txPDSet: PermissionedDomainSet = {
```
mvadari, line 41: nit:
```suggestion
      const pdSet: PermissionedDomainSet = {
```

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        // fetch the PD `index` from the previous account_objects RPC response
+        index: pd.index,
+      }
+      const ledgerEntryResult = await testContext.client.request(
+        ledgerEntryRequest,
+      )
+      assert.deepEqual(pd, ledgerEntryResult.result.node)
+
+      // Step-3: Test the PDDelete transaction
+      const txPDDelete: PermissionedDomainDelete = {
```
mvadari, line 75: nit:
```suggestion
      const pdDelete: PermissionedDomainDelete = {
```

packages/xrpl/src/models/ledger/PermissionedDomain.ts
```diff
@@ -0,0 +1,29 @@
...
+  /* The ledger object's type (PermissionedDomain). */
+  LedgerEntryType: 'PermissionedDomain'
+
+  /* The account that controls the settings of the domain. */
+  Owner: string
+
+  /* The credentials that are accepted by the domain.
+  Ownership of one of these credentials automatically
+  makes you a member of the domain. */
+  AcceptedCredentials: AuthorizeCredential[]
```
mvadari, line 17: This type name is really confusing here - can we alias it or something?

packages/xrpl/test/models/permissionedDomainSet.test.ts
mvadari: It's easy to check both at the same time, and doesn't really cost anything, so we should be testing both. This has the potential to catch issues if there _are_ discrepancies.

packages/xrpl/test/integration/transactions/permissionedDomain.test.ts
khancode: We should still create a ticket to eventually refactor the existing tests to the new convention. It's worth the time/effort to make the tests run more efficient and stick to a new convention. @ckeshava can you create it?

ckeshava: ok, I've created a ticket

ckeshava:
Hello all, I'd like to merge this PR soon. Please give another review of this PR. 

If you have pending concerns, I'll address them in a separate PR.

packages/xrpl/src/models/ledger/PermissionedDomain.ts
ckeshava: I had to use the `AuthorizeCredential` name because I imported it from the Credential-related file. Please check the earlier PR review comments from @achowdhry-ripple in this page

ckeshava: Do you have a preference on the alias name?

packages/xrpl/src/models/transactions/common.ts
ckeshava: @achowdhry-ripple  If I rename this to AcceptedCredentials, will it confuse the Credential's usage? Should I rename it to `isValidCredential` instead?

packages/xrpl/test/integration/requests/serverState.test.ts
ckeshava: The integration tests use a standalone docker instance of rippled. This rippled does not use a custom `network_id` field in its configuration. Hence, this field is not returned by `server_state` RPC.

ckeshava: It is difficult to predict the `hash` and the `branch` of the `git` sub-field.

khancode (approved):

packages/xrpl/test/integration/requests/serverState.test.ts
mvadari: `git` I understand, `network_id` is unnecessary.

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
mvadari: Bump

packages/xrpl/src/models/transactions/common.ts
mvadari: Bump

packages/xrpl/test/integration/requests/serverState.test.ts
ckeshava: the tests won't pass without this ignore directive

ckeshava: Alternatively, I'll have to update the rippled.cfg file

packages/xrpl/src/models/transactions/permissionedDomainSet.ts
ckeshava: [3dfeaf1](https://github.com/XRPLF/xrpl.js/pull/2874/commits/3dfeaf1d68fb42f48f0c86840c4aec4af2af46be)

packages/xrpl/src/models/transactions/common.ts
ckeshava: [3dfeaf1](https://github.com/XRPLF/xrpl.js/pull/2874/commits/3dfeaf1d68fb42f48f0c86840c4aec4af2af46be)

packages/xrpl/src/models/transactions/common.ts
ckeshava: [3dfeaf1](https://github.com/XRPLF/xrpl.js/pull/2874/commits/3dfeaf1d68fb42f48f0c86840c4aec4af2af46be)

packages/xrpl/src/models/transactions/common.ts
ckeshava: @mvadari I'd like to discuss this issue offline. I'll merge this PR at the moment because explorer changes need the client library

packages/xrpl/test/integration/requests/serverState.test.ts
ckeshava: @mvadari I'd like to discuss this issue offline. I'll merge this PR at the moment because explorer changes need the client library

PR #2875, opened 2025-01-10
Support NFTokenMintOffer

tequdev:

Support [NFTokenMintOffer](https://github.com/XRPLF/XRPL-Standards/blob/master/XLS-0052-NFTokenMintOffer/README.md)

### Context of Change

add fields to NFTokenMint transaction
- Amount
- Destination
- Expiration

### Type of Change

- [x] Yes

## Test Plan

add unit test


mvadari (commented):
Please add an integration test

packages/xrpl/src/models/transactions/NFTokenMint.ts
```diff
@@ -140,4 +165,16 @@ export function validateNFTokenMint(tx: Record<string, unknown>): void {
   if (tx.NFTokenTaxon == null) {
     throw new ValidationError('NFTokenMint: missing field NFTokenTaxon')
   }
+
```
ckeshava, line 168:
```suggestion
  if (!((tx.Amount == null && tx.Expiration == null && tx.Destination == null) || (tx.Amount != null && tx.Expiration != null && tx.Destination != null))) {
    if (tx.Expiration != null || tx.Destination != null) {
      throw new ValidationError(
        'NFTokenMint: All three fields Amount, Expiration, Destination must be present',
      )
    }
  }
```

We will need all three fields to be present (or) absent to make logical sense, isn't it? Isn't this a more comprehensive check?

I'm guessing this is already checked in rippled, so perhaps there is no need for a rigorous check here.

ckeshava:
@tequdev  if you would like to fix the integration test, please check this file: https://github.com/XRPLF/xrpl.js/pull/2874/files#diff-aed00401282930e11f525e026ee0128d91284d95586aaee7c741f9ba0c8d409e

The `server_info` and `server_state` RPC has been updated to include a new `git` field.

packages/xrpl/src/models/transactions/NFTokenMint.ts
```diff
@@ -140,4 +165,16 @@ export function validateNFTokenMint(tx: Record<string, unknown>): void {
...
+    if (tx.Expiration != null || tx.Destination != null) {
+      throw new ValidationError(
+        'NFTokenMint: Amount is required when Expiration or Destination is present',
+      )
+    }
+  }
+
+  validateOptionalField(tx, 'Amount', isAmount)
+  validateOptionalField(tx, 'Expiration', isNumber)
+  validateOptionalField(tx, 'Destination', isAccount)
```
achowdhry-ripple, line 179: nit: move these to the top so we validate all fields before any other logic that might use them

packages/xrpl/src/models/transactions/NFTokenMint.ts
```diff
@@ -104,12 +107,34 @@ export interface NFTokenMint extends BaseTransaction {
    * set to `undefined` value if you do not use it.
    */
   URI?: string | null
+  /**
+   * Indicates the amount for the Token.
```
achowdhry-ripple, line 111:
```suggestion
   * Indicates the amount expected or offered for the Token.
```

packages/xrpl/HISTORY.md
```diff
@@ -10,6 +10,9 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 ### Changed
 * Deprecated `setTransactionFlagsToNumber`. Start using convertTxFlagsToNumber instead
 
+### Added
+* Support for XLS-52 (NFTokenMintOffer)
```
achowdhry-ripple, line 14:
```suggestion
* Support for NFTokenMintOffer (XLS-52)
```

packages/xrpl/src/models/transactions/NFTokenMint.ts
achowdhry-ripple: line 169 seems unnecessary as well

tequdev: `Expiration` and `Destination` can only be set if `Amount` is present, but they are still optional fields.

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
```diff
@@ -91,4 +96,63 @@ describe('NFTokenMint', function () {
...
+      assert.equal(response.type, 'response')
+
+      const txRequest: TxRequest = {
+        command: 'tx',
+        transaction: hashSignedTx(response.result.tx_blob),
+      }
+      const txResponse = await testContext.client.request(txRequest)
+
+      assert.equal(
+        (txResponse.result.meta as TransactionMetadata).TransactionResult,
```
mvadari, line 126: This change to the `tx` model ought to fix your typing issues here:
```
  meta_blob?: Version extends typeof RIPPLED_API_V2 ? string : never
  meta?: Version extends typeof RIPPLED_API_V2
    ? TransactionMetadata<T>
    : TransactionMetadata<T> | string
```

packages/xrpl/test/models/NFTokenMint.test.ts
```diff
@@ -109,4 +109,56 @@ describe('NFTokenMint', function () {
...
+    } as any
+
+    assert.throws(
+      () => validate(invalid),
+      ValidationError,
+      'NFTokenMint: Amount is required when Expiration or Destination is present',
+    )
+  })
+
+  it(`verifies valid NFTokenMint with Amount, Destination and Expiration`, function () {
```
mvadari, line 149: nit: this test should go at the top

packages/xrpl/src/models/transactions/NFTokenMint.ts
ckeshava: What happens if a developer specifies `Destination` field but not the `Amount` field?

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
```diff
@@ -91,4 +96,63 @@ describe('NFTokenMint', function () {
...
+      assert.isTrue(
+        existsOffer,
+        `Expected to exist an offer for NFT with NFTokenID ${nftokenID} but did not find it.
+      \n\nHere's what was returned from 'nft_sell_offers': ${JSON.stringify(
+        sellOffers,
+      )}`,
+      )
+    },
+    TIMEOUT,
+  )
```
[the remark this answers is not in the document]
ckeshava: these cases would have been covered in the rippled cpp implementation. Usually, client libraries forward these transactions to the rippled node.

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
ckeshava: @mvadari  This tests uses rippled APIv2 (default version, if `api_version` field is not specified). Is there a need to multiplex between the APIv1 and APIv2 types? How is this useful?

packages/xrpl/src/models/transactions/NFTokenMint.ts
tequdev: it will be `temMALFORMED`

packages/xrpl/src/models/transactions/NFTokenMint.ts
tequdev: It is probably better for developers to check for cases where values should not exist before checking for values. 
(In other words, you may have seen an error message (invalid field) and fixed the value, but you shouldn't have set that field in the first place.)

But I have no objection to making the change. What do you think?

packages/xrpl/src/models/transactions/NFTokenMint.ts
ckeshava: Yes, but that error is returned from the `rippled` node. I feel we can catch that error inside client-libraries itself.

However, this is just a suggestion. I like the PR even without this change.

ckeshava (approved):
thank you for the PR

packages/xrpl/src/models/transactions/NFTokenMint.ts
achowdhry-ripple: More of a code cleanup to stay consistent with the repo, and makes the logic a bit easier to follow. I don't feel strongly about this though so ultimately up to you

achowdhry-ripple (approved):
thanks -- appreciate the work here, lgtm

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
mvadari: The `tx` model is slightly incorrect - the casting shouldn't be necessary, and this change should fix that.

tequdev:
> @tequdev  Can we merge this PR? Are you happy with it?

yes

packages/xrpl/test/integration/transactions/nftokenMint.test.ts
ckeshava: @tequdev if you'd like to address this comment, we can wait.

PR #2893, opened 2025-02-05
feat: add AMMClawback support

### LINKED ISSUE #2850: Support AMMClawback

khancode:

Adds `AMMClawback` transaction.

Resources:
- [XLS-74 spec](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0074d-amm-clawback) 
- [XRPL docs](https://xrpl.org/docs/references/protocol/transactions/types/ammclawback)

Fixes #2850 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Adds unit and integ tests for `AMMClawback` transaction.


packages/xrpl/package-lock.json
achowdhry-ripple, line 1: why are the package locks changing? Was there a dependency update

packages/ripple-binary-codec/package-lock.json
achowdhry-ripple, line 1: actually on second thought, these shouldn't be committed at all right?

packages/xrpl/src/models/transactions/AMMClawback.ts
```diff
@@ -0,0 +1,123 @@
...
+export function validateAMMClawback(tx: Record<string, unknown>): void {
+  validateBaseTransaction(tx)
+
+  validateRequiredField(tx, 'Holder', isAccount)
+
+  validateRequiredField(tx, 'Asset', isCurrency)
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment -- used for comparing tx.Asset.issuer
+  // @ts-expect-error -- used for comparing tx.Asset.issuer
+  if (tx.Holder === tx.Asset.issuer) {
```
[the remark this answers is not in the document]
achowdhry-ripple: I agree that stronger typing should be used to avoid these error suppressions

packages/xrpl/test/integration/transactions/ammClawback.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      Flags: AMMDepositFlags.tfSingleAsset,
+    }
+
+    await testTransaction(testContext.client, ammDepositTx, holderWallet)
+
+    const ammClawback: AMMClawback = {
+      TransactionType: 'AMMClawback',
+      Account: issuerWallet.address,
+      Holder: holderWallet.address,
+      // @ts-expect-error -- this is an IssuedCurrency
```
achowdhry-ripple, line 43: this seems like it doesn't need to be suppressed if you type check it

packages/xrpl/test/integration/transactions/ammClawback.test.ts
achowdhry-ripple, line 1: could be useful to add an integration test for the two asset path as well

achowdhry-ripple (changes requested):
First pass looks solid, just requesting some typing cleanups and minor changes

packages/xrpl/package-lock.json
ckeshava: these are new package-lock files. I'm guessing @khancode  installed these npm packages separately?

Do we need these files in the repository? Can't users do `npm i` and recreate these files?

packages/xrpl/test/integration/transactions/ammClawback.test.ts
```diff
@@ -0,0 +1,50 @@
...
+      Asset: asset,
```
ckeshava, line 44: I don't understand this test. How does it not conflict with this validation check?

```
if (tx.Amount.issuer !== tx.Asset.issuer) {
      throw new ValidationError(
        'AMMClawback: issuer must be identical for both Amount and Asset',
      )
```

`tx.Asset` is `{'currency': 'XRP'}`. It does not have an `issuer` specified right?

Does this transaction succeed if you specify the `Amount` field?

packages/ripple-binary-codec/package-lock.json
khancode: Yes they shouldn't. My VS Code editor generated them automatically 😄

packages/xrpl/package-lock.json
khancode: The sub packages shouldn't have package-lock.json since this is a monorepo. Only the root directory should which it does. My VS Code editor automatically generated them by mistake. I just removed them.

packages/xrpl/test/integration/transactions/ammClawback.test.ts
khancode: It's resolved now.

packages/xrpl/test/integration/transactions/ammClawback.test.ts
khancode: You're right, I just fixed it.

packages/xrpl/test/integration/transactions/ammClawback.test.ts
khancode: I did the same as [xrpl-py](https://github.com/XRPLF/xrpl-py/pull/802/files#r1945297078) and added a unit test for it

packages/xrpl/src/models/transactions/AMMClawback.ts
```diff
@@ -0,0 +1,118 @@
...
+ * Verify the form and type of an AMMClawback at runtime.
+ *
+ * @param tx - An AMMClawback Transaction.
+ * @throws When the AMMClawback is Malformed.
+ */
+export function validateAMMClawback(tx: Record<string, unknown>): void {
+  validateBaseTransaction(tx)
+
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- required
+  const txAMMClawback = tx as unknown as AMMClawback
```
achowdhry-ripple, line 83: I don't think the base check is enough to safely cast the tx to an AMMClawback -- It would be consistent with other transactions in the repo to not have this override and exhaust the needed checks here

khancode: Coderabbit suggested to do this override and you agreed with it in https://github.com/XRPLF/xrpl.js/pull/2893#discussion_r1945158820

achowdhry-ripple: My comment was just about avoiding error suppressions, not agreeing to add any more. My point there was we should follow good typing practices instead of all of the overrides that were previously there. Seeing as there is no need for overrides in the other transactions, unless there is a specific reason we need it here I'd prefer we don't use it for code cleanliness

khancode: But that's not true, `payment.ts` and other models that do use error suppressions in the codebase when needed. This suppression is needed because we're asserting from `Record<string, unknown>` and TS think this is an error but it's actually not.

ckeshava: Other transaction types do not make this type-assertion nor do they have the  `eslint-disable` directive. Here is an example:
image

ckeshava: I don't have a lot of experience with the type-system in Typescript. I'll defer to the two of you.

ckeshava (approved):

packages/xrpl/src/models/transactions/AMMClawback.ts
achowdhry-ripple: @khancode I do understand that type conversions are needed and used in other parts of the code, but I don't see any examples of a transaction validation casting the entire tx record before any validations. My understanding was that the intent of the validate function was to verify the shape by each property, which is why we don't force the conversion right away like that and instead add checks for each value?

khancode: Also the purpose of validate functions is for developers who don't use TS. Doing suppressions becomes inevitable when the function allows for any JSON object parameter.

achowdhry-ripple: if you feel strongly, I won't block this PR -- but I do think converting the whole tx is a bit unclean and wouldn't want this to become the convention unless others agree

achowdhry-ripple (approved):

packages/xrpl/src/models/transactions/AMMClawback.ts
khancode: @achowdhry-ripple I'm happy to go with a better alternative. Do you have anything explicit in mind? Otherwise, for this case specifically, this whole tx conversion does seem reasonable because it saves us from doing multiple suppressions as described in the Coderrabit thread I already shared.

achowdhry-ripple: My alternative would be to specifically cast just `const asset = tx.Asset as Currency` and `const amount = tx.Amount as Amount` after the validate checks on those properties, which would solve the nested property typing issues and would be consistent with payment/other transaction models in the rep. If you would prefer not to do that since it will need two overrides, then I'd say to at least change the order so that all of the field validations are done before you type cast the tx object

khancode: Makes sense, thanks! I just pushed the refactor.

PR #2896, opened 2025-02-14
4.2.0 release

khancode:

4.2.0 release

### Type of Change

- [ ] Documentation Updates
- [x] Release

- [x] Yes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14507,7 +14507,7 @@
       }
     },
     "packages/ripple-binary-codec": {
-      "version": "2.2.0",
+      "version": "2.3.0",
       "license": "ISC",
       "dependencies": {
         "@xrplf/isomorphic": "^1.0.1",
@@ -14540,7 +14540,7 @@
       }
     },
     "packages/xrpl": {
-      "version": "4.1.0",
+      "version": "4.2.0",
       "license": "ISC",
       "dependencies": {
         "@scure/bip32": "^1.3.1",
@@ -14550,7 +14550,7 @@
         "bignumber.js": "^9.0.0",
         "eventemitter3": "^5.0.1",
         "ripple-address-codec": "^5.0.0",
-        "ripple-binary-codec": "^2.2.0",
+        "ripple-binary-codec": "^2.3.0",
         "ripple-keypairs": "^2.0.0"
       },
       "devDependencies": {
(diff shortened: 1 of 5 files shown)


packages/ripple-binary-codec/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ripple-binary-codec",
-  "version": "2.2.0",
+  "version": "2.3.0",
```
ckeshava, line 3: Why do we need to increment the version of `ripple-binary-codec` package?

packages/ripple-binary-codec/HISTORY.md
```diff
@@ -2,8 +2,11 @@
 
 ## Unreleased
 
+## 2.3.0 (2025-2-13)
+
 ### Added
 * Support for the AMMClawback amendment (XLS-73)
+* Support for the Permissioned Domains amendment (XLS-80).
```
ckeshava, line 9: `PermissionedDomain` amendment does not affect `ripple-binary-codec` anymore than every other amendment. Is it required to include this line in the CHANGELOG ?

package-lock.json
```diff
@@ -14507,7 +14507,7 @@
       }
```
ckeshava, line 14507: should we place the `package-lock.json` file under `.gitignore` ? It's diff should not be pushed into remotes.

packages/ripple-binary-codec/package.json
khancode: For `definitions.json` changes.

packages/ripple-binary-codec/HISTORY.md
khancode: Yes it's required because `definitions.json` change is within `ripple-binary-codec`

package-lock.json
khancode: Only the root package-lock.json should be updated so this change is correct.

ckeshava: hmm, other developers can always do `npm install` inside the repository root. Do we need to push this lock file?

ckeshava (approved):

package-lock.json
khancode: This is required because `package-lock.json` installs the exact, versioned dependency tree rather than using starred versioning. Starred versioning can install newer versions of packages that can potentially break the project.

achowdhry-ripple (approved):

PR #2913, opened 2025-02-26
fix #2911: Update OracleSet transaction model, unit, integ tests

### LINKED ISSUE #2911: Fix OracleSet transaction model's AssetPrice field
OracleSet transaction needs to be modified such that AssetPrice field accepts a `str` type. This is required for specifying very large numbers. In the current version of the code, `AssetPrice` accepts only `number` type.

Refer to this conversation for more context: https://github.com/XRPLF/xrpl.js/pull/2795#discussion_r1968316704

ckeshava:
Please refer to this issue: https://github.com/XRPLF/xrpl.js/issues/2911

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

### DIFF
--- a/packages/xrpl/src/models/transactions/oracleSet.ts
+++ b/packages/xrpl/src/models/transactions/oracleSet.ts
@@ -10,6 +10,8 @@ import {
   validateRequiredField,
 } from './common'
 
+import {HEX_REGEX} from 'ripple-binary-codec/dist/types/uint-64'
+
 const PRICE_DATA_SERIES_MAX_LENGTH = 10
 const SCALE_MAX = 10
 
@@ -146,7 +148,7 @@ export function validateOracleSet(tx: Record<string, unknown>): void {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- we are validating the type
         'AssetPrice' in priceData.PriceData &&
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- we are validating the type
-        !isNumber(priceData.PriceData.AssetPrice)
+        !(isNumber(priceData.PriceData.AssetPrice) || (typeof priceData.PriceData.AssetPrice === 'string' && HEX_REGEX.test(priceData.PriceData.AssetPrice)))
       ) {
         throw new ValidationError('OracleSet: invalid field AssetPrice')
       }
--- a/packages/xrpl/test/models/oracleSet.test.ts
+++ b/packages/xrpl/test/models/oracleSet.test.ts
@@ -167,7 +167,14 @@ describe('OracleSet', function () {
   })
 
   it(`throws w/ invalid AssetPrice of PriceDataSeries`, function () {
-    tx.PriceDataSeries[0].PriceData.AssetPrice = '1234'
+    tx.PriceDataSeries[0].PriceData.AssetPrice = 'ghij' // value cannot be parsed as hexadecimal number
+    const errorMessage = 'OracleSet: invalid field AssetPrice'
+    assert.throws(() => validateOracleSet(tx), ValidationError, errorMessage)
+    assert.throws(() => validate(tx), ValidationError, errorMessage)
+  })
+
+  it(`throws w/ invalid AssetPrice type in PriceDataSeries`, function () {
+    tx.PriceDataSeries[0].PriceData.AssetPrice = ['sample', 'invalid', 'type']
     const errorMessage = 'OracleSet: invalid field AssetPrice'
     assert.throws(() => validateOracleSet(tx), ValidationError, errorMessage)
     assert.throws(() => validate(tx), ValidationError, errorMessage)
--- a/packages/ripple-binary-codec/src/types/uint-64.ts
+++ b/packages/ripple-binary-codec/src/types/uint-64.ts
@@ -129,4 +129,4 @@ class UInt64 extends UInt {
   }
 }
 
-export { UInt64 }
+export { UInt64, HEX_REGEX }
--- a/packages/xrpl/test/integration/transactions/oracleSet.test.ts
+++ b/packages/xrpl/test/integration/transactions/oracleSet.test.ts
@@ -39,6 +39,17 @@ describe('OracleSet', function () {
               Scale: 3,
             },
           },
+          {
+            PriceData: {
+              BaseAsset: 'XRP',
+              QuoteAsset: 'INR',
+              // Upper bound admissible value for AssetPrice field
+              // large numeric values necessarily have to use str type in Javascript
+              // number type uses double-precision floating point representation, hence represents a smaller range of values
+              AssetPrice: 'ffffffffffffffff',
+              Scale: 3,
+            },
+          },
         ],
         Provider: stringToHex('chainlink'),
         URI: '6469645F6578616D706C65',
@@ -62,12 +73,18 @@ describe('OracleSet', function () {
       assert.equal(oracle.Owner, testContext.wallet.classicAddress)
       assert.equal(oracle.AssetClass, tx.AssetClass)
       assert.equal(oracle.Provider, tx.Provider)
-      assert.equal(oracle.PriceDataSeries.length, 1)
+      assert.equal(oracle.PriceDataSeries.length, 2)
       assert.equal(oracle.PriceDataSeries[0].PriceData.BaseAsset, 'XRP')
       assert.equal(oracle.PriceDataSeries[0].PriceData.QuoteAsset, 'USD')
       assert.equal(oracle.PriceDataSeries[0].PriceData.AssetPrice, '2e4')
       assert.equal(oracle.PriceDataSeries[0].PriceData.Scale, 3)
       assert.equal(oracle.Flags, 0)
+
+      // validate the serialization of large AssetPrice values
+      assert.equal(
+        oracle.PriceDataSeries[1].PriceData.AssetPrice,
+        'ffffffffffffffff',
+      )
     },
     TIMEOUT,
   )


packages/xrpl/test/models/oracleSet.test.ts
```diff
@@ -167,7 +167,14 @@ describe('OracleSet', function () {
...
   it(`throws w/ invalid AssetPrice of PriceDataSeries`, function () {
-    tx.PriceDataSeries[0].PriceData.AssetPrice = '1234'
+    tx.PriceDataSeries[0].PriceData.AssetPrice = 'ghij' // value cannot be parsed as hexadecimal number
+    const errorMessage = 'OracleSet: invalid field AssetPrice'
+    assert.throws(() => validateOracleSet(tx), ValidationError, errorMessage)
+    assert.throws(() => validate(tx), ValidationError, errorMessage)
+  })
+
+  it(`throws w/ invalid AssetPrice type in PriceDataSeries`, function () {
+    tx.PriceDataSeries[0].PriceData.AssetPrice = ['sample', 'invalid', 'type']
```
khancode, line 177: Can you also add a positive test case with AssetPrice using a valid string value?

khancode (changes requested):

packages/xrpl/src/models/transactions/oracleSet.ts
```diff
@@ -10,6 +10,8 @@ import {
   validateRequiredField,
 } from './common'
 
+import {HEX_REGEX} from 'ripple-binary-codec/dist/types/uint-64'
```
mvadari, line 13: Isn't there a `HEX_REGEX` somewhere in xrpl.js? I don't like the idea of importing from ripple-binary-codec like this

packages/xrpl/src/models/transactions/oracleSet.ts
```diff
@@ -146,7 +148,7 @@ export function validateOracleSet(tx: Record<string, unknown>): void {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- we are validating the type
         'AssetPrice' in priceData.PriceData &&
         // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- we are validating the type
-        !isNumber(priceData.PriceData.AssetPrice)
+        !(isNumber(priceData.PriceData.AssetPrice) || (typeof priceData.PriceData.AssetPrice === 'string' && HEX_REGEX.test(priceData.PriceData.AssetPrice)))
```
mvadari, line 151: This regex test should potentially be included in `isNumber`

packages/xrpl/src/models/transactions/oracleSet.ts
khancode: Yes, it's actually wrapped in a utility function called `isHex()` in `packages/xrpl/src/models/utils/index.ts`

packages/xrpl/test/models/oracleSet.test.ts
ckeshava: [35286b0](https://github.com/XRPLF/xrpl.js/pull/2913/commits/35286b03cf9c5e80e8622dfb532f59ce4f41204a)

packages/xrpl/src/models/transactions/oracleSet.ts
ckeshava: There are many `unsigned int` values which aren't `std::uint64_t` amongst `sfABCD` values. Such values do not accept a `str` representation. There are 37 usages of this method in `packages/xrpl/`. Apart from a case-by-case decision, I don't know of a sweeping way to disambiguate this usage.

packages/xrpl/src/models/transactions/oracleSet.ts
ckeshava: These are the two other definitions of HEX_REGEX outside of `ripple-binary-codec` package. Both of them do not fit the bill. Perhaps a suitable regex is defined with a slightly different name? 

https://github.com/XRPLF/xrpl.js/blob/0a45c58934620d8cad127a8d5652e7bc9f377836/packages/isomorphic/src/utils/shared.ts#L3

https://github.com/XRPLF/xrpl.js/blob/0a45c58934620d8cad127a8d5652e7bc9f377836/packages/xrpl/src/models/utils/index.ts#L1

ckeshava: @khancode  That is the second regex in my above message, it does not satisfy our requirements.

packages/xrpl/src/models/transactions/oracleSet.ts
mvadari: Is that just true of the library or does rippled also not accept hex for smaller int values?

packages/xrpl/src/models/transactions/oracleSet.ts
khancode: How do they not fit it? What are the requirements?

packages/xrpl/src/models/transactions/oracleSet.ts
ckeshava: Based on a perfunctory glance at the code base, rippled does not have any constructors which convert a hexa-decimal string into a `std::uint<>_t` type. All of them accept a `std::uint<>_t` type or a byte-string.

c++ code does not need a `hexadecimal string -> UINT` conversion because the `std::uint<>_t` types sufficiently represent the values. My understanding is that (non-c++) client libraries rely on strings to preserve the precision of values.

Source: https://github.com/XRPLF/rippled/blob/af018c7b0bd2d019a009e4405cfb7c3923c3a078/include/xrpl/protocol/STInteger.h#L38

achowdhry-ripple (approved):
lgtm

packages/xrpl/src/models/transactions/oracleSet.ts
mvadari: If the `isHex` helper function doesn't work here, then it should be modified. We shouldn't need 5 different ways of checking if a string is hex.

ckeshava: @khancode  The `HEX_REGEX` in this [file](https://github.com/XRPLF/xrpl.js/blob/0a45c58934620d8cad127a8d5652e7bc9f377836/packages/xrpl/src/models/utils/index.ts#L1) accepts any non-zero length string comprising of the specified characters. However, the `std::uint64_t` cannot accommodate arbitrarily large strings. This is due to the range restrictions of the data type.

@mvadari  There are multiple `HEX_REGEX` values in the code base because rippled has different uses of the hexa-decimal representation. Reconciling all the existing `HEX_REGEX` values is not possible. For instance, Currency representation uses [this regex](https://github.com/XRPLF/xrpl.js/blob/0a45c58934620d8cad127a8d5652e7bc9f377836/packages/ripple-binary-codec/src/types/currency.ts#L6)

I am happy to rename this variable as `HEX_REGEX_UINT64`, if that helps with readability.

mvadari: Why not use `isHex` and do a separate length check, then?

khancode: I agree with @mvadari . You could also update `isHex` to take in a length parameter.

packages/xrpl/test/models/oracleSet.test.ts
```diff
@@ -167,7 +167,21 @@ describe('OracleSet', function () {
...
   it(`throws w/ invalid AssetPrice of PriceDataSeries`, function () {
-    tx.PriceDataSeries[0].PriceData.AssetPrice = '1234'
+    // value cannot be parsed as hexadecimal number
+    tx.PriceDataSeries[0].PriceData.AssetPrice = 'ghij'
+    const errorMessage = 'OracleSet: invalid field AssetPrice'
+    assert.throws(() => validateOracleSet(tx), ValidationError, errorMessage)
+    assert.throws(() => validate(tx), ValidationError, errorMessage)
+  })
+
+  it(`throws w/ valid AssetPrice of PriceDataSeries`, function () {
```
khancode, line 177: typo
```suggestion
  it(`valid AssetPrice of PriceDataSeries`, function () {
```

packages/xrpl/src/models/transactions/oracleSet.ts
ckeshava: acbfd3db

packages/xrpl/src/models/transactions/oracleSet.ts
```diff
@@ -142,14 +143,25 @@ export function validateOracleSet(tx: Record<string, unknown>): void {
...
+            isHex(priceData.PriceData.AssetPrice) &&
+            priceData.PriceData.AssetPrice.length <=
+              _MAXIMUM_ASSET_PRICE_LENGTH &&
+            priceData.PriceData.AssetPrice.length >=
+              _MINIMUM_ASSET_PRICE_LENGTH)
+        )
       ) {
         throw new ValidationError('OracleSet: invalid field AssetPrice')
       }
+      /* eslint-enable @typescript-eslint/no-unsafe-member-access */
```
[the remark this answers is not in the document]
ckeshava: This helps me address typescript linter errors. Without these directives, the compiler does not understand that priceData array members have the `AssetPrice` field.

packages/xrpl/src/models/transactions/oracleSet.ts
```diff
@@ -173,4 +185,5 @@ export function validateOracleSet(tx: Record<string, unknown>): void {
     }
     return true
   })
+  /* eslint-enable max-statements, max-lines-per-function */
```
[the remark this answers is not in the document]
ckeshava: I'm re-enabling the directive after this method. I'd like this directive to be enforced after this function definition.

packages/xrpl/src/models/transactions/oracleSet.ts
```diff
@@ -142,14 +143,25 @@ export function validateOracleSet(tx: Record<string, unknown>): void {
...
+          isNumber(priceData.PriceData.AssetPrice) ||
+          (isString(priceData.PriceData.AssetPrice) &&
+            isHex(priceData.PriceData.AssetPrice) &&
+            priceData.PriceData.AssetPrice.length <=
+              _MAXIMUM_ASSET_PRICE_LENGTH &&
+            priceData.PriceData.AssetPrice.length >=
+              _MINIMUM_ASSET_PRICE_LENGTH)
+        )
       ) {
         throw new ValidationError('OracleSet: invalid field AssetPrice')
```
khancode, line 162: This conditional looks messy. It would be better to do 3 separate conditionals with their own error messages:
1. isNumber
2. isString & isHex
3. in range

ckeshava: how does this look? [e1ae16f](https://github.com/XRPLF/xrpl.js/pull/2913/commits/e1ae16fa566f9d7dc99a3d46cb94a9a695705bf4)

khancode: Looks better but the asset price length condition doesn't have to be nested.

khancode: Oh nevermind, it also has to be nested. The new change looks good now.

packages/xrpl/src/models/transactions/oracleSet.ts
```diff
@@ -142,14 +143,35 @@ export function validateOracleSet(tx: Record<string, unknown>): void {
...
 
-      if (
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- we are validating the type
-        'AssetPrice' in priceData.PriceData &&
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access -- we are validating the type
-        !isNumber(priceData.PriceData.AssetPrice)
-      ) {
-        throw new ValidationError('OracleSet: invalid field AssetPrice')
+      const _MINIMUM_ASSET_PRICE_LENGTH = 1
+      const _MAXIMUM_ASSET_PRICE_LENGTH = 16
```
khancode, line 147: Prefixing with `_` is Pythonic. Also move these constant vars outside to the top of the module.
```suggestion
      const MINIMUM_ASSET_PRICE_LENGTH = 1
      const MAXIMUM_ASSET_PRICE_LENGTH = 16
```

ckeshava: I was under the belief that private variables are prefixed with the `_` character in Python convention. We use this convention in other constant-variable definitions.

ckeshava: I want to limit the scope of these variables, hence I defined them inside this method. I don't see their usage in the other methods of this file. Why should I move it to the top of the file?

khancode: But we're not using Python, it's TypeScript. I prefer having constants defined at the top of the module for visibility. For example - https://github.com/XRPLF/xrpl.js/blob/c97bbabc9d5bdf0a888b2e82098746f24d6c691a/packages/xrpl/src/models/transactions/oracleSet.ts#L13-L14

khancode (approved):

PR #2919, opened 2025-03-04
fix: add MPTAmount support in Issue model

achowdhry-ripple:

Add MPTAmount support in the Issue object. Parallel to https://github.com/XRPLF/xrpl-py/pull/809

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes


packages/ripple-binary-codec/test/issue.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      currency: '0123456789ABCDEF0123456789ABCDEF01234567',
+      issuer: 'rG1QQv2nh2gr7RCZ1P8YYcBUKCCN633jCn',
+    }
+    const iouIssue = Issue.from(iouJson)
+    expect(iouIssue.toJSON()).toMatchObject(iouJson)
+  })
+
+  it(`test from value mpt`, () => {
+    const mptJson = {
+      // value: '100', // MPT amounts must be an integer string (no decimal point)
```
achowdhry-ripple, line 31: @khancode saw value in the python tests, but wasn't sure whether to add it here based on our discussions. Seems like it should be here, but that would require me to change the isMPT logic in Issue to something different than what we discussed

ckeshava: `Issue.from` accepts both types in the input parameter.

> `static from<T extends Issue | IssueObject>(value: T): Issue`

It should work either ways

packages/ripple-binary-codec/test/issue.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  it(`test from parser mpt`, () => {
+    const mptJson = {
+      // value: '100', // MPT amounts must be an integer string (no decimal point)
+      mpt_issuance_id: 'BAADF00DBAADF00DBAADF00DBAADF00DBAADF00DBAADF00D',
+    }
+    const mptIssue = Issue.from(mptJson)
+    const parser = new BinaryParser(mptIssue.toHex())
+    const parserIssue = Issue.fromParser(parser, 24)
+    expect(parserIssue.toJSON()).toMatchObject(mptJson)
+  })
```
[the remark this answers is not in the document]
ckeshava: The `hint` parameter is not used in the `Issue.from_parser` method. Specifying a hint won't affect the output of the function. I suspect it has been retained for inheritance purposes.

ckeshava (approved):

packages/ripple-binary-codec/src/types/issue.ts
```diff
@@ -39,20 +50,34 @@ class Issue extends SerializedType {
    *
    * @param value An Amount, object representing an IOU, or a string
```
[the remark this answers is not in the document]
khancode: "an IOU, MPTAmount or string"

packages/ripple-binary-codec/src/types/issue.ts
```diff
@@ -62,9 +87,16 @@ class Issue extends SerializedType {
...
    *
    * @param parser BinaryParser to read the Amount from
-   * @returns An Amount object
+   * @param hint The number of bytes to consume from the parser.
+   * For an MPT amount, pass 24 (the fixed length for Hash192).
+   *
+   * @returns An Issue object
    */
-  static fromParser(parser: BinaryParser): Issue {
+  static fromParser(parser: BinaryParser, hint?: number): Issue {
```
khancode, line 95: In xrpl-js, the hint parameter isn't used; it's specific to xrpl-py. We may have to find another way to detect 'mpt_issuance_id'.

packages/ripple-binary-codec/test/issue.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  it(`test from value issued currency`, () => {
+    const iouJson = {
+      currency: 'USD',
+      issuer: 'rG1QQv2nh2gr7RCZ1P8YYcBUKCCN633jCn',
+    }
+    const iouIssue = Issue.from(iouJson)
+    expect(iouIssue.toJSON()).toMatchObject(iouJson)
+  })
+
+  it(`test from value nonstandard currency`, () => {
```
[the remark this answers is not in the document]
achowdhry-ripple: added in commit

packages/ripple-binary-codec/src/types/issue.ts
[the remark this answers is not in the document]
achowdhry-ripple: added in commit

packages/ripple-binary-codec/src/types/issue.ts
```diff
@@ -4,24 +4,35 @@ import { BinaryParser } from '../serdes/binary-parser'
...
  * Interface for JSON objects that represent amounts
  */
-interface IssueObject extends JsonObject {
-  currency: string
-  issuer?: string
-}
+type IssueObject = XRPIssue | IOUIssue | MPTIssue
 
 /**
  * Type guard for AmountObject
```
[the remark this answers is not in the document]
achowdhry-ripple: added in commit

packages/ripple-binary-codec/src/types/issue.ts
achowdhry-ripple: It is inherited from the parent interface. It was not used previously since both options required 20 bytes, but I don't see why it would be a problem to use it here for mpt (it is optional so does not break old code). Am open to suggestions if there is a cleaner way to check if something is MPT or not without passing in an additional parameter to fromParser

packages/ripple-binary-codec/test/issue.test.ts
achowdhry-ripple: @khancode when you say normally called do you mean the implementations for IOU and XRP? there were no tests before so hard to tell but based off of the isIssueObject check it seems like those have no value param. However, in your python implementation there was no value parameter for IOU and XRP either, but you added it for MPT. I'm assuming this should be the same in JS so I should add value as a param

packages/ripple-binary-codec/test/issue.test.ts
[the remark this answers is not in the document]
khancode: @achowdhry-ripple thanks for checking, LGTM.

packages/ripple-binary-codec/test/issue.test.ts
khancode: Yes I meant the implementations. Ok cool, if it still works when adding value then LGTM.

khancode: Looks like `value` is excluded by looking at the console.log inside of `Issue.from` so this looks good 👍

khancode (approved):
LGTM

PR #2920, opened 2025-03-06
chore: remove more mentions of elliptic

mvadari:

Title says it all.

### Context of Change

https://slowmist.medium.com/private-key-leakage-in-ecdsa-signatures-analysis-of-malformed-input-vulnerability-in-the-elliptic-24f73c05cac1

There were a few mentions left after #2273 

### Type of Change

- [x] Documentation Updates

## Test Plan

CI passes.

### DIFF
--- a/packages/ripple-keypairs/README.md
+++ b/packages/ripple-keypairs/README.md
@@ -1,7 +1,7 @@
 # ripple-keypairs [![NPM](https://img.shields.io/npm/v/ripple-keypairs.svg)](https://npmjs.org/package/ripple-keypairs) [![Build Status](https://img.shields.io/travis/ripple/ripple-keypairs/master.svg)](https://travis-ci.org/ripple/ripple-keypairs) ![Codecov](https://img.shields.io/codecov/c/github/ripple/ripple-keypairs)
 
 An implementation of XRP Ledger keypairs & wallet generation using
-[elliptic](https://github.com/indutny/elliptic) which supports rfc6979 and
+[noble](https://paulmillr.com/noble) which supports rfc6979 and
 eddsa deterministic signatures.
 
 [![NPM](https://nodei.co/npm/ripple-keypairs.png)](https://www.npmjs.org/package/ripple-keypairs)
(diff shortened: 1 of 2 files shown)


sublimator:
Your welcome

achowdhry-ripple (approved):

ckeshava (changes requested):
This line needs to b e removed: https://github.com/XRPLF/xrpl.js/blob/c68902dbab80e5fb3a6ec528d57c321b550f2d59/packages/ripple-keypairs/src/signing-schemes/secp256k1/utils.ts#L26

mvadari:
> This line needs to be removed:
> 
> https://github.com/XRPLF/xrpl.js/blob/c68902dbab80e5fb3a6ec528d57c321b550f2d59/packages/ripple-keypairs/src/signing-schemes/secp256k1/utils.ts#L26

I left that in on purpose - it's just a helpful comment

ckeshava:
> > This line needs to be removed:
> > https://github.com/XRPLF/xrpl.js/blob/c68902dbab80e5fb3a6ec528d57c321b550f2d59/packages/ripple-keypairs/src/signing-schemes/secp256k1/utils.ts#L26
> 
> I left that in on purpose - it's just a helpful comment

We  are not using that library anymire. Moreover, that line is pointing to raw-bytes which are not relevant to the context at all.

I don't see how that is helpful.

ckeshava (approved):

mvadari:
> > > This line needs to be removed:
> > > https://github.com/XRPLF/xrpl.js/blob/c68902dbab80e5fb3a6ec528d57c321b550f2d59/packages/ripple-keypairs/src/signing-schemes/secp256k1/utils.ts#L26
> > 
> > 
> > I left that in on purpose - it's just a helpful comment
> 
> We are not using that library anymire. Moreover, that line is pointing to raw-bytes which are not relevant to the context at all.
> 
> I don't see how that is helpful.

I don't either, but I figured the comment served a purpose and I didn't look into it enough to matter. It definitely doesn't matter for the purpose of this PR, which is just to ensure people aren't confused about what packages we are and aren't using (a comment doesn't affect that).

PR #2923, opened 2025-03-09
fix: remove index signature from base request

interc0der:

- Removed the index signature (e.g., `[key: string]: any`) from the base `request` method's TypeScript interface in the XRPL.js library.
- Updated all downstream instances where the index signature was previously used for accessing properties dynamically (e.g., replaced with explicit property definitions or more specific types).
- This change improves type safety and enhances compatibility with TypeScript utility types like `Omit<>` and `Partial<>`, making it easier for third-party developers to extend or manipulate the `request` type in their own projects.

## Context of Change

- Previously, the `request` method likely used an index signature to allow arbitrary key-value pairs, which provided flexibility but sacrificed type safety and discoverability. For example, a definition like `interface Request { [key: string]: any }` would allow any string key with any value, making it harder for developers to catch errors at compile time and limiting the utility of TypeScript’s tooling (e.g., autocompletion, refactoring).
- This change is a **refactor** with bug-fix-like benefits: it eliminates the overly permissive index signature, replacing it with a more explicit structure (e.g., specific fields like `method: string`, `params?: object`). This aligns with TypeScript best practices, as index signatures should only be used when the keys are truly dynamic and unknown ahead of time.
- The motivation stems from enhancing portability and usability for third-party developers. With a precise interface, developers can now leverage utility types (e.g., `Omit<Request, 'params'>` or `Partial<Request>`) to customize the `request` type without running into issues caused by the loose index signature.
- No specific bug is being fixed (i.e., no prior crash or runtime error), but this prevents potential type-related issues in downstream usage. The change likely emerged from feedback or observed friction in third-party integrations with the library.
- Alternatives considered might have included keeping the index signature with a more constrained value type (e.g., `[key: string]: string | number`), but removing it entirely was chosen to enforce stricter typing and better align with the library’s intended usage patterns.
- No design document is referenced, but this aligns with broader TypeScript ecosystem trends toward stricter, more explicit type definitions.

## Type of Change

  *(Prevents potential type misuse, though not tied to a specific runtime bug.)*  
  *(Assumed non-breaking since existing valid usage should still work with explicit types.)*  
  *(Primary intent is to improve type structure, making this a refactor as well.)*  
- [ ] Documentation Updates  
- [ ] Release  

## Did you update HISTORY.md?

- [ ] Yes  
  *(The change is internal to the type system and transparent to end users interacting with the library’s runtime behavior. It mainly benefits developers working with the typings, so no user-facing HISTORY.md update is needed.)*

## Test Plan

- No specific unit tests were added for this change, as it is a type-level refactor that does not alter runtime behavior. The TypeScript compiler itself enforces the correctness of the updated typings during the build process.
- Validation was performed by publishing a modified build to npm and testing it in a production-like environment. The updated package is available here: [https://www.npmjs.com/package/@interc0der/xrpl](https://www.npmjs.com/package/@interc0der/xrpl).
- Existing tests (if any) should still pass, assuming they were written to align with the `request` method’s intended usage. If the library has a CI pipeline, it likely caught any downstream type errors introduced by this change.


mvadari:
I'm a bit confused as to what problem this is solving... Can you provide an example?

interc0der:
Here is a quick and dirty example: https://github.com/interc0der/xrpl-example

```ts
import type * as xrpl from 'xrpl'

// Type is not restructured correctly on the AccountChannelsParams
// due to the signature indexing on the base request method
export type AccountChannelsParams = Omit<xrpl.AccountChannelsRequest, 'command'>;
```

ckeshava:
@interc0der  I'm able to run the below snippet without the typescript compiler throwing any warnings (or) errors. It looks like type manipulation is possible in the current version of the code. What exactly is missing?

```
import { AccountChannelsRequest } from '../../src/models/methods/accountChannels'

type AccountChannelsMinusCommand = Omit<AccountChannelsRequest, 'command'>
type AccountChannelsOnlyCommand = Pick<AccountChannelsRequest, 'command'>

const accountChannels: AccountChannelsMinusCommand = {
  account: 'r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59',
  limit: 10,
  ledger_index: 'validated',
}

const command: AccountChannelsOnlyCommand = {
  command: 'account_channels',
}

console.log(accountChannels)
console.log(typeof accountChannels)
console.log(command)
console.log(typeof command)
```

I get the below output:
image

interc0der:
Pick<> type utility is fine and will work despite the signature index.
Omit<> type utility is the one that is the problem. While this is not breaking the ts compiler, you are losing type safety.

Screenshot 2025-03-12 at 4 05 41 PM

ckeshava:
> Omit<> type utility is the one that is the problem. While this is not breaking the ts compiler, you are losing type safety.

Okay, I found this snippet which is problematic.
```
import { AccountChannelsRequest } from '../../src/models/methods/accountChannels'

type AccountChannelsMinusCommand = Omit<AccountChannelsRequest, 'command'>

const accountChannels: AccountChannelsMinusCommand = {
  account: 'r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59',
  limit: 10,
  ledger_index: 'validated',
  // compiler does not throw an error here
  command: 'account_channels',
}

console.log(accountChannels)
```

interc0der:
@ckeshava Unit testing for type utilities has been add within the models test directory. We are checking a few models. 

It should be noted that Chai / Jest unit testing is not designed to pick up non-runtime issues.

packages/xrpl/src/models/methods/subscribe.ts
```diff
@@ -454,9 +454,20 @@ export interface PathFindStream extends BaseStream {
...
 }
 
+/**
+ * The manifest stream sends messages when a new manifest has been processed or updated.
+ * The message contains the public key, either the master public key or ephemeral public key.
+ *
+ * @category Streams
+ */
+export interface ManifestStream extends BaseStream, ManifestRequest {
+  type: 'manifestReceived'
```
ckeshava, line 464: I could not find the documentation for the `ManifestStream` in this [docs](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/subscription-methods/subscribe) page.

Where did you come across this `manifestRecieved` keyword? Are you using this in your application?

packages/xrpl/src/models/methods/subscribe.ts
```diff
@@ -454,9 +454,20 @@ export interface PathFindStream extends BaseStream {
...
+}
+
 /**
  * @category Streams
  */
+// TODO: Add stream type for manifestReceived
```
ckeshava, line 470: This comment can be removed

packages/xrpl/test/models/typeUtility.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    assert.notProperty(empty, 'api_version')
+
+    // Verify optional nature at compile-time
+    // This should compile without error
+    const minimal: OptionalRequest = {}
+    assert.isUndefined(minimal.command)
+  })
+})
+
+describe('Stream', function () {
```
ckeshava, line 142: Since there is no update to the `Stream` type, I don't think these tests are needed. Am I missing something?

I understand that `Stream` inherits from `BaseRequest`, however there are many classes which inherit from `BaseRequest`.

ckeshava (changes requested):
Some of the tests use the `subcommand` key in the Request input. I suspect the original design wanted to accomodate these liberties. Here is an example: https://github.com/XRPLF/xrpl.js/actions/runs/13847785489/job/38914627395?pr=2923#step:9:72

Would you be willing to fix these discrepancies?

Thanks for bringing this to our attention. How did you come across this type-check requirement? What are you using xrpl.js for?

packages/xrpl/src/client/index.ts
```diff
@@ -343,7 +343,7 @@ class Client extends EventEmitter<EventTypes> {
     const request = {
       ...req,
       account:
-        typeof req.account === 'string'
+        'account' in req && typeof req.account === 'string'
```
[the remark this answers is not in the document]
ckeshava: oh, I misread the variables. You are correct.

packages/xrpl/src/models/methods/subscribe.ts
interc0der: This stream was already in the codebase, but was not typed. We can do a deep drive to figure out why this was here in the first place, but I figure we should at least type something here and avoid the usage of index signatures.

We can remove this addition if its going to create a fuss.

packages/xrpl/test/models/typeUtility.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    assert.property(omitted, 'api_version')
+    assert.typeOf(omitted.api_version, 'number')
+
+    type HasCommand = 'command' extends keyof BaseRequest ? true : false
+    type HasNoCommand = 'command' extends keyof RequestMinusCommand
+      ? false
+      : true
+
+    assertType<true>({} as HasCommand)
+    assertType<true>({} as HasNoCommand)
```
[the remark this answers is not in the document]
interc0der: Well, without an additional dependency, its kind tough the check correct type assertions that are not caught during runtime. Index signature removes type-checking of foreign keys and is not caught during runtime. Trying to write unit tests that are runtime-based is quite difficult for this scenerio.

packages/xrpl/test/models/typeUtility.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    assert.notProperty(empty, 'api_version')
+
+    // Verify optional nature at compile-time
+    // This should compile without error
+    const minimal: OptionalRequest = {}
+    assert.isUndefined(minimal.command)
+  })
+})
+
+describe('AccountInfoRequest', function () {
```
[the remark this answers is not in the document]
interc0der: The intention to to prevent signature index globally. Yes the index signature should not be on the BaseRequest class, but it should also not be on anything extended of the of the BaseRequest.

packages/xrpl/test/models/typeUtility.test.ts
interc0der: Okay I am just going to remove it the testing for everything, except the BaseRequest.

interc0der:
The integration testing leverages index signatures in a few cases to iterate through typed keys and assert types. This can be easily fixed by add the type signatures where required, locally. 


---

For anyone that is facing a similar issue (the rare few), here is a workaround type hack to extract the index signature from any type interface. 

```
type RemoveIndexSignature<T, TPropKey = PropertyKey> = {
  [K in keyof T as TPropKey extends K ? never : K extends TPropKey ? K : never]: T[K];
};
```

You can then use the Omit<> and the type structure is as expected.
```
export type AccountChannelsParams = Omit<
  RemoveIndexSignature<methods.AccountChannelsRequest>,
  'command'
>;
```

packages/xrpl/src/models/methods/subscribe.ts
ckeshava: I found this [reference](https://github.com/XRPLF/xrpl.js/blob/7fe27f3d9c8cdb92a7f9ad97daa9cf7bb0344637/packages/xrpl/src/models/methods/subscribe.ts#L477) to `manifestReceived` event trigger. Furthermore, that file uses the `ManifestRequest` type. Do you still see the need for `ManifestStream` type ?

ckeshava: ignore the previous comment. Yes, I see the need for the `ManifestStream` type.

ckeshava: I found a relevant [request-for-feature](https://github.com/XRPLF/xrpl-dev-portal/issues/958) issue in the XRPL docs repository.

packages/xrpl/test/integration/requests/serverState.test.ts
```diff
@@ -135,10 +135,12 @@ describe('server_state', function () {
         key,
       ) {
         assert.equal(
+          // @ts-expect-error -- non-iterative type
```
ckeshava, line 138: Why do we need all these declarations?

packages/xrpl/test/models/typeUtility.test.ts
```diff
@@ -0,0 +1,74 @@
+import { assert } from 'chai'
+
+import type { BaseRequest } from '../../src/models'
+
+// Helper type assertion function (only for TS, no runtime impact)
+function assertType<T>(value: T): void {
```
ckeshava, line 6: Since `value` is unused, we could perhaps use `_` ?

packages/xrpl/test/integration/requests/serverInfo.test.ts
```diff
@@ -138,29 +144,24 @@ describe('server_info (rippled)', function () {
...
-        omit(response.result.info.validated_ledger, 'hash'),
-      )) {
-        assert.equal(
-          typeof response.result.info.validated_ledger?.[key],
-          'number',
-        )
+      const validated_ledger = response.result.info.validated_ledger
+      assert.equal(typeof validated_ledger?.hash, 'string')
+      for (const key of Object.keys(omit(validated_ledger, 'hash'))) {
+        // @ts-expect-error -- non-iterative type
```
ckeshava, line 163: This type of @ts-expect-error statements can be removed.

packages/xrpl/test/integration/requests/serverInfo.test.ts
```diff
@@ -138,29 +144,24 @@ describe('server_info (rippled)', function () {
...
-        assert.equal(
-          typeof response.result.info.state_accounting[key].duration_us,
-          'string',
-        )
-        assert.equal(
-          typeof response.result.info.state_accounting[key].transitions,
-          'string',
-        )
+      const state_accounting: StateAccountingFinal & {
+        [index: string]: unknown
```
ckeshava, line 150: Why do we need this catch-all type here? I have found that we can iterate over the keys without using this additional key-value pair.

packages/xrpl/test/integration/requests/ledger.test.ts
```diff
@@ -55,6 +55,8 @@ describe('ledger', function () {
       assert.typeOf(ledgerResponse.result.ledger_index, 'number')
 
       const ledger = ledgerResponse.result.ledger as Ledger & {
+        // Add index signature to request to iterate through keys and asset types
+        [index: string]: unknown
```
ckeshava, line 59: Similar to my comments in other test files, I don't see the need for this additional kv pair

ckeshava (commented):
Hello! thanks for your work. I have a few comments regarding the linter and one test case.

I have incorporated my suggestions into this commit: https://github.com/XRPLF/xrpl.js/commit/193490bf87d25623afe2a4c71eb95709d2155224. Feel free to use it.

ckeshava:
@interc0der  Clarification: Are you trying to remove the `[x: string]: unknown` key-value pair from the `BaseRequest` ? I'm looking at the work-around that you suggested here:

> 
```
type RemoveIndexSignature<T, TPropKey = PropertyKey> = {
  [K in keyof T as TPropKey extends K ? never : K extends TPropKey ? K : never]: T[K];
};
```

ckeshava:
@interc0der  Hello, did you get a chance to go through my comments?

ckeshava:
Hello @interc0der , can you address my pending comments above?

ckeshava:
I'd like to close this PR. 

Please look at my comments and let me know if you would like to address/refute them. The PR can be re-opened at an appropriate time.

Thank you.

PR #2927, opened 2025-03-11
fix(rbc): throw error instead of throwing away unknown fields

mvadari:

This PR fixes an issue in ripple-binary-codec to throw an error instead of silently throwing away unknown fields and their values when serializing transactions.

It only errors on `CapitalizedFields` and still silently ignores `lowercaseFields`, as the latter are likely from copy-pasting `tx` outputs or something, rather than an actual error (since all rippled fields must be Capitalized).

### Context of Change

I came across this issue when working on a separate PR for a new feature, where my tests would fail for unknown reasons. I had to go into the rippled logs and decode the transaction rippled was receiving to figure out what was happening.

### Type of Change

- [x] Yes

## Test Plan

Added a test to make sure this works properly, and fixed existing tests that had the previous assumption.

### DIFF
--- a/packages/ripple-binary-codec/src/types/st-object.ts
+++ b/packages/ripple-binary-codec/src/types/st-object.ts
@@ -116,14 +116,25 @@ class STObject extends SerializedType {
       return Object.assign(acc, handled ?? { [key]: val })
     }, {})
 
-    let sorted = Object.keys(xAddressDecoded)
-      .map((f: string): FieldInstance => definitions.field[f] as FieldInstance)
-      .filter(
-        (f: FieldInstance): boolean =>
-          f !== undefined &&
-          xAddressDecoded[f.name] !== undefined &&
-          f.isSerialized,
+    function isValidFieldInstance(
+      f: FieldInstance | undefined,
+    ): f is FieldInstance {
+      return (
+        f !== undefined &&
+        xAddressDecoded[f.name] !== undefined &&
+        f.isSerialized
       )
+    }
+
+    let sorted = Object.keys(xAddressDecoded)
+      .map((f: string): FieldInstance | undefined => {
+        if (!(f in definitions.field)) {
+          if (f[0] === f[0].toLowerCase()) return undefined
+          throw new Error(`Field ${f} is not defined in the definitions`)
+        }
+        return definitions.field[f] as FieldInstance
+      })
+      .filter(isValidFieldInstance)
       .sort((a, b) => {
         return a.ordinal - b.ordinal
       })
--- a/packages/ripple-binary-codec/test/tx-encode-decode.test.ts
+++ b/packages/ripple-binary-codec/test/tx-encode-decode.test.ts
@@ -116,4 +116,13 @@ describe('encoding and decoding tx_json', function () {
       encode(my_tx)
     }).toThrow()
   })
+
+  it('throws when there is an unknown field', function () {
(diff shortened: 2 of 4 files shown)


packages/ripple-binary-codec/src/types/st-object.ts
```diff
@@ -116,14 +116,25 @@ class STObject extends SerializedType {
...
     }, {})
 
-    let sorted = Object.keys(xAddressDecoded)
-      .map((f: string): FieldInstance => definitions.field[f] as FieldInstance)
-      .filter(
-        (f: FieldInstance): boolean =>
-          f !== undefined &&
-          xAddressDecoded[f.name] !== undefined &&
-          f.isSerialized,
+    function isValidFieldInstance(
```
mvadari, line 119: This function is needed for typing

packages/ripple-binary-codec/src/types/st-object.ts
```diff
@@ -116,14 +116,25 @@ class STObject extends SerializedType {
...
+        f !== undefined &&
+        xAddressDecoded[f.name] !== undefined &&
+        f.isSerialized
       )
+    }
+
+    let sorted = Object.keys(xAddressDecoded)
+      .map((f: string): FieldInstance | undefined => {
+        if (!(f in definitions.field)) {
+          if (f[0] === f[0].toLowerCase()) return undefined
```
ckeshava, line 132: I don't understand this condition -- Can you point me to the documentation which explains the capitalization of the first character?

I observe examples where the first character of the field-name is both upper and lower case. Ex: https://github.com/XRPLF/xrpl.js/blob/effe7571f38827a7a44145e7163e1298ad2a262c/packages/ripple-binary-codec/src/enums/definitions.json#L44

Is there any explanation here? https://xrpl.org/docs/references/protocol/binary-format

packages/ripple-binary-codec/test/tx-encode-decode.test.ts
```diff
@@ -116,4 +116,13 @@ describe('encoding and decoding tx_json', function () {
       encode(my_tx)
     }).toThrow()
   })
+
+  it('throws when there is an unknown field', function () {
+    const my_tx = Object.assign({}, tx_json, {
+      BadField: 1,
```
ckeshava, line 122: If this field was titled `badField` (beginning with small letter), then the code does not throw. I understand the implementation of the code. But what is rationale for this behavior?

packages/ripple-binary-codec/src/types/st-object.ts
mvadari: Lower-case fields are synthetic fields that aren't serialized, like `hash` or `date`. They're included in `tx` outputs, so if you're re-feeding that into the binary codec, you want those to be discarded instead of erroring.

packages/ripple-binary-codec/test/tx-encode-decode.test.ts
mvadari: SFields are always capitalized. Synthetic fields (non-serialized fields that are returned in RPC outputs) are lower-case.

packages/ripple-binary-codec/src/types/st-object.ts
ckeshava: okay thanks. This is useful information, it will be helpful if its mentioned in the docs. Perhaps I missed it?

ckeshava (approved):

PR #2946, opened 2025-03-24
docs: add details for experimental releases

khancode:

Add details for experimental releases.

### Type of Change

- [x] Documentation Updates
- [ ] Release

- [ ] Yes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -255,6 +255,8 @@ This should almost always be done using the [`xrpl-codec-gen`](https://github.co
    NOW YOU HAVE PUBLISHED! But you're not done; we have to notify people!
 
 1. Run `git tag <tagname> -m <tagname>`, where `<tagname>` is the new package and version (e.g. `xrpl@2.1.1`), for each version released.
+
+   For experimental releases, the git tagname should include its own version number along with feature name. Here's an example for [MPT beta release](https://github.com/XRPLF/xrpl.js/releases/tag/xrpl%401.0.1-mpt-beta).
 1. Run `git push --follow-tags`, to push the tags to Github.
 1. On GitHub, click the "Releases" link on the right-hand side of the page.
 


Patel-Raj (approved):

CONTRIBUTING.md
```diff
@@ -255,6 +255,8 @@ This should almost always be done using the [`xrpl-codec-gen`](https://github.co
    NOW YOU HAVE PUBLISHED! But you're not done; we have to notify people!
 
 1. Run `git tag <tagname> -m <tagname>`, where `<tagname>` is the new package and version (e.g. `xrpl@2.1.1`), for each version released.
+
+   For experimental releases, the git tagname should include its own version number along with feature name. Here's an example for [MPT beta release](https://github.com/XRPLF/xrpl.js/releases/tag/xrpl%401.0.1-mpt-beta).
```
ckeshava, line 259: I'm trying to understand the differences between a usual release and an experimental release. Why was the MPT feature not release with the normal semver nomenclature? Why do we need to do experimental releases at all?

Should we follow the `rc-1, 2, ...` system of "release candidate" names instead?

khancode: It prevents disruption of the versioning of a normal release. For example, normal version is 2.4.0, we add an experimental feature with the version 2.5.0-beta. However, if we decide to do a normal release without the experimental feature being 2.5.0, then it wouldn't be accurate. That's why it's better for the feature to use its own version.

ckeshava: I'm still confused. After the release of the experimental feature (titled, say `2.5.0-mpt`), what would be the next version? How do we indicate the continuity of the versioning system? Suppose the next release is `2.5.1`, users do not know if `mpt` feature is included in the `2.5.1` release.

Why do we need to consider experimental features? A feature will either be included (if it gains 80% UNL majority vote on the mainnet for 2 weeks) or not. Since this is a binary decision, why should we account for an "experimental feature" ? In this example, the `MPT` (or) `AMM` feature would definitely be enabled on the XRPL Mainnet within a pre-determined time frame.

khancode: The 2.5.0 is a counter-example. An experimental feature would have its own version like the MPT example that's in this PR.

There have been cases where experimental features (AMM, Sidechains) would take awhile to develop such that we would do incremental beta releases for them. They usually would have their own devnets. In the meantime, we would still do the normal releases with other features/fixes. Therefore, having separate versions fixes this issue.

pdp2121: I think it's fine to have experimental feature with its own version in case the feature will not be on the next release

pdp2121, line 259: nit: maybe we should mention that the feature name should not be beta or rc to avoid confusion. I will leave it up to you.

pdp2121 (approved):

achowdhry-ripple (approved):

PR #2950, opened 2025-03-28
refactor: use `NetworkID` in faucet calculations

### LINKED ISSUE #2701: Use Network ID to determine the correct faucet to use
Code reference: https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/Wallet/fundWallet.ts

Currently, the faucet generation code guesses the correct faucet to use based on substrings in the URL. This isn't a very robust system. It would be better (and more robust) to use the Network ID.

mvadari:

This PR refactors the faucet code to use the `NetworkID` to determine which faucet to use.

### Context of Change

Resolves #2701

Previously, the code was making assumptions about what the URL would be, which isn't very robust.

### Type of Change

- [x] Yes

## Test Plan

CI still passes.


packages/xrpl/test/wallet/fundWallet.test.ts
```diff
@@ -45,20 +45,20 @@ describe('Get Faucet host ', function () {
...
-    const expectedFaucetPath = FaucetNetworkPaths[FaucetNetwork.Devnet]
+    const expectedFaucetPath = faucetNetworkPaths[FaucetNetwork.Devnet]
     // @ts-expect-error Intentionally modifying private data for test
     testContext.client.connection.url = FaucetNetwork.Devnet
 
     assert.strictEqual(
-      getDefaultFaucetPath(getFaucetHost(testContext.client)),
+      getFaucetPath(getFaucetHost(testContext.client)),
       expectedFaucetPath,
     )
```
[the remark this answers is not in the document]
mvadari: Fixed

packages/xrpl/src/Wallet/defaultFaucets.ts
```diff
@@ -29,21 +34,20 @@ export const FaucetNetworkPaths: Record<string, string> = {
...
 export function getFaucetHost(client: Client): FaucetNetwork | undefined {
-  const connectionUrl = client.url
-
-  // 'altnet' for Ripple Testnet server and 'testnet' for XRPL Labs Testnet server
-  if (connectionUrl.includes('altnet') || connectionUrl.includes('testnet')) {
-    return FaucetNetwork.Testnet
-  }
-
-  if (connectionUrl.includes('sidechain-net2')) {
+  console.log(client.networkID)
```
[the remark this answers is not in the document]
mvadari: fixed

packages/xrpl/src/Wallet/defaultFaucets.ts
```diff
@@ -29,21 +34,19 @@ export const FaucetNetworkPaths: Record<string, string> = {
...
-  // 'altnet' for Ripple Testnet server and 'testnet' for XRPL Labs Testnet server
-  if (connectionUrl.includes('altnet') || connectionUrl.includes('testnet')) {
-    return FaucetNetwork.Testnet
-  }
-
-  if (connectionUrl.includes('sidechain-net2')) {
+  if (client.networkID == null) {
     throw new XRPLFaucetError(
-      'Cannot fund an account on an issuing chain. Accounts must be created via the bridge.',
+      'Cannot create faucet URL without networkID or the faucet_host information',
```
ckeshava, line 39:
```suggestion
      'Cannot create faucet URL without networkID',
```

packages/xrpl/src/Wallet/defaultFaucets.ts
```diff
@@ -29,21 +34,19 @@ export const FaucetNetworkPaths: Record<string, string> = {
  * @throws When the client url is not on altnet or devnet.
```
ckeshava, line 34:
```suggestion
 * @throws When the client.networkID is not 1 (Testnet) or 2 (Devnet).
```

packages/xrpl/src/Wallet/defaultFaucets.ts
```diff
@@ -16,11 +16,16 @@ export enum FaucetNetwork {
   Devnet = 'faucet.devnet.rippletest.net',
 }
 
-export const FaucetNetworkPaths: Record<string, string> = {
+export const faucetNetworkPaths: Record<string, string> = {
   [FaucetNetwork.Testnet]: '/accounts',
   [FaucetNetwork.Devnet]: '/accounts',
 }
```
ckeshava, line 22: Can we delete this map? All the faucets that I'm aware of use the `/accounts` endpoint. We can introduce this map if there is a change to that policy.

This is also an opportunity to simplify `getFaucetPath` method below.

packages/xrpl/src/Wallet/defaultFaucets.ts
```diff
@@ -29,21 +34,19 @@ export const FaucetNetworkPaths: Record<string, string> = {
...
  */
 export function getFaucetHost(client: Client): FaucetNetwork | undefined {
-  const connectionUrl = client.url
-
-  // 'altnet' for Ripple Testnet server and 'testnet' for XRPL Labs Testnet server
-  if (connectionUrl.includes('altnet') || connectionUrl.includes('testnet')) {
-    return FaucetNetwork.Testnet
-  }
-
-  if (connectionUrl.includes('sidechain-net2')) {
```
ckeshava, line 39: We need to mention that support for `sidechain-net2` is being deprecated in the ReleaseNotes. Has this sidechain been down for long enough time?

packages/xrpl/src/Wallet/fundWallet.ts
```diff
@@ -190,31 +186,24 @@ async function processSuccessfulResponse(
...
-    if (err instanceof Error) {
-      throw new XRPLFaucetError(err.message)
-    }
-    throw err
   }
+  throw new XRPLFaucetError(
+    `Unable to fund address with faucet after waiting ${
+      INTERVAL_SECONDS * MAX_ATTEMPTS
+    } seconds`,
+  )
```
ckeshava, line 206: This error message is not relevant for the `processSuccessfulResponse` method. I do not see any retry-mechanism (or) back-off algorithm used here.

However, it is relevant for `getUpdatedBalance` method. These two error messages appear to have been swapped.

packages/xrpl/src/Wallet/defaultFaucets.ts
mvadari: Why do you want to remove the extra clause? If someone provides the `faucet_host`, the function doesn't need the network ID.

packages/xrpl/src/Wallet/defaultFaucets.ts
mvadari: Historically, some faucets have been set up differently with different endpoints. All of Ripple's faucets may use the same setup, but faucets other people set up may not have the same endpoint.

packages/xrpl/src/Wallet/fundWallet.ts
mvadari: No, the errors are correct. The faucet can also be used to add funds to an existing wallet, in which case the error should fall through to here, not throw in `getUpdatedBalance`.

packages/xrpl/src/Wallet/defaultFaucets.ts
ckeshava: How will `faucet_host` information help? This method will throw an error if `client.networkID == null`

packages/xrpl/src/Wallet/fundWallet.ts
ckeshava: Can you point me to the code where we do the waiting (or) retry mechanism? Most of the usages of the faucet have been to generate a new wallet.

Even while funding an existing wallet, the operation is tried exactly once. I don't see the retry-operation.

packages/xrpl/src/Wallet/defaultFaucets.ts
mvadari: True, but the error is more reasonable from a user perspective, who is never directly calling `getFaucetHost` but will only ever call `fundWallet`

packages/xrpl/src/Wallet/fundWallet.ts
mvadari: `setInterval` in `getUpdatedBalance`

packages/xrpl/src/Wallet/defaultFaucets.ts
ckeshava: I don't understand the relevance of `faucet_host` in the error message. Suppose a user invokes fundWallet method and then get back this message: `Cannot create faucet URL without networkID or the faucet_host information`.

They will need to go through XRPL documentation about `faucet_host`.

I don't feel too strongly about it. However, the error message is not intuitive.

packages/xrpl/src/Wallet/fundWallet.ts
ckeshava: Yes, that's what I mentioned in my [first comment](https://github.com/XRPLF/xrpl.js/pull/2950#discussion_r2029453153). The `getUpdatedBalance` method is the correct method to place this error message, not `processSuccessfulResponse` method.

packages/xrpl/src/Wallet/defaultFaucets.ts
mvadari: `faucet_host` is a parameter in `fundWallet`...?

packages/xrpl/src/Wallet/fundWallet.ts
mvadari: `getUpdatedBalance` already has its own error message.

packages/xrpl/src/Wallet/defaultFaucets.ts
ckeshava: Sure, but `client.networkID` is the deciding parameter in the success/error of this method. I'm a bit thrown off by the mention of `faucet_host`, that is all.

ckeshava (approved):

packages/xrpl/src/Wallet/defaultFaucets.ts
mvadari: Have you read through all the options on `fundWallet`? If you're using a custom devnet and have set up your own faucet for that, this function won't work unless you specify `faucetHost` in the params.

ckeshava: > Have you read through all the options on fundWallet? 

I had not.

achowdhry-ripple (approved):

PR #2895, opened 2025-02-12
refactor: clean up transaction model validations

mvadari:
## High Level Overview of Change

This PR cleans up the validation functions for transactions, to use newer helper functions. It also adds some helper functions for easier model validation test writing.

This process uncovered a few bugs, which are fixed.

The integration tests have not been touched at all, which shows that nothing major was changed.

Note for reviewers: the vast majority of the changes were made with regex find-and-replace, so while there are a lot of changes, most of them (especially in the test files) are pretty formulaic.

### Context of Change

The `validateRequiredField` and `validateOptionalField` helper functions were added somewhat recently, and the older models don't use it.

### Type of Change

- [x] Refactor (non-breaking change that only restructures code)

### Did you update HISTORY.md?

- [x] Yes

## Test Plan

Fixed tests. CI passes.

PR #2965, opened 2025-04-16
refactor: refactor model tests to use a helper function

mvadari:

This PR refactors the transaction model tests to be more consistent via a couple of helper functions, and fixes a couple of minor bugs along the way.

While the diff is rather large, most of the changes are fairly systematic.

### Context of Change

Pulled out of #2895 to make that PR simpler

### Type of Change

- [x] Yes

## Test Plan

CI passes.


packages/xrpl/test/models/depositPreauth.test.ts
```diff
@@ -56,104 +57,56 @@ describe('DepositPreauth', function () {
...
+      depositPreauth,
       "DepositPreauth: Account can't unauthorize its own address",
     )
-    assert.throws(
-      () => validate(depositPreauth),
-      ValidationError,
+    assertInvalid(
+      depositPreauth,
       "DepositPreauth: Account can't unauthorize its own address",
     )
```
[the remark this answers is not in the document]
mvadari: fixed

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
```diff
@@ -69,7 +69,7 @@ export function validateMPTokenIssuanceSet(tx: Record<string, unknown>): void {
   validateOptionalField(tx, 'Holder', isAccount)
 
   // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Not necessary
-  const flags = tx.Flags as number | MPTokenIssuanceSetFlagsInterface
+  const flags = (tx.Flags ?? 0) as number | MPTokenIssuanceSetFlagsInterface
```
achowdhry-ripple, line 72: why is the default needed here?

achowdhry-ripple (approved):
good cleanup

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
mvadari: The ternary below breaks otherwise. Not really sure what changed in this PR such that it didn't break before, tbh - maybe just more consistent testing/that's set somewhere else?

packages/xrpl/test/models/CredentialAccept.test.ts
```diff
@@ -1,16 +1,20 @@
...
-import { assert } from 'chai'
 
-import { validate, ValidationError } from '../../src'
 import { validateCredentialAccept } from '../../src/models/transactions/CredentialAccept'
+import { assertTxIsValid, assertTxValidationError } from '../testUtils'
+
+const assertValid = (tx: any): void =>
+  assertTxIsValid(tx, validateCredentialAccept)
+const assertInvalid = (tx: any, message: string): void =>
+  assertTxValidationError(tx, validateCredentialAccept, message)
```
achowdhry-ripple, line 9: continuing off of the thread in the other PR, I like the new assertTxIsValid methods, but I'm not sure about aliasing it here again as assertInvalid, since it seems to not simplify anything besides the function name

mvadari: It passes in the `validate[TxType]` function as a parameter, so that you don't have to pass it in every time you run it.  This makes tests a little cleaner - a basic `assertValid` and `assertInvalid`.

khancode (approved):
Net 2,361 lines deleted. Awesome refactor 👍

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
mvadari: I think it's this line: https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/models/transactions/transaction.ts#L243

If you're only testing `validate` and not `validate[TransactionType]` you will miss some errors

PR #2966, opened 2025-04-17
refactor: improve typing of `validateRequiredField`, `validateOptionalField`, and `DeliverMax`

mvadari:

This PR allows `validateRequiredField` and `validateOptionalField` to actually validate the fields from a TypeScript perspective - i.e. the types of the fields will be defined after they are called. This simplifies a lot of the typing in the library. A lot of other typing simplifications are included in this PR as well, as a downstream effect.

### Context of Change

Pulled out of #2801 

### Type of Change

## Test Plan

CI passes.


packages/xrpl/src/models/transactions/AMMBid.ts
```diff
@@ -86,7 +88,7 @@ export function validateAMMBid(tx: Record<string, unknown>): void {
   }
 
   if (tx.AuthAccounts != null) {
-    if (!Array.isArray(tx.AuthAccounts)) {
+    if (!isArray(tx.AuthAccounts)) {
```
achowdhry-ripple, line 91: what advantage does this have over the standard function?

packages/xrpl/src/models/transactions/AMMBid.ts
```diff
@@ -65,15 +67,15 @@ export function validateAMMBid(tx: Record<string, unknown>): void {
     throw new ValidationError('AMMBid: missing field Asset')
   }
 
-  if (!isCurrency(tx.Asset)) {
+  if (!isIssuedCurrency(tx.Asset)) {
```
achowdhry-ripple, line 70: Why is this change needed?

packages/xrpl/src/models/transactions/AMMBid.ts
achowdhry-ripple: nvm, saw your changes to common after

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -99,18 +103,28 @@ export function isNumber(num: unknown): num is number {
...
 }
 
+/**
+ * Verify the form and type of a Currency at runtime.
+ *
+ * @param input - The input to check the form and type of.
+ * @returns Whether the Currency is properly formed.
+ */
+export function isCurrency(input: unknown): input is Currency {
+  return isString(input) || isIssuedCurrency(input)
```
achowdhry-ripple, line 117: what is the case that we would need this function as opposed to isissuedCurrency?

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -538,7 +558,7 @@ export function validateCredentialsList(
   if (credentials == null) {
     return
   }
```
achowdhry-ripple, line 564: since we now use isArray, could we delete the null check?

packages/xrpl/src/models/transactions/transaction.ts
```diff
@@ -226,29 +231,31 @@ export function validate(transaction: Record<string, unknown>): void {
...
-    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- needed here
-    ;(tx.Memos as Array<Memo | null>).forEach((memo) => {
-      if (memo?.Memo == null) {
+    if (!isArray(tx.Memos)) {
+      throw new ValidationError('Memo must be array')
+    }
+    tx.Memos.forEach((memo) => {
+      if (!isRecord(memo)) {
+        throw new ValidationError('Memo must be an object')
+      }
```
achowdhry-ripple: can we use isMemo here?

khancode (approved):

packages/xrpl/src/models/transactions/common.ts
mvadari: Issued Currency is only IOUs, not XRP. So usually you'd need `isCurrency`, but some cases you may need specifically just IOU

packages/xrpl/src/models/transactions/common.ts
mvadari: Good catch

packages/xrpl/src/models/transactions/transaction.ts
mvadari: Good catch, was able to simplify a fair bit with `validateBaseTransaction`

packages/xrpl/src/models/transactions/common.ts
mvadari: No, we need to return if it's null, otherwise it'll fail the `isArray` check and throw an error

Issue #3011, opened 2025-06-14
breaking change in the latest implementation of ripple-binary-codec 2.4.0


ihomp:
We found a breaking change in the latest implementation of ripple-binary-codec 2.4.0
after Permission Delegation implementation
new FIELDS were added to the definitions file
```
    [
      "PermissionValue",
      {
        "isSerialized": true,
        "isSigningField": true,
        "isVLEncoded": false,
        "nth": 52,
        "type": "UInt32"
      }
    ],
```
And without this field, no previous versions of the definitions files are going to be loaded, or working with the latest version of ripple-binary-codec
`this.field['PermissionValue'].associatedType = this.delegatablePermissions`

https://github.com/XRPLF/xrpl.js/pull/2981

It would be nice to check if PermissionValue exists before adding it.

@ildaruz

khancode:
@ihomp thanks for raising this issue and your point is valid.

Since `definitions.json` schema can be customized by developers and we did a minor version bump, it should've been backwards compatible. We plan to do a quick fix to add a conditional check for `PermissionValue` and then do another minor version bump for `ripple-binary-codec`.

Patel-Raj11:
@ihomp I was under the opinion that ripple-binary-codec 2.4.0 is released as a whole package and the definitions.json file is baked into it. How are you able to provide a custom definitions.json file while using ripple-binary-codec 2.4.0 in your code?

An example use case would help me understand how is this a breaking change.

khancode:
@ihomp can you also share the error log you're getting?

khancode:
@ihomp The patch fix is now published for ripple-binary-codec v2.4.1

PR #3018, opened 2025-06-18
fix: add conditional check for `PermissionValue` definition

khancode:

Adds conditional check for `PermissionValue` so custom definitions (based on previous v2.x versions) don't break.

### Context of Change

Bithomp reported this [issue](https://github.com/XRPLF/xrpl.js/issues/3011).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Deleted `PermissionValue` from defintions.json and verified it doesn't throw an error anymore.


achowdhry-ripple (approved):

Patel-Raj11:
Can we add a test here that removes `PermissionValue` and makes sure that encode and decode of a transaction works fine?
https://github.com/XRPLF/xrpl.js/blob/ea649338e7962300a0d60f1e358f5f4e36e51d85/packages/ripple-binary-codec/test/definitions.test.ts

Patel-Raj11 (approved):
LGTM!

PR #3032, opened 2025-07-02
XLS-81D Permissioned DEX

ckeshava:

This PR implements support for the [specification defined here](https://github.com/XRPLF/XRPL-Standards/blob/56ba122be9f127c940fd37d433faaa2de91a4eb6/XLS-0081d-permissioned-dex/README.md).

The rippled C++ PR can be [found here](https://github.com/XRPLF/rippled/pull/5404/files).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Appropriate tests have been included.


ckeshava:
@coderabbit review

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
```diff
@@ -0,0 +1,359 @@
...
+        TakerPays: '1000',
+        TakerGets: {
+          currency: 'USD',
+          issuer: testContext.wallet.classicAddress,
+          value: '10',
+        },
+        DomainID: pd_ledger_object.index,
+      }
+
+      await testTransaction(testContext.client, offerCrossTx, wallet2)
```
ckeshava: @shawnxie999  Hello Shawn, can you help me with this test case? I expect this Offer to cross with another Offer that I created in the `beforeAll` method, at the beginning of this file.

However, these offers do not cross with each other. I have ensured that they both use the same `DomainID`. The `TakerPays` and `TakerGets` values have been configured to ensure that both the offers are complementary to each other.

Do you know why they do not cross? Am I missing something?

shawnxie999: not sure if it will fix the issue, but can you trying making the first offer non-hybrid?

ckeshava: I tried your suggestion -- it did not help ://

shawnxie999: does normal offer crossing work without domain?

packages/xrpl/src/models/ledger/Offer.ts
```diff
@@ -2,6 +2,17 @@ import { Amount } from '../common'
 
 import { BaseLedgerEntry, HasPreviousTxnID } from './BaseLedgerEntry'
 
+export interface Book {
```
mvadari, line 5:
```suggestion
export interface AdditionalBook {
```

packages/xrpl/src/models/transactions/offerCreate.ts
```diff
@@ -140,4 +174,19 @@ export function validateOfferCreate(tx: Record<string, unknown>): void {
   if (tx.OfferSequence !== undefined && typeof tx.OfferSequence !== 'number') {
     throw new ValidationError('OfferCreate: invalid OfferSequence')
   }
+
+  if (!validateDomainID(tx.DomainID)) {
+    throw new ValidationError('OfferCreate: invalid DomainID')
+  }
```
mvadari:
```suggestion
  requireOptionalField(tx, 'DomainID', validateDomainID)
```

packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -199,6 +213,10 @@ export function validatePayment(tx: Record<string, unknown>): void {
     throw new ValidationError('PaymentTransaction: InvoiceID must be a string')
   }
 
+  if (!validateDomainID(tx.DomainID)) {
```
mvadari: ditto here

packages/xrpl/src/models/utils/flags.ts
```diff
@@ -101,7 +101,7 @@ export function setTransactionFlagsToNumber(tx: Transaction): void {
  * @param tx - A Transaction to parse Flags for
  * @returns A numerical representation of a Transaction's Flags
  */
-export function convertTxFlagsToNumber(tx: Transaction): number {
+export function convertTxFlagsToNumber(tx: BaseTransaction): number {
```
mvadari: Why is this changed?

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
shawnxie999: you may have to enable defaultRipple on ur account root. that might be the issue

packages/xrpl/src/models/utils/flags.ts
ckeshava: I want to use the `convertTxFlagsToNumber` method inside one of the transaction validation. Without this change, the type-system of TypeScript expects to see all the PseudoTransaction-fields. 

This change makes `convertTxFlagsToNumber` more universally applicable.

packages/xrpl/src/models/transactions/offerCreate.ts
ckeshava: Isn't it easier to read the code with an explicit if-condition? With your suggestion, I'll need to understand the function definition of `requireOptionalField`

packages/xrpl/src/models/utils/flags.ts
mvadari: I recommend this structure: https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts#L73-L76

We shouldn't be mutating anything in `tx` in `validate`, as that will change the tx without the dev knowing.

packages/xrpl/src/models/transactions/offerCreate.ts
mvadari: It's fairly self-explanatory and simpler to read (and also results in a cleaner error message)

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: thank you very much Shawn, this suggestion helped a lot! I was missing the DefaultRipple flag. the cpp tests set this flag inside the test framework, so I missed it when reading the code

packages/xrpl/src/models/transactions/offerCreate.ts
ckeshava: You must be referring to the method `validateOptionalField` inside `packages/xrpl/src/models/transactions/common.ts` file.

Here is the error produced by TypeScript:
```
Argument of type '(domainID?: unknown) => boolean' is not assignable to parameter of type '(inp: unknown) => inp is unknown'.
  Signature '(domainID?: unknown): boolean' must be a type predicate.ts(2345)
```

All the existing uses of the `validateOptionalField` method make use of type-assertions. Here are some examples of the callback method: `isNumber`, `isString`, `isAccount`. I have not created an explicit `type`/`interface` for `DomainID` (for the sake of simplicity of the PR).

How can you write a type-predicate without a type/interface definition?

I'll need to do something like this:
```
type DomainID {
    value: string
}

export function validateDomainID(domainID?: unknown): domainID is DomainID {
  if (domainID === undefined || domainID === null || domainID.value === undefined || domainID.value === null) {
    return true
  }

  if (typeof domainID.value !== 'string') {
    return false
  }

  if (domainID.value.length !== _DOMAIN_ID_LENGTH) {
    return false
  }

  // possibly some more hex validation


  return true
}

// invoke inside validate method
validateOptionalField(tx, 'DomainID', validateDomainID)
```


Is this what you have in mind?

packages/xrpl/src/models/utils/flags.ts
ckeshava: ok, I have updated the validation in 58544b751
I have also reverted the changes to the `convertTxFlagsToNumber` method -- I realised that I don't need them anymore.

ckeshava: > We shouldn't be mutating anything in tx in validate, as that will change the tx without the dev knowing

I'm not mutating the `tx` at all. This change only accepts/rejects a transaction based on the input flags.

mvadari: `convertTxFlagsToNumber` mutates the transaction

packages/xrpl/src/models/transactions/offerCreate.ts
mvadari:
```
function validateDomainID(domainID: unknown): domainID is string {
    return isString(domainID) && domainID.length === _DOMAIN_ID_LENGTH
}
```
See the structure in all the existing `is...` functions - no need to check `null`, `validateOptionalField` already does that for you. It's all cleaned up a lot in #2895 FYI

packages/xrpl/src/models/utils/flags.ts
ckeshava: Yes, but I hadn't used `convertTxFlagsToNumber ` inside the OfferCreate validation. I planned to use it in one of my earlier drafts, but I chose otherwise in the final draft.

packages/xrpl/src/models/transactions/offerCreate.ts
ckeshava: okay thanks. Since domainID is not a rippled (or) TypeScript type (like `Account`, `string`, `number`), I felt that it is confusing for developers.

I've incorporated your suggestions in 6bec2eb54

I'll review your PR in the next few days

packages/xrpl/src/models/transactions/payment.ts
ckeshava: I've incorporated your suggestions in https://github.com/XRPLF/xrpl.js/commit/6bec2eb54e5e1051dd3759449484b9bed91fdf5f

packages/xrpl/src/models/transactions/offerCreate.ts
mvadari: IMO not everything _has_ to be a strict type, especially if it's just an alias for `string`. Might get confusing and people might think there's a type-level check for some things.

packages/xrpl/src/models/ledger/Offer.ts
ckeshava: I'm basing my changes off of [this file](https://github.com/XRPLF/rippled/pull/5404/files#diff-c4c9724bcf26e225b395e4218f52b07c123b1be58870d836bbe8b731f15d4cef). 

I couldn't find any reference to `sfAdditionalBook` in the rippled codebase.

Here is the `offer` ledger object produced from the `ledger_entry` RPC.

```
{
        Account: 'rMn7X1zJS8pv4R5APWLCrEkVPaSEqj8QPp',
        AdditionalBooks: [ { Book: [Object] } ],
        BookDirectory: '580F976574AC2FDB82181AA0EA5D8C6D7075D7E6F7A2D53053038D7EA4C68000',
        BookNode: '0',
        DomainID: '8BB4C6D27F469EDB167ECD684725DF4E09BE8E158958F5197B3A80AE053C5A65',
        // other fields below
```

`AdditionalBooks: [ { Book: [Object] } ],` is the observed output

packages/xrpl/src/models/ledger/Offer.ts
```diff
@@ -34,9 +45,20 @@ export default interface Offer extends BaseLedgerEntry, HasPreviousTxnID {
...
   Expiration?: number
+
+  /** The domain that the offer must be a part of. */
+  DomainID?: string
+
+  /**
+   * An additional list of order book directories that this offer belongs to.
+   * Currently this field only applicable to hybrid offers.
+   */
+  AdditionalBooks?: Book[]
```
Patel-Raj11, line 56: `AdditionalBooks` is not an array of `Book` objects, instead it's array of inner `Book` objects. For ref see: `AuthorizeCredential` in `DepositPreauth` ledger object.

packages/xrpl/src/models/transactions/offerCreate.ts
```diff
@@ -6,8 +6,23 @@ import {
...
+const _DOMAIN_ID_LENGTH = 64
+
+/**
+ * Utility method used across offerCreate and payment transactions to validate the domainID.
+ *
+ * @param domainID - The domainID is a 64-character string that is used to identify a domain.
+ *
+ * @returns true if the domainID is a valid 64-character string, false otherwise
+ */
+export function validateDomainID(domainID: unknown): domainID is string {
```
Patel-Raj11: I would keep this validator in `common.ts` since it's used in both OfferCreate and Payment transactions.

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
```diff
@@ -0,0 +1,360 @@
...
+    assert.equal(response.type, 'response')
+
+    // Note: The result is empty because no Offer has been created after the creation of the Subscription stream.
+    // This case is tested in the rippled code. To avoid the additional complexity, validating the contents
+    // of the response is skipped in this test.
+    // This test validates that domain_id is an acceptable input parameter to the subscribe command.
+    assert.isEmpty(response.result)
+  })
+
+  it(
```
Patel-Raj11: If this `it` gets executed first, then other tests like `Validate the bookOffers method` will fail. Easy solution to remove flaky test cases is to merge all the tests into one in the order that you have written them.

ckeshava: Image

[This section in the Jest documentation](https://jestjs.io/docs/setup-teardown#scoping) pages state that there is a well-defined order of execution of the tests.

Since the "crossing of the PermDEX offer" is the last test, it should not affect the execution of the other tests. 

I don't mind combining all the tests into one chunk -- this was my first draft :) 
I broke it up because I felt its easier to read in disparate `it(...)` blocks.

ckeshava: Empirically, I have not observed any flakiness of this integration test -- on my local machine (or) on the Github CI pipeline

packages/xrpl/src/models/ledger/Offer.ts
ckeshava: Hello Raj, I do not understand the point that you have raised here. [This conversation above](https://github.com/XRPLF/xrpl.js/pull/3032/files#r2193013035) is relevant to this discussion. 

1. The structure of 
```
AuthorizeCredentials : [ AuthorizeCredential: { /* appropriate fields */ }]
```
is different from:
```
AdditionalBooks: [ Book: { /* Book-related fields */ } ]
```

2. [Here is an example](https://github.com/XRPLF/rippled/blob/6534757d85879517cb825c8205b65b5d87943b65/src/xrpld/ledger/detail/View.cpp#L1505) where the AdditionalBooks array is parsed inside the rippled code. You can observe the `sfBookDirectory` and `sfBookNode` values inside the for-loop.

(tbh, the for-loop is not required because the `AdditionalBooks` array will contain exactly one element.)

3. The `AdditionalBooks` array is set here [inside the CreateOffer transaction](https://github.com/XRPLF/rippled/blob/6534757d85879517cb825c8205b65b5d87943b65/src/xrpld/app/tx/detail/CreateOffer.cpp#L1024).

`AdditionalBooks` array uses `Book` as the inner-object.

packages/xrpl/src/models/transactions/offerCreate.ts
ckeshava: refactored in 249b5ffea..73edaba97

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
Patel-Raj11: I see, if `it`'s get executed sequentially then we are good. Thanks for the details.

ckeshava: yes, that's my understanding as well

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -683,3 +683,16 @@ export function containsDuplicates(
...
+/**
+ * Utility method used across offerCreate and payment transactions to validate the domainID.
+ *
+ * @param domainID - The domainID is a 64-character string that is used to identify a domain.
+ *
+ * @returns true if the domainID is a valid 64-character string, false otherwise
+ */
+export function validateDomainID(domainID: unknown): domainID is string {
+  return isString(domainID) && domainID.length === _DOMAIN_ID_LENGTH
+}
```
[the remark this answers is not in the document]
ckeshava: this is not obvious from the XLS secification. The domainID is obtained as an "index" of a ledger-entry object. This is the `sha512Half` value of appropriate fields from the Permissioned Domain amendment. However, it is not clear to me if the values will always be hexa-decimal. 

If you find any resource to that effect, let me know.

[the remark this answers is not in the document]
ckeshava: the rippled cpp code does not enforce any regex check on the domainID value.

packages/xrpl/src/models/ledger/Offer.ts
Patel-Raj11: `Book` object is wrapped inside another object. It's quiet weird but that's how it is. This Offer ledger object is an example.

`curl --location --request GET 'https://s.devnet.rippletest.net:51234/' \
--header 'Content-Type: application/json' \
--data '{
  "method": "ledger_entry",
  "params": [{
    "offer": {
      "account": "rFEJs3GbJMKQ8s1A75bymhXdgCUqjFqcA",
      "seq": 4359946
    },
    "ledger_index": "validated"
  }]
}'`

ckeshava: @Patel-Raj11 The structure of `AdditionalBooks` in your example is identical to the one we observe in the [integ tests](https://github.com/XRPLF/xrpl.js/pull/3032/files#diff-f0891e287e6abb325026b593d4b8785e0bb2284affdb592189b6228c8ec048f3). In fact, all the STArray type values are represented in the following way:
```
outerFieldName: [
    {
        "innerField1": { /* some key-value pairs inside this JSON */ },
    },
    {
        "innerField2": { /* some key-value pairs inside this JSON */ },
    },
    {
        "innerField3": { /* some key-value pairs inside this JSON */ },
    },
    // arbitrary number of such items ...
]
``` 

This is the structure of all STArray objects. If it helps, other fields like `AffectedNodes` (inside a transaction's metadata), [`Memos`](https://xrpl.org/docs/references/protocol/transactions/common-fields#memos-field) (one of the common field in every transaction), etc also follow identical STArray format.

Here is the snippet of ripple-binary-codec library which [implements this structure](https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/types/st-array.ts#L104-L106) for the `STArray` objects. Perhaps, some unit tests from the ripple-binary-codec package might clarify this structure.

Patel-Raj11: Yes, the structure is the one you mentioned above. But the Typescript code that is currently implemented does not follow that. To verify, we can print the BookNode object in the IT and it would undefined. But it's actually present as per the JSON rippled response.
 
```
if (offer_ledger_object.AdditionalBooks) {
        console.log(offer_ledger_object.AdditionalBooks[0].BookNode)
}
```

Patel-Raj11: This should be the type instead:
```
export interface AdditionalBook {
  Book: {
    /** The ID of the offer directory that links to this offer. */
    BookDirectory: string

    /**
     * A hint indicating which page of the offer directory links to this entry,
     * in case the directory consists of multiple pages.
     */
    BookNode: string
  }
}
```

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
mvadari: I wouldn't necessarily count on that - there are lots of ways that those tests could run in parallel. It says "by default", not "guaranteed".

packages/xrpl/src/models/ledger/Offer.ts
ckeshava: Thanks for your explanation Raj, you are correct. I have rectified the mistake [in this commit](25715a345).
Furthermore, I have added [unit tests ](https://github.com/XRPLF/xrpl.js/pull/3032/files#diff-f0891e287e6abb325026b593d4b8785e0bb2284affdb592189b6228c8ec048f3R238-R250)that validate the structure+contents of the AdditionalBooks field.

cc @mvadari  you were correct in your suggestion regarding the structure of this nested object field

ckeshava: @Patel-Raj11  Post-fix, the expected way to access the data members would be:
```
console.log(offer_ledger_object.AdditionalBooks?.[0].Book.BookDirectory) // correct
console.log(offer_ledger_object.AdditionalBooks[0].BookNode) // your example from above, its missing "Book" member access
```

Patel-Raj11 (approved):
LGTM!

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: As per [jest documentation](https://jestjs.io/docs/api#testconcurrentname-fn-timeout): within one file, if you want to run tests in parallel, you need to use `test.concurrent` (or) `it.concurrent` methods. I did not find such methods inside the xrpl.js code base.

I could not find guarantees from Jest regarding the serial-order of execution, however I found [several Stackoverflow answers ](https://stackoverflow.com/questions/39831961/are-tests-inside-one-file-run-in-parallel-in-jest) to that effect.

ckeshava: I don't mind combining all the tests -- it will be one 400-line block, it might seem daunting for developers

packages/xrpl/src/models/transactions/common.ts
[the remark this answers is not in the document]
ckeshava: This validation is not necessary. It adds additional complexity to the validateDomainID method. Besides, if there is a mistake in the domainID value, users will not be able to serialize/deserialize such values. This is because domainID is translated as a Hash<length-parameter> rippled internal type.

Instead, I have added unit tests in the ripple-binary-codec package [in this commit](16fcb6c64). This validates that only hex-encoded strings are used in the constructors of `Hash`-types.

packages/xrpl/src/models/ledger/Offer.ts
```diff
@@ -34,9 +47,20 @@ export default interface Offer extends BaseLedgerEntry, HasPreviousTxnID {
...
```
khancode:
```suggestion
  /** The domain that the offer must be a part of. */
  DomainID?: string
  /**
   * An additional list of order book directories that this offer belongs to.
   * Currently this field only applicable to hybrid offers.
   */
  AdditionalBooks?: Book[]
```
To match format.

packages/xrpl/src/models/ledger/Offer.ts
```diff
@@ -34,9 +47,20 @@ export default interface Offer extends BaseLedgerEntry, HasPreviousTxnID {
...
 }
 
 export enum OfferFlags {
   lsfPassive = 0x00010000,
   lsfSell = 0x00020000,
+  // Indicates the offer is hybrid. (meaning it is part of both a domain and open order book)
+  lsfHybrid = 0x00040000,
```
khancode:
```suggestion
  lsfHybrid = 0x00040000,
```
Should probably remove the comment since this can be looked up on docs. The other flags don't have it too.

packages/xrpl/src/models/methods/bookOffers.ts
```diff
@@ -39,6 +39,14 @@ export interface BookOffersRequest extends BaseRequest, LookupByLedgerRequest {
...
    */
   taker_pays: BookOfferCurrency
+
+  /**
+   * The object ID of a PermissionedDomain object. If this field is provided,
+   * the response will include only valid domain offers associated with that
+   * specific domain. If omitted, the response will include only hybrid and open
+   * offers for the trading pair, excluding all domain-specific offers.
+   */
+  domain?: string
```
khancode:
```suggestion
  /**
   * The object ID of a PermissionedDomain object. If this field is provided,
   * the response will include only valid domain offers associated with that
   * specific domain. If omitted, the response will include only hybrid and open
   * offers for the trading pair, excluding all domain-specific offers.
   */
  domain?: string
```
Remove newline.

packages/xrpl/src/models/methods/pathFind.ts
```diff
@@ -30,6 +30,12 @@ export interface PathFindCreateRequest extends BasePathFindRequest {
    * about, or to check the overall cost to make a payment along a certain path.
    */
   paths?: Path[]
+
+  /**
```
khancode:
```suggestion
  /**
```

packages/xrpl/src/models/methods/pathFind.ts
```diff
@@ -97,6 +103,13 @@ export interface PathFindResponse extends BaseResponse {
      * Continues to send updates each time a new ledger closes.
      */
     full_reply: boolean
+
+    /**
+     * The object ID of a PermissionedDomain object, if the orderbook shown is
+     * for a specific domain.
+     */
+    domain?: string
+
```
khancode:
```suggestion
    /**
     * The object ID of a PermissionedDomain object, if the orderbook shown is
     * for a specific domain.
     */
    domain?: string
```

khancode: Also is the JSDoc description seems off and different than other domain variables in this PR.

packages/xrpl/src/models/methods/ripplePathFind.ts
```diff
@@ -38,6 +38,12 @@ export interface RipplePathFindRequest
    * and optional issuer field, like how currency amounts are specified.
    */
   source_currencies?: SourceCurrencyAmount[]
+
+  /**
+   * The object ID of a PermissionedDomain object. If this field is included,
+   * then only valid paths for this domain will be returned.
+   */
+  domain?: string
```
khancode:
```suggestion
  /**
   * The object ID of a PermissionedDomain object. If this field is included,
   * then only valid paths for this domain will be returned.
   */
  domain?: string
```

packages/xrpl/src/models/methods/subscribe.ts
```diff
@@ -39,6 +39,13 @@ export interface SubscribeBook {
   snapshot?: boolean
   /** If true, return both sides of the order book. The default is false. */
   both?: boolean
+
+  /**
+   * The object ID of a PermissionedDomain object. If this field is included,
+   * then the offers will be filtered to only show the valid domain offers for
+   * that domain.
+   */
+  domain?: string
```
khancode:
```suggestion
  /**
   * The object ID of a PermissionedDomain object. If this field is included,
   * then the offers will be filtered to only show the valid domain offers for
   * that domain.
   */
  domain?: string
```

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -683,3 +683,16 @@ export function containsDuplicates(
...
+const _DOMAIN_ID_LENGTH = 64
+
+/**
+ * Utility method used across offerCreate and payment transactions to validate the domainID.
+ *
+ * @param domainID - The domainID is a 64-character string that is used to identify a domain.
+ *
+ * @returns true if the domainID is a valid 64-character string, false otherwise
+ */
+export function validateDomainID(domainID: unknown): domainID is string {
```
khancode:
```suggestion
export function isDomainID(domainID: unknown): domainID is string {
```

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -683,3 +683,16 @@ export function containsDuplicates(
 
   return false
 }
+
+const _DOMAIN_ID_LENGTH = 64
+
+/**
+ * Utility method used across offerCreate and payment transactions to validate the domainID.
```
khancode:
```suggestion
 * Utility method used across OfferCreate and Payment transactions to validate the DomainID.
```

packages/xrpl/src/models/transactions/offerCreate.ts
```diff
@@ -106,6 +114,9 @@ export interface OfferCreate extends BaseTransaction {
   TakerGets: Amount
   /** The amount and type of currency being requested by the offer creator. */
   TakerPays: Amount
+
+  /** The domain that the offer must be a part of. */
+  DomainID?: string
```
khancode:
```suggestion
  /** The domain that the offer must be a part of. */
  DomainID?: string
```

packages/xrpl/src/models/transactions/offerCreate.ts
```diff
@@ -140,4 +151,23 @@ export function validateOfferCreate(tx: Record<string, unknown>): void {
...
+    paramName: 'DomainID',
+  })
+
+  if (tx.DomainID === undefined) {
+    if (
+      (typeof tx.Flags === 'number' &&
+        // eslint-disable-next-line no-bitwise -- flags require bitwise operations
+        (tx.Flags & OfferCreateFlags.tfHybrid) !== 0) ||
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Flags are stored as a number or an interface
+      (tx.Flags as OfferCreateFlagsInterface).tfHybrid
```
khancode: Replace this with `hasFlag()` from `packages/xrpl/src/models/utils/index.ts`

packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -161,6 +162,19 @@ export interface Payment extends BaseTransaction {
...
+   * be part of this domain. The DomainID can be included if the sender intends
+   * it to be a cross-currency payment (i.e. if the payment is going to interact
+   * with the DEX). The domain will only play it's role if there is a path that
+   * crossing an orderbook.
+   *
+   * Note: it's still possible that DomainID is included but the payment does
+   * not interact with DEX, it simply means that the DomainID will be ignored
+   * during payment paths.
+   */
+  DomainID?: string
```
khancode:
```suggestion
  /**
   * The domain the sender intends to use. Both the sender and destination must
   * be part of this domain. The DomainID can be included if the sender intends
   * it to be a cross-currency payment (i.e. if the payment is going to interact
   * with the DEX). The domain will only play it's role if there is a path that
   * crossing an orderbook.
   *
   * Note: it's still possible that DomainID is included but the payment does
   * not interact with DEX, it simply means that the DomainID will be ignored
   * during payment paths.
   */
  DomainID?: string
```

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
```diff
@@ -0,0 +1,373 @@
...
+  let wallet2: Wallet
+  let pd_ledger_object: PermissionedDomain
+  let offerCreateTxResponse: SubmitResponse
+  let offer_ledger_object: Offer
+
+  beforeAll(async () => {
+    // this section describes the pre-requisites for testing PermissionedDEX features
+    testContext = await setupClient(serverUrl)
+    wallet1 = await generateFundedWallet(testContext.client)
+    wallet2 = await generateFundedWallet(testContext.client)
```
khancode, line 53: Can you rename these wallets to their roles in this test?

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
```diff
@@ -0,0 +1,373 @@
...
+        offer_ledger_object.BookDirectory,
+      )
+    },
+    TIMEOUT,
+  )
+
+  it(
+    'Validate the properties of the DirectoryNode ledger object (contains the PermissionedDEX Offer object)',
+    async () => {
+      const directoryNode_ledger_object = await testContext.client.request({
```
khancode:
```suggestion
      const ledgerEntryResponse = await testContext.client.request({
```

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
```diff
@@ -0,0 +1,373 @@
...
+    assert.equal(
+      (response.result.offers[0].TakerPays as IssuedCurrencyAmount).issuer,
+      testContext.wallet.classicAddress,
+    )
+
+    assert.equal(response.result.offers[0].DomainID, pd_ledger_object.index)
+  })
+
+  it(`Validate the subscription stream for PermissionedDEX offers`, async () => {
+    const subscribe_request: SubscribeRequest = {
```
khancode:
```suggestion
    const request: SubscribeRequest = {
```

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
khancode: I don't mind combining all the tests to resolve this issue.

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
```diff
@@ -0,0 +1,373 @@
...
+    // of the response is skipped in this test.
+    // This test validates that domain_id is an acceptable input parameter to the subscribe command.
+    assert.isEmpty(response.result)
+  })
+
+  it(
+    'Crossing a PermissionedDEX Offer',
+    async () => {
+      // wallet2 "crosses" the offer within the domain
+      const offerCrossTx: OfferCreate = {
```
khancode:
```suggestion
      const offerCreateTx: OfferCreate = {
```

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
```diff
@@ -0,0 +1,373 @@
...
+          issuer: testContext.wallet.classicAddress,
+          value: '10',
+        },
+        DomainID: pd_ledger_object.index,
+      }
+
+      await testTransaction(testContext.client, offerCrossTx, wallet2)
+
+      // Validate that Offer ledger objects are consumed in both wallets
+      const wallet1_objects = await testContext.client.request({
```
khancode:
```suggestion
      const wallet1Objects = await testContext.client.request({
```

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
```diff
@@ -0,0 +1,373 @@
...
+        command: 'account_objects',
+        account: wallet1.classicAddress,
+        type: 'offer',
+      })
+      assert.isEmpty(wallet1_objects.result.account_objects)
+
+      const wallet2_objects = await testContext.client.request({
```
khancode:
```suggestion
      const wallet2Objects = await testContext.client.request({
```

packages/xrpl/test/models/payment.test.ts
```diff
@@ -37,13 +37,59 @@ describe('Payment', function () {
...
+      DomainID: '5'.repeat(64),
     } as any
   })
 
   it(`verifies valid Payment`, function () {
     assertValid(payment)
   })
 
+  it(`throws -- invalid DomainID type`, function () {
+    const payment_invalid_domain_id_type = {
```
khancode:
```suggestion
    const paymentTx = {
```

packages/xrpl/test/models/payment.test.ts
```diff
@@ -37,13 +37,59 @@ describe('Payment', function () {
...
+    } as any
+
+    assertInvalid(
+      payment_invalid_domain_id_type,
+      'PaymentTransaction: invalid field DomainID',
+    )
+  })
+
+  it(`throws -- invalid DomainID , exceeds expected length`, function () {
+    const payment_invalid_domain_id_length = {
```
khancode:
```suggestion
    const paymentTx = {
```

packages/xrpl/test/models/payment.test.ts
```diff
@@ -37,13 +37,59 @@ describe('Payment', function () {
...
+    } as any
+
+    assertInvalid(
+      payment_invalid_domain_id_length,
+      'PaymentTransaction: invalid field DomainID',
+    )
+  })
+
+  it(`throws -- invalid DomainID , falls short of expected length`, function () {
+    const payment_invalid_domain_id_length = {
```
khancode:
```suggestion
    const paymentTx = {
```

packages/xrpl/src/models/methods/pathFind.ts
ckeshava: hello Omar, my intention is to explain the context of this response field. The `domain` field is returned only if the response corresponds to an orderbook accessible within a specific `PermissionedDomain`

ckeshava: let me know if you'd prefer to rephrase it.

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: Hello Omar, 
I'm unable to identify `roles` for wallet1 and wallet2. Both of them participate in the PermissionedDEX orderbook. Both the wallets create permissioned-offers which cross each other.

Both of them are equivalent. Unlike the example of an escrow, there is no `creator` / `destination` in the case of Permissioned Offers.

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: this offer indicates that it is crossing the first offer. I'd like to differentiate this from a generic offer-creation transaction. It indicates that the first offer is being fulfilled by this second, matching offer.

packages/xrpl/test/models/payment.test.ts
ckeshava: changed variable name in 66ac58d27

packages/xrpl/test/models/payment.test.ts
ckeshava: changed variable name in 66ac58d27

packages/xrpl/test/models/payment.test.ts
ckeshava: changed variable name in 66ac58d27

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: changed variable name in 66ac58d27

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: changed variable name in 66ac58d27

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: changed variable name in 66ac58d27

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: changed variable name in 66ac58d27

packages/xrpl/src/models/transactions/offerCreate.ts
ckeshava: 66ac58d27

packages/xrpl/src/models/transactions/common.ts
ckeshava: changed function name in 66ac58d27

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: alright. I have combined all the tests into one function 66ac58d27..ecd85c7d1

packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -161,6 +162,18 @@ export interface Payment extends BaseTransaction {
...
```
khancode: nit: move DomainID above flags.

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
```diff
@@ -0,0 +1,369 @@
...
+// how long before each test case times out
+const TIMEOUT = 20000
+
+describe('PermissionedDEX', function () {
+  let testContext: XrplIntegrationTestContext
+  let wallet1: Wallet
+  let wallet2: Wallet
+  let pd_ledger_object: PermissionedDomain
+  let offerCreateTxResponse: SubmitResponse
+  let offer_ledger_object: Offer
```
khancode:
```suggestion
  let permDomainLedgerObject: PermissionedDomain
  let offerCreateTxResponse: SubmitResponse
  let offerLedgerObject: Offer
```

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
khancode: It's confusing because there is no transaction named `OfferCross`. The comment above it describes it's use too.

packages/xrpl/test/models/offerCreate.test.ts
```diff
@@ -86,6 +89,111 @@ describe('OfferCreate', function () {
     assertValid(offer3)
   })
 
+  it(`verfies valid offerCreate within permissioned domain`, function () {
+    const offer = {
```
khancode:
```suggestion
    const offerTx = {
```
Just to be consistent with naming transactions. Ditto for the remaining ones below this test too.

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: 75ae1c3d7

packages/xrpl/src/models/transactions/payment.ts
ckeshava: 75ae1c3d7

packages/xrpl/test/integration/transactions/permissionedDEX.test.ts
ckeshava: 75ae1c3d7

packages/xrpl/test/models/offerCreate.test.ts
ckeshava: 75ae1c3d7

ckeshava: 7c23dbe40

packages/ripple-binary-codec/test/hash.test.ts
```diff
@@ -32,6 +32,12 @@ describe('Hash128', function () {
       new Error('Invalid Hash length 17'),
     )
   })
+
+  it(`throws when constructed from non-hexadecimal string`, () => {
+    expect(() => Hash128.from('Z'.repeat(16))).toThrow(
+      new Error('Invalid hash string ' + 'Z'.repeat(16)),
+    )
+  })
```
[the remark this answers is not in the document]
ckeshava: fixed in ac6c83cb2

packages/xrpl/src/models/transactions/offerCreate.ts
```diff
@@ -140,4 +151,18 @@ export function validateOfferCreate(tx: Record<string, unknown>): void {
...
     throw new ValidationError('OfferCreate: invalid OfferSequence')
   }
+
+  validateOptionalField(tx, 'DomainID', isDomainID, {
+    txType: 'OfferCreate',
+    paramName: 'DomainID',
+  })
+
+  if (
+    tx.DomainID === undefined &&
```
khancode, line 161:
```suggestion
    tx.DomainID != null &&
```
It's better to use this which covers both `null` and `undefined` values.

ckeshava: Hello Omar, I'm not sure if the `!= null` condition is equivalent to `=== undefined`. I incorporated your suggestion, however it causes these unit test failures: https://github.com/XRPLF/xrpl.js/actions/runs/16324764374/job/46111811992?pr=3032#step:8:551

ckeshava: image

As shown here in the Node.js interpreter, `'' != null // true`. An empty `tx.DomainID` is not evaluated truthy, in comparison with `null`.

I haven't looked at the internals of Javascript to be able to show the specification. I suspect it has to do with type-fiddling in `==` operator versus the type+value comparison in `===` operator.

I'll not accept your suggestion for this PR. If you come across any disparity in my observations, let me know. I can create a fix PR.

ckeshava:
Thank you all the reviewers for your feedback!

packages/xrpl/src/models/transactions/offerCreate.ts
khancode: @ckeshava oh, my mistake. I meant to write `== null`. This will cover both `undefined` and `null` values. Otherwise `null === undefined` resolves to `false` which is invalid.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -683,3 +683,16 @@ export function containsDuplicates(
...
+
+/**
+ * Utility method used across OfferCreate and Payment transactions to validate the DomainID.
+ *
+ * @param domainID - The domainID is a 64-character string that is used to identify a domain.
+ *
+ * @returns true if the domainID is a valid 64-character string, false otherwise
+ */
+export function isDomainID(domainID: unknown): domainID is string {
+  return isString(domainID) && domainID.length === _DOMAIN_ID_LENGTH
```
khancode, line 697: We should add a hexadecimal check here too.

packages/xrpl/src/models/transactions/offerCreate.ts
ckeshava: `null === undefined` resolves to `false` because of the difference in the operands' types. However, why is that pertinent to our circumstance?

Do you have a unit-test in mind which could slip through the cracks? I don't understand where `null` adds more value to the validation process, over using `undefined`.

packages/xrpl/src/models/transactions/common.ts
ckeshava: I have addressed your concern with code-rabbit: https://github.com/XRPLF/xrpl.js/pull/3032#discussion_r2201900797

packages/xrpl/src/models/transactions/offerCreate.ts
khancode: It's generally best practice to account for both `null` and `undefined` cases. There could be a case were users could set it to `null` too. There's no drawback and more benefit in doing this broader check.

packages/xrpl/src/models/transactions/common.ts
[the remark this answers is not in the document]
khancode: @ckeshava since `DomanID` is required to be in hex format, it should be done at the `xrpl` level as other hex fields are. For reference, it's checked in:
- [`validateCredentialType()`](https://github.com/XRPLF/xrpl.js/blob/2309d591ec0c07d40ed00de69f8490d033bbbbcd/packages/xrpl/src/models/transactions/common.ts#L586-L590)
- [`validateMPTokenIssuanceCreate()`](https://github.com/XRPLF/xrpl.js/blob/2309d591ec0c07d40ed00de69f8490d033bbbbcd/packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts#L140-L144)

packages/xrpl/src/models/transactions/offerCreate.ts
ckeshava: image
 
 :)

mvadari: Agreed with @khancode, `== null` is best practice - can we fix this?

ckeshava: Quoting from popular [Stackoverflow answers](https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript): 
> null is an assignment value. It can be assigned to a variable as a representation of no value. undefined means a variable has been declared but has not yet been assigned a value.

Based on the archetypical usage of the xrpl.js client library, users are much more likely not set the `tx_json.DomainID` field. The alternative possibility of users setting `tx_json.DomainID = null;` appears to be rare.

What is your recommendation? Should I do something like:
`if ((tx.DomainID === undefined || tx.DomaindID === null) && <tfHybrid flag check>)`

Is this what you have in mind?

ckeshava: The xrpl.js client library appears to have a higher number of `=== undefined` (49 results) as compared against `=== null` (4 results).

I prefer the `===` over the `==` for the more stringent comparison.

ckeshava: @mvadari  says who?

> Agreed with @khancode, `== null` is best practice - can we fix this?

As [per](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#difference_between_null_and_undefined) MDN docs, `undefined == null // true`. I can update the code to use `==` (which performs implicit type conversion). This should placate your concerns.

However, I'm not convinced that validating against `null` is better than `undefined`.

mvadari: > The xrpl.js client library appears to have a higher number of `=== undefined` (49 results) as compared against `=== null` (4 results).

There are 107 results for `== null`. Most of the `=== undefined` usages will be removed in #2895.

> I prefer the `===` over the `==` for the more stringent comparison.

In this case, the less stringent requirements are better, because it'll cover additional cases, like `null` and `NaN`.

ckeshava: Alright, I have added the `== null` check in this [fix-PR](https://github.com/XRPLF/xrpl.js/pull/3050).

As per my understanding of the docs, both `operand == null` and `operand == undefined` perform identical functionality. Since the type of the operand is irrelevant, both of them evaluate to `false`. I don't see why you would prefer `null` over `undefined`.

mvadari: It's just the norm to use `== null` instead of `== undefined`. My guess is that it's because it's clearer

PR #3035, opened 2025-07-07
fix: improve Batch inner transaction typing

mvadari:

This PR swaps out the use of `BatchInnerTransaction` for runtime checking instead, allowing for better TS usage with Batch inner transactions.

Note: while this is technically a breaking change (since `BatchInnerTransaction` is being removed), it doesn't change any top-level exports.

### Context of Change

Feedback from Batch users suggested that this would be a better, more usable setup

### Type of Change

- [x] Yes

## Test Plan

Adjusted integration tests to check that it works correctly

### DIFF
--- a/packages/xrpl/test/models/batch.test.ts
+++ b/packages/xrpl/test/models/batch.test.ts
@@ -50,7 +50,7 @@ describe('Batch', function () {
             Fee: '0',
             Flags: 0x40000000,
             NetworkID: 21336,
-            Sequence: 0,
+            Sequence: 1,
             SigningPubKey: '',
             TransactionType: 'Payment',
           },


packages/xrpl/src/models/transactions/common.ts
```diff
@@ -111,6 +111,32 @@
...
 
+/**
+ * Verify the form and type of a null value at runtime.
+ *
+ * @param inp - The value to check the form and type of.
+ * @returns Whether the value is properly formed.
+ */
+export function isNull(inp: unknown): inp is null {
+  return inp == null
+}
```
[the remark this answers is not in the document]
mvadari: I want it to check for both `null` and `undefined` - is there a better name to call this function in that case?

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -111,6 +111,32 @@ export function isNumber(num: unknown): num is number {
...
+ * @param lower The lower bound (inclusive).
+ * @param upper The upper bound (inclusive).
+ * @param value
+ * @returns Whether the number is properly formed and within the bounds.
+ */
+export function isValue<V>(value: V): (inp: unknown) => inp is V {
+  // eslint-disable-next-line func-style -- returning a function
+  const isValueInternal = (inp: unknown): inp is V => inp === value
+  return isValueInternal
+}
```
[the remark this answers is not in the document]
mvadari: Fixed

packages/xrpl/src/models/transactions/common.ts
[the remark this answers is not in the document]
ckeshava: I like the name isNullish

packages/xrpl/src/models/transactions/batch.ts
```diff
@@ -73,10 +63,48 @@ export interface Batch extends BaseTransaction {
...
+  if (!hasFlag(tx, GlobalFlags.tfInnerBatchTxn, 'tfInnerBatchTxn')) {
+    throw new ValidationError(
+      `Batch: RawTransactions[${index}] must contain the \`tfInnerBatchTxn\` flag.`,
+    )
+  }
+  validateOptionalField(tx, 'Fee', isValue('0'), {
+    paramName: `RawTransactions[${index}].RawTransaction.Fee`,
+    txType: 'Batch',
+  })
+  validateOptionalField(tx, 'SigningPubKey', isValue(''), {
```
ckeshava, line 90:
```suggestion
  validateOptionalField(tx, 'SigningPubKey', () => tx.SigningPubKey === '', {
```

(another suggestion for a few lines below):
```suggestion
validateOptionalField(tx, 'TxnSignature', tx.TxnSignature == null, {
```

Do we really need the `isNull` and the `isValue` methods?

packages/xrpl/test/models/batch.test.ts
```diff
@@ -90,7 +90,7 @@ describe('Batch', function () {
             Fee: '0',
             Flags: 0x40000000,
             NetworkID: 21336,
-            Sequence: 0,
+            Sequence: 1,
```
ckeshava, line 93: There is a similar error on line-53. Two inner batch transactions have the identical sequence number.

ckeshava: I was of the understanding that rippled fills a "phantom" sequence number for the inner batch-transactions, is that not the case? Do users need to fill up the seq numbers for all the inner batch transactions?

Patel-Raj11 (approved):
LGTM!

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -111,6 +111,28 @@ export function isNumber(num: unknown): num is number {
...
 }
 
+/**
+ * Verify the form and type of a null value at runtime.
+ *
+ * @param inp - The value to check the form and type of.
+ * @returns Whether the value is properly formed.
+ */
+export function isNull(inp: unknown): inp is null {
+  return inp == null
```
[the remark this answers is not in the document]
mvadari: That is the desired behavior

packages/xrpl/test/models/batch.test.ts
mvadari: Correct

packages/xrpl/src/models/transactions/batch.ts
mvadari: Do you have a strong reason for not having them? I find it makes the code much cleaner and easier to read, and it doesn't take up much space.

packages/xrpl/test/models/batch.test.ts
ckeshava: can you also rectify this error within this PR? its a small change
> There is a similar error on line-53. Two inner batch transactions have the identical sequence number.

ckeshava (approved):
looks alright except for one small suggestion.

PR #3036, opened 2025-07-07
fix: add missing `MPTAmount` type to `VaultDeposit` & `VaultWithdraw`

khancode:

Adds missing `MPTAmount` type to `VaultDeposit` & `VaultWithdraw`

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

## Test Plan

No tests needed since it's a TS fix.

### DIFF
--- a/packages/xrpl/src/models/transactions/vaultDeposit.ts
+++ b/packages/xrpl/src/models/transactions/vaultDeposit.ts
@@ -1,4 +1,4 @@
-import { Amount } from '../common'
+import { Amount, MPTAmount } from '../common'
 
 import {
   BaseTransaction,
@@ -24,7 +24,8 @@ export interface VaultDeposit extends BaseTransaction {
   /**
    * Asset amount to deposit.
    */
-  Amount: Amount
+  // TODO: remove MPTAmount when MPTv2 is released
+  Amount: Amount | MPTAmount
 }
 
 /**
--- a/packages/xrpl/src/models/common/index.ts
+++ b/packages/xrpl/src/models/common/index.ts
@@ -29,6 +29,7 @@ export interface MPTAmount {
   value: string
 }
 
+// TODO: add MPTAmount to Amount once MPTv2 is released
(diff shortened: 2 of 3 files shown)


Patel-Raj11 (approved):
LGTM!

packages/xrpl/src/models/transactions/vaultDeposit.ts
```diff
@@ -24,7 +24,8 @@ export interface VaultDeposit extends BaseTransaction {
   /**
    * Asset amount to deposit.
    */
-  Amount: Amount
+  // TODO: remove MPTAmount when MPTv2 is released
+  Amount: Amount | MPTAmount
```
ckeshava, line 28: Its confusing to keep track of what `Amount` really means -- this could change with time.

I'd prefer to spell out all the constituent types, instead of using the `Amount` alias. This improves readability vastly and reduces one more indirection.
```suggestion
  Amount: MPTAmount | IssuedCurrencyAmount | string
```

khancode: I would keep `Amount` because it's usage is already ubiquitous.

ckeshava: However, that is the dangerous part isn't it? Different parts of XRPL allow for different types of Amounts. For example, `Check`s do not support MPTs, however `Offer`s work with both `MPT` and `IOU` (after the Permissioned DEX amendment). These are just two examples.

By encapsulating different types of values under `Amount`, we run the risk of making mistakes in the client library.

ckeshava (approved):

packages/xrpl/src/models/transactions/vaultDeposit.ts
khancode: That's fair but these other transactions will support MPT when its v2 is released. We can just add it as a union to `Amount` when that happens (that's what my code comments refer to).

PR #3041, opened 2025-07-11
Add warnings for MPTokenMetadata - XLS-89d

Patel-Raj11:

This PR adds console warnings to `MPTokenIssuanceCreate` transaction if the `MPTokenMetadata` does not follow [XLS-89d](https://github.com/XRPLF/XRPL-Standards/pull/293) standard.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan
Added unit tests.


packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -68,6 +99,30 @@ export interface MPTokenIssuanceCreateFlagsInterface
...
+  ticker: string
+  name: string
+  desc: string
+  icon: string
+  asset_class: string
+  asset_subclass?: string
+  issuer_name: string
+  urls?: MPTokenMetadataUrl[]
+  additional_info?: string
+}
```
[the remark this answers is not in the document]
Patel-Raj11: Good catch. coderabbitai is really useful 💯

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -178,5 +233,164 @@ export function validateMPTokenIssuanceCreate(
...
+
+  if (
+    !(
+      jsonMetaData != null &&
+      typeof jsonMetaData === 'object' &&
+      !Array.isArray(jsonMetaData)
+    )
+  ) {
+    console.warn(
+      'MPTokenMetadata is not properly formatted as JSON as per the XLS-89d standard. ',
```
ckeshava: Could you rephrase the warning message?
```suggestion
      'MPTokenMetadata is not properly formatted as JSON as per the XLS-89d standard. While adherence to this standard is not mandatory, such non-compliant MPToken's might not be discoverable by Explorers and Indexers in the XRPL ecosystem.',
```

ckeshava (commented):
Hello Raj, this is good work. 

1. What is the decision on the `MPTIssuanceCreate` transactions whose meta-data exceed the expected size i.e. 1024 bytes? In this version of the PR, rippled will reject them with an appropriate error-code. Is that the expected behavior?

Here is a [verbatim snippet](https://github.com/XRPLF/XRPL-Standards/pull/293#issuecomment-3062735435) from the final design choices for the XLS-89 spec:
> If the metadata is too large, developers will receive a pre-submission warning before anything is sent to rippled


2. This solution is repetitive. You need to implement this solution in all the client libraries. Despite all this effort, if a [user submits a transaction directly to a rippled node](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/transaction-methods/submit), they are not supplied with any warnings at all. If companies/organizations use non-SDK pipelines, they will not be notified of the XLS-89d structure. Do you have any thoughts on this issue?

For reference, [here is the associated validation](https://github.com/XRPLF/rippled/blob/8aa94ea09af1eef8df54b0b2b3a0373400a09b34/src/xrpld/app/tx/detail/MPTokenIssuanceCreate.cpp#L51) in rippled.

3. Users can set the MPTokenMetadata in the `VaultCreate` transaction (Single-Asset-Vault amendment) too. Are you planning to add this validation into that transaction too? 

4. I'm concerned about the number of console warnings that will be printed due to this PR -- Can a user choose to opt out of adhering to the XLS-89D standard? [This comment](https://github.com/XRPLF/XRPL-Standards/pull/293#discussion_r2181169205) from @sappenin  is helpful in detecting if users are trying to adhere to XLS-89.

5. Can you include all of this validation inside an exportable method? A method with the one-of-the-below signature is helpful. 

Otherwise, this logic will need to be repeated across `explorer` and indexers in the XRPL ecosystem. If users can import this method, that will be very helpful.
```
validateMPTokenMetadata(metadata: unknown): metadata is MPTokenMetadata { ... }
isMetadata_XLS_89_compatible(metadata: unknown): boolean { ... }
isMPTMetadataDiscoverable(metadata: unknown): boolean { ... }
```

Patel-Raj11:
@ckeshava Thanks for the detailed feedback:

> What is the decision on the MPTIssuanceCreate transactions whose meta-data exceed the expected size i.e. 1024 bytes? In this version of the PR, rippled will reject them with an appropriate error-code. Is that the expected behavior?

Yes, it's expected behaviour for rippled to reject it. Client library will just warn the user. 

> This solution is repetitive. You need to implement this solution in all the client libraries. Despite all this effort, if a [user submits a transaction directly to a rippled node](https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/transaction-methods/submit), they are not supplied with any warnings at all. If companies/organizations use non-SDK pipelines, they will not be notified of the XLS-89d structure. Do you have any thoughts on this issue?

I brought this same point to Julian and it was concluded that if the project/company is really serious about their project, they will take the warnings seriously if they are using our SDKs. If they are directly submitting the transactions to rippled, then we can't do much and I don't have any other solution in mind but to have rippled enforce such metadata structure and error out if the transaction does not adhere to it. Since MPTokensV1 is out for voting, I think we can have another amendment to enforce such metadata structure.

> Users can set the MPTokenMetadata in the VaultCreate transaction (Single-Asset-Vault amendment) too. Are you planning to add this validation into that transaction too?

Thanks for highlighting this. I overlooked it. To me it makes sense to give warning messages here as well.

> I'm concerned about the number of console warnings that will be printed due to this PR -- Can a user choose to opt out of adhering to the XLS-89D standard? https://github.com/XRPLF/XRPL-Standards/pull/293#discussion_r2181169205 from @sappenin is helpful in detecting if users are trying to adhere to XLS-89.

Yes, there would be many warnings for `VaultCreate` and `MPTokenIssuanceCreate` transactions. But those are for good. We can combine those into one single warning message. Do you have any better suggestion other than having rippled enforce at their end? Yes, user can opt out, they are just warnings and not hard failures.

> Can you include all of this validation inside an exportable method? A method with the one-of-the-below signature is helpful.

I will move these validations to a common place now since `VaultCreate` needs to be validated as well. However, since we want to export this, we need to have an extra parameter for whether to log warnings or not. Something like this:
`validateMPTokenMetadata(metadata: unknown, logWarnings: boolean): metadata is MPTokenMetadata { ... }
`

ckeshava:
> @ckeshava Thanks for the detailed feedback:
> 
> > What is the decision on the MPTIssuanceCreate transactions whose meta-data exceed the expected size i.e. 1024 bytes? In this version of the PR, rippled will reject them with an appropriate error-code. Is that the expected behavior?
> 
> Yes, it's expected behaviour for rippled to reject it. Client library will just warn the user.

Where have you included the warning in the xrpl-py client library, appropos the size of the meta-data field?

> 
> > Users can set the MPTokenMetadata in the VaultCreate transaction (Single-Asset-Vault amendment) too. Are you planning to add this validation into that transaction too?
> 
> Thanks for highlighting this. I overlooked it. To me it makes sense to give warning messages here as well.
> 
Okay, let me know once the `VaultCreate` transaction has been updated with the warning message.

> > I'm concerned about the number of console warnings that will be printed due to this PR -- Can a user choose to opt out of adhering to the XLS-89D standard? [XRPLF/XRPL-Standards#293 (comment)](https://github.com/XRPLF/XRPL-Standards/pull/293#discussion_r2181169205) from @sappenin is helpful in detecting if users are trying to adhere to XLS-89.
> 
> Yes, there would be many warnings for `VaultCreate` and `MPTokenIssuanceCreate` transactions. But those are for good. We can combine those into one single warning message. Do you have any better suggestion other than having rippled enforce at their end? Yes, user can opt out, they are just warnings and not hard failures.
> 
Hmm apart from implementing this inside rippled and David Fuelling's suggestion, I don't have any other ideas.

> > Can you include all of this validation inside an exportable method? A method with the one-of-the-below signature is helpful.
> 
> I will move these validations to a common place now since `VaultCreate` needs to be validated as well. However, since we want to export this, we need to have an extra parameter for whether to log warnings or not. Something like this: `validateMPTokenMetadata(metadata: unknown, logWarnings: boolean): metadata is MPTokenMetadata { ... } `

Are console warnings the best way to inform the user? If the `xrpl` package is used inside a `browser` (or) `node`, these console-warnings will be printed in the "developer-console". However, will developers read all the logs of their application? I'm trying to think of a better alternative to inform the user.

Patel-Raj11:
> Where have you included the warning in the xrpl-py client library, appropos the size of the meta-data field?

This will be implemented once, xrpl.js review progresses beyond the fact of deciding whether console warnings are the right thing to do, in the interest of development effort.

> Hmm apart from implementing this inside rippled and David Fuelling's suggestion, I don't have any other ideas.

David Fuelling's suggestion is marked as resolved and is up for future versions of the meta data spec. This PR pertains to XLS-89d. We can revisit this implementation once the spec is updated with that suggestion in future. Let me know if you think otherwise.

> Are console warnings the best way to inform the user? If the xrpl package is used inside a browser (or) node, these console-warnings will be printed in the "developer-console". However, will developers read all the logs of their application? I'm trying to think of a better alternative to inform the user.

For now, to me it seems console warnings is the best way forward. xrpl.org should be updated for even better communication. Developer might or might not view the console warnings. Let me know once you have a better suggestion for this, I'll resume implementation after that, in the interest of development effort.

ckeshava:
> > Where have you included the warning in the xrpl-py client library, appropos the size of the meta-data field?
> 
> This will be implemented once, xrpl.js review progresses beyond the fact of deciding whether console warnings are the right thing to do, in the interest of development effort.

My bad, I am referring to metada-sized based warnings in this PR? Where are such validations?

> > Hmm apart from implementing this inside rippled and David Fuelling's suggestion, I don't have any other ideas.
> 
> David Fuelling's suggestion is marked as resolved and is up for future versions of the meta data spec. This PR pertains to XLS-89d. We can revisit this implementation once the spec is updated with that suggestion in future. Let me know if you think otherwise.

Yes, I understand; I'm only pointing out that his suggestion -- adding a `spec_version` field will make it easier for SDK to detect if a user wants to adhere to XLS-89d. The absence of such a field could have been used to silence these console-warnings.

> > Are console warnings the best way to inform the user? If the xrpl package is used inside a browser (or) node, these console-warnings will be printed in the "developer-console". However, will developers read all the logs of their application? I'm trying to think of a better alternative to inform the user.
> 
> For now, to me it seems console warnings is the best way forward. xrpl.org should be updated for even better communication. Developer might or might not view the console warnings. Let me know once you have a better suggestion for this, I'll resume implementation after that, in the interest of development effort.

I agree that xrpl.org should be updated with the XLS-89d standard. Additionally, I recommend using the Typedoc/JSDoc-strings in the `VaultCreate`, `MPTokenIssuanceCreate` transactions to document the existence of a new metadata standard.

I don't have any other objections pertaining to this implementation approach. Please proceed.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
Patel-Raj11: Handled this suggestion as console message header in the updated PR.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -131,15 +142,13 @@ export function validateMPTokenIssuanceCreate(
...
+  ) {
     throw new ValidationError(
-      'MPTokenIssuanceCreate: MPTokenMetadata must not be empty string',
-    )
-  }
-
-  if (typeof tx.MPTokenMetadata === 'string' && !isHex(tx.MPTokenMetadata)) {
-    throw new ValidationError(
-      'MPTokenIssuanceCreate: MPTokenMetadata must be in hex format',
+      `MPTokenIssuanceCreate: MPTokenMetadata must be in hex format and max ${MAX_MPT_META_BYTE_LENGTH} bytes.`,
```
ckeshava: The validation on size is present inside the `validateMPTokenMetadata` method inside the `common.ts` file. Why do you need it here?

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -696,3 +733,170 @@ const _DOMAIN_ID_LENGTH = 64
...
+    return { validationMessages, isValid: false }
+  }
+
+  let jsonMetaData: unknown
+
+  try {
+    jsonMetaData = JSON.parse(hexToString(input))
+  } catch (_err) {
+    validationMessages.push(
+      'MPTokenMetadata is not properly formatted as JSON.',
```
ckeshava: Can you also append the `_err` error-string in the response to the user? This could be helpful for debugging purposes.

Besides, we will not leak any sensitive/confidential information during the `JSON.parse(...)` step.

packages/xrpl/test/models/MPTokenIssuanceCreate.test.ts
```diff
@@ -40,7 +48,7 @@ describe('MPTokenIssuanceCreate', function () {
 
     assertInvalid(
       invalid,
-      'MPTokenIssuanceCreate: MPTokenMetadata must not be empty string',
+      'MPTokenIssuanceCreate: MPTokenMetadata must be in hex format and max 1024 bytes.',
```
ckeshava: This error message is confusing because the metadata string is empty. Can we exclude the console-warnings, if an empty string is used in the metadata field?

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
Patel-Raj11: This is a hard validation error and rippled enforces this size validation. So it's required here to throw an error.

`validateMPTokenMetadata` is validating the metadata structure and the size validation is present there as well because it's an exported method that other would use to collect all the validation messages.

packages/xrpl/test/models/MPTokenIssuanceCreate.test.ts
Patel-Raj11: MPTokenMetadata cannot be an empty string. rippled fails the transaction with `temMALFORMED`. Let me update the error message instead.

Patel-Raj11: Updated [2269745](https://github.com/XRPLF/xrpl.js/pull/3041/commits/2269745ffd972dfddeffab46301dd3845ab99dd6)

packages/xrpl/src/models/transactions/common.ts
Patel-Raj11: Updated here [2269745](https://github.com/XRPLF/xrpl.js/pull/3041/commits/2269745ffd972dfddeffab46301dd3845ab99dd6)

packages/xrpl/src/models/common/index.ts
```diff
@@ -218,3 +218,27 @@ export interface PriceData {
...
+export interface MPTokenMetadata {
+  ticker: string
+  name: string
+  desc?: string
+  icon: string
+  asset_class: string
+  asset_subclass?: string
+  issuer_name: string
+  urls?: MPTokenMetadataUrl[]
+  additional_info?: string
```
khancode: Can you move optional fields after required ones?

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -104,10 +107,18 @@ export interface MPTokenIssuanceCreate extends BaseTransaction {
...
-   * Arbitrary metadata about this issuance, in hex format.
+   * Optional arbitrary metadata about this issuance, encoded as a hex string and limited to 1024 bytes.
+   *
+   * The decoded value must be a UTF-8 encoded JSON object that adheres to the
+   * XLS-89d MPTokenMetadata standard.
+   *
+   * While adherence to the XLS-89d format is not mandatory, non-compliant metadata
+   * may not be discoverable by ecosystem tools such as explorers and indexers.
    */
   MPTokenMetadata?: string | null
```
khancode:
```suggestion
  MPTokenMetadata?: string
```
I know you didn't make this change but there shouldn't be `| null` since it's an optional field.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -696,3 +733,170 @@ const _DOMAIN_ID_LENGTH = 64
...
+  validationMessages: string[]
+} {
+  const validationMessages: string[] = []
+
+  if (!isHex(input) || input.length / 2 > MAX_MPT_META_BYTE_LENGTH) {
+    validationMessages.push(
+      `MPTokenMetadata must be in hex format and max ${MAX_MPT_META_BYTE_LENGTH} bytes.`,
+    )
+    return { validationMessages, isValid: false }
+  }
```
khancode: I would separate these error messages so it's cleaner.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -696,3 +733,170 @@ const _DOMAIN_ID_LENGTH = 64
...
+
+/**
+ * Validates if MPTokenMetadata adheres to XLS-89d standard.
+ *
+ * @param input - Hex encoded MPTokenMetadata.
+ * @returns Validation error messages and boolean denoting if MPTokenMetadata adheres to XLS-89d standard.
+ */
+export function validateMPTokenMetadata(input: string): {
+  isValid: boolean
+  validationMessages: string[]
```
khancode:
```suggestion
  validationMessages?: string[]
```
We can infer validity by checking if `validationMessages` is returned.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -696,3 +733,170 @@ const _DOMAIN_ID_LENGTH = 64
...
+    )
+    return { validationMessages, isValid: false }
+  }
+
+  if (
+    !(
+      jsonMetaData != null &&
+      typeof jsonMetaData === 'object' &&
+      !Array.isArray(jsonMetaData)
+    )
```
khancode:
```suggestion
    jsonMetaData == null ||
    typeof jsonMetaData !== 'object' ||
    Array.isArray(jsonMetaData)
```
Simplified and easy to read.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -696,3 +733,170 @@ const _DOMAIN_ID_LENGTH = 64
...
+    return { validationMessages, isValid: false }
+  }
+
+  if (
+    obj.additional_info != null &&
+    !(
+      isString(obj.additional_info) ||
+      (typeof obj.additional_info === 'object' &&
+        !Array.isArray(obj.additional_info))
+    )
```
khancode:
```suggestion
    obj.additional_info != null &&
    !isString(obj.additional_info) &&
    (
      typeof obj.additional_info !== 'object' ||
      Array.isArray(obj.additional_info)
    )
```
Nested negation is difficult to read.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -696,3 +733,170 @@ const _DOMAIN_ID_LENGTH = 64
...
+  if (
+    obj.urls != null &&
+    (!Array.isArray(obj.urls) ||
+      !obj.urls.every(isValidMPTokenMetadataUrlStructure))
+  ) {
+    validationMessages.push(
+      `urls field is not properly structured as per the XLS-89d standard.`,
+    )
+    return { validationMessages, isValid: false }
+  }
```
khancode:
```suggestion
  if (obj.urls != null) {
    if (!Array.isArray(obj.urls)) {
      validationMessages.push('urls must be an array as per XLS-89d.');
      return { validationMessages, isValid: false };
    }
    if (!obj.urls.every(isValidMPTokenMetadataUrlStructure)) {
      validationMessages.push('One or more urls are not structured per XLS-89d.');
      return { validationMessages, isValid: false };
    }
  }

```
Separate error messages here too.

Patel-Raj11: @khancode Added your suggestions in [756a81a](https://github.com/XRPLF/xrpl.js/pull/3041/commits/756a81a3242a1fb8f7435c8f9da763e7cc9ae959).
Also, updated the validation function to check for field count as well, so that users are warned if they add other fields that are not mentioned in the spec. It's not an issue to have such fields present since tools can ignore them, but good to have warnings.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -696,3 +737,181 @@ const _DOMAIN_ID_LENGTH = 64
...
+    validationMessages.push(`asset_subclass must be a string.`)
+    return validationMessages
+  }
+
+  if (
+    obj.additional_info != null &&
+    typeof obj.additional_info !== 'string' &&
+    (typeof obj.additional_info !== 'object' ||
+      Array.isArray(obj.additional_info))
+  ) {
```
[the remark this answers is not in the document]
Patel-Raj11: Good suggestion!

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -700,3 +741,180 @@ export function isDomainID(domainID: unknown): domainID is string {
...
+  }
+
+  if (
+    mptMPTokenMetadata.urls != null &&
+    !mptMPTokenMetadata.urls.every((ele) => ele.url.startsWith('https://'))
+  ) {
+    validationMessages.push(`url should be a valid https url.`)
+  }
+
+  return validationMessages
```
khancode, line 901:
```suggestion
  return validationMessages.length > 0 ? validationMessages : null
```
When there are no errors, I prefer returning `null` rather than returning an empty array to signal no errors. Then the calling function would only have to do a truthy check `if (validationMessages)` rather than `if (validationMessages.length > 0)`.

khancode: I would also rename it to `warningMessages` so it's more explicit.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -700,3 +741,180 @@ export function isDomainID(domainID: unknown): domainID is string {
...
 }
+
+/* eslint-disable max-lines-per-function -- Required here as structure validation is verbose. */
+/* eslint-disable max-statements -- Required here as structure validation is verbose. */
+
+/**
+ * Validates if MPTokenMetadata adheres to XLS-89d standard.
+ *
+ * @param input - Hex encoded MPTokenMetadata.
+ * @returns Validation error messages if MPTokenMetadata does not adheres to XLS-89d standard.
```
khancode:
```suggestion
 * @returns Validation warning messages if MPTokenMetadata does not adheres to XLS-89d standard.
```
These are warning messages rather than errors right?

packages/xrpl/src/models/transactions/common.ts
Patel-Raj11: I think in general an empty list is better design than a function returning an optional list. The consumers of the functions are guaranteed to get a list as opposed to null or a list. This avoids mistakes at consumer end of iterating, accessing elements of a list that is not even present. In xrpl4j, this convention is followed for simplicity. (i.e. Empty array/list is as good as no array/list).

packages/xrpl/src/models/transactions/common.ts
Patel-Raj11: This function just returns validation messages. Consumers decide whether to throw error or warnings. Updated to depict this.

packages/xrpl/src/models/transactions/common.ts
khancode: That's fair, LGTM 👍

PR #2857, opened 2024-12-19
Autofill function in Client not validating `DeliverMax` and `Amount` correctly 

JordiParraCrespo:
## Autofill function in Client not validating `DeliverMax` and `Amount` correctly 

Fixes an issue where the autofill function throws an error when passing amounts as objects.

### Context of Change

In JavaScript, objects are compared by reference, not by their property values. Here's an example to illustrate this behavior:

```
const a = {
  currency: 'USD',
  value: '31415',
  issuer: 'r9vbV3EHvXWjSkeQ6CAcYVPGeq7TuiXY2X',
};

const b = {
  currency: 'USD',
  value: '31415',
  issuer: 'r9vbV3EHvXWjSkeQ6CAcYVPGeq7TuiXY2X',
};

console.log(a === b); // Output: false
```
Because objects are compared by reference, even identical objects like a and b are considered different. This was causing incorrect validation in the autofill function when dealing with amounts represented as objects.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### Did you update HISTORY.md?

- [x] Yes
- [ ] No, this change does not impact library users

## Test Plan

I have added tests to verify that the bug is resolved.

PR #2689, opened 2024-05-03
Aliter: Implement DeliverMax alias in Payment transactions, through autofill method

ckeshava:
## High Level Overview of Change

This is an alternative solution for the DeliverMax alias, inspired from this discussion: https://github.com/XRPLF/xrpl.js/pull/2684#discussion_r1588500925

Note: Only one of these two PRs must be merged.

Both the PRs seek to achieve the same objectives, albeit through different means.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### Did you update HISTORY.md?

- [x] Yes
- [ ] No, this change does not impact library users

## Test Plan

Added unit tests

PR #3044, opened 2025-07-16
fix: allow `DeliverMax` in `Payment` transaction

khancode:

Reverts bug in rejecting `DeliverMax` param in `Payment` transaction

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

## Test Plan

Added unit test.

### DIFF
--- a/packages/xrpl/src/models/transactions/payment.ts
+++ b/packages/xrpl/src/models/transactions/payment.ts
@@ -208,12 +208,6 @@ export function validatePayment(tx: Record<string, unknown>): void {
   }
 
   checkPartialPayment(tx)
-
-  if (tx.DeliverMax != null) {
-    throw new ValidationError(
-      'PaymentTransaction: Cannot have DeliverMax in a submitted transaction',
-    )
-  }
 }
 
 function checkPartialPayment(tx: Record<string, unknown>): void {
--- a/packages/xrpl/test/models/payment.test.ts
+++ b/packages/xrpl/test/models/payment.test.ts
@@ -44,6 +44,11 @@ describe('Payment', function () {
     assertValid(payment)
   })
 
+  it(`verifies DeliverMax is valid`, function () {
+    payment.DeliverMax = '1234'
+    assertValid(payment)
+  })
+
   it(`Verifies memos correctly`, function () {
     payment.Memos = [
       {


packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -208,12 +208,6 @@ export function validatePayment(tx: Record<string, unknown>): void {
   }
 
   checkPartialPayment(tx)
-
```
ckeshava, line 229: Hello, I had worked on DeliverMax change in 2024. I had added a validation which compared the `Amount` and `DeliverMax` fields. If both of the fields were present and both had different values, we need to throw an error.

This validation appears to have been removed in the current version of the codebase.

Can you add this validation back in the code?

For further context, this PR builds upon this validation: https://github.com/XRPLF/xrpl.js/pull/2857

khancode: @ckeshava can you share the commit of when you added this validation?

Patel-Raj11 (approved):
LGTM!

Patel-Raj11:
@mvadari Why did we add `DeliverMax` in payment transaction now? Was this to accommodate some changes for Batch transactions? Also, I don't see this in xrpl-py.

mvadari:
> @mvadari Why did we add `DeliverMax` in payment transaction now? Was this to accommodate some changes for Batch transactions? Also, I don't see this in xrpl-py.

I didn't add it

ckeshava:
@mvadari : I believe @Patel-Raj11  is asking why `DeliverMax` was included in [this PR change](https://github.com/XRPLF/xrpl.js/pull/2966). 

I wrote the changes pertaining to `DeliverMax` more than one year ago, [in this commit](https://github.com/XRPLF/xrpl.js/pull/2689).

packages/xrpl/src/models/transactions/payment.ts
ckeshava: Hello Omar, this is the [PR for the original change](https://github.com/XRPLF/xrpl.js/pull/2689)

ckeshava: The integ-tests in this file evaluate those aspects of the validation: https://github.com/XRPLF/xrpl.js/pull/2689/files#diff-4fbead92a8c0e3529d7c7a8b04d552b22a79264ea9147c0f7e79bf2ee35e16a2

khancode: @ckeshava the validation is still here but it was refactored - https://github.com/XRPLF/xrpl.js/blob/2309d591ec0c07d40ed00de69f8490d033bbbbcd/packages/xrpl/src/client/index.ts#L694-L696

mvadari:
> @mvadari : I believe @Patel-Raj11 is asking why `DeliverMax` was included in [this PR change](https://github.com/XRPLF/xrpl.js/pull/2966).

A mistake

packages/xrpl/src/models/transactions/payment.ts
khancode: Integ tests are also still here.

ckeshava: @khancode  The refactor has two mistakes.
- The `handleDeliverMax` method is only invoked inside the [autofill-related function](https://github.com/XRPLF/xrpl.js/blob/2309d591ec0c07d40ed00de69f8490d033bbbbcd/packages/xrpl/src/client/index.ts#L695). Consequently, non-autofill code-paths will not validate the `DeliverMax` field at all. In the original implementation, this validation was performed inside the `Payment` transaction model.

More importantly, what is the benefit offered by this refactored? What is the goal of that PR change?

- In the `handleDeliverMax` method, [the `DeliverMax` field is deleted here](https://github.com/XRPLF/xrpl.js/blob/2309d591ec0c07d40ed00de69f8490d033bbbbcd/packages/xrpl/src/sugar/autofill.ts#L421). What is the need for removing the `DeliverMax` field? There is no harm in retaining it.

From an end-user's perspective, this causes confusion. When users look at the transaction response, it will not match their inputs while constructing the transaction object.

ckeshava: > Integ tests are also still here.

Did you look at [this branch of mine](https://github.com/ckeshava/xrpl.js/pull/new/checkoutDeliverMax)? I had referenced it in the context of the Jira issue. That test case in that branch is failing. Clearly, that cannot be present in the current code base. If it was, this error would have been caught by the CI.

ckeshava: Furthermore, the Payment integration tests are missing one corner case. This case was present in my original commit, however it appears to have been removed in recent history.

The below case needs to be included inside: `packages/xrpl/test/integration/transactions/payment.test.ts`

```
it(
    'Validate Payment transaction API v2: Payment Transaction: identical DeliverMax and Amount fields',
    async () => {
      paymentTx.DeliverMax = '2000'
      paymentTx.Amount = '1000'

      const result = await testTransaction(
        testContext.client,
        paymentTx,
        senderWallet,
      )

      // you need to assert for a validation-error
    },
    TIMEOUT,
  )
```

_Note: This test case is different from what I present in my custom branch. This code tests the `autofill` code-path, unlike the unit-test._

Please go through the original commit for details of the implementation.

packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -226,12 +226,6 @@ export function validatePayment(tx: Record<string, unknown>): void {
...
-  if (tx.DeliverMax != null) {
```
ckeshava, line 230: @khancode  You still need to invoke handleDeliverMax method at this line. In the present state, the `Payment` transaction model does not do any validation on `DeliverMax` at all.

packages/xrpl/src/models/transactions/payment.ts
ckeshava: In the current PR, the deliver_max field is being validated inside the `autofill` method. However, if users do not make use of the autofill functionality, then the DeliverMax field will not be validated for compatibility.

In the interest of time/developer-effort, this is an acceptable compromise. If any users need/want to use the `Payment` transaction-validation without the autofill-functonality, please raise an issue.

ckeshava (approved):

PR #3059, opened 2025-08-08
Add method to derive a wallet from PrivateKey

interc0der:

As there are more wallets natively supporting the XRP Ledger, private keys are being passed around in different context and there should be a method in this codebase to derive a wallet from a private key. Since we have (2) algorithms and this package correctly handles both, this is a necessary addition to the package and offers immense value.

### Context of Change

- Adds a new static method to the Wallet class to derive an XRPL wallet using a private key.

```
  /**
   * Derives a wallet from a private key.
   *
   * @param privateKey - A string used to generate a keypair (publicKey/privateKey) to derive a wallet.
   * @returns A Wallet derived from a private key.
   */
  public static fromPrivateKey(privateKey: string): Wallet {
    return new Wallet(derivePublicKey(privateKey), privateKey)
  }
```

### Type of Change

## Test Plan

- Added a few tests in the Wallet folder to test derivations from mocked keypairs. These tests ensure that the correct public address and public key are generated from the private key.
- Ensure that the library correctly handles a malformed private key and properly throws an error


packages/ripple-keypairs/src/signing-schemes/ed25519/index.ts
```diff
@@ -19,6 +19,27 @@ const ed25519: SigningScheme = {
...
+    )
+
+    const normalizedPrivateKey = privateKey.startsWith(ED_PREFIX)
+      ? privateKey.slice(2)
+      : privateKey
+
+    const buffer = Buffer.from(normalizedPrivateKey, 'hex')
+
+    const publicKey = ED_PREFIX + bytesToHex(nobleEd25519.getPublicKey(buffer))
+    return { privateKey, publicKey }
```
ckeshava: Would you consider adding the 'ED' prefix to the returned `privateKey` ? In this file, the `deriveKeypair` method returns both the `publicKey` and the `privateKey` along with the `ED` prefix.

ckeshava: Having said that, I'm aware that the `deriveKeypair` method inside the `secp256k1` (`packages/ripple-keypairs/src/signing-schemes/secp256k1/index.ts`) module does not explicitly add the `00` prefix. 

Do you know what is the expected convention in cryptographic methods?

packages/xrpl/test/wallet/index.test.ts
```diff
@@ -437,6 +437,51 @@ describe('Wallet', function () {
...
   })
 
+  describe('from PrivateKey', function () {
+    describe('using secp256k1 private key', function () {
+      const mockWallet_secp256k1 = {
+        address: 'rhvh5SrgBL5V8oeV9EpDuVszeJSSCEkbPc',
+        publicKey:
+          '030E58CDD076E798C84755590AAF6237CA8FAE821070A59F648B517A30DC6F589D',
+        privateKey:
+          '00141BA006D3363D2FB2785E8DF4E44D3A49908780CB4FB51F6D217C08C021429F',
```
ckeshava: Hello, how did you generate this mocked `privateKey` and `publicKey` data? Which library/tool did you use?

packages/ripple-keypairs/src/index.ts
```diff
@@ -95,6 +95,15 @@ function deriveAddress(publicKey: string): string {
   return deriveAddressFromBytes(hexToBytes(publicKey))
 }
 
+function derivePublicKey(privateKey: string): string {
+  const algorithm = getAlgorithmFromPrivateKey(privateKey)
+
+  if (algorithm === 'ecdsa-secp256k1') {
+    return secp256k1.deriveKeypairFromPrivateKey(privateKey).publicKey
+  }
+  return ed25519.deriveKeypairFromPrivateKey(privateKey).publicKey
```
ckeshava: Can you add an explicit if-condition that does:

```
if (algorithm === 'ed25519') {
    return ed25519.deriveKeypairFromPrivateKey(privateKey).publicKey
  } else
throw new Error('Unknown signing scheme algorithm')
```

In the future, if support for new SigningSchemes are provided in the xrpl.js library, then this code is protected from incorrect default behavior.

ckeshava (changes requested):
Thanks for this work. Please address the linter failures.

How do you envision the usage of your method? Will users generate their privateKeys on a trusted execution environment and use it to sign/verify the transactions?

Since there are existing methods to generate key-pairs from Seed, Entropy and Mnemonic, I'd like to understand the need for this method.

interc0der:
> Thanks for this work. Please address the linter failures.
> 
> How do you envision the usage of your method? Will users generate their privateKeys on a trusted execution environment and use it to sign/verify the transactions?
> 
> Since there are existing methods to generate key-pairs from Seed, Entropy and Mnemonic, I'd like to understand the need for this method.

Exactly - there are methods to instantiate a wallet from Seed, Entropy and Mnemonic. There should also be a method for PrivateKey, not only for completeness, but for wallets that support both generating and deriving public addresses using a pk.

Usage - Hmm, same usage as the methods above (Seed, Entropy and Mnemonic). Users can generate/derive a wallet using a private key. This method could be used for verifying that a private key is valid, and it can also be used to instantiate a Wallet for future signing.

The linting issues seem to be coming from the binary-codec package which I did not modify. I am not going to correct an issue related to another change. I will continue to sync my branch with corrective commits.

packages/ripple-keypairs/src/signing-schemes/ed25519/index.ts
interc0der: > Would you consider adding the 'ED' prefix to the returned `privateKey` ? In this file, the `deriveKeypair` method returns both the `publicKey` and the `privateKey` along with the `ED` prefix.

https://github.com/interc0der/xrpl.js/blob/bed93562ebde0e13a2c919e74a5f3ab12d99ee0b/packages/ripple-keypairs/src/signing-schemes/ed25519/index.ts#L11-L20
The `deriveKeypair` method does not return the prefix. What are you referencing?

interc0der: > Having said that, I'm aware that the `deriveKeypair` method inside the `secp256k1` (`packages/ripple-keypairs/src/signing-schemes/secp256k1/index.ts`) module does not explicitly add the `00` prefix.
> 
> Do you know what is the expected convention in cryptographic methods?

This is why the `deriveKeypairFromPrivateKey` matches the conventions in the `sign` method. The function allows for a prefixed, or non-prefixed key - and derives the public address with the normalized private key.

Not sure what you mean by the expected convention in cryptographic methods. 
- Secp256k1 expects/generates a 33 byte public key and a 32 byte private key. 
- Ed25519 expects/generates a 32 byte public key and and 32 byte private key.

Key are normalized and padded according to algorithm. See this nice reference table: https://github.com/interc0der/xrpl.js/blob/main/packages/ripple-keypairs/src/utils/getAlgorithmFromKey.ts

packages/xrpl/test/wallet/index.test.ts
interc0der: I believe I used this library to generate them. Why? - The unit check confirms that it is a valid keypair...

packages/ripple-keypairs/src/index.ts
interc0der: I didn't do this, because it was causing a linting issue. 

I have added it and suppressed the warning.
https://github.com/interc0der/xrpl.js/blob/d6ba366155e56cbd9e736efcd9073717c01b7b69/packages/ripple-keypairs/src/index.ts#L98-L108

interc0der:
I mistakenly reverted changes breaking the commit tree. Moved the same commits over to a fresh pull request.
https://github.com/XRPLF/xrpl.js/pull/3063

@ckeshava If there are any comments here that you feel have not been resolved, can you move them over?

packages/xrpl/test/wallet/index.test.ts
ckeshava: Okay. As long as the newly introduced method was not used to generate the test case, it is acceptable.

PR #3065, opened 2025-08-25
Fix node.js process termination due to delayed websocket error callback

Patel-Raj11:
Node.js process terminates, when an error is thrown from a promise rejection and not handled immediately. This PR catches the error and fixes the issue.

### Context of Change
In `validator-history-service` node.js process running inside `vhs-connections` container was terminating with these logs statements.

```
"message":"","source_type":"kubernetes_logs","stream":"stderr"}
"message":"","source_type":"kubernetes_logs","stream":"stderr"}
"message":"    at processTicksAndRejections (node:internal/process/task_queues:95:5)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"    at /validator-history-service/node_modules/xrpl/dist/npm/client/connection.js:160:37","source_type":"kubernetes_logs","stream":"stderr"}
"message":"    at reject (/validator-history-service/node_modules/xrpl/dist/npm/client/RequestManager.js:44:19)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"Thrown at:","source_type":"kubernetes_logs","stream":"stderr"}
"message":"}","source_type":"kubernetes_logs","stream":"stderr"}
"message":"  }","source_type":"kubernetes_logs","stream":"stderr"}
"message":"    }","source_type":"kubernetes_logs","stream":"stderr"}
"message":"          at Zlib.processCallback (node:zlib:588:8)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"          at Zlib.cb (node:internal/streams/transform:191:7)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"          at onwrite (node:internal/streams/writable:688:7)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"          at afterWrite (node:internal/streams/writable:710:5)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"          at /validator-history-service/node_modules/ws/lib/permessage-deflate.js:458:7","source_type":"kubernetes_logs","stream":"stderr"}
"message":"          at /validator-history-service/node_modules/ws/lib/permessage-deflate.js:326:9","source_type":"kubernetes_logs","stream":"stderr"}
"message":"          at /validator-history-service/node_modules/ws/lib/sender.js:510:21","source_type":"kubernetes_logs","stream":"stderr"}
"message":"      data: Error: The socket was closed while data was being compressed","source_type":"kubernetes_logs","stream":"stderr"}
"message":"        at Zlib.processCallback (node:zlib:588:8) {","source_type":"kubernetes_logs","stream":"stderr"}
"message":"        at Zlib.cb (node:internal/streams/transform:191:7)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"        at onwrite (node:internal/streams/writable:688:7)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"        at afterWrite (node:internal/streams/writable:710:5)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"        at /validator-history-service/node_modules/ws/lib/permessage-deflate.js:458:7","source_type":"kubernetes_logs","stream":"stderr"}
"message":"        at /validator-history-service/node_modules/ws/lib/permessage-deflate.js:326:9","source_type":"kubernetes_logs","stream":"stderr"}
"message":"        at /validator-history-service/node_modules/ws/lib/sender.js:514:9","source_type":"kubernetes_logs","stream":"stderr"}
"message":"        at callCallbacks (/validator-history-service/node_modules/ws/lib/sender.js:581:33)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"        at /validator-history-service/node_modules/xrpl/dist/npm/client/connection.js:45:28","source_type":"kubernetes_logs","stream":"stderr"}
"message":"    error: DisconnectedError: The socket was closed while data was being compressed","source_type":"kubernetes_logs","stream":"stderr"}
"message":"    type: 'reject',","source_type":"kubernetes_logs","stream":"stderr"}
"message":"  data: {","source_type":"kubernetes_logs","stream":"stderr"}
"message":"    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {","source_type":"kubernetes_logs","stream":"stderr"}
"message":"    at /validator-history-service/node_modules/xrpl/dist/npm/client/connection.js:160:37","source_type":"kubernetes_logs","stream":"stderr"}
"message":"    at RequestManager.reject (/validator-history-service/node_modules/xrpl/dist/npm/client/RequestManager.js:44:19)","source_type":"kubernetes_logs","stream":"stderr"}
"message":"XrplError: No existing promise with id 0","source_type":"kubernetes_logs","stream":"stderr"}
"message":"","source_type":"kubernetes_logs","stream":"stderr"}
"message":"                  ^","source_type":"kubernetes_logs","stream":"stderr"}
"message":"            throw new errors_1.XrplError(`No existing promise with id ${id}`, {","source_type":"kubernetes_logs","stream":"stderr"}
"message":"/validator-history-service/node_modules/xrpl/dist/npm/client/RequestManager.js:44","source_type":"kubernetes_logs","stream":"stderr"}

```

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan
Added test to simulate the issue.

### DIFF
--- a/packages/xrpl/test/connection.test.ts
+++ b/packages/xrpl/test/connection.test.ts
@@ -981,8 +981,17 @@ describe('Connection', function () {
   )
 
   it(
-    'TimeoutError on send',
+    'Delayed websocket error callback on send',
     async () => {
+      const traceMessages: string[] = []
+      // @ts-expect-error -- Testing private member
+      clientContext.client.connection.trace = (
+        id: string,
+        message: string,
+      ): void => {
+        traceMessages.push(`${id}: ${message}`)
+      }
+
       // @ts-expect-error -- Testing private member
       clientContext.client.connection.ws.send = function (
         _ignore,
@@ -1008,6 +1017,11 @@ describe('Connection', function () {
       await new Promise((resolve) => {
         setTimeout(resolve, 2000)
       })
+
+      assert.includeMembers(traceMessages, [
+        'send: send errored after connection was closed: [XrplError(No existing promise with id 1, ' +
+          '{"type":"reject","error":{"name":"DisconnectedError","data":{"message":"some error"}}})]',
+      ])
     },
     TIMEOUT,
   )


pdp2121 (approved):
LGTM

packages/xrpl/src/client/connection.ts
```diff
@@ -301,7 +301,18 @@ export class Connection extends EventEmitter {
     >(request, timeout ?? this.config.timeout)
     this.trace('send', message)
     websocketSendAsync(this.ws, message).catch((error) => {
-      this.requestManager.reject(id, error)
+      try {
+        this.requestManager.reject(id, error)
+      } catch (err) {
+        if (err instanceof XrplError) {
+          this.trace(
```
ckeshava, line 308: Hello Raj, the default implementation of the `trace` method is empty.

Image

Are you able to see this trace messages in the log statements? 

The constructor of the `Connection` class accepts a `trace` callback, however I do not know if it is set appropriately in all the object-instances of the `Connection` class.

Image

packages/xrpl/test/connection.test.ts
```diff
@@ -979,4 +979,36 @@ describe('Connection', function () {
...
+        _ignore,
+        sendCallback,
+      ): void {
+        // server_info request will timeout in 0.5s, but we send an error after 1s
+        setTimeout(() => {
+          sendCallback({ message: 'some error' })
+        }, 1000)
+      }
+
+      await clientContext.client.connection
```
ckeshava, line 1006: In this change, what is the method under test? Is it `connection.request`? 

Shouldn't you test against the output of `this.trace` response? Your change prints the error into `this.trace` instead of throwing an `Error`.

packages/xrpl/src/client/connection.ts
Patel-Raj11: Ya, we can provide a callback function when instantiating a Client object in [ClientOptions](https://github.com/XRPLF/xrpl.js/blob/5c816e48d00c19b3b46dee62fe55a2a69f978cc4/packages/xrpl/src/client/index.ts#L248), if it's not provided it defaults to [console.log](https://github.com/XRPLF/xrpl.js/blob/5c816e48d00c19b3b46dee62fe55a2a69f978cc4/packages/xrpl/src/client/connection.ts#L137). That will be invoked for every trace message we send in the client's code. Now, they can do whatever they want to do with those messages, like console log it, or store somewhere etc.

They might use something like this.
```
function logger(id, message) {
  console.log(`[${id}] ${message}`);
}

const client = new Client("wss://s.devnet.rippletest.net:51233/", {
    trace: logger,
  });
```

packages/xrpl/test/connection.test.ts
Patel-Raj11: The thing I am testing here is that, we should not abruptly terminate the test in case when `ws.send` results in error and `promisesAwaitingResponse` does not have the id of that request. (Have induced via earlier request closer due to short request timeout and erroring `ws.send` after that).

So, if we keep the test and revert my code change, it will abruptly terminate the test with error message XrplError: `No existing promise with id 1`. 

To be explicit, I will push a change to assert for the above error message in trace messages as well. 💯

Patel-Raj11: Added assert for trace message [76c0b44](https://github.com/XRPLF/xrpl.js/pull/3065/commits/76c0b446c72e85a6679878a1cf818fa506254999)

packages/xrpl/src/client/connection.ts
ckeshava: Here is the relevant snippet from xrpl.js Connection class constructor: 
```
if (options.trace) {
      // eslint-disable-next-line no-console -- Used for tracing only
      this.trace = console.log
    }
```



image

In the VHS, we are initializing all the Client objects like: `new Client(url)`. Hence, `options === {}` during the initialization of the `Connection` class. This is the default initialization of the trace method: (this is what I pointed out in my previous message)
```
  public readonly trace: (id: string, message: string) => void = () => {}
```

Without explicit `options.trace` initialization in the VHS source code, we will not get appropriate `console.log` messages.

ckeshava: Note: The unit test below mocks the `connection.trace` method, hence it does not traverse the code-path that I raised in the above message.

I'm only concerned about losing the `this.trace` messages in the logs. However, if you are observing such caught-errors in the VHS `stg` instance (for ex: `processTicksAndRejections: `), then I will withdraw this comment and investigate further myself. I do not want to delay this PR.

Patel-Raj11: Make sense, we could do that in VHS when initializing the client. But it will print whole lot of trace messages for each connection, anyways this discussion can happen while improving VHS codebase.

I choose to send trace messages in this PR's error handling because, throwing error was not an option here. Maybe console.error can be done instead of pushing it into trace messages.

Patel-Raj11: The codepath highlighted in above image is already tested here with true, false, and a function. You are right it's not tested for undefined value completely missing out of the trace messages.

https://github.com/XRPLF/xrpl.js/blob/5c816e48d00c19b3b46dee62fe55a2a69f978cc4/packages/xrpl/test/connection.test.ts#L182

ya, it would be helpful if further testing for trace messages is done on another PR. and as far as VHS is concerned it's ok to not trace this message as we would know if container stops with error logs.

ckeshava: alright

PR #3081, opened 2025-09-15
Support for Dynamic MPTs (XLS-94D)

ckeshava:

Specification is located here: https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0094-dynamic-MPT
CPP implementation can be found here: https://github.com/XRPLF/rippled/pull/5705

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

A few unit and integration tests have been added.


packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -53,6 +53,15 @@ export enum MPTokenIssuanceCreateFlags {
...
   tfMPTCanClawback = 0x00000040,
+
+  tfMPTCanMutateCanLock = 0x00000002,
+  tfMPTCanMutateRequireAuth = 0x00000004,
+  tfMPTCanMutateCanEscrow = 0x00000008,
+  tfMPTCanMutateCanTrade = 0x00000010,
+  tfMPTCanMutateCanTransfer = 0x00000020,
+  tfMPTCanMutateCanClawback = 0x00000040,
+  tfMPTCanMutateMetadata = 0x00010000,
+  tfMPTCanMutateTransferFee = 0x00020000,
```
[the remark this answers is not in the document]
ckeshava: This is not a disadvantage because these flags will be used in different contexts. It has been clarified here: https://github.com/XRPLF/rippled/pull/5705#discussion_r2350098121

Furthermore, this is out of scope of the client library implementation.

ckeshava:
I'm working on fixing the integration and browser tests, however I'd love to get a round of code-reviews. I'm not entirely sure why the integ-tests are not connecting to the local rippled standalone instance.

.github/workflows/nodejs.yml
```diff
@@ -154,6 +154,16 @@ jobs:
...
+      - name: Check if Docker container is running
+        run: |
+          docker ps
+          docker ps -a
+          docker inspect rippled-service
+          if ! docker ps | grep -q rippled-service; then
+            echo "rippled-service Docker container is not running"
+            exit 1
+          fi
+
```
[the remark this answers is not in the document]
ckeshava: It is not useful to poll for the existence of the rippled docker container. If the docker container failed to start-up correctly, there is no chance for it to restart in a healthy manner.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -69,6 +78,15 @@ export interface MPTokenIssuanceCreateFlagsInterface
   tfMPTCanTrade?: boolean
   tfMPTCanTransfer?: boolean
   tfMPTCanClawback?: boolean
+
+  tmfMPTCanMutateCanLock?: boolean
```
pdp2121: These flags are only available for MutableFlags field. Should it be separated into its own interface, similar to what we did with `SetFlag` options in `AccountSet`?

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -120,6 +138,7 @@ export interface MPTokenIssuanceCreate extends BaseTransaction {
   MPTokenMetadata?: string
 
   Flags?: number | MPTokenIssuanceCreateFlagsInterface
```
pdp2121, line 226: Flags cannot accept the mutable flags so the above change will be needed

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
```diff
@@ -37,6 +65,31 @@ export enum MPTokenIssuanceSetFlags {
 export interface MPTokenIssuanceSetFlagsInterface extends GlobalFlagsInterface {
   tfMPTLock?: boolean
   tfMPTUnlock?: boolean
+  // The below flags are used in the MutableFlags field
```
pdp2121: ditto

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
ckeshava: 1. The flags associated with Dynamic MPTs have `tmf...` in their prefix. This is used to indicate `transaction-mutable-flag`. This difference in names (`tf...` versus `tmf...`) should help users differentiate between the intent behind the flags.
2. If we use a separate interface for these flags, it could potentially cause confusion. The xrpl.org documentation (and the rippled cpp code, other programming language clients) organize these flags under the `MPTokenIssuanceCreate` transaction's flags. This change would have been more impactful if it was done at the XLS level, instead of only xrpl.js deviating from the specification.

.github/workflows/nodejs.yml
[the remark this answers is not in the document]
ckeshava: Why do we need to test the `curl` command? If the `docker ps` response does not list the `rippled-service` container, why do I need other ways to detect the existence of the container?

Jasonradford51 (approved):

Jasonradford51 (commented):
complete

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
```diff
@@ -11,4 +11,92 @@ export interface MPTokenIssuance extends BaseLedgerEntry, HasPreviousTxnID {
   MPTokenMetadata?: string
   OwnerNode?: string
   LockedAmount?: string
+  DomainID?: string
```
kuan121, line 73: What do we need DomainID for this object?

ckeshava:
@Patel-Raj11  The relevant tests in Github CI have passed.

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
```diff
@@ -11,4 +11,92 @@ export interface MPTokenIssuance extends BaseLedgerEntry, HasPreviousTxnID {
   MPTokenMetadata?: string
   OwnerNode?: string
   LockedAmount?: string
```
kuan121, line 68: nit: Do we typically take the opportunity to remove or deprecate unused fields? `LockedAmount` was removed according to [this]( https://github.com/XRPLF/xrpl.js/blob/d6ec04a89cb204484faa1bdfdb41e8c8ff3ae6d2/.github/workflows/nodejs.yml).

packages/ripple-binary-codec/src/enums/definitions.json
```diff
@@ -680,6 +680,16 @@
         "type": "UInt32"
       }
     ],
+    [
+      "MutableFlags",
+      {
+        "isSerialized": true,
+        "isSigningField": true,
+        "isVLEncoded": false,
+        "nth": 53,
```
kuan121, line 717: For my understanding, how is the value 53 determined?

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
```diff
@@ -11,4 +11,92 @@ export interface MPTokenIssuance extends BaseLedgerEntry, HasPreviousTxnID {
...
+  DomainID?: string
+  MutableFlags: number
+}
+
+export interface MPTokenIssuanceFlagsInterface {
```
kuan121, line 90: MPTokenIssuanceFlagsInterface is exported, but where it's used?

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
```diff
@@ -11,4 +11,92 @@ export interface MPTokenIssuance extends BaseLedgerEntry, HasPreviousTxnID {
...
+   * Allows field MPTokenMetadata to be modified
+   */
+  lmfMPTCanMutateMetadata?: boolean
+  /**
+   * Allows field TransferFee to be modified
+   */
+  lmfMPTCanMutateTransferFee?: boolean
+}
+
+export enum MPTokenIssuanceFlags {
```
kuan121, line 135: Where is `MPTokenIssuanceFlags` used?

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -53,6 +53,15 @@ export enum MPTokenIssuanceCreateFlags {
    * to clawback value from individual holders.
    */
   tfMPTCanClawback = 0x00000040,
+
+  tmfMPTCanMutateCanLock = 0x00000002,
```
kuan121: No documentation/comment like the above-mentioned flags?

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
kuan121: >The xrpl.org documentation (and the rippled cpp code, other programming language clients) organize these flags under the MPTokenIssuanceCreate transaction's flags

Is the xrpl.org documentation updated already?

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -141,6 +160,7 @@ export function validateMPTokenIssuanceCreate(
   validateOptionalField(tx, 'MPTokenMetadata', isString)
   validateOptionalField(tx, 'TransferFee', isNumber)
   validateOptionalField(tx, 'AssetScale', isNumber)
+  validateOptionalField(tx, 'MutableFlags', isNumber)
```
kuan121, line 254: Do we have a test case where we provide an invalid `MutableFlags` value, e.g., 63, as mentioned in https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0094-dynamic-MPT#23-failure-conditions?

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
```diff
@@ -75,4 +75,65 @@ describe('MPTokenIssuanceDestroy', function () {
...
+        // set the transfer fee to a non-zero value
+        TransferFee: 200,
+      }
+
+      await testTransaction(
+        testContext.client,
+        setTransferFeeTx,
+        testContext.wallet,
+      )
+
```
kuan121, line 203: Should we verify the transferFee is set successfully?

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
```diff
@@ -75,4 +75,65 @@ describe('MPTokenIssuanceDestroy', function () {
...
+        Account: testContext.wallet.classicAddress,
+        MPTokenIssuanceID: mptID!,
+        MutableFlags: MPTokenIssuanceSetFlags.tmfMPTClearCanTransfer,
+      }
+
+      await testTransaction(
+        testContext.client,
+        clearTransferFlagTx,
+        testContext.wallet,
+      )
```
kuan121, line 470: Should we verify that the transaction fail with a certain code?

packages/xrpl/test/models/MPTokenIssuanceSet.test.ts
```diff
@@ -75,4 +86,93 @@ describe('MPTokenIssuanceSet', function () {
 
     assertInvalid(invalid, 'MPTokenIssuanceSet: flag conflict')
   })
+
```
kuan121, line 121: Do we usually write tests for all the failure conditions mentioned in the standard, e.g., https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0094-dynamic-MPT#42-failure-conditions

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
ckeshava: `DomainID` was a part of the `MPTokenIssuance` ledger object, however the previous developer forgot to include it in the xrpl.js model.

Source: https://github.com/XRPLF/rippled/blob/294e03ecf53e00043c801da4cbe8a2133b91c8e8/include/xrpl/protocol/detail/ledger_entries.macro#L417

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
ckeshava: Hello Kuan, the commit that you referenced points to a github workflow file. I don't understand how that relates to this piece of code.

LockedAmount is still a part of the MPTokenIssuance ledger-object. [Here is the source]( https://github.com/XRPLF/rippled/blob/294e03ecf53e00043c801da4cbe8a2133b91c8e8/include/xrpl/protocol/detail/ledger_entries.macro#L413). Why do you think it has been removed?

packages/ripple-binary-codec/src/enums/definitions.json
ckeshava: Every field used in XRPL transactions/ledger-objects is associated with a `type-code` + a `field-code`. This is used to serialize/deserialize the values in a unique manner. [More explanation can be found here](https://xrpl.org/docs/references/protocol/binary-format#field-codes).

The developer of an amendment [picks the next available field-code](https://github.com/XRPLF/rippled/blob/294e03ecf53e00043c801da4cbe8a2133b91c8e8/include/xrpl/protocol/detail/sfields.macro#L117). As long there is no conflict in the use of field-code, any unique number should suffice.

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
ckeshava: We expect the users of the library to make use of this interface. However, I do not have data regarding its historical usage.

[Here is a utility method](https://github.com/XRPLF/xrpl.js/blob/006ef0f7b614a351ada204bb10529546d9758b17/packages/xrpl/src/models/utils/flags.ts#L37) that translates `AccountRoot` flags from numeric representation into a Typescript interface. We had envisioned to write such utility methods for all the ledger-object + transaction-types, however due to lack of staff, that has not been accomplished yet.

Furthermore, I vaguely remember that [one of the flags-utility methods](https://github.com/XRPLF/xrpl.js/blob/006ef0f7b614a351ada204bb10529546d9758b17/packages/xrpl/src/models/utils/flags.ts#L143) make use of the FlagsInterface/FlagsEnum. However, I don't remember the details.

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
ckeshava: Same answer as this one: https://github.com/XRPLF/xrpl.js/pull/3081#discussion_r2392366948
Ideally, we would like to write the `flags (number) -> flags (interface)` transformation. 

Even without this transformation, downstream users (like block-explorers, custody wallets, etc) can use this enum to parse/display the flags on the MPTokenIssuance object.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
ckeshava: I'll add comments after I change the flag names in my next commit

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
ckeshava: `testTransaction` verifies that the [transaction made its way into a validated ledger](https://github.com/XRPLF/xrpl.js/blob/006ef0f7b614a351ada204bb10529546d9758b17/packages/xrpl/test/integration/utils.ts#L288).
image

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
ckeshava: what do you mean? which error code are you talking about?

packages/xrpl/test/models/MPTokenIssuanceSet.test.ts
ckeshava: We can only check for `tem` error codes, we cannot validate the other failure cases without reading the state of the blockchain.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
ckeshava: No, the XLS specification is a good source of info for unreleased amendments. Usually, you can find the pre-release documentation for amendments on https://opensource.ripple.com/ as well, however LendingProtocol docs have not been published yet. You can look at pre-release docs for other amendments on that website.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
ckeshava: updated in 565402bda

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
ckeshava: added testcase in 565402bda

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
ckeshava: [The `MPTokenIssuanceSet` transaction](https://github.com/XRPLF/rippled/blob/develop/src/xrpld/app/tx/detail/MPTokenIssuanceSet.cpp) has at least 13 (`tem` error codes) + 23 (`tec` error codes) = 36 cases of unique failure possibilities.

To perform an exhaustive test of all these cases would be a lot effort. Besides, these cases are covered under rippled unit-tests + integration tests from the QE team.

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
kuan121: Sorry, this is the [link](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033-multi-purpose-tokens#a117-why-was-lockedamount-removed-from-the-spec) I meant to provide. It says
> A.1.17. Why was LockedAmount removed from the spec?

Note that this isn't blocker though.

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
kuan121: OK. I've done my review based on the [spec](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033-multi-purpose-tokens) and the [documentation](https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/mptokenissuance#mptokenissuance-fields), and `DomainID` is not mentioned in either. What is the source of truth? What is the DomainID of MPTokenIssuance?

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
pdp2121: We were separating the flags that can be set with `Flags` and `SetFlags` for `AccountSet` transactions:
https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/models/transactions/accountSet.ts
I think it's clearer that way, and also allows users to distinguish which flags they can set with `Flags` and `MutableFlags`. With this setup user can set a mutable flag with `Flags` field and there will be no warning, so I think it's more confusing

packages/ripple-binary-codec/src/enums/definitions.json
kuan121: Thanks Chenna. This's very useful information.

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
kuan121: Thanks for the context!

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
kuan121: What I meant was should we read the MPTokenIssuance object after the transaction is submitted and verify that the fee is 200 — [similar to how we verify the MaximumAmount after creating an MPT?](https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/test/integration/transactions/mptokenIssuanceCreate.test.ts#L46)

packages/xrpl/test/models/MPTokenIssuanceSet.test.ts
kuan121: Right — in this case, should we test all the tem failure conditions? If we haven’t done that before, should we start? I agree we don’t need to be exhaustive in integration tests, but for unit tests, I don’t see why not — especially since the spec developer has already listed the conditions for us. We could even use AI to translate them into tests.

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
kuan121: > what do you mean? which error code are you talking about?

I might’ve been looking at the wrong test case. What I meant was should we verify that `MutableFlags` was actually updated by reading it back and checking the value? In general, when we write a test that updates state on a remote server, the steps will be something like:
```
1. Send a request to update X from 1 → 2
2. Read the state back and verify X is now 2
3. Done
```
rather than 
```
1. Send a request to update X from 1 → 2
2. Done
```

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
ckeshava: @pdp2121  ok. I have introduced separate `interface` and `enum` for the Mutable Flags: ce166161d0b9fe9d60ddd246e7997f0ef5c0bd48

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
ckeshava: code has been changed in ce166161d0b9fe9d60ddd246e7997f0ef5c0bd48

packages/xrpl/test/models/MPTokenIssuanceSet.test.ts
ckeshava: Presently, I have not written validations for all the possible failure cases (`tem` codes). That is a prerequisite for writing unit tests.

I'm not convinced that any validation is required from the client libraries at all. If there is an error in the transaction, it will be failed by rippled with the appropriate error code. The basic validation of the transaction occurs with minimal network overhead, i.e. transactions which fail with a `tem` code are never "gossip"ed by the XRPL network. What are we gaining by implementing these exhaustive validations in the client libraries?

This is a repetition of engineering effort across three teams. The rippled cpp binary already implements validations+unit-tests for the `tem` codes. We will be [duplicating the code](https://github.com/XRPLF/rippled/blob/5d79bfc5317093b00b787dfd5e50ce3bee1de0ca/src/xrpld/app/tx/detail/MPTokenIssuanceCreate.cpp#L51) in `Transactor::preflight` method for each transaction, in all the client libraries. Furthermore, I'm given to understand that the QA team also performs exhaustive tests of a new amendment. 

Lastly, The existing set of ~70 transactions do not have exhaustive validations/unit-tests. Are we planning to extend this idea to the other transactions as well? Older transactions like `Payment` and `TrustSet` are used a lot more frequently than the other transactions. Hence, validations+unit-tests on older transactions provide the most exposure to the users.

kuan121: Based on your argument, should we stop writing unit tests for our client libraries just because other teams are already doing so? That doesn’t sound like a good idea. If we shouldn’t stop, then what should our unit tests focus on — the failure conditions that are easier to test or write? Maybe you could elaborate on what you chose to unit test or integration test in this PR and the reasons?

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
ckeshava: A return value of `tesSUCCESS` in a validated-ledger necessarily indicates that the transaction's effects are correctly applied to the blockchain state. It is the responsibility of the XRPL client (rippled) to [return the correct return code](https://xrpl.org/docs/references/protocol/transactions/transaction-results) (0 or non-zero return codes).

I agree with your strategy in a general distributed system. However, the architecture of XRPL guarantees that `tesSUCCESS` is returned if and only if >80% of the validators are in consensus about the final state of the blockchain. The other non-conforming validators will need to catch up with the rest of the network. (likely due to network issues / byzantine behavior).

The integ tests are run against a standalone XRPL network i.e. with one node. Hence, we will never see instances of data inconsistency (even if that were the case)

packages/xrpl/test/models/MPTokenIssuanceSet.test.ts
kuan121: >Lastly, The existing set of ~70 transactions do not have exhaustive validations/unit-tests. Are we planning to extend this idea to the other transactions as well?

We don’t need to apply this approach to all other transactions — I see the absence of those tests as technical debt. As the client libraries evolve, we should focus on making things better through incremental improvements. We can start improving now and maintain a higher standard going forward.

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
ckeshava: This is the [code that is deployed in the production XRPL mainnet](https://github.com/XRPLF/rippled/blob/5d79bfc5317093b00b787dfd5e50ce3bee1de0ca/include/xrpl/protocol/detail/ledger_entries.macro#L417) and devnet. I am using this as the source of truth.

The documentation needs to be updated.

packages/xrpl/test/models/MPTokenIssuanceSet.test.ts
kuan121: > This is a repetition of engineering effort across three teams. The rippled cpp binary already implements validations+unit-tests for the tem codes. We will be [duplicating the code](https://github.com/XRPLF/rippled/blob/5d79bfc5317093b00b787dfd5e50ce3bee1de0ca/src/xrpld/app/tx/detail/MPTokenIssuanceCreate.cpp#L51) in Transactor::preflight method for each transaction, in all the client libraries. Furthermore, I'm given to understand that the QA team also performs exhaustive tests of a new amendment.

To be clear, my main concern isn’t about unit testing every failure condition — it’s about ensuring we have the right validations in our model objects to prevent obvious failure conditions on the client side. The `tem` failure conditions listed in the spec seem like a solid reference, which is why I think we should have proper validations or tests covering them, instead of relying on developers to pick and choose which cases to test based on personal judgment.

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
kuan121: Thanks for taking the time to explain! I agree that we don’t need to read the value and verify. I only asked because I noticed some other integration tests were doing it.

packages/xrpl/test/integration/transactions/mptokenIssuanceSet.test.ts
kuan121: Resolved since I saw the explanation in https://github.com/XRPLF/xrpl.js/pull/3081/files#r2389388334

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
ckeshava: updated in 42049ac99

packages/xrpl/test/models/MPTokenIssuanceSet.test.ts
ckeshava: > Based on your argument, should we stop writing unit tests for our client libraries just because other teams are already doing so?

that is my personal opinion, yes.

The USP of the client libraries is the autofill feature, the serializability of transaction, the ease of communicating with the XRPL blockchain. I feel we shouldn't duplicate the engineering effort of testing for the correctness of transactions in both rippled-cpp-code and all the client libraries (py, js, java, go).

Although I disagree with you, I can write more tests for the failure modes of DynamicMPT amendment in the next few days, I understand the benefit of your suggestion.

kuan121: Thanks @ckeshava! I’ll add this as a topic for our next sprint retro and see what the rest of the team thinks.

ckeshava: I have added all the validations and unit tests for the `MPTokenIssuance{Create, Set}` transactions. Please refer to this commit: 7260f2f8f

kuan121: Thanks! I will review it again today.

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
kuan121: As discussed offline, it's re-introduced for [Token Escrow](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0085-token-escrow#152-mptoken-ledger-object)

ckeshava: yes

Jasonradford51:
i am trying to complete and send on to yu  people it will not allow me

kuan121 (dismissed):

ckeshava:
@pdp2121  Please let me know if you have any comments on this PR.

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -120,13 +227,15 @@ export interface MPTokenIssuanceCreate extends BaseTransaction {
   MPTokenMetadata?: string
 
   Flags?: number | MPTokenIssuanceCreateFlagsInterface
+  MutableFlags?: number
```
pdp2121, line 229: Can you add docstring for this?

pdp2121 (dismissed):
LGTM! Left a small comment

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
ckeshava: I added a recent commit to fix this. However, I could not find the documentation for the `domain_id` field on the xrpl.org website ://

ckeshava: Update: I have added a docstring for the `DomainID` field as well, based on my understanding of the topic. I could not find any other documentation to that effect. Commit: 0c228cdef

packages/xrpl/src/models/transactions/MPTokenIssuanceCreate.ts
```diff
@@ -157,7 +294,7 @@ export function validateMPTokenIssuanceCreate(
       throw new ValidationError('MPTokenIssuanceCreate: Invalid MaximumAmount')
     } else if (
       BigInt(tx.MaximumAmount) > BigInt(MAX_AMT) ||
-      BigInt(tx.MaximumAmount) < BigInt(`0`)
+      BigInt(tx.MaximumAmount) <= BigInt(`0`)
```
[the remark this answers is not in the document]
ckeshava: Rebuttal evidence: https://github.com/XRPLF/rippled/blob/f8b4f692f1cb3b971138b71912bcd2c2eb9ee4d5/src/xrpld/app/tx/detail/MPTokenIssuanceCreate.cpp#L90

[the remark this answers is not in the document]
ckeshava: can you resolve this comment, if there are no other concerns?

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
```diff
@@ -67,6 +156,23 @@ export function validateMPTokenIssuanceSet(tx: Record<string, unknown>): void {
...
+      'MPTokenIssuanceSet: Cannot set both DomainID and Holder fields.',
+    )
+  }
+  if (
+    tx.MutableFlags != null &&
+    // eslint-disable-next-line no-bitwise -- Need bitwise operations to replicate rippled behavior
+    tx.MutableFlags & tmfMPTokenIssuanceSetMutableMask
+  ) {
+    throw new ValidationError('MPTokenIssuanceSet: Invalid MutableFlags value')
+  }
```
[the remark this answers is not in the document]
ckeshava: Thanks — the inconsistency you flagged between the raw `&` on line 172 and `isFlagEnabled()` (which calls `BigInt()` and throws on non-integer / out-of-range values) is real, but I'm going to defer it.

Across this SDK, every `Flags`/`MutableFlags` field is validated via `isNumber` + bitwise mask. That mirrors how `rippled` reads these as `uint32_t` on the wire — a malformed value is rejected at deserialization, not at the model layer. Adding a hard `Number.isSafeInteger && 0 ≤ v ≤ 0xFFFFFFFF` guard only on `MutableFlags` would diverge from the rest of the codebase without changing what actually reaches the chain.

Aligning the integer-range check across **all** flag fields is worth doing, but it belongs in a follow-up that touches `validateOptionalField` / `isNumber` consistently, not scoped to XLS-94D. Leaving this thread open for visibility.

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
```diff
@@ -11,4 +11,92 @@ export interface MPTokenIssuance extends BaseLedgerEntry, HasPreviousTxnID {
...
+  lsfMPTRequireAuth = 0x00000004,
+  lsfMPTCanEscrow = 0x00000008,
+  lsfMPTCanTrade = 0x00000010,
+  lsfMPTCanTransfer = 0x00000020,
+  lsfMPTCanClawback = 0x00000040,
+
+  /**
+   * Indicates flag lsfMPTCanLock can be changed
+   */
+  lsmfMPTCanMutateCanLock = 0x00000002,
```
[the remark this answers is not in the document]
ckeshava: Already done — see commit [96f319ada](https://github.com/XRPLF/xrpl.js/pull/3081/commits/96f319ada) ("feat(xrpl): split MPTokenIssuance flags into lsf*/lsmf* views"). The single enum has been split into `MPTokenIssuanceFlags` (lsf*) and `MPTokenIssuanceMutableFlags` (lsmf*), matching the pattern in `MPTokenIssuanceCreate.ts`. Resolving.

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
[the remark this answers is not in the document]
ckeshava: sure

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
```diff
@@ -83,4 +189,112 @@ export function validateMPTokenIssuanceSet(tx: Record<string, unknown>): void {
...
   }
+
+  if (tx.Holder != null && tx.Holder === tx.Account) {
+    throw new ValidationError(
+      'MPTokenIssuanceSet: Holder cannot be the same as the Account.',
+    )
+  }
+
+  const isMutate =
+    tx.MutableFlags != null ||
```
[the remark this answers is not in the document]
ckeshava: If a user sets MutableFlags: 0  for an intentional no-op transaction, that is a code smell. I would prefer the user hit an error instead of silently passing the transaction to the rippled server. An error would incentivize the user take another look at the transaction payload.

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
```diff
@@ -55,8 +115,14 @@ export interface MPTokenIssuanceSet extends BaseTransaction {
    */
   Holder?: Account
   Flags?: number | MPTokenIssuanceSetFlagsInterface
+
+  MPTokenMetadata?: string
+  TransferFee?: number
+  MutableFlags?: number
+  DomainID?: string
```
ckeshava, line 143: Doc-strings need to be added for these attributes

ckeshava:
@kuan121 @Patel-Raj11  please review this PR again with the latest changes ^

packages/xrpl/src/models/ledger/MPTokenIssuance.ts
```diff
@@ -77,4 +77,72 @@ export interface MPTokenIssuance extends BaseLedgerEntry, HasPreviousTxnID {
    * vaults. Absent for XRP-backed vaults.
    */
   ReferenceHolding?: string
+
+  MutableFlags: number
```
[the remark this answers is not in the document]
ckeshava: Good catch — confirmed against rippled (`MPTokenIssuanceCreate.cpp` only sets `sfMutableFlags` when supplied). Made `MutableFlags` optional on the ledger interface and hardened `parseMPTokenIssuanceMutableFlags` to return an empty view for nullish input (avoids a `BigInt(undefined)` throw on pre-amendment objects). Addressed in f54cdb6e2.

packages/xrpl/HISTORY.md
```diff
@@ -26,6 +26,7 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 * `Client.getServerInfo()` and `Client.connect()` now throw if the `server_info` request fails, or if the response succeeds but does not include a `network_id`. Previously, these failures were swallowed and only logged via `console.error`, leaving `client.networkID` undefined and causing `autofill()` to omit the `NetworkID` field — producing transactions valid on the wrong network. Servers running rippled <1.11 (which do not return `network_id`) will now fail to connect; upgrade to rippled 1.11+ or set `client.networkID` manually after construction.
 
 ### Added
+* Support for `Dynamic MPT` (XLS-94D)
```
Patel-Raj11: Need to be moved to Unreleased section

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
```diff
@@ -83,4 +190,71 @@ export function validateMPTokenIssuanceSet(tx: Record<string, unknown>): void {
...
+      throw new ValidationError(
+        `MPTokenIssuanceSet: TransferFee must be between 0 and ${MAX_TRANSFER_FEE}`,
+      )
+    }
+  }
+
+  if (tx.MPTokenMetadata != null) {
+    if (
+      typeof tx.MPTokenMetadata === 'string' &&
+      (!isHex(tx.MPTokenMetadata) ||
```
Patel-Raj11: Empty blob is valid, as it clears the MPTokenMetadata as per rippled

ckeshava: Yes, you are right. Here is that source code in rippled: https://github.com/XRPLF/rippled/blob/8abbd1ba3a194a652a158fcb6f3a0e9a13e4c9c1/src/libxrpl/tx/transactors/token/MPTokenIssuanceSet.cpp#L365

I will make this change.

packages/xrpl/HISTORY.md
ckeshava: Moved the `Dynamic MPT` (XLS-94D) entry from the released 5.0.0 section to **Unreleased → Added** in 9864c2f31. Thanks!

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
ckeshava: Good catch — confirmed in rippled `MPTokenIssuanceSet.cpp` doApply (`metadata->empty() → makeFieldAbsent`). An empty `MPTokenMetadata` is now accepted as a clear; hex/length/XLS-89 validation runs only for non-empty values, and the error message was reworded. Fixed in ae05c3436.

Patel-Raj11 (approved):

kuan121 (approved):

Issue #3070, opened 2025-09-02
Release workflow improvement


Patel-Raj11:
Release workflow [failed](https://github.com/XRPLF/xrpl.js/blob/main/.github/workflows/release.yml) to publish to npm registry when trying to release `xrpl@4.4.1.` `xrpl@4.4.1` was released manually due to that, however we should incorporate these items in the release workflow based on my observations from doing one release:

1. Remove ability to do a `draft` release as the actual release provides a stopping point to review `artifacts` and `security reports`. This will simplify the release workflow.

2. Currently, the workflow asks for which branch's `release.yml` to use to run the workflow. We always select main branch. So remove that option altogether and always use `release.yml` from `main` branch.

3. Provide a text box to specify the tag for npm package. Keeping it blank means latest release. This is to make sure that when releasing the release candidate or feature specific release (For instance - https://www.npmjs.com/package/xrpl/v/4.5.0-smartescrow.0), we explicitly publish that package using the tag provided instead of marking it as `latest` (which is the default). Factor in this text box content during npm publish step.

4. Just before the approval step, create a pull request from `release-branch` -> `main` and keep the pull request link handy. Only create a pull request for latest releases. Don't create a pull request for release candidate or beta releases. (check npm tag input, raise PR if tag is empty or latest)

5. Create a Github issue with the content of the [vulnerability-report](https://github.com/XRPLF/xrpl.js/actions/runs/17413666450/artifacts/3909673000) if it's not empty, so that DGE team can work on fixing. If it's empty, skip this step.

6. During the approval phase, send a slack message to `#xrpl-js` channel with the following content:
```
<Workflow executor> is releasing <package name with version>. Two approvers from <List of members of official-release environment> need to take following actions:
1. Review the release artifacts and approve/reject the release. (Provide link to Github action run here)
2. Review the package update PR and provide two approvals. DO NOT MERGE - <Workflow executor> will verify the package on npm registry and merge this approved PR. (Provide link to PR raised in step 4). (Skip this step if it's a release candidate). (only add second point if it npm tag is empty or latest)
```

7. Execute below mentioned steps after the npm package is successfully published to npm registry as opposed to before it. This will prevent creation of post release stuff in case of npm publish step fails due to network error or API key expiry. Otherwise the developer will need to unwind these steps manually and due to presence of git tag, the release workflow will fail on subsequent execution.
7.1. Creating and pushing a git tag.
7.2. Generating release notes.
[...]

PR #3085, opened 2025-09-18
Release with trusted publisher

shichengripple001:

Made changes according to https://github.com/XRPLF/xrpl.js/issues/3070
Removed NPM token, the new workflow will use trusted publisher instead.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes


.github/workflows/release.yml
```diff
@@ -31,45 +27,69 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          PACKAGE_NAME="${{ github.event.inputs.package_name }}"
+          PKG_JSON="packages/${PACKAGE_NAME}/package.json"
+          if [[ ! -f "$PKG_JSON" ]]; then
+            echo "package.json not found at $PKG_JSON. Check 'package_name' input." >&2
+            exit 1
+          fi
+          VERSION=$(jq -er .version "$PKG_JSON")
+          if [[ -z "$VERSION" || "$VERSION" == "null" ]]; then
```
Patel-Raj11, line 98: @shichengripple001 Can we add suggestions from @mvadari to fail the workflow if `npmjs_dist_tag` is latest and `VERSION` is not of the form `a.b.c`.

.github/workflows/release.yml
```diff
@@ -96,161 +115,317 @@ jobs:
...
+          curl -X POST https://slack.com/api/chat.postMessage \
+            -H "Authorization: Bearer $SLACK_TOKEN" \
+            -H "Content-Type: application/json" \
+            -d "$(jq -n \
+                  --arg channel "#xrpl-js" \
+                  --arg text "$MESSAGE" \
+                  '{channel: $channel, text: $text}')"
+
+      - name: Install cyclonedx-npm
+        run: npm install -g @cyclonedx/cyclonedx-npm
```
Patel-Raj11: @shichengripple001 Can we install a specific version of `cyclonedx-npm`, to prevent supply chain attacks.
`npm install -g @cyclonedx/cyclonedx-npm`

.github/workflows/release.yml
```diff
@@ -260,94 +435,98 @@ jobs:
...
+          node-version: 20
+          registry-url: 'https://registry.npmjs.org/'
+
+      - name: Publish to npm
+        run: |
+          cd dist
+          PKG=$(ls *.tgz)
+          echo $PKG
+          NPM_DIST_TAG="${{ github.event.inputs.npmjs_dist_tag }}"
+          if [ -z "$NPM_DIST_TAG" ]; then
```
Patel-Raj11, line 558: @shichengripple001 Can we add suggestions from @mvadari to fail the workflow if `npmjs_dist_tag` is latest and `VERSION` is not of the form `a.b.c` here as well?

.github/workflows/release.yml
```diff
@@ -260,94 +435,98 @@ jobs:
...
+            NPM_DIST_TAG="latest"
+          fi
+          npm i -g npm@11.6.0
+          npm publish "$PKG" --provenance --access public --registry=https://registry.npmjs.org/ --tag "$NPM_DIST_TAG"
```
Patel-Raj11, line 566: Let us fail the workflow if this step fails. We don't want to create Github tags and Github Release if publishing fails.

.github/workflows/release.yml
```diff
@@ -96,161 +115,317 @@ jobs:
...
+      - name: Pack tarball
+        run: |
+          set -euo pipefail
+          echo "Packaging ${{ env.PACKAGE_NAME }}"
+          find "packages/${{ env.PACKAGE_NAME }}" -maxdepth 1 -name '*.tgz' -delete || true
+          FULL_PACKAGE_NAME="$(jq -er '.name' packages/${{ env.PACKAGE_NAME }}/package.json)"
+          TARBALL=$(npx lerna exec --scope "$FULL_PACKAGE_NAME" -- npm pack --json | jq -r '.[0].filename')
+          echo "TARBALL=packages/${{ env.PACKAGE_NAME }}/${TARBALL}" >> "$GITHUB_ENV"
+
+      - name: Upload tarball as artifact
```
Patel-Raj11, line 286: @shichengripple001 Are we sure that the `TARBALL` will be overwritten when same workflow run is executed multiple times due to some failures and that latest overwritten `TARBALL` will be fetched in the release step to actually publish?

Patel-Raj11: If thats not the case, let us make this entire `release.yml` workflow not re-runnable. If it fails for any reason, then we must run a new instance of that workflow to be on safer side.

shicheng001sg: Artifacts are scoped to the specific workflow run attempt. Each rerun of the workflow (or of a job) runs in a fresh workspace and publishes a new artifact set; the release job downloads artifacts from the same attempt that produced them.
but i think it is still good idea to not allow 2nd attempt to avoid any unintended issue. 
change has been made

.github/workflows/release.yml
shicheng001sg: pined to a version for all the actions that are used by the workflow

.github/workflows/release.yml
shicheng001sg: made the suggested change

.github/workflows/release.yml
shichengripple001: it actually will fail the workflow and will not run the rest of the steps, if the publish command had failed.
Here is an example: https://github.com/xpring-eng/xrpl.js/actions/runs/17906433748/job/50908647506

.github/workflows/release.yml
```diff
@@ -96,161 +127,317 @@ jobs:
...
+        uses: actions/upload-artifact@de65e23aa2b7e23d713bb51fbfcb6d502f8667d8
+        with:
+          name: npm-package-tarball
+          path: ${{ env.TARBALL }}
 
   review:
     runs-on: ubuntu-latest
     needs: [get_version, run_faucet_test, run_tests, pre_release]
+    permissions:
+      pull-requests: write
```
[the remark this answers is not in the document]
Patel-Raj11: @shichengripple001  We should keep these Github action versioning as it is. Only npm versions need to be pinned.

.github/workflows/release.yml
Patel-Raj11: @shichengripple001
I think the additional checks were missed here

.github/workflows/release.yml
Patel-Raj11: Only `npm` package version should be pinned. Github actions versions were fine.

shichengripple001: it is better to pin github action version. it was a recommendation from code review bot also.

.github/workflows/release.yml
shichengripple001: https://github.com/XRPLF/xrpl.js/blob/30143e20b3f70a6f8cd3ef82da8a0853c0c7c5f5/.github/workflows/release.yml#L106 the change was made here

.github/workflows/release.yml
Patel-Raj11: Bot can sometimes be wrong. v4 is already a good pinned version. The SHA as a version is not a good practice.

.github/workflows/release.yml
Patel-Raj11: Ya, but we should have checks at release step as well to be consistent. It is the main step where publishing happens.

.github/workflows/release.yml
shichengripple001: changed back to version instead of commits

Patel-Raj11 (approved):
LGTM! 
nit: `#test-alert` needs to reverted once testing is done.

Issue #3083, opened 2025-09-16
VaultCreate fails when Asset is MPT


Patel-Raj11:
There are two issues in SAV implementation:

1. `validateVaultCreate` does not allow creation of Vault that has MPT as an asset.
2. `ripple-binary-codec` throws this error when parsing VaultCreate transaction that has MPT as an asset.

Lending Protocol IT › Successful LoanBroker creation and loan payment

      61 |   read(n: number): Uint8Array {
      62 |     if (n > this.bytes.byteLength) {
    > 63 |       throw new Error()
         |             ^
      64 |     }
      65 |
      66 |     const slice = this.bytes.slice(0, n)

      at BinaryParser.read (../ripple-binary-codec/src/serdes/binary-parser.ts:63:13)
      at Function.read [as fromParser] (../ripple-binary-codec/src/types/issue.ts:104:49)
      at BinaryParser.fromParser [as readFieldValue] (../ripple-binary-codec/src/serdes/binary-parser.ts:208:24)
      at Function.readFieldValue [as fromParser] (../ripple-binary-codec/src/types/st-object.ts:77:38)
      at BinaryParser.fromParser [as readType] (../ripple-binary-codec/src/serdes/binary-parser.ts:181:17)
      at readType (../ripple-binary-codec/src/binary.ts:47:11)
      at readJSON (../ripple-binary-codec/src/binary.ts:60:18)
      at binaryToJSON (../ripple-binary-codec/src/index.ts:34:10)
      at hashSignedTx (src/utils/hashes/hashLedger.ts:80:22)
      at Wallet.sign (src/Wallet/index.ts:443:25)
      at sign (src/sugar/submit.ts:258:17)
      at fulfilled (src/sugar/submit.ts:2354:24)

VaultCreate transaction example:
```
 
const vaultCreateTx: VaultCreate = {
    TransactionType: 'VaultCreate',
    Asset: {
      mpt_issuance_id: mptIssuanceId,
    },
    Account: vaultOwnerWallet.address,
  }
```

PR #3086, opened 2025-09-19
Add hashOracle utility hash function

pkcs8:
## High Level Overview of Change

Added `hashOracle` utility hashing function for Oracle objects.

### Context of Change

* Added utility hash function `hashOracle`
* Added oracle space key to `ledgerSpaces.ts`
* Added a test for `hashOracle`

### Type of Change

- [x] New feature (non-breaking change which adds functionality)

### Did you update HISTORY.md?

- [x] No, this change does not impact library users

## Test Plan

This chance closely follows `hashEscrow` utility function and implements similar tests for `hashOracle` function implementation.

PR #3090, opened 2025-09-23
Fix serialization of Issue type when asset is MPT

Patel-Raj11:

1. Fixes https://github.com/XRPLF/xrpl.js/issues/3083
2. Fixes serialization of Asset object when asset is MPT. Perform Big-endian and Little-endian conversions where necessary.
3. Fix incorrect type checking in `validateVaultCreate` that prevented vault with MPT as an asset.
4. Fixes flaking Oracle integration testcases.

### Context of Change

This bug prevented creation of Vault that has MPT as an asset.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan
Added unit and integration test cases.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15346,7 +15346,7 @@
       }
     },
     "packages/xrpl": {
-      "version": "4.4.1",
+      "version": "4.4.2",
       "license": "ISC",
       "dependencies": {
         "@scure/bip32": "^1.3.1",
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -7,6 +7,11 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 ### Fixed
 * Fix incorrect type checking in `validateVaultCreate` that prevented vault creation with MPT as an asset.
 
+## 4.4.2 (2025-09-25)
+
+### Fixed
+* improve Batch inner transaction typing
+
 ## 4.4.1 (2025-08-29)
 
 ### Fixed
--- a/packages/xrpl/package.json
+++ b/packages/xrpl/package.json
@@ -1,6 +1,6 @@
 {
   "name": "xrpl",
-  "version": "4.4.1",
+  "version": "4.4.2",
   "license": "ISC",
   "description": "A TypeScript/JavaScript API for interacting with the XRP Ledger in Node.js and the browser",
   "files": [


packages/xrpl/src/models/transactions/common.ts
```diff
@@ -203,7 +208,16 @@ export function isXRPLNumber(value: unknown): value is XRPLNumber {
...
-  return isString(input) || isIssuedCurrency(input)
+  return (
+    isRecord(input) &&
+    ((Object.keys(input).length === ISSUE_CURRENCY_SIZE &&
+      isString(input.issuer) &&
+      isString(input.currency)) ||
+      (Object.keys(input).length === XRP_CURRENCY_SIZE &&
+        input.currency === 'XRP') ||
+      (Object.keys(input).length === MPT_CURRENCY_SIZE &&
+        isString(input.mpt_issuance_id)))
```
ckeshava, line 219: this could throw an unhandled-error if `mpt_issuance_id` is not a part of the `Object.keys(input)`.

`(Object.keys(input).length == 1 && input['mpt_issuance_id'] != undefined && isString(input.mpt_issuance_id)` might help resolve that case.

packages/ripple-binary-codec/src/types/st-object.ts
```diff
@@ -151,13 +151,7 @@ class STObject extends SerializedType {
             ? STArray.from(xAddressDecoded[field.name], definitions)
             : field.type.name === 'UInt64'
               ? UInt64.from(xAddressDecoded[field.name], field.name)
-              : field.associatedType?.from
-                ? field.associatedType.from(xAddressDecoded[field.name])
-                : (() => {
-                    throw new Error(
```
ckeshava, line 157: This error is useful when rippled has a new internal type but xrpl.js is missing an implementation for such a new type.

packages/ripple-binary-codec/test/fixtures/codec-fixtures.json
```diff
@@ -33,7 +33,9 @@
         "LedgerEntryType": "DirectoryNode",
         "Flags": 0,
         "RootIndex": "059D1E86DE5DCCCF956BF4799675B2425AF9AD44FE4CCA6FEE1C812EEF6423E6",
-        "Indexes": ["908D554AA0D29F660716A3EE65C61DD886B744DDF60DE70E6B16EADB770635DB"]
+        "Indexes": [
```
ckeshava, line 36: Can you confirm that this file only has formatting changes?

packages/ripple-binary-codec/src/types/issue.ts
```diff
@@ -76,27 +79,37 @@ class Issue extends SerializedType {
...
-    if (hint === Hash192.width) {
-      const mptBytes = parser.read(Hash192.width)
+  static fromParser(parser: BinaryParser): Issue {
+    // MPT
+    if (parser.size() === MPT_WIDTH) {
+      const mptBytes = parser.read(MPT_WIDTH)
       return new Issue(mptBytes)
     }
+
+    // XRP/IOU
```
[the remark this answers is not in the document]
Patel-Raj11: `fromParser` gets called based on the `associatedType` of the field. In this case it's `Issue`. The Issue sub-classes `fromParser` only gets called once we identify the `associatedType` is Issue. So the binary parser should exactly point to either MPT/XRP/IOU asset and these checks are in-place while parsing.

packages/xrpl/src/models/transactions/common.ts
Patel-Raj11: isString will return false, if that's the case.

packages/ripple-binary-codec/src/types/issue.ts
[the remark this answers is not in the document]
ckeshava: @Patel-Raj11  I agree with the rabbit's suggestion. Since the `Issue` type is shared by IOU/XRP/MPT types, how can you determine the appropriate size of the parser?

If we are provided with a sufficiently long buffer of bytes (say a serialized transaction), the size of the parser will mislead us, because we can read 44 bytes (for an MPT), 40 bytes ( for an IOU) or 20 bytes (for XRP representation).

This suggestion from the AI is more robust: 
> Or detect MPT by peeking bytes (e.g., check bytes[20..40] === NO_ACCOUNT) without consuming, then read 44 if matched.

packages/ripple-binary-codec/src/types/st-object.ts
Patel-Raj11: We already throw an error from base class SerializedType if the subclass does not overrides this method.
```

static from(value: SerializedType | JSON | bigint): SerializedType {
    throw new Error('from not implemented')
    return this.from(value)
  }
```

packages/ripple-binary-codec/test/fixtures/codec-fixtures.json
Patel-Raj11: Ya, there is a VaultCreate test case with MPT added at the bottom.

packages/ripple-binary-codec/src/types/issue.ts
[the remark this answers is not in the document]
Patel-Raj11: Makes sense, addressed here [1169182](https://github.com/XRPLF/xrpl.js/pull/3090/commits/116918237f9e8b4a409208d772c8a9bcb49324f3)

packages/xrpl/test/integration/transactions/singleAssetVault.test.ts
```diff
@@ -37,7 +43,7 @@ describe('Single Asset Vault', function () {
 
   beforeEach(async () => {
     testContext = await setupClient(serverUrl)
-    issuerWallet = testContext.wallet
+    issuerWallet = await generateFundedWallet(testContext.client)
```
achowdhry-ripple, line 46: why was this change needed?

achowdhry-ripple:
LGTM, one small question for understanding

ckeshava (approved):

packages/xrpl/test/integration/transactions/singleAssetVault.test.ts
ckeshava: hmm, I'm not particularly sure myself. I believe any valid wallet on the network suffices for the purposes of this test.

Patel-Raj11: To reuse a wallet and write integration tests that are not flaky, we need to move away from querying account objects and picking up the first one in our integration tests.

We should write integration tests that computes LedgerIndex using such example methods introduced here - https://github.com/XRPLF/xrpl.js/pull/3086

Somehow we have stopped writing these utility functions and we should write those for all the missing ledger objects and refactor all test cases to deterministically compute ledger index.

PR #3100, opened 2025-10-02
Trigger from release branch

shichengripple001:

Trigger deployment from release branch to fix the wrong provenance issue

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,10 +17,6 @@ concurrency:
   group: release
   cancel-in-progress: true
 
-env:
-  RELEASE_REF: ${{ github.ref }}
-  RELEASE_BRANCH: ${{ github.ref_name }}
-
 jobs:
   get_version:
     runs-on: ubuntu-latest
@@ -35,7 +31,7 @@ jobs:
 
       - name: Validate inputs
         env:
-          RELEASE_BRANCH: ${{ env.RELEASE_BRANCH }}
+          RELEASE_BRANCH: ${{ github.ref_name }}
         run: |
           set -euo pipefail
           # Validate package_name
@@ -111,7 +107,7 @@ jobs:
     needs: [get_version]
     uses: ./.github/workflows/faucet_test.yml
     with:
-      git_ref: ${{ env.RELEASE_REF }}
+      git_ref: ${{ github.ref }}
     secrets: inherit
 
   run_tests:
@@ -123,7 +119,7 @@ jobs:
     needs: [get_version]
     uses: ./.github/workflows/nodejs.yml
     with:
-      git_ref: ${{ env.RELEASE_REF }}
+      git_ref: ${{ github.ref }}
     secrets: inherit
 
   pre_release:
@@ -135,7 +131,7 @@ jobs:
     env:
       PACKAGE_VERSION: "${{ needs.get_version.outputs.package_version }}"
       PACKAGE_NAME: "${{ github.event.inputs.package_name }}"
-      RELEASE_BRANCH: "${{ env.RELEASE_BRANCH }}"
+      RELEASE_BRANCH: "${{ github.ref_name }}"
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -241,7 +237,7 @@ jobs:
           REPO: ${{ github.repository }}
           PKG_NAME: ${{ env.PACKAGE_NAME }}
           PKG_VER: ${{ env.PACKAGE_VERSION }}
-          REL_BRANCH: ${{ env.RELEASE_BRANCH }}
+          REL_BRANCH: ${{ github.ref_name }}
           VULN_ART_URL: ${{ steps.vuln_art.outputs.art_url }}
           LABELS: security
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -30,10 +27,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.inputs.release_branch }}
           fetch-depth: 0
 
       - name: Validate inputs
+        env:
+          RELEASE_BRANCH: ${{ github.ref_name }}
```
Patel-Raj11: @shichengripple001 Can you use this updated workflow and test if xrpl release is working on your personal npm registry?
Also, I see both `github.ref_name `and `github.ref` being used at various places, is this difference intended?

Patel-Raj11: Also, let us update the RELEASE.md with the updated information.

Patel-Raj11: @shichengripple001 Let us also mention the release branch name convention in the RELEASE.md file.

shichengripple001: - it has been tested. 

https://github.com/xpring-eng/xrpl.js/actions/runs/18192475125 
https://www.npmjs.com/package/cheng-xrpl/v/4.4.73-beta.2 the provenance commit id is match the commit of the release branch

- github.ref_name is used to get the branch name only for display purpose only.

- updated RELEASE.md for branch name

RELEASE.md
```diff
@@ -36,13 +37,12 @@ You can manually trigger the release workflow from the [GitHub Actions UI](https
...
 |------------------|-----------------------|
 | package_name     | xrpl                  |
-| release_branch   | release/xrpl@4.3.8    |
 | npmjs_dist_tag   | latest                |
 
 
 ### **Reviewing the release details and scan result**
 
-1. The pipeline will pause at the "Review test and security scan result" step, at least 2 approvers are required to review and approve the release.
+1. The pipeline will pause at the "Print Test/Security scan result and invite Dev team to review" step and also "Release Pipeline for 4.4.73-beta.2" step, relevant team should review the release details and scan result.
```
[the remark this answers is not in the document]
Patel-Raj11: cc @shichengripple001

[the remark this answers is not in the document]
shichengripple001: fixed

Patel-Raj11 (approved):
LGTM!

RELEASE.md
```diff
@@ -10,21 +10,22 @@ You can manually trigger the release workflow from the [GitHub Actions UI](https
 
 ### **Before triggering a release**
 
-1. Create a release branch and update the **`version`** field in `packages/<package_name>/package.json` to the intended release version.
+1. Create a release branch. A qualified branch name should start with "release-" or "release/", case-insensitive. e.g: `release/xrpl@4.3.8`, `release-xrpl-4.3.8`, `Release/xrpl@4.3.8`.
```
ckeshava, line 13: Hello, 
1. Is the updated version name copied into the `main` branch as well? All the new features are developed and merged into the `main` branch of the xrpl.js repository. The version info needs to be updated on both `main` and `release/<>` branches.

2. Suppose there is a critical bug in one of the releases. Can I look at the procedure for handling that scenario? Is it documented in any README file? "

Do we need to patch the hotfix to both the `release` branch and the `main` branch ? (or) Should we delete the current `release` branch and cut a `rectified-released` branch?

.github/workflows/release.yml
ckeshava: `env.RELEASE_BRANCH` value is not used in this step. What is the need to define this env variable?

.github/workflows/release.yml
```diff
@@ -139,11 +131,11 @@ jobs:
     env:
       PACKAGE_VERSION: "${{ needs.get_version.outputs.package_version }}"
       PACKAGE_NAME: "${{ github.event.inputs.package_name }}"
+      RELEASE_BRANCH: "${{ github.ref_name }}"
```
ckeshava: Similar to the previous point, I don't see any usages of `env.RELEASE_BRANCH` in the pre_release job. Why did you define this variable?

.github/workflows/release.yml
ckeshava, line 447: I was able to trigger a release workflow from a tag-name here: https://github.com/XRPLF/xrpl.js/actions/runs/18320084793

In this case, the `github.ref`/`$RELEASE_BRANCH` value points to the tag name, rather than the branch name. The workflow eventually failed in the final step.

For the purposes of debugging, it could be more useful to differentiate between the two cases.

RELEASE.md
shichengripple001: #1. A pr is automatically generated for non beta release. for example https://github.com/XRPLF/xrpl.js/pull/3093
#2. I think a new branch should be used, since it is gonna to be a new version. @Patel-Raj11 what do you think?

.github/workflows/release.yml
shichengripple001: thanks for identifying this, i have removed it.

.github/workflows/release.yml
shichengripple001: thanks for identifying this, i have removed it.

.github/workflows/release.yml
shichengripple001: added check at valid input step to make the branch name has to start with release- or release/ regardless of casing. the environment protection rule only allow review and release step to run when the branch name qualifies.

RELEASE.md
Patel-Raj11: > I think a new branch should be used, since it is gonna to be a new version. @Patel-Raj11 what do you think?

Yes

ckeshava (approved):

PR #3131, opened 2025-11-10
Beta release workflow

shichengripple001:

1, Workflow needs to be triggered from main branch for both beta and stable release
2, For stable release, A PR from release branch to main branch is required before triggering the release pipeline
3, Security champion(Chenna and Cheng) are added to Security reviewers. Beta sec review will be done by Security champion.
4, "-experimental" will automatically appended to dist tag for beta release
5, Only unit test needs to be run for beta release

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes


.github/workflows/release.yml
```diff
@@ -54,43 +65,87 @@ jobs:
...
-          # Empty → default to 'latest'
-          if [ -z "$NPM_DIST_TAG" ]; then
-            NPM_DIST_TAG="latest"
+          # validate dist tag(s)
+          RAW_DIST_TAG="${{ github.event.inputs.npmjs_dist_tag }}"
+          if [ -z "$RAW_DIST_TAG" ]; then
+            TAG="latest"
             echo "ℹ️ npmjs_dist_tag empty → defaulting to 'latest'."
+          else
+            TAG="$(printf '%s' "$RAW_DIST_TAG" | tr -d '[:space:]')"
```
ckeshava:
```suggestion
            TAG="$(printf '%s' "$RAW_DIST_TAG" | tr -d '[:blank:]')"
```
nit: This removes both spaces and tabs from the inputs, whereas `:space:` does not recognize tab characters.

.github/workflows/release.yml
```diff
@@ -104,39 +159,34 @@ jobs:
             echo "Version is empty or missing in $PKG_JSON" >&2
             exit 1
           fi
-          NPM_DIST_TAG="${{ github.event.inputs.npmjs_dist_tag }}"
-          if [ -z "$NPM_DIST_TAG" ]; then
-            NPM_DIST_TAG="latest"
-          fi
-          if [[ "$NPM_DIST_TAG" == "latest" ]] && ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+          NPM_DIST_TAG="${NPM_DIST_TAG:-latest}"
```
ckeshava: Is this step necessary? `NPM_DIST_TAG` is already written into `GITHUB_ENV` in lines 109-112 (`get_version.steps.validate_inputs` section) of this updated file.

.github/workflows/release.yml
```diff
@@ -104,39 +159,34 @@ jobs:
...
     needs: [get_version]
     uses: ./.github/workflows/nodejs.yml
     with:
-      git_ref: ${{ github.ref }}
+      git_ref: ${{ needs.get_version.outputs.release_branch }}
     secrets: inherit
 
   pre_release:
     runs-on: ubuntu-latest
+    if: ${{ always() && needs.get_version.result == 'success' && (needs.run_faucet_test.result == 'success' || needs.get_version.outputs.is_beta == 'true') && (needs.run_tests.result == 'success' || needs.get_version.outputs.is_beta == 'true') }}
```
ckeshava:
```suggestion
    if: ${{ needs.get_version.result == 'success' && (needs.run_faucet_test.result == 'success' || needs.get_version.outputs.is_beta == 'true') && (needs.run_tests.result == 'success' || needs.get_version.outputs.is_beta == 'true') }}
```

Since there are other conditional expressions, `always()` is not useful in this context.

.github/workflows/release.yml
ckeshava, line 197: Hypothetically, suppose that there are SBOM vulnerabilities reported for a beta release. Do we have a policy for accepting/rejecting the beta-release, for the role of the security champion? What are the tolerable types of SBOM vulnerabilities?

ckeshava, line 228: Can we increase the retention policy for these uploaded vulnerability-report artifacts? I believe the default is 10 days, however a 30-day period is useful to perform retroactive analysis.

.github/workflows/release.yml
```diff
@@ -297,14 +348,15 @@ jobs:
 
   ask_for_dev_team_review:
     runs-on: ubuntu-latest
+    if: ${{ always() && needs.pre_release.result == 'success' && (needs.run_tests.result == 'success' || needs.get_version.outputs.is_beta == 'true') }}
```
ckeshava:
```suggestion
    if: ${{ needs.pre_release.result == 'success' && (needs.run_tests.result == 'success' || needs.get_version.outputs.is_beta == 'true') }}
```

.github/workflows/release.yml
```diff
@@ -620,28 +649,68 @@ jobs:
...
+    name: Generate and Publish documentation for ${{ needs.get_version.outputs.package_version }}
+    if: ${{ needs.get_version.outputs.is_beta != 'true' }}
+    uses: ./.github/workflows/generate-documentation.yml
+    needs: [get_version, release]
+    permissions:
+      contents: read
+      pages: write
+      id-token: write
+    with:
+      git_ref: ${{ github.ref_name }}
```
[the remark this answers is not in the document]
ckeshava: ^^ @shichengripple001  can you please update the `git_ref` variable? I wrote the generate-documentation workflow based on the current version of the release workflow.

.github/workflows/release.yml
```diff
@@ -620,28 +649,68 @@ jobs:
...
+      - name: Notify Slack about workflow failure
         env:
           SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+          PACKAGE_NAME: ${{ github.event.inputs.package_name }}
+          PACKAGE_VERSION: ${{ needs.get_version.outputs.package_version }}
+          NEEDS_JSON: ${{ toJson(needs) }}
         run: |
-          MESSAGE="❌ Release failed for ${{ env.PACKAGE_NAME }}@${{ env.PACKAGE_VERSION }}. Check the logs: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
-          curl -X POST https://slack.com/api/chat.postMessage \
+          set -euo pipefail
```
ckeshava, line 757: The `notify_failures` job is not pinned to run on any specific shell. However, the pipefail option is configured to work only on `bash` and `ksh` shells. It will not work on `dash` (or) `sh` shells.

Discussion: https://askubuntu.com/questions/886537/set-e-o-pipefail-not-working-on-bash-script-on-ubuntu-16

Some of the jobs in this workflow use `bash` as the pinned shell, is that sufficient for all the jobs?

.github/workflows/release.yml
```diff
@@ -54,43 +65,87 @@ jobs:
...
           fi
 
-          # Disallow version-like prefixes (avoid semver/range confusion)
-          if [[ "$NPM_DIST_TAG" =~ ^v[0-9] || "$NPM_DIST_TAG" =~ ^[0-9] ]]; then
-            echo "❌ Invalid npm dist-tag '$NPM_DIST_TAG'. Must not start with 'v' + digit or a digit (e.g., 'v1', '1.2.3')." >&2
+          if [ "$TAG" = "latest" ]; then
+            IS_BETA="false"
+          else
+            IS_BETA="true"
+            TAG="${TAG}-experimental"
```
kuan121: Can we do `TAG="${TAG}-beta"` instead?

.github/workflows/release.yml
```diff
@@ -54,43 +65,87 @@ jobs:
...
+          fi
+
+          if [ "$IS_BETA" != "true" ] && [[ ! "${RELEASE_BRANCH,,}" =~ ^release[-/] ]]; then
```
kuan121: Can we also validate branch name for beta release to ensure that the branch is `main`? I don't think we want people to release beta using any branch.

.github/workflows/release.yml
```diff
@@ -9,6 +9,9 @@ on:
       package_name:
```
kuan121, line 9: > 1, Workflow needs to be triggered from main branch for both beta and stable release
> 2, For stable release, A PR from release branch to main branch is required before triggering the release pipeline

Why do we want to check that a PR to merge `release-xxx` into `main` exists? Can we still automatically create a PR to merge `release-xxx` into `main` for stable release?

> 5, Tests are allowed to fail for beta release

If test failures are acceptable, can we skip the test steps entirely for beta releases? What value do the tests provide in this case?

.github/workflows/release.yml
```diff
@@ -9,6 +9,9 @@ on:
...
         description: 'Package folder (Name of the package directory under packages/ folder. e.g., xrpl, ripple-address-codec)'
         required: true
+      release_branch_name:
```
kuan121, line 12: Can you update https://github.com/XRPLF/xrpl.js/blob/main/RELEASE.md to reflect the final release process before you merge this PR?

.github/workflows/release.yml
```diff
@@ -54,43 +65,87 @@ jobs:
...
           fi
 
-          # validate dist tag
-          NPM_DIST_TAG="${{ github.event.inputs.npmjs_dist_tag }}"
-
-          # Empty → default to 'latest'
-          if [ -z "$NPM_DIST_TAG" ]; then
-            NPM_DIST_TAG="latest"
+          # validate dist tag(s)
+          RAW_DIST_TAG="${{ github.event.inputs.npmjs_dist_tag }}"
```
kuan121: FYI - https://github.com/XRPLF/xrpl.js/pull/3133, we need to use env vars.

.github/workflows/release.yml
```diff
@@ -507,7 +517,7 @@ jobs:
         run: |
           set -euo pipefail
 
-          MSG="${EXECUTOR} is releasing ${PACKAGE_NAME}@${PACKAGE_VERSION}. A member from the infosec team (${SEC_REVIEWERS}) needs to take the following action:\n Review the release artifacts and approve/reject the release. (${RUN_URL})"
+          MSG="${EXECUTOR} is releasing ${PACKAGE_NAME}@${PACKAGE_VERSION}. A sec reviewer from (${SEC_REVIEWERS}) needs to take the following action:\nReview the release artifacts and approve/reject the release. (${RUN_URL})"
```
kuan121: >3, Security champion(Chenna and Cheng) are added to Security reviewers. Beta sec review will be done by Security champion.

For the beta release, will ${SEC_REVIEWERS} include Infosec members? They may find it confusing or inconvenient to be mentioned in the Slack notification when they don’t actually need to review beta releases. Would it make sense to use a separate notification mechanism, or maintain two security reviewer lists—one for stable releases and another for beta?

.github/workflows/release.yml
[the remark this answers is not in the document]
shichengripple001: @ckeshava updated to ${{ needs.get_version.outputs.release_branch }}

.github/workflows/release.yml
shichengripple001: because we need to allow pipeline continue to run for beta follow when tests fail. this is the only way to make it possible

.github/workflows/release.yml
shichengripple001: [:space:] should remove all those space, tab, vertical tab, form feed, newline, and carriage return
[:blank:] would strip those two characters but leave other whitespace such as newlines, carriage returns, vertical tabs

.github/workflows/release.yml
shichengripple001: made fix for it

.github/workflows/release.yml
shichengripple001: can we agree that no Critial/High vulnerabilities are allowed even for beta release?
@mvadari @kuan121 @Patel-Raj @pdp2121

shichengripple001: they are uploaded to owasp dependency track, we can view the vulnerabilities for each version of the package from web UI. I am also going to setup another pipeline for quickly audit the dependencies when a comprised npm attack is reported.

.github/workflows/release.yml
shichengripple001: added the following. so all steps will use bash as default
defaults:
  run:
    shell: bash

.github/workflows/release.yml
shichengripple001: there is already beta in version number, having experimental in dist tag can give an extra sense of immaturity.

.github/workflows/release.yml
shichengripple001: thanks. added check to only allow the the pipeline to be run from main branch

.github/workflows/release.yml
shichengripple001: we want to the change to be review before doing a release. infosec team often need to ask what are the changes for the release, there was a discussion here in the doc
https://docs.google.com/document/d/1JWx3kJYGqTisrOubiKuo59PIX85r9AQpz7mEp3PkLAc/edit?disco=AAABtjqnE2c 

I am fine to skip the test steps for beta release, it does make the workflow simpler, if we need test result, we can still run the test pipelines. Are we ok with skipping test step for beta release? @mvadari @Patel-Raj @ckeshava

.github/workflows/release.yml
shichengripple001: change has been made

.github/workflows/release.yml
shichengripple001: npmjs.com only allow one Github protected environment. so only thing that we can do is to add beta reviewers to the same group. 
For pypi, we are able to implement the mentioned approach.

RELEASE.md
```diff
@@ -10,21 +10,36 @@ You can manually trigger the release workflow from the [GitHub Actions UI](https
...
      "version": "x.y.z"
    }
    ```
-3. Run npm i to update the package-lock with the updated versions and commit the lock file to the release branch
+4. Set `npm distribution tag` to `latest`.
+5. Run npm i to update the package-lock with the updated versions and commit the lock file to the release branch
+
+**Beta release **
+1. Create a release branch. There is no restriction for branch name.
+2. 2. Update the **`version`** field in `packages/<package_name>/package.json` to the intended beta release version.
```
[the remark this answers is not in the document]
achowdhry-ripple: this is a valid nitpick

RELEASE.md
```diff
@@ -34,15 +49,16 @@ You can manually trigger the release workflow from the [GitHub Actions UI](https
...
+|---------------------|-----------------------|
+| release_branch_name | release/xrpl@4.3.8   |
+| package_name        | xrpl                  |
+| npmjs_dist_tag      | latest                |
 
 
 ### **Reviewing the release details and scan result**
 
-1. The pipeline will pause at the "Print Test/Security scan result and invite Dev team to review" step and also before the final release step, relevant team should review the release details and scan result.
+1. The pipeline will pause at the "Print Test/Security scan result and invite Dev team to review" step and also before the final release step, relevant team should review the release details and scan result. Stable release release will be reviewed by infosec team as Sec reviewer. Beta release will be reviewed by security champions from Dev team.
```
achowdhry-ripple:
```suggestion
1. The pipeline will pause at the "Print Test/Security scan result and invite Dev team to review" step and also before the final release step, relevant team should review the release details and scan result. Stable release will be reviewed by infosec team as Sec reviewer. Beta release will be reviewed by security champions from Dev team.
```

.github/workflows/release.yml
kuan121: In theory, we shouldn’t — after all, we wouldn’t intentionally release software with critical or high-severity vulnerabilities. However, it often comes down to the specifics, since some vulnerabilities may not be relevant or practically exploitable. It’s also worth noting that we released xrpl@4.4.2 and xrpl@4.4.3 despite known critical/high vulnerabilities (see the reported [issues](https://github.com/XRPLF/xrpl.js/issues?q=is%3Aissue%20Security)), and I don’t believe we evaluated whether those issues would have any practical impact.

The best approach, in my view, is to prioritize fixing these security vulnerabilities and, going forward, apply stricter scrutiny to high- and critical-severity issues during the release process.

.github/workflows/release.yml
```diff
@@ -211,7 +263,7 @@ jobs:
           curl -X POST \
             -H "X-Api-Key: ${OWASP_TOKEN}" \
             -F "autoCreate=true" \
-            -F "projectName=xrpl-js" \
+            -F "projectName=test-alert" \
```
kuan121: Forget to revert the change?

.github/workflows/release.yml
kuan121: >npmjs.com only allow one Github protected environment.

Could you clarify this a bit? My understanding is that reviewer lists are a GitHub feature. Wouldn’t it be possible to define three separate reviewer groups and use each where appropriate?
1. Dev Reviewers
2. Security Reviewers
3. Beta Security Reviewers

RELEASE.md
```diff
@@ -10,21 +10,36 @@ You can manually trigger the release workflow from the [GitHub Actions UI](https
...
+1. Create a release branch. There is no restriction for branch name.
+2. 2. Update the **`version`** field in `packages/<package_name>/package.json` to the intended beta release version.
+   ```json
+   {
+     "name": "<package_name>",
+     "version": "x.y.z-<beta|rc>.a"
+   }
+   ```
+3. Provide a non-`latest` `npm distribution tag` and not starting with `v` + digit or a digit. The workflow will automatically append `-experimental`, as `<tag>-experimental`.
+4. Run `npm i` to refresh `package-lock.json` and commit it.
```
kuan121: Don't we need to merge the branch to main first since Beta release needs to be trigger from `main`?

RELEASE.md
```diff
@@ -10,21 +10,36 @@ You can manually trigger the release workflow from the [GitHub Actions UI](https
...
+3. Update the **`version`** field in `packages/<package_name>/package.json` to the intended release version.
    ```json
    {
      "name": "<package_name>",
      "version": "x.y.z"
    }
    ```
-3. Run npm i to update the package-lock with the updated versions and commit the lock file to the release branch
+4. Set `npm distribution tag` to `latest`.
+5. Run npm i to update the package-lock with the updated versions and commit the lock file to the release branch
```
kuan121: Can we make this line and L36 the same? Let's not describe the same thing differently in two processes.

shichengripple001:
Re: https://github.com/XRPLF/xrpl.js/pull/3131#discussion_r2565474413 

The reviewers are fetched from a GitHub protected environment. However, npmjs supports only a single protected environment for trusted publishing, which must be the designated “official-release” environment. As a result, reviewers configured in other protected environments cannot grant approval; only reviewers defined within the “official-release” environment are permitted to authorize the release.
@kuan121

kuan121:
Overall, LGTM. We can merge once:
1. You address the unresolved comments regarding text updates and unreverted changes.
2. You align with the other reviewers on whether we should skip tests for the Beta release.

RELEASE.md
shichengripple001: for L36, I meant to tell the the npm distribution tag can't be "latest",  so it will be treated as beta release @kuan121

RELEASE.md
```diff
@@ -10,21 +10,36 @@ You can manually trigger the release workflow from the [GitHub Actions UI](https
 
 ### **Before triggering a release**
 
+**Stable release **
 1. Create a release branch. A qualified branch name should start with "release-" or "release/", case-insensitive. e.g: `release/xrpl@4.3.8`, `release-xrpl-4.3.8`, `Release/xrpl@4.3.8`.
```
Patel-Raj11: As per changes to `release.yml`, upper case release branch names are not allowed. `RELEASE.md` should be updated.

.github/workflows/release.yml
```diff
@@ -57,45 +70,89 @@ jobs:
...
+
+          {
+            echo "NPM_DIST_TAG=$NPM_DIST_TAG"
+            echo "RELEASE_BRANCH=$RELEASE_BRANCH"
+          } >> "$GITHUB_ENV"
+
+          PR_NUMBER=""
+          PR_URL=""
+
+          # For stable releases, check that a PR exists from the release branch to main
```
Patel-Raj11: I was under the impression that for stable releases, we would first merge the PR into main branch and then release it from main branch itself so that provenance commit details match. There won't be any need for `RELEASE_BRANCH` for stable releases. So, checking release -> main PR would be unnecessary for stable releases. 

For beta releases, we would release from `RELEASE_BRANCH` since, we can't merge beta branch into main. 

Do you see any issue with this approach?

RELEASE.md
```diff
@@ -10,21 +10,36 @@ You can manually trigger the release workflow from the [GitHub Actions UI](https
...
+2. Raise a PR from the release branch to main branch
+3. Update the **`version`** field in `packages/<package_name>/package.json` to the intended release version.
    ```json
    {
      "name": "<package_name>",
      "version": "x.y.z"
    }
    ```
-3. Run npm i to update the package-lock with the updated versions and commit the lock file to the release branch
+4. Set `npm distribution tag` to `latest`.
```
Patel-Raj11: This is a Github Action step and should be mentioned somewhere else (or at the end of the list), since other steps are the one developer performs on their local machine.

RELEASE.md
Patel-Raj11: Environment protection rules for `offical-review`, `github-pages `and `first-review` etc should be updated accordingly to reflect these changes.

.github/workflows/release.yml
Patel-Raj11: Another benefit of releasing stable releases from main branch is that the Git tags are tagged on the commit of main branch as opposed to the commit of release branch.

RELEASE.md
kuan121: This is nit, but I meant these two lines
1. Run npm i to update the package-lock with the updated versions and commit the lock file to the release branch
2. Run `npm i` to refresh `package-lock.json` and commit it.

Are they trying to describe the same thing?

shichengripple001: that make sense. fixed it.

RELEASE.md
shichengripple001: thanks for spotting it. I have made change to release.yml to match the environment protection rule.

kuan121 (dismissed):

RELEASE.md
shichengripple001: rewrote this section.

.github/workflows/release.yml
shichengripple001: One concern with releasing directly from the main branch is that the workflow becomes unreliable if any unrelated commits are pushed to main after the intended release commit. In that scenario, the workflow might release a wrong version.

I suggest we defer this change to a later PR if we really want to release from main. And I'd rather keep the release flow tied to the merge event, so the merge automatically triggers the release and guarantees that the workflow runs against the exact commit we intend to publish.

.github/workflows/release.yml
ckeshava: > If test failures are acceptable, can we skip the test steps entirely for beta releases? What value do the tests provide in this case?

From a record-keeping point of view, it is best to track which specific test suites failed in the CI run. Personally, I would run the test suite for all types of releases.

.github/workflows/release.yml
```diff
@@ -211,7 +267,7 @@ jobs:
           curl -X POST \
             -H "X-Api-Key: ${OWASP_TOKEN}" \
             -F "autoCreate=true" \
-            -F "projectName=xrpl-js" \
+            -F "projectName=test-alert" \
```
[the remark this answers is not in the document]
ckeshava: @shichengripple001  this appears to be a relevant concern ^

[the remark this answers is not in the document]
shichengripple001: fixed. thank for finding the issue

.github/workflows/release.yml
shichengripple001: @ckeshava it will now only run unit test for beta release now. based on the discussion on slack.

.github/workflows/release.yml
Patel-Raj11: Ya, this is a valid concern if we are releasing from `main` branch. So if we keep the current workflow as it is (releasing from release branch), then shouldn't we just keep automated PR creation in the workflow? What value it adds by removing it from the workflow?

Also, if we revert to automated PR creation from workflow, can you make sure to add the following in RELEASE.md file?
1. Close and open PR to make sure the tests run. (Unless we find a way to make sure that tests run when Github automation creates a PR).
2. Restore the release branch once it's merged (its gets auto deleted right now). This will make sure that the commit exists on some branch that was tagged by the automation.

RELEASE.md
```diff
@@ -10,39 +10,53 @@ You can manually trigger the release workflow from the [GitHub Actions UI](https
 
 ### **Before triggering a release**
 
-1. Create a release branch. A qualified branch name should start with "release-" or "release/", case-insensitive. e.g: `release/xrpl@4.3.8`, `release-xrpl-4.3.8`, `Release/xrpl@4.3.8`.
-2. Update the **`version`** field in `packages/<package_name>/package.json` to the intended release version.
+**Stable release **
+1. Create a release branch. A qualified branch name should start with "release-" or "release/", **case-insensitive**. e.g: `release/xrpl@4.3.8`, `release-xrpl-4.3.8`, `Release/xrpl@4.3.8`.
+2. Raise a PR from the release branch to main branch
```
Patel-Raj11: This should be the last step.

.github/workflows/release.yml
mvadari: The advantage of releasing from `main` (with the exception of betas) is that you can ensure that nobody has messed with the pipeline code when it's running.

shichengripple001: The pipeline will always be executed using workflow yaml from main branch. It then will checkout the release branch or feature branch specified by workflow input for packaging. The protected environment makes sure nothing can be messed on the workflow code. 
And regardless how stable release is packaged, beta release will be packaged from a beta release branch. 
so I think let's address stable release packaging with another PR when needed.
@mvadari

shichengripple001: I have unchecked "Automatically delete head branches" and checked "Enable release immutability" from repo setting, so the release branch will be kept. @Patel-Raj11

Patel-Raj11: > checked "Enable release immutability" from repo setting

1. Can we edit release notes of an immutable release?
2. Do we have any solution for this? - `Close and open PR to make sure the tests run. (Unless we find a way to make sure that tests run when Github automation creates a PR).` If not then lets add this in RELEASE.md.

shichengripple001: I will look into 1st point
for 2nd point, we can raise PR as draft, as the person running a release will convert it to "ready for review", so CI will get triggered. I am testing this.

shichengripple001: I have tried, release note can still be edited
and could you review this PR https://github.com/xpring-eng/xrpl.js/pull/132 for the requested changes @Patel-Raj11

Patel-Raj11: @shichengripple001 Thanks for the change! Can you merge the changes here (essentially the changes should be reverting back to raising the PR), so that I can review here and provide suggestions if any at a single place.

shichengripple001: @Patel-Raj11 It has been merged.

.github/workflows/faucet_test.yml
```diff
@@ -4,16 +4,24 @@ on:
   push:
     branches: [main]
   workflow_dispatch:
+  pull_request:
```
Patel-Raj11: Faucet test trigger was refactored to only run when push to `main` branch, to avoid, faucet rate limiting. Will synchronize make it run on every commit push post raising a pull request?

.github/workflows/nodejs.yml
```diff
@@ -13,13 +13,29 @@ on:
     tags:
       - '**'
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
Patel-Raj11, line 16: Why did we had to add these?

shichengripple001: added the line, because we want the test to be run for the PR automatically rasied by the pipeline. so when the PR is set to "ready for review" the test will be run, same for faucet test. @Patel-Raj11

.github/workflows/faucet_test.yml
shichengripple001: same reason as the change made on nodejs.yml @Patel-Raj11

Patel-Raj11: Then can we just have `ready_for_review` in the list? To avoid this being run on every pull request open/synchronize?

shichengripple001: made the suggested change

.github/workflows/nodejs.yml
Patel-Raj11: Ok, I am assuming that the following behaviour is not changed:
1. `node.js` tests still run for each commit on a PR + during release.
2. `faucet_test` tests run only run when PR is merged into `main` + during release.

kuan121 (approved):

Patel-Raj11 (approved):

PR #3133, opened 2025-11-17
chores: Set free-form CI inputs as env vars

pdp2121:

Changes made:

- Moved all `${{ github.* }}`, `${{ inputs.* }}`, and `${{ secrets.* }}` references from inline usage in bash scripts to the `env:` block of each step
- Referenced these values in scripts using Shell-style pattern with double quotes.
- This prevents script injection because GitHub Actions treats env: values as literal strings (preprocessor-style substitution) rather than executable code, ensuring user-controlled data cannot break out of string context to execute commands.

### Context of Change

Semgrep scanning indicated:

> Using variable interpolation `$...` with `github` context data in a `run:` step could allow an attacker to inject their own code into the runner. This would allow them to steal secrets and code. `github` context data can have arbitrary user input and should be treated as untrusted. Instead, use an intermediate environment variable with `env:` to store the data and use the environment variable in the `run:` script. Be sure to use double-quotes the environment variable, like this: "$ENVVAR".

Similar fix:  [rippled PR #5822](https://github.com/XRPLF/rippled/pull/5822)

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

## Test Plan

CI releases pass.


.github/workflows/release.yml
```diff
@@ -173,10 +180,13 @@ jobs:
         if: failure()
         env:
           SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+          PACKAGE_VERSION: ${{ env.PACKAGE_VERSION }}
```
bthomee: Here and elsewhere, this environment variable (re)definition shouldn't be necessary as it's already an env var.

.github/workflows/release.yml
```diff
@@ -343,7 +365,7 @@ jobs:
               --arg title "Release $VERSION: $RELEASE_BRANCH → main" \
               --arg head  "$RELEASE_BRANCH" \
               --arg base  "main" \
-              --arg body  "Automated PR for release **$VERSION** from **$RELEASE_BRANCH** → **main**. Workflow Run: https://github.com/$REPO/actions/runs/${{ github.run_id }}" \
+              --arg body  "Automated PR for release **$VERSION** from **$RELEASE_BRANCH** → **main**. Workflow Run: https://github.com/$REPO/actions/runs/$RUN_ID" \
```
bthomee, line 360: FWIW there are multiple ways of referring to environment variables. One way is by using GitHub's `${{ env.XXX }}` and the other is by referring to them via the Shell's `$XXX` or `${XXX}`. You can use GitHub-style of referencing them in both `if:` statements and in `run:` statements, while Shell-style can only be used in the `run:` statements.

Ayaz strongly recommends using Shell-style in `run:` statements. I'm fine either way, although I would then make sure you apply this choice consistently (i.e. either use GitHub-style or Shell-style in those statements, but do not mix and match). If you do use Shell-style, please use `${XXX}` as that can avoid all kinds of annoying errors once you want to do some string concatenation.

bthomee (commented):
Thanks for doing this - it's looking good so far.

.github/workflows/release.yml
pdp2121: Great call! Yes I was using Shell-style initially but then I saw you used Github style in your PR so I got back and fixed, but seems that I missed some places. I think I can use Shell style if it is strongly recommended and also is already used elsewhere. Will fix in a later commit

bthomee: I just checked and we only have one place left in rippled where we still use `${{ env.XXX }}` in a `run:` statement, everything else already has been migrated to using `${XXX}` per Ayaz' request.

Personally I think it's a style preference, but if you're open to it, let's follow Ayaz so we use `${XXX}` consistently wherever possible. It's entirely up to you, however. It's definitely more compact though!

.github/workflows/release.yml
pdp2121: Great catch! I removed redundant redefinition here: [9b731b7](https://github.com/XRPLF/xrpl.js/pull/3133/commits/9b731b7175a1ee176ee50e89bdcc6d5f2a2dcdb3)

.github/workflows/release.yml
pdp2121: Reverted to Shell-style here: [9b731b7](https://github.com/XRPLF/xrpl.js/pull/3133/commits/9b731b7175a1ee176ee50e89bdcc6d5f2a2dcdb3)

shichengripple001 (approved):
LGTM

.github/workflows/release.yml
```diff
@@ -30,32 +30,36 @@ jobs:
           fetch-depth: 0
 
       - name: Validate inputs
+        env:
+          REF_NAME: ${{ github.ref_name }}
+          PACKAGE_NAME_INPUT: ${{ github.event.inputs.package_name }}
```
bthomee, line 35: Some optional feedback, which is easy to fix but not a blocker in any way:

* In `env:` you now have `PACKAGE_NAME_INPUT: ${{ github.event.inputs.package_name }}`. 
* In `run:` you then do `PKG_NAME="${PACKAGE_NAME_INPUT}"`.

You can easily combine this by:
* In `env:` do `PKG_NAME: ${{ github.event.inputs.package_name }}`.
* In `run:` remove the assignment statement. The `PKG_NAME` can be used directly in the script.

Separately, I noticed you sometimes refer to `PACKAGE_XXX` and sometimes to `PKG_XXX` for some `XXX`. You could consider doing a find-and-replace to make them all consistent. I'd suggest replacing all occurrences of `PACKAGE` by `PKG`  since it's shorter and its meaning is clear.

.github/workflows/release.yml
```diff
@@ -30,32 +30,36 @@ jobs:
...
+          NPMJS_DIST_TAG_INPUT: ${{ github.event.inputs.npmjs_dist_tag }}
```
bthomee, line 36: The same applies to this one, where you define `NPMJS_DIST_TAG_INPUT` here and then in the code you assign it to `NPM_DIST_TAG`; you may as well use `NPM_DIST_TAG` here directly.

.github/workflows/release.yml
```diff
@@ -200,12 +209,14 @@ jobs:
           severity: CRITICAL,HIGH
 
       - name: Upload sbom to OWASP
+        env:
+          OWASP_TOKEN: ${{ secrets.OWASP_TOKEN }}
         run: |
           curl -X POST \
-            -H "X-Api-Key: ${{ secrets.OWASP_TOKEN }}" \
+            -H "X-Api-Key: ${OWASP_TOKEN}" \
```
bthomee, line 216: FWIW I believe it's not a concern to directly reference secrets via `${{ secrets.XXX }}` in a `run:` statement, since end users cannot manipulate them (unlike the GitHub object, according to Semgrep).

Either way, assigning it to an env var is perfectly fine, so you can leave this change as -is.

.github/workflows/release.yml
```diff
@@ -67,53 +71,56 @@ jobs:
...
           fi
-          if [[ "$NPM_DIST_TAG" == "latest" ]] && ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-            echo "With npmjs_dist_tag 'latest', version must be of the form x.y.z. Found '$VERSION'." >&2
+          if [[ "${NPM_DIST_TAG}" == "latest" ]] && ! [[ "${VERSION}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "With npmjs_dist_tag 'latest', version must be of the form x.y.z. Found '${VERSION}'." >&2
             exit 1
           fi
-          echo "PACKAGE_VERSION=$VERSION" >> "$GITHUB_ENV"
-          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+          echo "PACKAGE_VERSION=${VERSION}" >> "$GITHUB_ENV"
```
bthomee, line 122:
```suggestion
          echo "package_version=${VERSION}" >> "$GITHUB_OUTPUT"
```

In follow-up jobs you refer to `${{ needs.get_version.outputs.package_version }}` but that will always be empty, since you're writing the version as an env var and not as an output.

.github/workflows/release.yml
```diff
@@ -491,36 +506,55 @@ jobs:
...
         shell: bash
         env:
-          SLACK_TOKEN:     ${{ secrets.SLACK_TOKEN }}
-          CHANNEL:         "#ripplex-security"
-          EXECUTOR:        ${{ github.triggering_actor || github.actor }}
-          PACKAGE_NAME:    ${{ needs.get_version.outputs.package_version && github.event.inputs.package_name }}
+          SLACK_TOKEN: ${{ secrets.SLACK_TOKEN }}
+          CHANNEL: "#ripplex-security"
+          EXECUTOR: ${{ github.triggering_actor || github.actor }}
+          PACKAGE_NAME: ${{ needs.get_version.outputs.package_version && github.event.inputs.package_name }}
```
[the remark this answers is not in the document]
bthomee: This seems a bit questionable indeed.

I think - after making the fix I proposed in another comment - that using `{{ needs.get_version.outputs.package_version }}` should be enough, since now the value will get correctly set.

If the output isn't always set (or can be empty), then you are more likely to want to use `||` here than `&&`. However, in that case you will want to apply the same login in other places too, e.g. when running the faucet test.

PR #3134, opened 2025-11-18
chores: assorted fixes for `release.yaml`

pdp2121:

Address some outstanding comments in https://github.com/XRPLF/xrpl.js/pull/3133.
Changes include:
- Use env variable names `PKG_NAME`, `PKG_VERSION`, and `NPM_DIST_TAG` consistently within the script and remove unnecessary intermediate variables.
- Write package version in line https://github.com/XRPLF/xrpl.js/blob/main/.github/workflows/release.yml#L122 into `$GITHUB_OUTPUT` instead of `$GITHUB_ENV` to be used in other jobs (instead of the same job).
- Fix line https://github.com/XRPLF/xrpl.js/blob/main/.github/workflows/release.yml#L526 to just use `${{ needs.get_version.outputs.package_name }}` directly since this should be available and && operator does not make much sense in this context.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Get release version from package.json
     outputs:
-      package_version: ${{ steps.get_version.outputs.version }}
+      package_version: ${{ steps.get_version.outputs.package_version }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4


pdp2121:
@shichengripple001 would you mind taking a look at these changes to make sure nothing will break? Thanks!

kuan121 (approved):

.github/workflows/release.yml
```diff
@@ -32,8 +32,8 @@ jobs:
       - name: Validate inputs
         env:
           REF_NAME: ${{ github.ref_name }}
-          PACKAGE_NAME_INPUT: ${{ github.event.inputs.package_name }}
-          NPMJS_DIST_TAG_INPUT: ${{ github.event.inputs.npmjs_dist_tag }}
+          PKG_NAME: ${{ github.event.inputs.package_name }}
+          NPM_DIST_TAG: ${{ github.event.inputs.npmjs_dist_tag }}
```
bthomee, line 36: You can just name this `DIST_TAG` directly, since in the script below you call `DIST_TAG="${NPM_DIST_TAG}"` and don't use `NPM_DIST_TAG` for anything else.

pdp2121: In line 76 we reassign the value to `latest` which I believe is not allowed for env variable, that's why we need an intermediate value. However, this is only used for printing out messages so we can use something like `${NPM_DIST_TAG:-latest}`. Let me know what you think

pdp2121:
> Line 25 must be updated to reference the correct output key:
> 
> ```diff
>   outputs:
> -   package_version: ${{ steps.get_version.outputs.version }}
> +   package_version: ${{ steps.get_version.outputs.package_version }}
> ```

Great catch! Fixed in a later commit

.github/workflows/release.yml
bthomee: > In line 76 we reassign the value to latest which I believe is not allowed for env variable, that's why we need an intermediate value.

Are you saying that if `${{ github.event.inputs.npmjs_dist_tag }}` is "latest" then GitHub will not execute the workflow?

Assuming you'd rename `NPM_DIST_TAG` to `DIST_TAG` here in the `env:` section, then below you can just do:

```
# Empty → default to 'latest'
if [ -z "${DIST_TAG}" ]; then
  DIST_TAG="latest"
  echo "ℹ️ npmjs_dist_tag empty → defaulting to 'latest'."
fi
```

no? I've never heard of the value "latest" not being allowed as an env var, that would be very odd. (then again, Gmail doesn't allow you to use a label "Finance", so sometimes things one sees in the world are just strange).

pdp2121: Actually, you are right, it's just a good practice not to reassign env variable since it is shared within the job, but you can certainly do so. Also, in our case it does not affect anything since these steps are at the end of the job. However, since this is pretty minor, I would prefer to keep it as is just to follow good practice

bthomee: That's fair. I don't know if next steps will use the original or the modified env var.

bthomee: btw you can also just do:

```
DIST_TAG: ${{ github.event.inputs.npmjs_dist_tag || "latest" }}
```

and then you don't need the check in the script.

bthomee (approved):

Issue #3152, opened 2025-12-08
SAV integration test cases are failing post release of new docker image


Patel-Raj11:
Failing CI - https://github.com/XRPLF/xrpl.js/actions/runs/20033510776
New docker images - https://hub.docker.com/r/rippleci/rippled/tags

PR #3153, opened 2025-12-08
Mark AssetsAvailable, AssetsTotal and LossUnrealized of Vault object as Optional

kuan121:

Mark the AssetsAvailable, AssetsTotal, and LossUnrealized fields of the Vault object as optional to match the behavior introduced by the recent rippled changes ([git blame](https://github.com/XRPLF/rippled/blame/develop/include/xrpl/protocol/detail/ledger_entries.macro#L483-L486)), which change those fields from `soeREQUIRED` to `soeDEFAULT`.

### Context of Change
Fix https://github.com/XRPLF/xrpl.js/issues/3152

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

* Updated existing integration tests

### DIFF
--- a/packages/xrpl/HISTORY.md
+++ b/packages/xrpl/HISTORY.md
@@ -7,6 +7,7 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 ### Fixed
 * Update ripple-binary-codec to 2.5.1 to address serialization/deserialization issues in `Issue` serialized type for `MPTIssue`.
 * Better faucet error handling
+* Mark the `AssetsAvailable`, `AssetsTotal`, and `LossUnrealized` fields of the Vault object as optional.
 
 ## 4.4.3 (2025-11-07)
 


packages/xrpl/src/models/methods/vaultInfo.ts
```diff
@@ -48,12 +48,12 @@ export interface VaultInfoResponse extends BaseResponse {
       /**
        * Amount of assets currently available for withdrawal.
        */
-      AssetsAvailable: string
+      AssetsAvailable?: string
```
Patel-Raj11, line 51: It seems like a breaking change is introduced in rippled that makes these properties optional that were once required. If someone if accessing them directly as we did in integration tests, it would break their code.

Not to block this, but should we confirm this in `cpp-informal` or `xrpl-lending-protocol-core` slack channel?

kuan121: Thanks for bringing this up. I asked in #cpp-informal. The conclusion is this is fine since SAV is not even up for voting yet and breaking changes are allowed.

Patel-Raj11: Saw the note from Mayukha, since we are releasing library even before the amendment goes live on Mainnet, we should allow such breaking changes as bug fixes OR keep releasing, the features as release candidates and only release a stable version once the amendment is live on mainnet (to follow Semver). Some thoughts...

kuan121: What you said makes sense. Feel free to champion this, and we should adopt it if it brings benefits to the team and the community.

ckeshava: > we are releasing library even before the amendment goes live on Mainnet, we should allow such breaking changes as bug fixes

Where can people use the client library features for Single-Asset-Vault? Apart from standalone nodes on their local nodes, is it available for use on public networks? (The XRPLF developers have the right to introduce breaking changes into the `devnet`, because features are experimental, by definition on the `devnet`.)

packages/xrpl/src/models/ledger/Vault.ts
```diff
@@ -49,17 +49,17 @@ export default interface Vault extends BaseLedgerEntry, HasPreviousTxnID {
   /**
    * The total value of the vault.
    */
-  AssetsTotal: string
+  AssetsTotal?: string
```
ckeshava, line 52: Note: These fields have been marked as "Default", not "Optional". Hence, we need to introduce a Default value for these fields. I'm not sure if there is a precedent in the xrpl.js library, we'll need to look into it.

packages/xrpl/test/integration/requests/vaultInfo.test.ts
```diff
@@ -80,12 +80,12 @@ describe('Single Asset Vault', function () {
       VaultWithdrawalPolicy.vaultStrategyFirstComeFirstServe,
     )
     assert.equal(
-      vault.AssetsTotal,
+      vault.AssetsTotal ?? '0',
```
ckeshava, line 83: Why is the test file supplying a default value for these fields? The default value of `0` must have been provided by rippled automatically. 

It is preferable to remove the `??` operator because it will inform readers about the choices for default values.
```suggestion
      vault.AssetsTotal,
```

packages/xrpl/HISTORY.md
```diff
@@ -7,6 +7,7 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 ### Fixed
 * Update ripple-binary-codec to 2.5.1 to address serialization/deserialization issues in `Issue` serialized type for `MPTIssue`.
 * Better faucet error handling
+* Mark the `AssetsAvailable`, `AssetsTotal`, and `LossUnrealized` fields of the Vault object as optional.
```
ckeshava, line 10:
```suggestion
* Mark the `AssetsAvailable`, `AssetsTotal`, and `LossUnrealized` fields of the Vault object as Default.
```

packages/xrpl/src/models/methods/vaultInfo.ts
```diff
@@ -48,12 +48,12 @@ export interface VaultInfoResponse extends BaseResponse {
...
 
       /**
        * Total amount of assets in the vault.
        */
-      AssetsTotal: string
+      AssetsTotal?: string
```
ckeshava, line 56: Why are these fields marked as Optional in the `vaultInfo` RPC method? As per my understanding, the `vaultInfo` RPC response prints the Default values for these fields, even if the genesis `VaultCreate` transaction did not explicitly specify these fields.

In your experiments, did you observe that `vaultInfo` does not print these fields?

packages/xrpl/src/models/ledger/Vault.ts
kuan121: Those fields of the Vault ledger entry were updated from soeREQUIRED to soeDEFAULT, and, by definition, soeDEFAULT is [optional, if present, must not have default value](https://github.com/XRPLF/rippled/blob/42a432c5dc2af2754f4083b93f0ee0d616092681/include/xrpl/protocol/SOTemplate.h#L18). The correct TypeScript representation is simply making the field optional with ?, which is what we've done in this library, e.g., `AccountRoot.AccountTxnID?`, `AccountRoot.AMMID?`, `AccountRoot.BurnedNFTokens?`, `AccountRoot.BurnedNFTokens?`

packages/xrpl/src/models/methods/vaultInfo.ts
kuan121: > In your experiments, did you observe that vaultInfo does not print these fields?

Correct. The existing integration tests are failing because vaultInfo RPC doesn't include those fields in the response when they have default values. This match the definition of an `soeDEFAULT` field, which is [optional, if present, must not have default value](https://github.com/XRPLF/rippled/blob/42a432c5dc2af2754f4083b93f0ee0d616092681/include/xrpl/protocol/SOTemplate.h#L18). In other words, If they have the default values, they won't be present in rippled, and thus Vault related RPC responses.

packages/xrpl/test/integration/requests/vaultInfo.test.ts
kuan121: > The default value of 0 must have been provided by rippled automatically.

This is no longer true, and why those tests are failing. Without the change, we are performing `assert.equal(vault.AssetsTotal=undefined, '0', 'New Vault should have zero total assets')`

packages/xrpl/HISTORY.md
kuan121: I don't think we need to make this change. In rippled, `soeDEFAULT` is an internal serialization style. In xrpl.js/Typescript, fields are either required or optional (marked with ?). There is no concept of "Default" fields in TypeScript.

ckeshava (approved):
Thanks for the clarification: I was under the misapprehension that rippled stores ledger-object fields with "Default" values, that is not true.

I have created another ticket to track the misconfiguration of other ledger-objects, with respect to the `DEFAULT` parameter. https://github.com/XRPLF/xrpl.js/issues/3154

PR #3156, opened 2025-12-15
Upgrade lerna to v8

ckeshava:
This PR aims to accomplish the following:
1. Upgrade lerna to v8.
2. Update parse-url, parse-path, lodash.template, form-data dependencies to newer versions.

This PR succeeds the following dependabot PR: https://github.com/XRPLF/xrpl.js/pull/3051

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

The existing tests pass. No new tests are required for this upgrade.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "eslint-plugin-n": "^17.21.3",
     "eslint-plugin-prettier": "^5.5.4",
     "eslint-plugin-tsdoc": "^0.4.0",
-    "expect": "^29.3.1",
+    "expect": "^30.2.0",
     "jest": "^29.3.1",
     "jest-mock": "^29.3.1",
     "lerna": "^8.2.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
         "eslint-plugin-n": "^17.21.3",
         "eslint-plugin-prettier": "^5.5.4",
         "eslint-plugin-tsdoc": "^0.4.0",
-        "expect": "^29.3.1",
+        "expect": "^30.2.0",
         "globals": "^16.3.0",
         "jest": "^29.3.1",
         "jest-mock": "^29.3.1",
@@ -99,6 +99,7 @@
       "version": "7.28.0",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.27.1",
@@ -602,6 +603,7 @@
       "version": "4.5.0",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "escape-string-regexp": "^4.0.0",
         "ignore": "^5.2.4"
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -48,7 +48,7 @@
     "expect": "^29.3.1",
     "jest": "^29.3.1",
     "jest-mock": "^29.3.1",
-    "lerna": "^4.0.0",
+    "lerna": "^8.2.3",
```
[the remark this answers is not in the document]
ckeshava: We will consider v9 at a later date. For the purposes of resolving security vulnerability issues, lerna v8 suffices.

Patel-Raj11, line 51: It would be good idea to test releasing a package using lerna commands by following manual release steps. Apart from that, this change looks good!

For ref: https://github.com/XRPLF/typescript-style/pull/24#issuecomment-2815848817

mvadari: Yes, definitely do that - the last time we tried upgrading a major version of lerna (v6), the release failed, and we had to emergency downgrade

ckeshava: I have published three versions of a testing xrpl package with the new lerna version: https://www.npmjs.com/package/xrpl-chenna-testing-1?activeTab=versions

I used [this git-branch](https://github.com/ckeshava/xrpl.js/pull/new/testLernaPublish) to publish these versions. This git branch is identical to this PR, except for top three commits which have synthetically incremented the versions for the `xrpl` package inside the monorepo.

ckeshava: > Yes, definitely do that - the last time we tried upgrading a major version of lerna (v6), the release failed, and we had to emergency downgrade

That sounds like a big issue for all users of Lerna. I'm guessing lerna provided a subsequent fix.

In any case, can't we always use `npm publish`? I understand that lerna makes our life easier with the child packages, however is it a critical blocker?

Patel-Raj11: > In any case, can't we always use npm publish? 


I think with lerna in place, `npm pack` does not work out of the box. So, the release pipeline had to package it using lerna before publishing - https://github.com/XRPLF/xrpl.js/blob/d6c89f93f8b2e4398b226324ae589a414b7402db/.github/workflows/release.yml#L298

Thus lerna needs to work for npm publish to work.

Patel-Raj11 (approved):

package.json
mvadari: > That sounds like a big issue for all users of Lerna. I'm guessing lerna provided a subsequent fix.

Probably was an issue with our setup, not Lerna (it was a Dependabot update so it wasn't really checked that hard)

mvadari: > In any case, can't we always use `npm publish`? I understand that lerna makes our life easier with the child packages, however is it a critical blocker?

There was some talk about switching to npm workspaces a while back, perhaps that's worth reinvestigating. If npm has sufficient functionality built-in, then we wouldn't need the complexity of lerna.

ckeshava: > I think with lerna in place, npm pack does not work out of the box. So, the release pipeline had to package it using lerna before publishing -

`npm pack` works for me. You need to go into the `packages/xrpl/` folder and execute this command. It successfully produces a tarball for me. However, the disadvantage is that we need to release all the children packages seperately.

kuan121 (approved):

ckeshava:
Thank you very much for the comments and reviews 🙇

PR #3157, opened 2025-12-16
feat: export more helper functions

mvadari:

This PR exports:
* `signMultiBatch`
* `combineBatchSigners`
* `multisign`
* `authorizeChannel`
* `encodeForSigningBatch`

in xrpl.js.

### Context of Change

These functions are completely inaccessible if you install the minified version xrpl.js and use it in a webpage.

### Type of Change

- [x] Yes

## Test Plan

Works locally.

### DIFF
--- a/packages/xrpl/src/Wallet/index.ts
+++ b/packages/xrpl/src/Wallet/index.ts
@@ -515,6 +515,6 @@ function removeTrailingZeros(tx: Transaction): void {
 
 export { signMultiBatch, combineBatchSigners } from './batchSigner'
 
-export { multisign } from './signer'
+export { multisign, verifySignature } from './signer'
 
 export { authorizeChannel } from './authorizeChannel'


kuan121 (approved):

packages/xrpl/src/Wallet/index.ts
```diff
@@ -512,3 +512,9 @@ function removeTrailingZeros(tx: Transaction): void {
     tx.Amount.value = new BigNumber(tx.Amount.value).toString()
   }
 }
+
+export { signMultiBatch, combineBatchSigners } from './batchSigner'
```
Patel-Raj11, line 516: 1. `signMultiBatch` and `combineBatchSigners` are already exported directly from src/index.ts using `export * from './Wallet/batchSigner'` and I can see them [here](https://unpkg.com/xrpl@4.4.3/build/xrpl-latest-min.js).
2. Similarly `multisign` is already exported.
3. Should we add export for `authorizeChannel` in `src/index.ts` ?

mvadari: Huh, looks like I was looking at 4.4.1, not 4.4.3 - that's probably why I didn't catch it earlier. I pushed a new commit that cleans that up so we don't need to add every single thing to `src/index.ts`.

kuan121 (approved):

PR #3191, opened 2026-01-23
Fix STNumber serialization and add counterparty signing helper functions

Patel-Raj11:

Recent Number type refactor in rippled changed the following:
1. mantissa range from [10^15, 10^16-1] to [10^18, 10^19-1].
2. Shrinking mantissa if it goes beyond max signed 64-bit integer (2^63 - 1). Occurs due to increased mantissa range.

This PR implements above two changes and fixes toJSON (string representation) to match with rippled's behaviour.

Additional changes in PR:
1. Adds helper functions `signLoanSetByCounterparty` and `combineLoanSetCounterpartySigners` helper functions.
2. Adds temporary fix to docker container so that rippled node runs. This can be reverted once Michael confirms the root cause and provides a permanent fix.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Added unit tests that takes expected value from standalone rippled instance.

### DIFF
--- a/packages/ripple-binary-codec/src/types/st-number.ts
+++ b/packages/ripple-binary-codec/src/types/st-number.ts
@@ -140,6 +140,18 @@ function normalize(
 
   if (lastDigit != null && lastDigit >= BigInt(5)) {
     m += BigInt(1)
+    // After rounding, mantissa may exceed MAX_INT64 again
+    if (m > MAX_INT64) {
+      if (exponent >= MAX_EXPONENT) {
+        throw new Error('Exponent overflow: value too large to represent')
+      }
+      lastDigit = m % BigInt(10)
+      exponent += 1
+      m /= BigInt(10)
+      if (lastDigit >= BigInt(5)) {
+        m += BigInt(1)
+      }
+    }
   }
 
   if (isNegative) m = -m
--- a/packages/ripple-binary-codec/test/st-number.test.ts
+++ b/packages/ripple-binary-codec/test/st-number.test.ts
@@ -180,6 +180,12 @@ describe('STNumber', () => {
     expect(sn.toJSON()).toEqual('1e22')
   })
 
+  it('mantissa > MAX_INT64', () => {
+    const value = '92233720368547758079'
+    const sn = STNumber.from(value)
+    expect(sn.toJSON()).toEqual('922337203685477581e2')
+  })
+
   it('throws on exponent overflow (value too large)', () => {
     // 1e40000 has exponent 40000, after normalization exponent = 40000 - 18 = 39982
     // which exceeds MAX_EXPONENT (32768)


packages/ripple-binary-codec/src/types/st-number.ts
```diff
@@ -87,16 +88,41 @@ function normalize(
   let m = mantissa < BigInt(0) ? -mantissa : mantissa
   const isNegative = mantissa < BigInt(0)
 
-  while (m !== BigInt(0) && m < MIN_MANTISSA && exponent > MIN_EXPONENT) {
+  if (m > MAX_INT64) {
```
shawnxie999: just want to make sure, `m` here is already the mantissa with all trailing zeroes stripped, right?

packages/ripple-binary-codec/src/types/st-number.ts
```diff
@@ -87,16 +88,41 @@ function normalize(
...
+  // Handle overflow: if mantissa exceeds MAX_INT64 (2^63-1) after growing.
+  // Rounding is not required because last digit is always 0 after growing.
+  if (m > MAX_INT64) {
+    if (exponent >= MAX_EXPONENT) {
+      throw new Error('Exponent overflow: value too large to represent')
+    }
     exponent += 1
     m /= BigInt(10)
   }
+
```
shawnxie999, line 156: just want to make sure, do we really need to divide here
```
    exponent += 1
    m /= BigInt(10)
```

Take `99e20` as example. Initially the beginning of function, `m` is `99` and exponent is `20`.

After `m` is brought to at least `MIN_MANTISSA`, `m` becomes `990,000,000,000,000,000` (18 digits) and `exponent` is `4`.

 `m = 990,000,000,000,000,000` and `exponent = 4` is already an acceptable/serializable input. So I don't believe we need to check if it's greater than `INT64_MAX` and to divide it further

Patel-Raj11: Actually after bringing `m` to be at-least `MIN_MANTISSA`, it's value becomes `9,900,000,000,000,000,000` and that is > `INT64_MAX`. So we are reducing mantissa by 10 and increasing exponent by 1. I added a test where the `binary` is taken from standalone rippled instance and `json` is what we get if we sign the same transaction using xrpl.js  - [0e31a83](https://github.com/XRPLF/xrpl.js/pull/3191/commits/0e31a832bee26aee4d4691436acffcb963ab2993)

packages/ripple-binary-codec/src/types/st-number.ts
Patel-Raj11: No, if the user enters the input as `990000000000000`. Then `m` here would be `990000000000000`.  But, I think it should be changed to the following to provide more room for other digits:

1. 99000 -> m = 99, exponent = 3
2. 99.99000 -> m = 9999, exponent = -2
3. 0.0025 -> m = 25, exponent = -4
4. 0.002500 -> m = 25, exponent = -4
5. 0.250025 -> m = 250025, exponent = -6

I will make that change and add some tests.

packages/ripple-binary-codec/src/types/st-number.ts
shawnxie999: sorry my bad, yes 9,900,000,000,000,000,000 would be m, but I believe that is still acceptable as long as it's less than MAX_MANTISSA

shawnxie999: what would happen if we removed this block of code? Would any tests fail?

Patel-Raj11: Ya, the transaction fails with `temMALFORMED` if we remove the `if` block.

packages/ripple-binary-codec/src/types/st-number.ts
Patel-Raj11: I removed trailing zeroes from mantissa and added tests that passes (Otherwise would have failed due to `m > MAX_INT64`) in [9e3c80e](https://github.com/XRPLF/xrpl.js/pull/3191/commits/9e3c80ef06dc4b7deb5c9a77468739a24a879c24)

packages/ripple-binary-codec/src/types/st-number.ts
shawnxie999: I just tested in rippled. A number with `m = 99` and `exponent = 20` would turn into `m = 9,900,000,000,000,000,000` (19 digits) and `exponent = 3` after normalization.

So `m` can actually be greater than INT64_MAX, as long as the significant digits (`99`) is less than INT64_MAX

shawnxie999: I'm wondering what removing this if-block would do, since you just pushed a change that removes trailing zeroes from the mantissa

Patel-Raj11: If we remove the if-block (do not bring back increased manitssa <= INT64_MAX) the transaction fails with `temMALFORMED` because the transaction that got submitted does not have properly encoded `AssetsMaximum` because we use `writeInt64BE` to write the mantissa into bytes and now it's value is beyond `INT64_MAX`.

```
const bytes = new Uint8Array(12)
writeInt64BE(bytes, normalizedMantissa, 0)
writeInt32BE(bytes, normalizedExponent, 8)
```

This is the transaction that essentially gets submitted to rippled and fails.
```
{
        TransactionType: 'VaultCreate',
        Sequence: 65,
        Fee: '5000000',
        SigningPubKey: 'ED8455056CF70CAFB7F569C4319EAD3875295B6B4DA0AD5AA57C4E1A690B0AC3E5',
        TxnSignature: '32A5DD46982A7F2FEDD06CE85C95B2B5158FB26787D279FB36DF16C592FB7F576EDAA59A71D392362E255EB108B342D06B1E2010CFEA8198FB93529B6082170E',
        Account: 'rHHJLEQ42gJrakkCfNebAkKFiYZPmFw6u3',
        AssetsMaximum: '-8546744073709551616e3',
        Asset: { currency: 'USD', issuer: 'rGertFz66AA5zJKMijfBZ8UBuWWMjLRbHo' }
}
```

However, if we truncate the manitssa to `99` (so that significant digits are less than INT64_MAX) and move all the trailing zeroes to exponent to just before encoding it to bytes, the transaction fails at rippled with `fails local checks: Invalid signature.`.

Can you help in confirming that if we need to use `writeUInt64BE` for encoding so that we can go beyond signed INT64_MAX but stay within `MAX_MANTISSA` while encoding?

Also, if we keep the current logic (truncating if it goes beyond INT64_MAX) this is the bytes representation `0DBD2FC137A3000000000004`  for 99e20 when encoded. Can you check if you get the same representation in rippled when a request is parsed when received using `sign and submit` mode of `submit` command?

packages/ripple-binary-codec/src/types/st-number.ts
```diff
@@ -87,16 +94,41 @@ function normalize(
   let m = mantissa < BigInt(0) ? -mantissa : mantissa
   const isNegative = mantissa < BigInt(0)
 
-  while (m !== BigInt(0) && m < MIN_MANTISSA && exponent > MIN_EXPONENT) {
+  if (m > MAX_INT64) {
```
ckeshava: Hello Raj, rippled allows users to encode larger value inside the Number type. From a compatibility PoV, we need to do the same in client libraries.

I tested in this file: `src/test/protocol/STNumber_test.cpp`
```
BEAST_EXPECT(
                numberFromJson(sfNumber, "9223372036854775808") == STNumber(sfNumber, Number(9'223'372'036'854'775'807) + Number(1)));
```

ckeshava: Even if the user input exceeds, `9223372036854775807`, the representation is managed here: https://github.com/XRPLF/rippled/blob/fe9c8d568fcf6ac21483024e01f58962dd5c8260/src/libxrpl/basics/Number.cpp#L414

packages/ripple-binary-codec/src/types/st-number.ts
shawnxie999: My apologies, yes you are right. The current implementation works and we need to keep the if block. 

My initial confusion came from mixing up the two. The requirement to fit the mantissa into `largeRange` only applies to rippled’s _internal representation_. During serialization, however, the mantissa is serialized as a 64-bit value. Because of that, the mantissa must be reduced (by dividing by 10 as needed) so that it fits within `INT64_MAX`.

After taking another look at the `STNumber` serialization [function](https://github.com/XRPLF/rippled/blob/fe9c8d568fcf6ac21483024e01f58962dd5c8260/src/libxrpl/protocol/STNumber.cpp#L59) , it first retrieves the [mantissa](https://github.com/XRPLF/rippled/blob/fe9c8d568fcf6ac21483024e01f58962dd5c8260/src/libxrpl/protocol/STNumber.cpp#L65), which has been [divided](https://github.com/XRPLF/rippled/blob/fe9c8d568fcf6ac21483024e01f58962dd5c8260/include/xrpl/basics/Number.h#L552) it by 10 so that the mantissa is less that `INT64_MAX`. This is exactly what you are doing and it works. Sorry for the confusion.

Patel-Raj11: Thanks for confirming this specific logic and going out of your way and helping out with the serialization part.

packages/ripple-binary-codec/src/types/st-number.ts
Patel-Raj11: So, if the users input has manitssa with significant digits that goes beyond `9223372036854775807` we will have to truncate it and do rounding in client library. This would mean we would silently change the number that user provided initially and submit it to rippled. To avoid that behaviour we are throwing an error straight away if the significant digits goes beyond `9223372036854775807`. This does not mean inputs greater than `9223372036854775807` won't be allowed. (i.e. numbers like `99e40` will work just fine).

I discussed this part with @shawnxie999 when implementing and understanding the rounding part. But we can discuss further if you feel strongly that we should do rounding in client library and change the approach if it makes sense.

shawnxie999: Consider the value `9,999,999,999,999,999,123,000,000,000` as an IOU amount. In this case, rounding could cause it to be represented as `9,999,999,999,999,999,120,000,000,000`, resulting in a loss of `3,000,000,000 `purely due to rounding.

Although rippled technically allows this behavior, it is acknowledged to be extremely undesirable and potentially dangerous. This is precisely one of the reasons we are moving away from IOUs toward MPTs, so that numbers are represented in integers without loss of precision.

While we could mimic rippled’s rounding behavior in the client library for consistency, IMO it is better to proactively safeguard against these type of scenarios.

shawnxie999: @Patel-Raj11 it might be worth putting a comment where the exception thrown when it's checking if `m > INT64_MAX` to explain why it is different from rippled.

Patel-Raj11: Makes sense, I'll add a code comment once @ckeshava goes through it and we reach a conclusion.

packages/ripple-binary-codec/src/types/st-number.ts
ckeshava: @shawnxie999  -- I agree with your characterization of internal implementation details of the `Number::mantissa()` and `Number::exponent()` methods. However, from an external API perspective, a user will be able to serialize really large values using the Number class in rippled. (I'm referring to values larger than `9'223'372'036'854'775'807`).

I want to clarify that this is not possible with the current design of the client libraries. In the client libraries, such large inputs will invoke an overflow_error, rather than the rounding behavior observed in rippled.

packages/ripple-binary-codec/src/types/st-number.ts
ckeshava: okay, this makes sense to me 👍 . I will follow a similar behavior in the python library.

packages/ripple-binary-codec/src/types/st-number.ts
ckeshava: okay, I followed the conversation in the below thread just now. I think we are all expressing the same situation in different ways. It makes sense to me now 👍

packages/ripple-binary-codec/test/st-number.test.ts
```diff
@@ -4,21 +4,58 @@ import { coreTypes } from '../src/types'
 const { Number: STNumber } = coreTypes
 
 describe('STNumber', () => {
-  it('should encode and decode integers', () => {
-    const value = '9876543210'
+  it('+ve normal value', () => {
```
ckeshava, line 7: thanks for writing extensive test cases. They are very useful.

packages/xrpl/test/integration/transactions/lendingProtocol.test.ts
```diff
@@ -62,6 +60,131 @@ describe('Lending Protocol IT', () => {
     await teardownClient(testContext)
   }, TIMEOUT)
 
+  it(
```
achowdhry-ripple, line 63: this is very useful, thank you for adding the integration test

achowdhry-ripple (approved):
code+syntax lgtm, approving to not block release -- am less familiar with the context on the serialization changes but did not see an issue

packages/ripple-binary-codec/src/types/st-number.ts
```diff
@@ -79,7 +80,7 @@ function extractNumberPartsFromString(val: string): {
 /**
  * Normalize the mantissa and exponent to XRPL constraints.
  *
- * Ensures that after normalization, the mantissa is between MIN_MANTISSA and MAX_INT64.
+ * Ensures that after normalization, the mantissa is between MIN_MANTISSA and MAX_MANTISSA.
```
ckeshava: The original comment is still valid. After normalization, if the mantissa is larger than MAX_INT64, then the mantissa will be reduced again. So, there is no need to update this comment.

packages/ripple-binary-codec/src/types/st-number.ts
```diff
@@ -94,16 +95,8 @@ function normalize(
...
-  // the 64-bit signed integer range, but we will throw an error as we don't want to silently
-  // change the user's input before serialization.
-  // For numbers where mantissa > 2^63-1, (e.g. 9,999,999,999,999,999,123,000,000,000)
-  // The mantissa would be 9,999,999,999,999,999,123 and since it is > 2^63-1, we would need
-  // to round it to fit within the 64-bit signed integer range. This would mean that mantissa
-  // would be rounded down to 9,999,999,999,999,999,120, since 9,999,999,999,999,999,123 % 10 < 5.
-  // This results in total loss of 9,999,999,999,999,999,123,000,000,000 - 9,999,999,999,999,999,120 = 3,000,000,000
-  // Which is a significant change in overall value.
-  if (m > MAX_INT64) {
+  // Mantissa is too large, throw error as rippled fails transaction with invalidParams
```
ckeshava: Do we need this if-condition anymore? The below while loop (`while (m > MAX_MANTISSA) {...}`) indicates that larger values also undergo normalization. So  we should not throw this error in this if-condition.

packages/ripple-binary-codec/src/types/st-number.ts
```diff
@@ -118,6 +111,15 @@ function normalize(
     m *= BigInt(10)
   }
 
+  // Shrink mantissa until it fits within MAX_MANTISSA
+  while (m > MAX_MANTISSA) {
+    if (exponent >= MAX_EXPONENT) {
+      throw new Error('Mantissa and exponent are too large')
+    }
+    exponent += 1
+    m /= BigInt(10)
```
ckeshava, line 118: Before we remove the last digit in the decimal representation of `m`, we need to record it inside the rounding calculator. Like `lastDigit` in the below code.

packages/ripple-binary-codec/src/types/st-number.ts
Patel-Raj11: Yes, we need this since, rippled throws an error straight away when mantissa (without trailing zeroes) goes beyond MAX_MANTISSA. If we remove this check then we would be truncating, so better to throw an error.

Sample response from rippled:

```
{
    "api_version": 2,
    "deprecated": "Signing support in the 'submit' command has been deprecated and will be removed in a future version of the server. Please migrate to a standalone signing tool.",
    "error": "invalidParams",
    "error_code": 31,
    "error_message": "Field 'tx_json.AssetsMaximum' has invalid data.",
    "request": {
        "api_version": 2,
        "command": "submit",
        "secret": "<masked>",
        "tx_json": {
            "Account": "rQpNo3BWw2VBfhcpX9fBTvZ9CrjkFePMvd",
            "Asset": {
                "currency": "USD",
                "issuer": "rNpF5g1PQzommmihmcZNWN44kQkSoeyCgn"
            },
            "AssetsMaximum": "99999999999999999991",
            "Fee": "5000000",
            "Sequence": 68,
            "TransactionType": "VaultCreate"
        }
    },
    "status": "error",
    "type": "response"
}
```

packages/ripple-binary-codec/src/types/st-number.ts
Patel-Raj11: Since we throw an error if m > MAX_MANTISSA, any increased mantissa here would be due to multiplication by 10. So we can safely truncate it.

packages/ripple-binary-codec/src/types/st-number.ts
Patel-Raj11: @shawnxie999 FYI - We have introduced the rounding similar to `rippled` so that we stay consistent with rippled and QA team can test with potentially high/low values without being blocked by client libraries. - [41099a7](https://github.com/XRPLF/xrpl.js/pull/3191/commits/41099a7aa23d67882cf64a372cf94a2b0352fff8)

packages/ripple-binary-codec/src/types/st-number.ts
Patel-Raj11: @ckeshava I removed the checks in https://github.com/XRPLF/xrpl.js/pull/3191/commits/739708ef43941030ebffe0c44d86f2f8e148b3ff

Patel-Raj11: Also, addressed an edge case - [46f0507](https://github.com/XRPLF/xrpl.js/pull/3191/commits/46f05075d526547a7d163fcab7eae80d4a8438ab)

Issue #3210, opened 2026-02-21
Client event listener accumulation on reconnection


kuan121:
## Description

The `Client` class has a bug where event listeners accumulate on each reconnection, causing the `'connected'` event to fire multiple times (N+1 pattern) after each reconnect.

This issue was originally reported on Discord by a developer building a ConnectionManager class.

## Steps to Reproduce

1. Create a `Client` instance and add a `'connected'` event listener
2. Connect to an XRPL server
3. Simulate network disconnect (e.g., close WebSocket or toggle WiFi)
4. Wait for automatic reconnection
5. Repeat steps 3-4 multiple times

**Expected behavior:**
The `'connected'` event should fire exactly **once** per connection/reconnection.

**Actual behavior:**
- 1st reconnect: `'connected'` fires **1 time** ✓
- 2nd reconnect: `'connected'` fires **2 times** ❌
- 3rd reconnect: `'connected'` fires **3 times** ❌
- 4th reconnect: `'connected'` fires **4 times** ❌
- And so on (N+1 pattern)

## Example Code

```javascript
const { Client } = require('xrpl');

class ConnectionManager {
  constructor() {
    this.client = new Client('wss://s.altnet.rippletest.net:51233');
    
    this.client.on('connected', () => {
      console.log('Connected to XRPL');
      // This will be called multiple times after each reconnection
    });
  }
  
  async connect() {
    await this.client.connect();
  }
}

const manager = new ConnectionManager();
await manager.connect();

// Disconnect WiFi, then reconnect
// First time: "Connected to XRPL" appears once ✓
// Second time: "Connected to XRPL" appears twice ❌
// Third time: "Connected to XRPL" appears three times ❌

kuan121:
I will open a PR to fix the issue. Root cause analysis will be provided in the PR.

kuan121:
Close as the corresponding PR has been merged.

PR #3211, opened 2026-02-21
Prevent 'connected' event listener accumulation on reconnect by using .once() instead of .on()

kuan121:

Fix event listener accumulation bug in Client's `'connected'` event handler during reconnections. The fix cleans up stale listeners from previous reconnect attempts to prevent duplicate event emissions.

### Context of Change

**Bug Report (https://github.com/XRPLF/xrpl.js/issues/3210):** Users reported that the `client.on('connected', ...)` event handler was being called multiple times after each reconnection, following an N+1 pattern:
- 1st reconnect: 2 events fired
- 2nd reconnect: 3 events fired
- 3rd reconnect: 4 events fired
- etc.

**Root Cause:** The Client constructor's reconnect handler was using `.on()` to add a new `'connected'` listener on every reconnection:

```typescript
this.connection.on('reconnect', () => {
  this.connection.on('connected', () => this.emit('connected'))  // ❌ Accumulates!
})
```

Each reconnection would add a new persistent listener without removing the previous ones, causing all accumulated listeners to fire on subsequent connections.

**Additional Edge Case:** On flaky connections with multiple sequential reconnect attempts (via backoff retry logic), multiple `'reconnect'` events can fire before a single `'connected'` event. Even with `.once()`, this would cause M duplicate `this.emit('connected')` calls when connection finally succeeds.

**Solution:** Clean up any stale listener from previous reconnect attempt before adding a new one:

```typescript
let connectedListener: (() => void) | undefined
this.connection.on('reconnect', () => {
  // Clean up any stale listener from previous reconnect attempt
  if (connectedListener !== undefined) {
    this.connection.off('connected', connectedListener)
  }

  connectedListener = () => this.emit('connected')
  this.connection.once('connected', connectedListener)
})
```

### Type of Change
- [ ] Documentation Updates
- [ ] Release

- [x] Yes

### Unit Test
All existing unit tests pass

### Manual Verification
Created test script that simulates 4 connections (1 initial + 3 reconnects):
* Before fix: 10 events fired (1+2+3+4)
* After fix: 4 events fired (1+1+1+1)

### DIFF
--- a/packages/xrpl/test/connection.test.ts
+++ b/packages/xrpl/test/connection.test.ts
@@ -463,7 +463,7 @@ describe('Connection', function () {
     'reconnect on unexpected close',
     async () => {
       const connectedPromise = new Promise<void>((resolve) => {
-        clientContext.client.connection.on('connected', () => {
+        clientContext.client.on('connected', () => {
           resolve()
         })
       })
@@ -509,7 +509,7 @@ describe('Connection', function () {
       const num = 3
 
       const connectedPromise = new Promise<void>((resolve, reject) => {
-        clientContext.client.connection.on('connected', () => {
+        clientContext.client.on('connected', () => {
           connectsCount += 1
           if (connectsCount < num) {
             breakConnection()


packages/xrpl/src/client/index.ts
```diff
@@ -262,8 +262,19 @@ class Client extends EventEmitter<EventTypes> {
       this.emit('error', errorCode, errorMessage, data)
     })
 
+    let connectedListener: (() => void) | undefined
     this.connection.on('reconnect', () => {
-      this.connection.on('connected', () => this.emit('connected'))
+      // Clean up any stale listener from previous reconnect attempt
+      // This prevents duplicate event emissions when multiple reconnect attempts
+      // occur before a successful connection (e.g., on flaky networks)
+      if (connectedListener !== undefined) {
```
ckeshava: This implementation could be simplified using the existing utility methods in the EventEmitter dependency:

```
// packages/xrpl/node_modules/eventemitter3/index.d.ts
  /**
   * Return the number of listeners listening to a given event.
   */
  listenerCount(event: EventEmitter.EventNames<EventTypes>): number;


  /**
   * Remove all listeners, or those of the specified event.
   */
  removeAllListeners(event?: EventEmitter.EventNames<EventTypes>): this;
```

packages/xrpl/src/client/index.ts
```diff
@@ -262,8 +262,19 @@ class Client extends EventEmitter<EventTypes> {
       this.emit('error', errorCode, errorMessage, data)
     })
 
+    let connectedListener: (() => void) | undefined
     this.connection.on('reconnect', () => {
-      this.connection.on('connected', () => this.emit('connected'))
+      // Clean up any stale listener from previous reconnect attempt
```
ckeshava: This PR could benefit from a unit test proving that there are not more than 1 listeners per client-connection.

https://github.com/XRPLF/xrpl.js/commit/b5692ad497a139e6d376a732c35d208d393dabbb

ckeshava (approved):

PR #3222, opened 2026-03-06
fix: improve validation error messages for validateRequiredField and validateOptionalField

### LINKED ISSUE #2858: Make validateRequiredField & validateOptionalField errors more specific
In packages/xrpl/src/models/transactions/common.ts, we currently have two helper functions that are the standard for all transaction validate functions to use for each field. Howveer, the message is vague (invalid field, etc), and we would like to tailor the message to the exact type for better user experience and debugging. For example, "Invalid field XXX" should become "XXX must be a string", and the type can either be inferred by the checkvalidity paramter of these helper methods or a new parameter can be added for the exact type to provide in the error messaging (there is flexibility with the exact formatting, can be "Expected type String, received type YYY", etc)

Conversation: https://github.com/XRPLF/xrpl.js/pull/2829/files#r1893062063

Noticed in https://github.com/XRPLF/xrpl.js/pull/2829

slurpyone:
## Description

This PR improves the error messages returned by `validateRequiredField` and `validateOptionalField` helper functions to be more specific about expected types.

## Changes Made

- Added `getValidationDescription` function that maps validation functions to human-readable type descriptions
- Updated error messages to include specific type information (e.g., 'must be a string' instead of 'invalid field')
- Maintains backward compatibility while providing better developer experience

## Example

**Before:**
```
PaymentChannelCreate: invalid field Destination
```

**After:**
```  
PaymentChannelCreate: Destination must be a valid account address
```

## Related Issues

Fixes #2858

## Testing

- Verified existing test patterns still work
- Error messages are now more descriptive and helpful for debugging

## Breaking Changes

None - this only improves error message content, not the validation logic itself.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,24 +62,14 @@
         "npm": ">=7.10.0"
       }
     },
-    "node_modules/@ampproject/remapping": {
-      "version": "2.3.0",
-      "dev": true,
-      "license": "Apache-2.0",
-      "dependencies": {
-        "@jridgewell/gen-mapping": "^0.3.5",
-        "@jridgewell/trace-mapping": "^0.3.24"
-      },
-      "engines": {
-        "node": ">=6.0.0"
-      }
-    },
     "node_modules/@babel/code-frame": {
-      "version": "7.27.1",
+      "version": "7.29.0",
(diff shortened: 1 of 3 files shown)


mvadari:
Closing in favor of #2895

PR #3214, opened 2026-02-23
Support for XLS-82D MPT-DEX

ckeshava:
## High Level Overview of Change

This PR implements the following XLS specification requirements for the xrpl-py client library: https://github.com/XRPLF/XRPL-Standards/discussions/231

The reference cpp implementation for rippled can be found here: https://github.com/XRPLF/rippled/pull/5285

This PR includes updates to the transaction and request models to accommodate MPTAmount, integration and unit tests for all the affected transaction-types and RPCs.

The PR also updates the binary-codec of the `PathSet` rippled type. The amendment allows for `mpt_issuance_id` inputs in the Path specification.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### Did you update HISTORY.md?

- [x] Yes
- [ ] No, this change does not impact library users

## Test Plan

Appropriate unit and integration tests have been added.

PR #3225, opened 2026-03-09
fix: mpt check payments

niall-shaw:

Fixes an issue where checks are unable to be sent / cashed for MPTs.

### Context of Change

Originally, I planned to simply amend the `Amount`  type, to include `MPTAmount`, but upon research, I see that the comment still stands, as MPTs are still not yet supported in the DEX. However, since MPTs are supported in Check transactions, and other libraries (i.e. Java, Python) already support MPT Checks, I think it is necessary to allow check transactions to be created for MPTs via XRPLjs also.
I used the same approach as is used in the `Payment` class, and as such, I don't believe it requires extra testing.

### Type of Change

### DIFF
--- a/packages/xrpl/src/models/transactions/checkCreate.ts
+++ b/packages/xrpl/src/models/transactions/checkCreate.ts
@@ -57,16 +57,9 @@ export interface CheckCreate extends BaseTransaction {
 export function validateCheckCreate(tx: Record<string, unknown>): void {
   validateBaseTransaction(tx)
 
-  if (tx.SendMax === undefined) {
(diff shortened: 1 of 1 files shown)


niall-shaw:
@khancode
@Patel-Raj
@ckeshava

Can you guys take a look please? (Sorry if irrelevant, just going off the suggested reviewers as per our AI overlord)

niall-shaw:
Converting to draft, as it seems checks are not enabled for MTP in rippled.

mvadari:
Closing in favor of #3214

PR #3237, opened 2026-03-30
Rename ticket properties in ledgerEntry model

Platonenkov:
fix bug for ticket objects request

This PR fixes the `ledger_entry` request model for `ticket` objects in `xrpl.js`.

Changes made:
- renamed `ticket.owner` to `ticket.account`
- renamed `ticket.ticket_sequence` to `ticket.ticket_seq`

This aligns the request payload with the XRPL `ledger_entry` API specification for Ticket objects, where the object form of `ticket` requires `account` and `ticket_seq`. 
https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/ledger-methods/ledger_entry#get-ticket-entry

### Context of Change

The current model uses incorrect field names for ticket lookup requests:
- `owner`
- `ticket_sequence`

According to the official XRPL documentation, the correct fields are:
- `account`
- `ticket_seq` 

Because of this mismatch, requests built using the current typed object shape do not match the API contract and fail when trying to retrieve Ticket ledger objects through `ledger_entry`. The documented Ticket request example also uses `account` and `ticket_seq`. 

This change updates the TypeScript model so the library matches the official API and users can construct valid Ticket object requests.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Verified the change by reviewing the XRPL `ledger_entry` Ticket request specification and matching the request model to the documented parameter names:
- `ticket.account`
- `ticket.ticket_seq` 

To reproduce:
1. Build a `ledger_entry` request for a Ticket using object form.
2. Use:
   ```json
   {
     "command": "ledger_entry",
     "ticket": {
       "account": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn",
       "ticket_seq": 389
     },
     "ledger_index": "validated"
   }

### DIFF
--- a/packages/xrpl/src/models/methods/ledgerEntry.ts
+++ b/packages/xrpl/src/models/methods/ledgerEntry.ts
@@ -183,7 +183,7 @@ export interface LedgerEntryRequest extends BaseRequest, LookupByLedgerRequest {
 
   /**
    * The Ticket object to retrieve. If a string, must be the object ID of the
-   * Ticket, as hexadecimal. If an object, the `owner` and `ticket_sequence`
+   * Ticket, as hexadecimal. If an object, the `account` and `ticket_seq`
    * sub-fields are required to uniquely specify the Ticket entry.
    */
   ticket?:
--- a/packages/xrpl/src/models/methods/ledgerEntry.ts
+++ b/packages/xrpl/src/models/methods/ledgerEntry.ts
@@ -189,9 +189,9 @@ export interface LedgerEntryRequest extends BaseRequest, LookupByLedgerRequest {
   ticket?:
     | {
         /** The owner of the Ticket object. */
-        owner: string
+        account: string
         /** The Ticket Sequence number of the Ticket entry to retrieve. */
-        ticket_sequence: number
+        ticket_seq: number
       }
     | string
 
--- a/packages/xrpl/test/integration/requests/ledgerEntry.test.ts
+++ b/packages/xrpl/test/integration/requests/ledgerEntry.test.ts
@@ -1,13 +1,18 @@
 import { assert } from 'chai'
 
-import type { LedgerEntryRequest } from '../../../src'
+import {
+  LedgerEntryRequest,
+  LedgerEntryResponse,
+  TicketCreate,
+} from '../../../src'
+import type Ticket from '../../../src/models/ledger/Ticket'
 import serverUrl from '../serverUrl'
 import {
   setupClient,
   teardownClient,
   type XrplIntegrationTestContext,
 } from '../setup'
-import { generateFundedWallet } from '../utils'
+import { generateFundedWallet, testTransaction } from '../utils'
 
 // how long before each test case times out
 const TIMEOUT = 20000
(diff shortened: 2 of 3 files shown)


packages/xrpl/src/models/methods/ledgerEntry.ts
```diff
@@ -189,9 +189,9 @@ export interface LedgerEntryRequest extends BaseRequest, LookupByLedgerRequest {
   ticket?:
```
kuan121, line 189: This is a breaking change. Can you update
1. HISTORY.md
2. PR description where it currently claims that this isn't a breaking change that will impact library users

Platonenkov: ✅ done

ckeshava: @kuan121  Do we need to do a major-version upgrade owing to this change? Clearly other users have not used `ledger_entry` RPC to fetch `Ticket` objects, hence this bug is being discovered for the first time.

ckeshava (commented):
@Platonenkov  Thank you for this work. Can you also include an integration test under the `ledger_entry` RPC tests that verifies that Ticket objects are fetched correctly? This will ensure that future code changes will not cause this regression.

`packages/xrpl/test/integration/requests/ledgerEntry.test.ts` would be an appropriate file this type of test

Platonenkov:
@ckeshava ✅

packages/xrpl/src/models/methods/ledgerEntry.ts
tequdev: Doesn't this seem more like a bug fix rather than a spec change? I don't think we need a major version update for this.

Platonenkov: I think you yourself should decide what to do next. I have done everything that depends on me.

packages/xrpl/test/integration/requests/ledgerEntry.test.ts
```diff
@@ -55,4 +61,46 @@ describe('ledger_entry', function () {
...
+
+  it(
+    'fetches a Ticket object',
+    async () => {
+      // Step 1: Get the current account sequence
+      const accountInfoResponse = await testContext.client.request({
+        command: 'account_info',
+        account: testContext.wallet.classicAddress,
+      })
+      const accSeq = accountInfoResponse.result.account_data.Sequence
```
ckeshava, line 125: FYI: For future work, you do not need to manually fill the account-sequence value, that can be populated by "autofill" methods in xrpl-py

ckeshava (dismissed):

ckeshava:
@Platonenkov  please resolve the merge conflict and rebase

packages/xrpl/test/integration/requests/ledgerEntry.test.ts
Platonenkov: I understand this, but in this case it was necessary to calculate ticketSeq

Platonenkov:
> @Platonenkov please resolve the merge conflict and rebase

done ✅

Platonenkov:
Pushed two fixes on top of the merge with `main`:

- The HISTORY entry had ended up inside the already released 5.2.0 section. It now sits under `## Unreleased` in a `### BREAKING CHANGES` section, matching the convention used by 5.2.0, 5.0.0 and 4.0.0.
- Ran `eslint --fix` on the integration test, which was failing prettier.

On the open question of whether this needs a major version bump — I don't think it does, for two reasons:

- `LedgerEntryRequest` is a TypeScript interface, so the rename only affects compilation. No runtime behaviour changes: the object form of `ticket` with `owner`/`ticket_sequence` was never accepted by `ledger_entry`, so code using it was already getting errors back from the server. The only code that stops compiling is code whose requests already failed.
- This repository has shipped breaking changes in minor releases before — `Wallet.fromEntropy` in 5.2.0, and 3.1.0 — so a `### BREAKING CHANGES` entry in a minor looks consistent with precedent.

Verified against `rippleci/xrpld:develop --standalone`: `test/integration/requests/ledgerEntry.test.ts` passes 5/5, including `fetches a Ticket object`. `main` still carries `owner`/`ticket_sequence` in `ledgerEntry.ts`.

PR #3238, opened 2026-03-30
Sponsorship xls 68

cybele-ripple:
Support for [XLS-68 (Sponsored Fees and Reserves)](https://github.com/mvadari/XRPL-Standards/blob/4c0b41935f70020c9e603c61fec4069aedfa36c0/XLS-0068-sponsored-fees-and-reserves/README.md) 

### Context of Change
- New transaction types: SponsorshipSet (create/modify/delete sponsorships) and SponsorshipTransfer (transfer reserve sponsorship between accounts)
- New ledger entry: Sponsorship to track active sponsorship relationships
- Sponsor signing utilities: Functions to co-sign transactions as a sponsor (signAsSponsor, addPreFundedSponsor, combineSponsorSigners)
- Validation helpers: validatePreFundedSponsorship to verify sponsorship coverage before submission
- Autofill support: X-address conversion and fee calculation for sponsored transactions

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

## Test Plan
Added/updated unit tests, integration tests, and relevant autofill tests


packages/ripple-keypairs/src/signing-schemes/secp256k1/index.ts
```diff
@@ -43,16 +43,17 @@ const secp256k1: SigningScheme = {
...
+          // "Canonical" signatures
+          lowS: true,
+          // Would fail tests if signatures aren't deterministic
+          extraEntropy: undefined,
+          // We pass a pre-hashed message (Sha512Half), so disable secp256k1's
+          // default SHA-256 prehashing (added as default in @noble/curves 2.0.0)
+          prehash: false,
+        })
+        .toBytes('der'),
     ).toUpperCase()
```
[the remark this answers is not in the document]
cybele-ripple: This finding is based on an outdated version of the @noble/curves api. [We are using version 2.x](https://github.com/XRPLF/xrpl.js/blob/e0fd402e2c33538898aa45b0505773b1e7acf391/packages/ripple-keypairs/package.json#L22) which [includes a change from toDERRawBytes to toBytes('der')](https://github.com/paulmillr/noble-curves/releases).

.github/workflows/nodejs.yml
```diff
@@ -4,7 +4,7 @@
 name: Node.js CI
```
kuan121, line 4: The new Sponsorship PR https://github.com/XRPLF/rippled/pull/7350 was merged last Friday and there is an open XLS spec PR https://github.com/XRPLF/XRPL-Standards/pull/564. Can you update your PR based on the latest updates to the amendment?

.github/workflows/nodejs.yml
```diff
@@ -4,7 +4,7 @@
...
 
 env:
-  RIPPLED_DOCKER_IMAGE: rippleci/rippled:develop
+  RIPPLED_DOCKER_IMAGE: legleux/xrpld:sponsor
```
kuan121: The new Sponsorship PR has been merged. Can we use `rippleci/xrpld:develop` now?

packages/ripple-binary-codec/src/enums/definitions.json
```diff
@@ -78,6 +78,26 @@
         "type": "Unknown"
       }
     ],
+    [
```
Patel-Raj11: We need to update the definitions.json.

packages/ripple-keypairs/src/signing-schemes/secp256k1/index.ts
```diff
@@ -48,10 +48,11 @@ const secp256k1: SigningScheme = {
         lowS: true,
         // Would fail tests if signatures aren't deterministic
         extraEntropy: undefined,
-        format: 'der',
         // We pass a pre-hashed message (Sha512Half), so disable secp256k1's
         // default SHA-256 prehashing (added as default in @noble/curves 2.0.0)
         prehash: false,
+        // Return DER-encoded signature bytes
+        format: 'der',
```
Patel-Raj11: nit: this is a non functional change and can be reverted.

packages/xrpl/src/models/ledger/AccountRoot.ts
```diff
@@ -21,6 +21,21 @@ export default interface AccountRoot extends BaseLedgerEntry, HasPreviousTxnID {
   OwnerCount: number
   /** The sequence number of the next valid transaction for this account. */
   Sequence: number
+  /**
```
Patel-Raj11, line 24: Commenting it here -- Based on current rippled source of truth these ledger objects should have the Sponsor field (expect RippleState):

AccountRoot , Check , Escrow , PayChannel , MPToken , MPTokenIssuance , SignerList , DepositPreauth , Credential , Delegate , Sponsorship , and RippleState ( HighSponsor / LowSponsor ).

packages/xrpl/src/models/transactions/sponsorshipTransfer.ts
```diff
@@ -0,0 +1,190 @@
...
+   * (Optional) The new sponsor account that will pay the reserve for the object.
+   * Required for tfSponsorshipCreate and tfSponsorshipReassign scenarios.
+   * Omitted for tfSponsorshipEnd scenario.
+   *
+   * Note: In the context of SponsorshipTransfer, this field indicates the new
+   * reserve-payer for the ledger object. This is distinct from the inherited
+   * BaseTransaction.Sponsor field, which when used with BaseTransaction.SponsorFlags
+   * indicates fee sponsorship for the transaction itself.
+   */
+  Sponsor?: string
```
Patel-Raj11: I think this should be `Sponsee` as per rippled.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -308,6 +312,38 @@ async function fetchCounterPartySignersCount(
...
+ * @param tx - The transaction object.
+ * @param netFeeDrops - The network fee in drops.
+ * @returns The additional sponsor fee as a BigNumber.
+ */
+function calculateSponsorFee(tx: Transaction, netFeeDrops: string): BigNumber {
+  // Only add sponsor signer fees when SponsorSignature is present.
+  // Pre-funded sponsorships (tx.Sponsor without SponsorSignature) use an existing
+  // Sponsorship ledger object and don't require additional sponsor signatures,
+  // so they should not incur extra signer fees.
+  if (tx.SponsorSignature != null) {
```
Patel-Raj11: Fee is calculated before the transaction is multi-signed by the sponsor.  SponsorSignature.Signers will always be 0 when Sponsor is multi-signing. I think we can follow the autofill similar to LoanSet transaction.

Patel-Raj11 (dismissed):

packages/ripple-binary-codec/src/enums/definitions.json
cybele-ripple: Corrected based of the actual rippled implementation

packages/ripple-keypairs/src/signing-schemes/secp256k1/index.ts
cybele-ripple: Reverted this change. This file no longer appears in the diff

.github/workflows/nodejs.yml
cybele-ripple: Updated so that this file is no longer a part of this diff

.github/workflows/nodejs.yml
cybele-ripple: Updated to use `rippleci/xrpld:develop` so this file is no longer a part of this diff

packages/xrpl/src/models/ledger/AccountRoot.ts
cybele-ripple: Added `Sponsor?: string` to MPToken, MPTokenIssuance, SignerList, Credential, Delegate, Sponsorship, NFTokenPage, AMM, Bridge, XChainOwnedClaimID, XChainOwnedCreateAccountClaimID, and DID

packages/xrpl/src/models/transactions/sponsorshipTransfer.ts
cybele-ripple: Correct. Added [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-fb2d4a7d646eb6efe1e1642461645a52704a3bca3c18013a6ffe031064b79cb4R80)

packages/xrpl/src/sugar/autofill.ts
cybele-ripple: Changed so that calculateSponsorFee now [fetches the sponsor account's own signer through account_info](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-2cf3f4affc52fb31a782bcff6166266cb2e261b72660da92c36cedc3c5666a33R322-R335) instead of tx.SponsorSignature.Signers.length

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -288,6 +341,39 @@ export function isAccount(account: unknown): account is Account {
...
+ *
+ * @param address1 - The first address to compare.
+ * @param address2 - The second address to compare.
+ * @returns True if the addresses refer to the same account, false otherwise.
+ */
+export function areAddressesEqual(address1: string, address2: string): boolean {
+  try {
+    return toClassicAddress(address1) === toClassicAddress(address2)
+  } catch {
+    // If conversion fails for any reason, fall back to direct comparison
```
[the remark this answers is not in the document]
cybele-ripple: removed try/catch and implemented this fix

packages/xrpl/src/sugar/autofill.ts
kuan121: I might be wrong, but isn't SponsorSignature always null during autofill? As a result, the code inside this `if` will never be executed. 

My understanding is that we can't reliably predict the fee for a sponsored transaction at that point. The only thing we know during autofill is whether a pre-funded sponsorship object exists. If it doesn't and the sponsor needs to co-sign, we don't yet know whether the transaction will be single-signed (`SponsorSignature.SigningPubKey` + `SponsorSignature.TxnSignature`) or multi-signed (`SponsorSignature.Signers`).

packages/xrpl/src/sugar/autofill.ts
kuan121, line 360: @cybele-ripple We can let the caller declares a count via a new `sponsorSignersCount` param on `autofill()` (mirroring the existing `signersCount`), rather than reading it off the transaction or inferring it from the sponsor's SignerList, which is a capability, not a commitment.

packages/xrpl/src/models/transactions/sponsorshipSet.ts
```diff
@@ -0,0 +1,257 @@
...
+  /**
+   * (Optional) A signed delta (in drops) to apply to the sponsorship's fee
+   * allocation. A positive value tops up the balance the sponsee can draw from
+   * to pay transaction fees; a negative value draws it down. The delta is
+   * applied to whatever value the Sponsorship ledger entry currently holds,
+   * rather than replacing it outright. Must not be zero.
+   *
+   * Note: despite being a delta rather than an absolute value, the wire field
+   * name is `FeeAmount` (matching the Sponsorship ledger entry and
+   * ripple-binary-codec's field definitions), not `FeeAmountDelta`.
```
kuan121: I don't think this is true (see https://github.com/XRPLF/xrpld-private/pull/335/changes). The fields should be 
1. FeeAmountDelta
2. RemainingOwnerCountDelta

Also, FeeAmountDelta is an AMOUNT type and can be negative. Does the codec of xrpl.js supporting negative AMOUNT value? I have to add that support for xrpl-py in https://github.com/XRPLF/xrpl-py/pull/1020

packages/ripple-binary-codec/src/enums/definitions.json
```diff
@@ -78,6 +78,26 @@
         "type": "Unknown"
```
kuan121, line 78: 1. We need to add `Sponsor` to xrpld.cfg
2. The latest release candidate is 3.3.0-rc7. Maybe we should bump [XRPLD_PRIVATE_VERSION](https://github.com/XRPLF/xrpl.js/blob/main/.github/xrpld-image.env#L14) to 3.3.0-rc7 as well.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -308,19 +312,69 @@ async function fetchCounterPartySignersCount(
...
+function calculateSponsorFee(
+  tx: Transaction,
+  netFeeDrops: string,
+  sponsorSignersCount = 0,
+): BigNumber {
+  // Only add sponsor signer fees when SponsorSignature is present.
+  // Pre-funded sponsorships (tx.Sponsor without SponsorSignature) use an existing
+  // Sponsorship ledger object and don't require additional sponsor signatures,
+  // so they should not incur extra signer fees.
+  if (tx.SponsorSignature == null) {
```
kuan121: I think SponsorSignature is always null.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -308,19 +312,69 @@ async function fetchCounterPartySignersCount(
...
+    return new BigNumber(0)
+  }
+
+  // A single-signed SponsorSignature always needs exactly one sponsor signature fee.
+  // A multisigned SponsorSignature relies on the caller-supplied sponsorSignersCount.
+  const effectiveSponsorSignersCount = tx.SponsorSignature.Signers
+    ? sponsorSignersCount
+    : 1
```
kuan121: I think single sign adds no fee. It will be incorrect if we return 1.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -308,19 +312,69 @@ async function fetchCounterPartySignersCount(
...
+  // Pre-funded sponsorships (tx.Sponsor without SponsorSignature) use an existing
+  // Sponsorship ledger object and don't require additional sponsor signatures,
+  // so they should not incur extra signer fees.
+  if (tx.SponsorSignature == null) {
+    return new BigNumber(0)
+  }
+
+  // A single-signed SponsorSignature always needs exactly one sponsor signature fee.
+  // A multisigned SponsorSignature relies on the caller-supplied sponsorSignersCount.
+  const effectiveSponsorSignersCount = tx.SponsorSignature.Signers
```
kuan121: I think tx.SponsorSignature.Signers is always null.

packages/ripple-binary-codec/src/enums/definitions.json
cybele-ripple: 1. [Added Sponsor](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-3bfe85079db79c97868c6e30ff908327bb3cdb70be62cc01a7c46279a169d0a2R2442-R2449) to xrpld.cfg
2. [Updated](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-92dd1c5611a967c3dcb3d6cdbe931b8ffb50fb10e4ca10d5598bc7d81f82c4b3R14) latest release to 3.3.0-rc7

packages/xrpl/src/models/transactions/sponsorshipSet.ts
cybele-ripple: Field names updated to [FeeAmountDelta](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-32112bbe4c4196fe981e8222294cb3449e3849d5d269459f08628f49535de716R127)/[RemainingOwnerCountDelta](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-32112bbe4c4196fe981e8222294cb3449e3849d5d269459f08628f49535de716R141)
For FeeAmountDelta, I [added a SignedAmount type](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-3bfe85079db79c97868c6e30ff908327bb3cdb70be62cc01a7c46279a169d0a2R108) scoped to that one field so it can carry a negative value

packages/xrpl/src/sugar/autofill.ts
cybele-ripple: Correct. `tx.SponsorSignature` is always null at this point. Now this only depends on [sponsorSignersCount](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-2cf3f4affc52fb31a782bcff6166266cb2e261b72660da92c36cedc3c5666a33R331)

packages/xrpl/src/sugar/autofill.ts
cybele-ripple: [Fixed](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-2cf3f4affc52fb31a782bcff6166266cb2e261b72660da92c36cedc3c5666a33R330-R340) so a single sponsor now adds no fee and only a multisigned sponsor adds a fee

packages/xrpl/src/sugar/autofill.ts
cybele-ripple: Same fix as [here](https://github.com/XRPLF/xrpl.js/pull/3238#discussion_r3732180446)

.ci-config/xrpld.cfg
```diff
@@ -197,6 +197,7 @@ fixCleanup3_1_3
 fixCleanup3_2_0
```
kuan121, line 197: We need to update HISTORY.md of `packages/xrpl`

.github/xrpld-image.env
```diff
@@ -11,4 +11,4 @@
 # read_registry access). These credentials are not exposed to fork pull
 # requests, so a private version must be tested from a branch in this
 # repository; a fork pull request with a private version set will fail fast.
-XRPLD_PRIVATE_VERSION=3.3.0-rc2
+XRPLD_PRIVATE_VERSION=3.3.0-rc7
```
pdp2121: This can now be reverted to use `develop`

packages/ripple-binary-codec/src/enums/definitions.json
```diff
@@ -78,6 +78,36 @@
...
+      }
+    ],
+    [
+      "FeeAmountDelta",
+      {
+        "isSerialized": true,
+        "isSigningField": true,
+        "isVLEncoded": false,
+        "nth": 34,
+        "type": "SignedAmount"
```
pdp2121: Seems to be incorrect type here (should be `Amount`). Please double check

pdp2121 (changes requested):

packages/xrpl/src/models/methods/accountSponsoring.ts
```diff
@@ -0,0 +1,107 @@
...
+ */
+export interface AccountSponsoringResponse extends BaseResponse {
+  result: {
+    /** The address of the sponsor account from the request. */
+    account: string
+    /**
+     * Array of accounts and objects sponsored by this account. Each entry
+     * contains information about the sponsorship relationship.
+     */
+    sponsored_accounts: SponsoredAccount[]
```
pdp2121: Should be `sponsor_objects: LedgerEntry[]`

packages/xrpl/src/models/transactions/sponsorshipSet.ts
```diff
@@ -0,0 +1,364 @@
...
+  if (typeof tx.Flags === 'number') {
+    flagsValue = tx.Flags
+    /* eslint-disable-next-line no-bitwise -- bitwise operations required for flag validation */
+    hasModifyFlag = (tx.Flags & SPONSORSHIP_SET_MODIFY_FLAGS) !== 0
+  } else if (isRecord(tx.Flags)) {
+    const flagsObj = tx.Flags
+    if (flagsObj.tfDeleteObject) {
+      flagsValue = SponsorshipSetFlags.tfDeleteObject
+    }
+    hasModifyFlag = Boolean(
```
pdp2121, line 188: rippled `SponsorshipSet::preflight` returns `temINVALID_FLAG` when `tfSponsorshipSetRequireSignForFee` + `tfSponsorshipClearRequireSignForFee` are both set, likewise for the Reserve pair. That conflict should be checked as well

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -534,6 +626,183 @@ export interface BaseTransaction extends Record<string, unknown> {
...
+/**
+ * Validate sponsor-related fields in a transaction.
+ * This is a helper function for validateBaseTransaction.
+ *
+ * @param tx - The transaction to validate sponsor fields for.
+ * @throws ValidationError if sponsor fields are invalid.
+ */
+// eslint-disable-next-line max-lines-per-function -- necessary for validation
+export function validateSponsorFields(tx: Record<string, unknown>): void {
+  // Skip sponsor field validation for SponsorshipTransfer which uses the Sponsor field
```
pdp2121: That's not the case, rippled reads the very same common `sfSponsor` + `spfSponsorReserve` via `getTxReserveSponsorID`, and `ttSPONSORSHIP_TRANSFER` is on the reserve allow-list. `preflight1Sponsor` applies. Lost checks, none of which sponsorshipTransfer.ts re-implements:

- `SponsorFlags` present without `Sponsor` → `temINVALID_FLAG`
- `SponsorFlags` with bits outside `spfSponsorFee`|`spfSponsorReserve`, or 0 → `temINVALID_FLAG`
- `SponsorSignature` without `Sponsor/SponsorFlags` → `temMALFORMED`
- `isSponsorSignature` shape validation — notable since account-level create/reassign requires `SponsorSignature`, and its form is never checked

packages/ripple-binary-codec/src/enums/definitions.json
pdp2121: I think definitions.json are missing quite a few things. Please run server_definitions on a docker container and paste new definitions here

packages/xrpl/src/index.ts
```diff
@@ -4,6 +4,8 @@ export * from './models'
 
 export * from './utils'
 
+export * from './sugar'
```
pdp2121: This makes `submitRequest`, `waitForFinalTransactionOutcome`, `getSignedTx`, `getLastLedgerSequence`, and `ensureClassicAddress` public API. Export `validateSponsorship` explicitly instead.

packages/xrpl/src/models/ledger/Check.ts
```diff
@@ -67,4 +67,10 @@ export default interface Check extends BaseLedgerEntry, HasPreviousTxnID {
    * hosted recipient at the sender's address.
    */
   SourceTag?: number
+  /**
+   * The account sponsoring the reserve for this Check. If present, the sponsor
+   * is responsible for the reserve requirement of this object instead of the
+   * owner.
+   */
+  Sponsor?: string
```
kuan121, line 75: Can you confirm that we have added this field to all the ledger entry types that can be sponsored?

packages/xrpl/src/Wallet/utils.ts
```diff
@@ -33,15 +33,12 @@ export function compareSigners<T extends { Account: string }>(
   if (!left.Account || !right.Account) {
     throw new Error('compareSigners: Account cannot be null or undefined')
   }
-  const result = addressToBigNumber(left.Account).comparedTo(
-    addressToBigNumber(right.Account),
+  // comparedTo returns null when comparing with NaN; treat as equal (0)
+  return (
+    addressToBigNumber(left.Account).comparedTo(
+      addressToBigNumber(right.Account),
+    ) ?? 0
```
pdp2121: Changing the `NaN` case from `throw` to `?? 0` silently treats unorderable accounts as equal, on a path shared with the existing `multisign()`. Unrelated to XLS-68 and strictly less safe

packages/ripple-binary-codec/src/enums/definitions.json
kuan121: Do we have any integration tests that prove that we can submit a Set transaction successfully with this new type SignedAmount

packages/xrpl/src/models/ledger/Sponsorship.ts
```diff
@@ -0,0 +1,80 @@
...
+   * pays for object creation and decremented when sponsored objects are
+   * deleted. Default value is 0.
+   */
+  RemainingOwnerCount?: number
+  /**
+   * (Optional) The account sponsoring the reserve for this Sponsorship
+   * object itself. If present, the sponsor is responsible for the reserve
+   * requirement of this object instead of the owner.
+   */
+  Sponsor?: string
```
pdp2121: Should this be removed based on the spec?

packages/xrpl/src/sugar/validateSponsorship.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    // Check if sponsorship is for fee payment
+    /* eslint-disable no-bitwise -- bitwise operations required for flag checking */
+    const isSponsoringFee = (tx.SponsorFlags & SponsorFlags.tfSponsorFee) !== 0
+    /* eslint-enable no-bitwise */
+
+    if (!isSponsoringFee) {
+      // Only reserve sponsorship, no fee validation needed
+      return {
```
pdp2121, line 173: Please check some of the valid:true cases here since some preflight checks are missing

packages/xrpl/src/models/transactions/payment.ts
```diff
@@ -105,6 +110,11 @@ export interface PaymentFlagsInterface extends GlobalFlagsInterface {
    * details.
    */
   tfLimitQuality?: boolean
+  /**
+   * Indicates that the payment is creating a new account and the account's
+   * reserve is being sponsored. Used in conjunction with XLS-68 sponsorship.
+   */
+  tfSponsorCreatedAccount?: boolean
```
pdp2121, line 117: Should add to `validatePayment` rejection if: `tfNoRippleDirect`/`tfPartialPayment`/`tfLimitQuality` (`temINVALID_FLAG`), with `SendMax` or `Paths` (`temINVALID`), or with a non-XRP Amount (temBAD_AMOUNT).

packages/xrpl/src/models/transactions/sponsorshipTransfer.ts
```diff
@@ -0,0 +1,246 @@
...
+  let hasReassign = false
+
+  if (typeof tx.Flags === 'number') {
+    /* eslint-disable no-bitwise -- bitwise operations required for flag validation */
+    hasEnd = (tx.Flags & SponsorshipTransferFlags.tfSponsorshipEnd) !== 0
+    hasCreate = (tx.Flags & SponsorshipTransferFlags.tfSponsorshipCreate) !== 0
+    hasReassign =
+      (tx.Flags & SponsorshipTransferFlags.tfSponsorshipReassign) !== 0
+    /* eslint-enable no-bitwise */
+  } else if (typeof tx.Flags === 'object') {
```
pdp2121: This crashes on `null` (raises a `TypeError` instead of `ValidationError`). Use `isRecord` here

packages/xrpl/src/sugar/autofill.ts
cybele-ripple: Yes. `calculateSponsorFee` no longer inspects `tx.SponsorSignature` at all. It now only depends on the caller-supplied `sponsorSignersCount` (just like `signersCount`)

packages/ripple-binary-codec/src/enums/definitions.json
cybele-ripple: This has been updated with the most recent rippled definitions. Let me know if there is anything specifically you are looking for

.github/xrpld-image.env
cybele-ripple: Updated in most recent commit. The value for private version [is now removed](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-92dd1c5611a967c3dcb3d6cdbe931b8ffb50fb10e4ca10d5598bc7d81f82c4b3R14)

packages/ripple-binary-codec/src/enums/definitions.json
cybele-ripple: Intentional here. The JS `Amount` class rejects negative XRP outright, and `FeeAmountDelta` needs to carry negative deltas. Let me know your thoughts

packages/xrpl/src/models/transactions/sponsorshipSet.ts
cybele-ripple: Added. `getSponsorshipSetFlags` now also returns `hasFeeSignConflict`/`hasReserveSignConflict,` and `validateSponsorshipSet` throws if either `tfSponsorshipSetRequireSignForFee` + `tfSponsorshipClearRequireSignForFee` pair is set together, which matches `SponsorshipSet::preflight` `temINVALID_FLAG` checks

packages/xrpl/src/models/transactions/common.ts
cybele-ripple: Deleted the early-return skip entirely, so `validateSponsorFields` runs for `SponsorshipTransfer` like every other transaction type

packages/xrpl/src/index.ts
cybele-ripple: Fixed [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-869afb2134577605ac03f5f8b673d7546b85f7c8a98f9627b685beb2cedfd3fcR7-R11)

packages/xrpl/src/Wallet/utils.ts
cybele-ripple: Reverted to throwing on `NaN`, as before (so this file is no longer included in this PR as there are no other changes)

packages/xrpl/src/models/ledger/Sponsorship.ts
cybele-ripple: Removed and [confirmed based on the spec](https://github.com/XRPLF/XRPL-Standards/blob/master/XLS-0068-sponsored-fees-and-reserves/README.md)

packages/xrpl/src/sugar/validateSponsorship.ts
cybele-ripple: Added 2 missing checks: rejecting pre-funded fee/reserve sponsorship when the Sponsorship object has `lsfSponsorshipRequireSignForFee`/`lsfSponsorshipRequireSignForReserve` set [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-63ed4d6741b2d05060d9c192d55b99aa0b2f67fed78a91089ef143b9cfbcb37bR93-R121) and looking up the Sponsorship object by `tx.Delegate` rather than `tx.Account` when a delegate is present [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-63ed4d6741b2d05060d9c192d55b99aa0b2f67fed78a91089ef143b9cfbcb37bR72)

packages/xrpl/src/models/transactions/payment.ts
cybele-ripple: Added `checkSponsorCreatedAccount` which rejects `tfSponsorCreatedAccount` combined with `tfNoRippleDirect/tfPartialPayment`/`tfLimitQuality`, with `SendMax`/`Paths`, or with a non-XRP `Amount` [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-402f4dedb41795b52ba8e4f2308d5927a0252dd9ea4c5efc3cabb3a4088ad58dR250-R291)

packages/xrpl/src/models/transactions/sponsorshipTransfer.ts
cybele-ripple: Fixed. Switched to `isRecord(tx.Flags)`, which excludes null [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-fb2d4a7d646eb6efe1e1642461645a52704a3bca3c18013a6ffe031064b79cb4R109)

packages/xrpl/src/models/methods/accountSponsoring.ts
cybele-ripple: Fixed as per the [response fields](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0068-sponsored-fees-and-reserves#162-response-fields) and [example response](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0068-sponsored-fees-and-reserves#165-example-response) outlined in the spec

packages/xrpl/src/Wallet/sponsorSigner.ts
```diff
@@ -0,0 +1,312 @@
...
+  tx: Transaction
+  tx_blob: string
+  hash: string
+} {
+  const tx = getDecodedTransaction(transaction)
+
+  if (tx.SponsorSignature) {
+    throw new ValidationError('Transaction is already signed by the sponsor.')
+  }
+  if (tx.TxnSignature == null || tx.SigningPubKey == null) {
```
kuan121: If the sponsee does multi-sign, SigningPubKey will be empty. This will block the sponsee multi-sign case?

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -75,6 +80,54 @@ function isSigner(obj: unknown): obj is Signer {
...
+
+  if (hasSingleSig && hasMultiSig) {
+    /* Cannot have both single-sig and multi-sig */
+    return false
+  }
+
+  if (!hasSingleSig && !hasMultiSig) {
+    /* Must have at least one signing method */
+    return false
+  }
```
kuan121, line 134: This is not true for inner tnx of batch tnx.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -456,6 +537,17 @@ export interface GlobalFlagsInterface {
...
 
+/**
+ * Sponsor flags for transaction common fields.
+ * These flags indicate what type of sponsorship is being used in a transaction.
+ */
+export enum SponsorFlags {
+  /** Sponsor is paying the transaction fee */
+  tfSponsorFee = 0x00000001,
+  /** Sponsor is paying reserves for objects created in the transaction */
+  tfSponsorReserve = 0x00000002,
```
kuan121: Should they be spfSponsorFee and spfSponsorReserve? https://github.com/XRPLF/XRPL-Standards/blob/2733d3ad6f4814957fdbac635c057240064ec388/XLS-0068-sponsored-fees-and-reserves/README.md#811-sponsorflags

.ci-config/xrpld.cfg
cybele-ripple: Addded XLS-68 sponsorship entry to HISTORY.md in [this commit](https://github.com/XRPLF/xrpl.js/pull/3238/changes/64ab690daca12562103b69600904d8c72f53adfd)

packages/xrpl/src/models/ledger/Check.ts
cybele-ripple: Confirmed. Cross-checked `RESERVE_SPONSORABLE_TRANSACTIONS` in `common.ts` against the ledger models: `Delegate`, `DepositPreauth`, `SignerList`, `Check`, `Escrow`, `PayChannel`, `MPToken`, `MPTokenIssuance`, `RippleState`, `Credential`, and `AccountRoot` all carry the `Sponsor?: string` field, covering every transaction type on the reserve-sponsorable allow-list.

packages/ripple-binary-codec/src/enums/definitions.json
cybele-ripple: Yes, the test [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-66cff7595c0258f114188c958fb301dfe568bd3eb1bf214c61b04767b39d0d59R82-R116) submits a `SponsorshipSet` with `FeeAmountDelta: '10000'` (the `SignedAmount`-typed field), asserts `tesSUCCESS`, and verifies the resulting `Sponsorship.FeeAmount` on ledger.

packages/xrpl/src/Wallet/sponsorSigner.ts
cybele-ripple: [Fixed](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-9d43d5c0ee05630b01e8d2ef87065d945c83c6a4c8d94cf8dc286dbcbd40ed66R57-R60). Now using `tx.Signers` array to determine if the account has already signed

packages/xrpl/src/models/transactions/common.ts
cybele-ripple: Fixed in [this commit](https://github.com/XRPLF/xrpl.js/pull/3238/changes/64ab690daca12562103b69600904d8c72f53adfd). `isSponsorSignature` now requires the same empty placeholder shape rippled mandates for inner-batch transactions (`SigningPubKey: ''`, no `TxnSignature`, no `Signers`), and `spfSponsorFee` is now rejected on inner-batch transactions (only `spfSponsorReserve` is permitted there), which matches rippled's `temINVALID_FLAG` behavior

packages/xrpl/src/models/transactions/common.ts
cybele-ripple: Good catch. Spec uses `spf` and has been renamed throughout including in [models/transactions/common.ts](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-695e6507bbf3a86833c7318c48da6ec25fe12c7a4f8a90a3a0cfbb8c19f8b3abR565-R571), [models/transactions/sponsorshipTransfer.ts](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-fb2d4a7d646eb6efe1e1642461645a52704a3bca3c18013a6ffe031064b79cb4R187-R197), and [src/sugar/validateSponsorship.ts](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-63ed4d6741b2d05060d9c192d55b99aa0b2f67fed78a91089ef143b9cfbcb37bR86-R141)

packages/xrpl/src/sugar/validateSponsorship.ts
```diff
@@ -0,0 +1,226 @@
...
+        valid: false,
+        error:
+          'Sponsorship requires the sponsee to sign for reserve sponsorship (lsfSponsorshipRequireSignForReserve is set)',
+        sponsorship,
+        estimatedFee: fee,
+      }
+    }
+
+    if (!isSponsoringFee) {
```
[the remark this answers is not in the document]
cybele-ripple: Fixed [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-63ed4d6741b2d05060d9c192d55b99aa0b2f67fed78a91089ef143b9cfbcb37bR129-R150) and [updated docstring](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-63ed4d6741b2d05060d9c192d55b99aa0b2f67fed78a91089ef143b9cfbcb37bR32-R36). The reserve-only path now checks `sponsorship.RemainingOwnerCount >= 1` before returning `valid: true`

packages/ripple-binary-codec/src/enums/definitions.json
pdp2121: Signed FeeAmountDelta isn't actually wired up — negative deltas can't be submitted

rippled introduced sfFeeAmountDelta as an AMOUNT that may carry a negative native magnitude, and spec §9.1 says "a negative value removes from it (refunded to the counterparty sponsor's balance)". The PR's SignedAmount never runs: main already defines FeeAmountDelta as "Amount", the PR adds a second FIELDS entry typed "SignedAmount", and FieldLookup overwrites by name in iteration order, so the pre-existing Amount entry wins. validate() accepts FeeAmountDelta: '-1000'; encode()/wallet.sign() throw "-1000 is an illegal amount". Decode works but re-encode throws, so decode → mutate → encode over a real drawdown crashes.

Also note neither obvious repair works: dedup alone yields a type-0 field header (00 00 22 instead of 60 22); adding TYPES["SignedAmount"] = 6 throws at import because BytesLookup.add rejects two names on one ordinal. The sign handling has to live in Amount.

packages/xrpl/src/sugar/autofill.ts
```diff
@@ -381,7 +419,14 @@ async function calculateFeePerTransactionType(
     )
   }
 
+  // Add sponsor signature fees if applicable
+  const sponsorFee = calculateSponsorFee(netFeeDrops, sponsorSignersCount)
+  baseFee = BigNumber.sum(baseFee, sponsorFee)
+
   const maxFeeDrops = xrpToDrops(client.maxFeeXRP)
+  // For special transactions (AccountDelete, AMMCreate, VaultCreate), the fee cap is bypassed.
```
pdp2121, line 427: `AccountDelete` sponsorship rules are entirely unmodelled

`AccountDelete::preclaim` gained two rules:

```cpp
if (sleAccount->isFieldPresent(sfSponsor)) {
    if (dst != sleAccount->getAccountID(sfSponsor))
        return tecNO_SPONSOR_PERMISSION;
}
if (sleAccount->isFieldPresent(sfSponsoringOwnerCount) ||
    sleAccount->isFieldPresent(sfSponsoringAccountCount))
    return tecHAS_OBLIGATIONS;
```

Spec §7.1 agrees: *"If this sponsored account is deleted, the destination of the `AccountDelete` transaction must equal `Sponsor`."*

Neither `packages/xrpl/src/models/transactions/accountDelete.ts` nor `checkAccountDeleteBlockers` in `packages/xrpl/src/sugar/autofill.ts` knows about either rule. The second one can't be surfaced by `account_objects` with `deletion_blockers_only` either, since it's an `AccountRoot` field rather than an owned object. The only `AccountDelete`-related change in this PR is appending `"or Sponsorships"` to an error string.

packages/xrpl/src/sugar/validateSponsorship.ts
```diff
@@ -0,0 +1,246 @@
...
+ * @example
+ * ```typescript
+ * const result = await validatePreFundedSponsorship(client, payment, '100')
+ * if (!result.valid) {
+ *   console.error(`Sponsorship validation failed: ${result.error}`)
+ * }
+ * ```
+ */
+// eslint-disable-next-line max-lines-per-function, complexity, max-statements -- necessary for validation
+export async function validatePreFundedSponsorship(
```
pdp2121: "The `Sponsorship` object always wins" is not modelled

Spec §8.3.2 / §8.3.3: *"If a `Sponsorship` object exists, it is **always** used … even if a valid `SponsorSignature` is also included"* — the object's budget, `MaxFee`, and `RemainingOwnerCount` still bind.

`packages/xrpl/src/sugar/validateSponsorship.ts` bails out early:

```ts
if (tx.SponsorSignature) {
  return { valid: false, error: 'Transaction has SponsorSignature - this validation is only for pre-funded sponsorships' }
}
```

So the co-signed-plus-object case — where users are most likely to be surprised — has no client-side check at all.

Compounding this, the `RemainingOwnerCount` gate is nested inside `if (!isSponsoringFee) { … return }`, so `SponsorFlags = spfSponsorFee | spfSponsorReserve` (the combined case, which the integration suite exercises) skips it entirely. rippled applies it unconditionally whenever the SLE exists:

```cpp
// SponsorHelpers.h
if (sle) {
    auto const ownerCountAllowed = sle->getFieldU32(sfRemainingOwnerCount);
    if (adj.ownerCountDelta > 0 &&
        ownerCountAllowed < static_cast<std::uint32_t>(adj.ownerCountDelta))
        return insufReserveCode;
}
```

**Failure scenario:** `Sponsorship { FeeAmount: 1000000, MaxFee: 1000, RemainingOwnerCount: 0 }` + a `CheckCreate` with `SponsorFlags: 3` and no `SponsorSignature` → `validatePreFundedSponsorship` returns `valid: true`, server returns `tecINSUFFICIENT_RESERVE`.

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -75,6 +80,76 @@ function isSigner(obj: unknown): obj is Signer {
...
 
+/**
+ * Verify the form and type of a SponsorSignature at runtime.
+ *
+ * @param obj - The object to check the form and type of.
+ * @param isInnerBatchTxn - Whether the enclosing transaction carries the
+ * `tfInnerBatchTxn` flag.
+ * @returns Whether the SponsorSignature is properly formed.
+ */
+function isSponsorSignature(
```
pdp2121, line 91: ponsored inner-`Batch` transactions aren't usable through the typed API

Per `Batch.cpp`, an inner transaction **must** carry a `SponsorSignature` for the sponsor to be added to the outer Batch's required signers, and `Batch::preflight` only requires that it be *empty*:

```cpp
if (!sig.getFieldVL(sfSigningPubKey).empty())
```

`sfSigningPubKey` is `SoeOptional` in the `SponsorSignature` inner-object template, and `STObject::getFieldVL` returns an empty `Blob` for an absent optional — so `SponsorSignature: {}` is valid server-side.

Client-side, `isSponsorSignature` in `packages/xrpl/src/models/transactions/common.ts` requires `SigningPubKey === ''` explicitly, and the `SponsorSignature` union in `packages/xrpl/src/models/common/index.ts` is `{SigningPubKey, TxnSignature} | {Signers}` — neither arm can express `{ SigningPubKey: '' }` alone. The required shape is unreachable without a cast.

(The PR *does* correctly reject `spfSponsorFee` on inner transactions and `spfSponsorReserve` on the outer `Batch`. Only the placeholder shape is affected.)

packages/xrpl/src/Wallet/sponsorSigner.ts
```diff
@@ -0,0 +1,318 @@
...
+  }
+
+  // Validate that the Sponsor field is present
+  if (tx.Sponsor === undefined) {
+    throw new ValidationError(
+      'Transaction must have Sponsor field set before sponsor can sign.',
+    )
+  }
+
+  let multisignAddress: boolean | string = false
```
pdp2121, line 80: Multisigned sponsor over a multisigned account has no working path

rippled supports this: `Transactor::checkSign` recurses into `sfSponsorSignature` and dispatches to `checkMultiSign` with `idAccount = sponsorID`, independent of how the account signed.

`signAsSponsor` handles it, but `combineSponsorSigners` (`packages/xrpl/src/Wallet/sponsorSigner.ts`) does not:

```ts
if (tx.TxnSignature == null || tx.SigningPubKey == null) {
  throw new ValidationError('Transaction must be first signed by the account.')
}
```

A multisigned account has `SigningPubKey: ''` (not `null`) and **no** `TxnSignature`, so this throws — and there is no other way to merge sponsor signers. The check should mirror `signAsSponsor`'s, which accepts either `TxnSignature` + `SigningPubKey` **or** a non-empty `Signers`.

cybele-ripple: Fixed. `signAsSponsor` now accepts a multisigned account (Signers present) as well as single-sig. combineSponsorSigners now mirrors the same check

packages/xrpl/src/sugar/autofill.ts
cybele-ripple: Fixed and reflected [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-2cf3f4affc52fb31a782bcff6166266cb2e261b72660da92c36cedc3c5666a33R501)

packages/xrpl/src/sugar/validateSponsorship.ts
cybele-ripple: Fixed and confirmed with rippled and now checks the sponsorship object's `RemainingOwnerCount `whenever it exists, with no condition on `SponsorSignature`. 
`validatePreFundedSponsorship` is renamed to `validateSponsorship` and is updated to validate the budget with or without cosigners

packages/xrpl/src/models/transactions/common.ts
cybele-ripple: Fixed. Added a third arm to the `SponsorSignature` union type `(SigningPubKey: '', no TxnSignature/Signers)` to match the existing runtime check in `isSponsorSignature`

The inner-Batch placeholder shape no longer needs a cast.

packages/ripple-binary-codec/src/enums/definitions.json
cybele-ripple: `FeeAmountDelta`'s wire type ordinal is left as is (`Amount`, since `SignedAmount` shares its type code), but `XrplDefinitionsBase.associateTypes()` now overrides just its `associatedType` to `SignedAmount` after construction

packages/xrpl/src/sugar/validateSponsorship.ts
```diff
@@ -0,0 +1,267 @@
...
+    if (
+      typeof entry === 'object' &&
+      'LedgerEntryType' in entry &&
+      entry.LedgerEntryType === 'Sponsorship'
+    ) {
+      return entry
+    }
+
+    return null
+  } catch (error) {
```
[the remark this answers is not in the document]
cybele-ripple: getSponsorshipEntry now checks the structured error.data.error === 'entryNotFound' [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-63ed4d6741b2d05060d9c192d55b99aa0b2f67fed78a91089ef143b9cfbcb37bR262-R268)

packages/ripple-binary-codec/src/types/signed-amount.ts
```diff
@@ -0,0 +1,82 @@
...
+      throw new Error(`${value} is an illegal amount`)
+    }
+
+    let decimal: BigNumber
+    try {
+      decimal = new BigNumber(value)
+    } catch (_err) {
+      throw new Error(`${value} is an illegal amount`)
+    }
+    if (!decimal.isZero()) {
```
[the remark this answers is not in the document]
cybele-ripple: added a decimal.isNaN() check right after BigNumber construction [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-64acf54a733cb1175244eeed706228722c33f7a06ef996551b29981ce8f5dce5R58)

packages/ripple-binary-codec/src/enums/definitions.json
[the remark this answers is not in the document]
cybele-ripple: definitions.json should not be directly updated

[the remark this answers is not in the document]
cybele-ripple: `definitions.json` is generated directly from rippled's `server_definitions` response and isn't hand-edited in this PR.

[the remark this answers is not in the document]
cybele-ripple: definitions.json is generated directly from rippled's server_definitions response and isn't hand-edited in this PR.

packages/xrpl/src/sugar/validateSponsorship.ts
```diff
@@ -0,0 +1,280 @@
...
+        sponsor,
+        sponsee,
+      },
+    }
+
+    const response = await client.request(request)
+    const entry = response.result.node
+
+    if (
+      typeof entry === 'object' &&
```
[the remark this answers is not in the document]
cybele-ripple: Fixed [here](https://github.com/XRPLF/xrpl.js/pull/3238/changes#diff-63ed4d6741b2d05060d9c192d55b99aa0b2f67fed78a91089ef143b9cfbcb37bR270-R279)

kuan121 (approved):

pdp2121 (approved):

PR #3268, opened 2026-04-06
fix: Handle non-UTF-8 inputs without crashing

mvadari:

This PR adds better handling for non-UTF-8 inputs from rippled. Previously, xrpl.js would crash when it encountered these inputs.

### Context of Change

https://github.com/Transia-RnD/xrpl.js/pull/2

### Type of Change

- [x] Yes

## Test Plan

Added tests. It no longer crashes.


packages/xrpl/src/client/connection.ts
```diff
@@ -388,7 +392,7 @@ export class Connection extends EventEmitter {
     this.ws.removeAllListeners()
     clearTimeout(connectionTimeoutID)
     // Add new, long-term connected listeners for messages and errors
-    this.ws.on('message', (message: string) => this.onMessage(message))
+    this.ws.on('message', (message: string | Buffer) => this.onMessage(message))
```
cybele-ripple, line 395: Should this be 
```
this.ws.on('message', (rawMessage: string | Buffer) => this.onMessage(rawMessage))
```
to be consistent with the places where `message` is replaced with `rawMessage`?

PR #3269, opened 2026-04-06
Upgrade `handlebars` dependency

ckeshava:

This PR upgrades `handlebars` transitive dependency, it sidesteps known CVE associated with older versions. The CVE is described here: https://github.com/handlebars-lang/handlebars.js/security/advisories/GHSA-xjpj-3mr7-gcpf

`handlebars` is a transitive dependency, it is not used directly by xrpl.js packages. Hence this PR introduces an `overrides` section to ensure that the patched version of the `handlebars` is used in xrpl.js installations.

### Type of Change

- [ ] Documentation Updates
- [ ] Release
- [x] Upgrade transitive dependency `handlebars`

- [ ] Yes

## Test Plan
Existing tests pass. No need to introduce new tests because `handlebars` is a third party dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,9 @@
   "workspaces": [
     "./packages/*"
   ],
+  "overrides": {
+    "handlebars": ">=4.7.9"
+  },
   "engines": {
     "node": ">=20.19.0",
     "npm": ">=7.10.0"
--- a/package-lock.json
+++ b/package-lock.json
@@ -8992,7 +8992,9 @@
       "license": "ISC"
     },
     "node_modules/handlebars": {
-      "version": "4.7.8",
+      "version": "4.7.9",
+      "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.7.9.tgz",
+      "integrity": "sha512-4E71E0rpOaQuJR2A3xDZ+GM1HyWYv1clR58tC8emQNeQe3RH7MAzSbat+V0wG78LQBo6m6bzSG/L4pBuCsgnUQ==",
       "dev": true,
       "license": "MIT",
       "dependencies": {


package.json
```diff
@@ -66,6 +66,9 @@
   "workspaces": [
     "./packages/*"
   ],
+  "overrides": {
```
kuan121, line 69: In general, overrides should be the last resort. 
Do we know which transitive dependencies use handlebars? Can we NOT upgrade them so that the newer versions pull in the safe version of handlebars?

ckeshava: image
`handlebars` is used by `lerna` and `ts-jest`. We can upgrade the latter dependency by 5 patch versions, however `lerna` is already on the latest version. 

Do you suggest we wait until `lerna` and its associated dependencies (shown in the above picture) publish new versions of themselves?

kuan121: > however `lerna` is already on the latest version

I see. If that's the case, we have to use overrides then, unless the practical risk of the vuln is negligible and we can wait until lerna upgrade their dependency trees.

kuan121 (approved):

package.json
Patel-Raj11: Looks like 4.7.9 was working without overrides but need to double check after rebasing this PR - https://github.com/XRPLF/xrpl.js/pull/3235

ckeshava: @Patel-Raj11  thank you. Yes, it looks like `npm update handlebars` sufficiently updates all the references to the old 4.7.8 version. I checked out the above dependabot PR and verified that it installs `handlebarsv4.7.9`. 

This PR is no longer necessary. Thank you.

ckeshava:
This PR is no longer necessary. https://github.com/XRPLF/xrpl.js/pull/3269#discussion_r3047997315

PR #3270, opened 2026-04-09
assortment of fixes in the xrpl packages

ckeshava:
## High Level Overview of Change

This PR includes a few improved validations and bug-fixes. Some of these changes are not warranted by active bugs, rather they are motivated a defense-in-depth perspective. Please read the respective `HISTORY.md` files for more context of the updates.

In addition to the above fixes, this PR also fixes the Github CI setup. The renaming of `rippled` -> `xrpld` was blocking the startup of the docker container. This PR updates most of the important references to rippled executable into `xrpld`.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

### Did you update HISTORY.md?

- [x] Yes
- [ ] No, this change does not impact library users

## Test Plan

Additional test cases have been added to ensure that future code changes do not result in regressions.

PR #3271, opened 2026-04-10
chore(deps): quarterly batch dependency upgrade 2026-Q2

kuan121:

Batches 30 open Dependabot dependency upgrade PRs into a single quarterly upgrade. Upgrades 23 packages across the monorepo, skips 5 blocked by peer dependency constraints or migration scope, and classifies 2 as no-ops.

### Context of Change

Dependabot PRs accumulate faster than they can be individually reviewed and merged. This batch upgrade applies all compatible upgrades at once, fixes breaking changes introduced by major version bumps, and validates the full test suite against the combined result.

Breaking dependency changes fixed:
- `bignumber.js` 9→10: constructor now throws on invalid input instead of returning NaN. Wrapped in try-catch in both `xrpl` (xrpConversion.ts, quality.ts) and `ripple-binary-codec` (quality.ts, amount.ts) to preserve existing error behavior.
- `https-proxy-agent` 7→9: now ESM-only. Updated Jest config to transpile ESM imports.
- `@scure/base` 1→2: stricter `Uint8Array` generics. Widened type annotations in `ripple-binary-codec` (amount.ts, uint-64.ts) to resolve compile errors.

This change is pure dependency maintenance — no public API of any published package changes.

## Test Plan

- `npm run build` — all 6 packages compile
- `npm run lint` — no lint errors
- `npm test` — 1032 unit tests pass
- `npm run test:integration` — 159 integration tests pass against xrpld Docker container
- `npm run test:browser` — 159 browser tests pass (also fixed a pre-existing TS1503 error in `ripple-keypairs` by adding `target: "es2018"` to the shared ts-loader config)
- `npm run test:faucet` — 4 faucet tests pass

## Superseded Dependabot PRs

| PR | Package | From | To | Status | MajorVersionUpgrade |
|---|---|---|---|---|---|
| [#3276](https://github.com/XRPLF/xrpl.js/pull/3276) | follow-redirects | 1.15.11 | 1.16.0 | Upgraded | No |
| [#3272](https://github.com/XRPLF/xrpl.js/pull/3272) | axios, nx | 1.12.0→1.15.0, 22.6.4→22.6.5 | (grouped) | Upgraded | No |
| [#3267](https://github.com/XRPLF/xrpl.js/pull/3267) | lodash | 4.17.23 | 4.18.1 | Upgraded | No |
| [#3266](https://github.com/XRPLF/xrpl.js/pull/3266) | serialize-javascript, terser-webpack-plugin | (grouped) | (grouped) | No-op (transitive, resolved by lockfile) | No |
| [#3265](https://github.com/XRPLF/xrpl.js/pull/3265) | eslint-plugin-jsdoc | 52.0.4 | 62.9.0 | Skipped (peer dep conflict: `@xrplf/eslint-config@^3` requires `eslint-plugin-jsdoc@^52`) | Yes ([v62](https://github.com/gajus/eslint-plugin-jsdoc/releases/tag/v62.0.0)) |
| [#3264](https://github.com/XRPLF/xrpl.js/pull/3264) | https-proxy-agent | 7.0.6 | 9.0.0 | Upgraded | Yes ([v8](https://github.com/TooTallNate/proxy-agents/releases/tag/https-proxy-agent%408.0.0), [v9](https://github.com/TooTallNate/proxy-agents/releases/tag/https-proxy-agent%409.0.0)) |
| [#3263](https://github.com/XRPLF/xrpl.js/pull/3263) | @types/chai | 4.3.20 | 5.2.3 | Upgraded | Yes ([chai v5](https://github.com/chaijs/chai/releases/tag/v5.0.0)) |
| [#3262](https://github.com/XRPLF/xrpl.js/pull/3262) | eslint-plugin-tsdoc | 0.5.0 | 0.5.2 | Upgraded | No |
| [#3261](https://github.com/XRPLF/xrpl.js/pull/3261) | ws | 8.19.0 | 8.20.0 | Upgraded | No |
| [#3260](https://github.com/XRPLF/xrpl.js/pull/3260) | webpack-bundle-analyzer | 5.1.1 | 5.3.0 | Upgraded | No |
| [#3259](https://github.com/XRPLF/xrpl.js/pull/3259) | typescript-eslint | 8.52.0 | 8.58.0 | Upgraded | No |
| [#3258](https://github.com/XRPLF/xrpl.js/pull/3258) | jest-mock | 30.2.0 | 30.3.0 | Upgraded | No |
| [#3257](https://github.com/XRPLF/xrpl.js/pull/3257) | @eslint/js | 9.39.2 | 10.0.1 | Skipped (peer dep conflict: `@xrplf/eslint-config@^3` requires `@eslint/js@^9`) | Yes ([v10](https://github.com/eslint/eslint/releases/tag/v10.0.0)) |
| [#3255](https://github.com/XRPLF/xrpl.js/pull/3255) | prettier | 3.7.4 | 3.8.1 | Upgraded | No |
| [#3253](https://github.com/XRPLF/xrpl.js/pull/3253) | expect | 30.2.0 | 30.3.0 | Upgraded | No |
| [#3252](https://github.com/XRPLF/xrpl.js/pull/3252) | bignumber.js | 9.3.0 | 10.0.2 | Upgraded | Yes ([v10](https://github.com/MikeMcl/bignumber.js/blob/master/CHANGELOG.md)) |
| [#3251](https://github.com/XRPLF/xrpl.js/pull/3251) | typedoc | 0.28.15 | 0.28.18 | Upgraded | No |
| [#3250](https://github.com/XRPLF/xrpl.js/pull/3250) | jest | 30.2.0 | 30.3.0 | Upgraded | No |
| [#3249](https://github.com/XRPLF/xrpl.js/pull/3249) | eslint | 9.39.2 | 10.1.0 | Skipped (peer dep conflict: `@xrplf/eslint-config@^3` requires `eslint@^9`) | Yes ([v10](https://github.com/eslint/eslint/releases/tag/v10.0.0)) |
| [#3248](https://github.com/XRPLF/xrpl.js/pull/3248) | react | 19.2.3 | 19.2.4 | Upgraded | No |
| [#3247](https://github.com/XRPLF/xrpl.js/pull/3247) | eslint-plugin-n | 17.23.1 | 17.24.0 | Upgraded | No |
| [#3246](https://github.com/XRPLF/xrpl.js/pull/3246) | eslint-plugin-array-func | 5.1.0 | 5.1.1 | Upgraded | No |
| [#3245](https://github.com/XRPLF/xrpl.js/pull/3245) | globals | 16.3.0 | 17.4.0 | Skipped (peer dep conflict: `@xrplf/eslint-config@^3` requires `globals@^16`) | Yes ([v17](https://github.com/sindresorhus/globals/releases/tag/v17.0.0)) |
| [#3244](https://github.com/XRPLF/xrpl.js/pull/3244) | typescript | 5.2.2 | 6.0.2 | Skipped (requires monorepo-wide moduleResolution migration) | Yes ([v6](https://github.com/microsoft/TypeScript/releases/tag/v6.0.2)) |
| [#3243](https://github.com/XRPLF/xrpl.js/pull/3243) | eslint-plugin-prettier | 5.5.4 | 5.5.5 | Upgraded | No |
| [#3242](https://github.com/XRPLF/xrpl.js/pull/3242) | webpack | 5.104.1 | 5.105.4 | Upgraded | No |
| [#3241](https://github.com/XRPLF/xrpl.js/pull/3241) | webpack-cli | 6.0.1 | 7.0.2 | Upgraded | Yes ([v7](https://github.com/webpack/webpack-cli/releases/tag/webpack-cli%407.0.0)) |
| [#3080](https://github.com/XRPLF/xrpl.js/pull/3080) | @scure/base | 1.2.6 | 2.0.0 | Upgraded | Yes ([v2](https://github.com/paulmillr/scure-base/releases/tag/2.0.0)) |
| [#3051](https://github.com/XRPLF/xrpl.js/pull/3051) | form-data, lerna | (grouped) | (grouped) | No-op (transitive, resolved by lockfile) | No |
| [#3013](https://github.com/XRPLF/xrpl.js/pull/3013) | ws, @types/ws | 8.18.1→8.18.2, 8.18.0→8.18.1 | (grouped) | Upgraded | No |

## Closing Instructions

After merging, close the following superseded PRs (Skipped ones remain open for future handling): #3276, #3272, #3267, #3266, #3264, #3263, #3262, #3261, #3260, #3259, #3258, #3255, #3253, #3252, #3251, #3250, #3248, #3247, #3246, #3243, #3242, #3241, #3080, #3051, #3013

The following PRs were Skipped and should remain open: #3265 (eslint-plugin-jsdoc), #3257 (@eslint/js), #3249 (eslint), #3245 (globals), #3244 (typescript).


package.json
```diff
@@ -28,40 +28,41 @@
...
-    "expect": "^30.2.0",
-    "jest": "^30.2.0",
-    "jest-mock": "^30.2.0",
+    "eslint-plugin-n": "^17.24.0",
+    "eslint-plugin-prettier": "^5.5.5",
+    "eslint-plugin-tsdoc": "^0.5.2",
+    "expect": "^30.3.0",
+    "jest": "^30.3.0",
+    "jest-mock": "^30.3.0",
     "lerna": "^9.0.4",
```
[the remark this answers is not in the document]
kuan121: > According to the team's documented decision, the repository intentionally uses v8 for now, with v9 planned for later consideration.

Where is this documented?

.claude/skills/batch-deps-upgrade/SKILL.md
```diff
@@ -0,0 +1,76 @@
...
+7. If any upgrade changes the public API of a package (new errors, changed return types, removed functionality) and result in a breaking change, add an entry under `## Unreleased` in that package's `HISTORY.md`.
+8. Verify completeness: every PR from step 1 must have a status (Upgraded, No-op, Removed, or Skipped). If any PR is unaccounted for, stop and report it before proceeding.
+
+## Step 3: Validate
+
+Run the full test suite in order:
+1. npm run build && npm run lint
+2. npm test
+3. Start xrpld Docker container (based on CONTRIBUTING.md, with `--detach` instead of `-it` for automation):
+   docker run --detach --rm -p 6006:6006 --volume "$PWD/.ci-config/":/etc/opt/xrpld/ --name rippled_standalone --entrypoint bash rippleci/xrpld:develop -c "xrpld --standalone"
```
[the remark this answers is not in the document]
kuan121: The name and the path have been updated in  https://github.com/XRPLF/xrpl.js/pull/3270

.claude/skills/batch-deps-upgrade/SKILL.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Step 4: Generate Outputs
+
+Do NOT commit or create a PR. Instead, generate the following outputs for the human to use:
+
+1. **Code changes explanation** — write a markdown file (`.claude/skills/batch-deps-upgrade/code-changes.md`) documenting every non-package.json source code change, explaining what broke, why, and the minimal fix applied.
+
+2. **Commit message** — output a concise commit message the human can copy-paste into `git commit -m "..."`. Format: `chore(deps): quarterly batch dependency upgrade YYYY-QN` followed by a brief summary of upgrades, skips, and removals.
+
+3. **PR description** — write a markdown file (`.claude/skills/batch-deps-upgrade/pr-description.md`) following the repo's PR template (.github/pull_request_template.md):
```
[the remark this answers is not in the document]
kuan121: The only occurrence of "github" on that line is .github/ (the literal directory name, which must be lowercase). There's no product reference to capitalize.

packages/ripple-binary-codec/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "@xrplf/isomorphic": "^1.0.1",
-    "bignumber.js": "^9.0.0",
+    "bignumber.js": "^10.0.2",
```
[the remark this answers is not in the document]
kuan121: Good point. The issue is addressed in https://github.com/XRPLF/xrpl.js/pull/3271/changes/876a5d47e8b78e40007895f66342fe32230d5933

.claude/skills/batch-deps-upgrade/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+   - Wait for port 6006 with a bounded timeout and halt on failure:
+     ```
+     SECONDS=0
+     until nc -z localhost 6006 || [ $SECONDS -gt 120 ]; do sleep 2; done
+     if ! nc -z localhost 6006; then
+       echo "Error: xrpld did not start within 120s"
+       docker logs xrpld-service
+       exit 1
+     fi
+     ```
```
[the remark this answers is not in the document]
kuan121: The issue is addressed in https://github.com/XRPLF/xrpl.js/pull/3271/changes/697aa98772c951d900445e563db82134f32e7e65

webpack.test.config.js
```diff
@@ -34,6 +34,7 @@ function webpackForTest(testFileName, basePath) {
               loader: "ts-loader",
               options: {
                 compilerOptions: {
+                  target: "es2018",
```
kuan121: The test file `packages/ripple-keypairs/test/getAlgorithmFromKey.test.ts` uses a named capture group: `/(?<tabs>\n)\s+/gu`. Named groups are an ES2018 feature. The shared test webpack config used ts-loader with `lib: ["esnext", "dom"]` but no explicit `target`, so TypeScript fell back to its default (below ES2018) and ts-loader raised `TS1503: Named capturing groups are only available when targeting 'ES2018' or later`. `npm run test:browser` failed for `ripple-keypairs`, which blocked downstream packages.

**Fix**: Added `target: "es2018"` to the ts-loader `compilerOptions` in the shared `webpack.test.config.js`. This only affects the browser test bundle; production builds continue to use each package's own tsconfig (`target: "es6"`), and the test file isn't part of the published artifact. ES2018 is the minimum that makes the regex compile — we could go higher but kept it at the threshold.

jest.config.base.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
       },
```
kuan121, line 12: The changes in this file are **caused by**: `https-proxy-agent` 7 → 9

v9 is ESM-only (`import` syntax), but Jest runs in CommonJS. Without this, Jest fails with `SyntaxError: Cannot use import statement outside a module` when tests import the xrpl client (which uses https-proxy-agent).

**Fix**: Added transform entries so ts-jest transpiles the ESM code, and excluded them from `transformIgnorePatterns` so Jest doesn't skip them.

packages/xrpl/src/utils/xrpConversion.ts
```diff
@@ -21,12 +21,12 @@ export function dropsToXrp(dropsToConvert: BigNumber.Value): number {
    * decimal point followed by zeros, e.g. '1.00'.
```
kuan121, line 21: The changes in this file are **caused by**: `bignumber.js` 9 → 10

v9: `new BigNumber("FOO")` returns a BigNumber with value `NaN`.
v10: `new BigNumber("FOO")` **throws** `[BigNumber Error] Not a number: FOO`.

The existing code passed invalid input to BigNumber first, then checked if the result was NaN to throw a custom `ValidationError`. In v10 the BigNumber constructor throws before the NaN check runs, so the user-facing error message changes from the expected `ValidationError` to a raw BigNumber error.

**Fix**: Wrap `new BigNumber()` calls in try-catch, re-throwing as the same `ValidationError` the code already intended. Also removed the now-dead NaN checks that followed (unreachable since the constructor throws first). Applied in both `dropsToXrp` and `xrpToDrops`.

packages/xrpl/src/utils/quality.ts
```diff
@@ -17,7 +17,11 @@ function percentToDecimal(percent: string): string {
     throw new ValidationError(`Value ${percent} contains too many % signs`)
```
kuan121, line 17: Similar to the changes in `xrpConversion.ts`, the changes in this file are **caused by**: same `bignumber.js` 9 → 10

Three functions accept user string input and construct a BigNumber: `percentToDecimal`, `decimalToTransferRate`, and `decimalToQuality`. All three previously relied on v9's NaN-return behavior to either propagate a `'NaN'` string downstream or trip an explicit `if (billionths === 'NaN')` check that threw `ValidationError("Value is not a number")`.

In v10, the constructor throws a raw `[BigNumber Error]` before any downstream check can run, so the expected `ValidationError` is never produced and tests (e.g., `percentToQuality throws with gibberish`, `decimalToQuality throws with gibberish` in [quality.test.ts](packages/xrpl/test/utils/quality.test.ts#L60-L74)) fail with a raw BigNumber error.

**Fix**:
- Wrap each `new BigNumber()` call in try-catch and re-throw `ValidationError("Value is not a number")` — the exact error class and message consumers already observed in v9.
- Remove the now-unreachable NaN checks in `decimalToTransferRate` and `decimalToQuality` (the stringified `'NaN'` path is dead in v10 because the constructor throws first).a

packages/ripple-binary-codec/src/quality.ts
```diff
@@ -13,8 +13,13 @@ class quality {
    * @returns Serialized quality
```
kuan121, line 13: The changes in this file are **caused by**: same `bignumber.js` 9 → 10

`encode(quality)` and `decode(quality)` construct BigNumbers from strings. In v10 both will throw raw `[BigNumber Error]` on invalid input instead of returning NaN.

**Fix**: Wrap both constructor calls in try-catch and re-throw a clear `Error: <value> is not a valid quality` instead.instead.

packages/ripple-binary-codec/src/types/amount.ts
```diff
@@ -92,7 +92,7 @@ class Amount extends SerializedType {
       return value
     }
 
-    let amount = new Uint8Array(8)
+    let amount: Uint8Array = new Uint8Array(8)
```
kuan121, line 95: This change is **caused by**: `@scure/base` 1 → 2

v2 introduced stricter `Uint8Array` generics. `new Uint8Array(8)` is inferred as `Uint8Array<ArrayBuffer>`, but `concat()` (via `@noble/hashes`) returns `Uint8Array<ArrayBufferLike>` — a wider type that includes `SharedArrayBuffer`. TypeScript won't assign the wider return type back to the narrower variable, causing a compile error. This is purely a compile-time issue — at runtime there is no difference.

**Fix**: Explicit type annotation `Uint8Array` (without generic parameter) so TypeScript uses the unparameterized base type that accepts both.

packages/ripple-binary-codec/src/types/amount.ts
```diff
@@ -110,7 +110,12 @@ class Amount extends SerializedType {
     }
```
kuan121, line 111: The following changes are **caused by**: same `bignumber.js` 9 → 10

Three constructor call sites accept user input and could throw in v10:
- Line 113: `new BigNumber(value.value)` inside `from()` for IOU amounts
- Line 264: `new BigNumber(amount)` inside `assertXrpIsValid`
- Line 304: `new BigNumber(amount)` inside `assertMptIsValid`

**Fix**: Wrap each constructor call in try-catch, re-throwing the existing `Error: <amount> is an illegal amount` message to preserve API behavior. Also removed the dead `if (decimal.isNaN())` check in `assertMptIsValid` — unreachable in v10 because the constructor throws first.

packages/ripple-binary-codec/src/types/uint-64.ts
```diff
@@ -51,7 +51,7 @@ class UInt64 extends UInt {
       return val
     }
 
-    let buf = new Uint8Array(UInt64.width)
+    let buf: Uint8Array = new Uint8Array(UInt64.width)
```
kuan121, line 54: This change is **caused by**: `@scure/base` 1 → 2

v2 introduced stricter `Uint8Array` generics. `new Uint8Array(8)` is inferred as `Uint8Array<ArrayBuffer>`, but `concat()` (via `@noble/hashes`) returns `Uint8Array<ArrayBufferLike>` — a wider type that includes `SharedArrayBuffer`. TypeScript won't assign the wider return type back to the narrower variable, causing a compile error. This is purely a compile-time issue — at runtime there is no difference.

**Fix**: Explicit type annotation `Uint8Array` (without generic parameter) so TypeScript uses the unparameterized base type that accepts both.

webpack.test.config.js
ckeshava: > so TypeScript fell back to its default (below ES2018) and ts-loader raised TS1503: Named capturing groups are only available when targeting 'ES2018' or later. npm run test:browser failed for ripple-keypairs, which blocked downstream packages

I haven't observed this error on historical CI runs. Was this necessitated by changes in this PR?

.claude/skills/batch-deps-upgrade/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+     - Otherwise, do not check any Type of Change — dependency upgrades are maintenance and don't fit "Refactor" (which means restructuring code without behavior change). Note in the PR body that the upgrade is maintenance.
+   - Include a "Superseded Dependabot PRs" section with a table: PR (linked), Package, From, To, Status, MajorVersionUpgrade
+     - Status values: Upgraded, No-op (reason), Skipped (peer dep conflict / CI failure: error)
+     - MajorVersionUpgrade: `No` if the major version number did not change. Otherwise `Yes` plus a link for each major version crossed. For example, 7.x → 9.x yields `Yes ([v8](url), [v9](url))`. Each link should point to the package's release notes or changelog for that major version. Verify each link returns HTTP 200 and has meaningful content (e.g., `curl -sL -o /dev/null -w "%{http_code}" <url>`); if a package doesn't publish per-version GitHub releases (e.g., TypeScript sometimes skips `x.0.0`/`x.0.1` tags, bignumber.js puts details in CHANGELOG.md), fall back to the CHANGELOG.md file or the closest valid release tag.
+   - Closing instructions: "After merging, close the following PRs: #X, #Y, #Z"
```
ckeshava: This bullet point is not clear to me. Does the PR author need to close the dependabot PRs manually?  
Alternatives:
1. We can run a `gh ...` script that closes all the mentioned PRs with a helpful comment.
2. We can add this task as an instruction for Claude i.e. Close the following PRs after the successful creation of the Batch-Dependency-Upgrade PR.

packages/ripple-binary-codec/src/quality.ts
```diff
@@ -13,8 +13,13 @@ class quality {
...
   static encode(quality: string): Uint8Array {
-    const decimal = BigNumber(quality)
-    const exponent = (decimal?.e || 0) - 15
+    let decimal: BigNumber
+    try {
+      decimal = new BigNumber(quality)
+    } catch (_err) {
+      throw new Error(`${quality} is not a valid quality`)
+    }
+    const exponent = (decimal.e || 0) - 15
```
ckeshava, line 22:
```suggestion
    const exponent = (decimal.e) - 15
```

Since the BIgNumber object is constructed successfully, the exponent attribute must always be present.

packages/ripple-binary-codec/src/quality.ts
```diff
@@ -13,8 +13,13 @@ class quality {
    * @returns Serialized quality
    */
   static encode(quality: string): Uint8Array {
-    const decimal = BigNumber(quality)
-    const exponent = (decimal?.e || 0) - 15
+    let decimal: BigNumber
+    try {
+      decimal = new BigNumber(quality)
+    } catch (_err) {
+      throw new Error(`${quality} is not a valid quality`)
```
ckeshava, line 20: nit: It could be useful to append the message in `_err` message inside the `throw new Error(...)` block. This could provide context/stacktrace for the user.

packages/ripple-binary-codec/src/types/amount.ts
ckeshava: nit: This is also a good place to wrap within a `try-catch` block. We accept unverified inputs from the `BinaryParser`. Consequently, the bytes in `mantissa` might be parsed as a `NaN`. 

Ideally, we should verify the inputs in the `constructor / fromParser` methods. However, with the current state of the code, this is a good place to add this try-catch block.

Patel-Raj11 (dismissed):

webpack.test.config.js
kuan121: Good catch. After digging in, the issue came from TypeScript drifting from 5.2.2 to 5.9.3 when I ran npm install to update the lockfile (both satisfy ^5.1.6). TypeScript 5.9 enforces TS1503, which 5.2 does not.

Since upgrading TypeScript is out of scope, I reverted the webpack.test.config.js change and pinned the lockfile back to 5.2.2 to match main.

.claude/skills/batch-deps-upgrade/SKILL.md
kuan121: Good callout. The plan is for the PR author to close the Dependabot PRs after the consolidated PR is merged. I don't prescribe how to do that in this version—they can close them manually or use Claude to bulk close.

I will update this bullet point to 
> Closing instructions with two paragraphs:
>1. "After merging, close the following superseded PRs (Skipped ones remain open for future handling): #X, #Y, #Z" — list only Upgraded and No-op PRs.
>2. "The following PRs were Skipped and should remain open: #A (package-a), #B (package-b), ..." — annotate each with the package name. These stay open so Dependabot keeps rebasing them.

As a result, AI will generate an output like the following for this PR
> ## Closing Instructions
> After merging, close the following superseded PRs (Skipped ones remain open for future handling): #3276, #3272, #3267, #3266, #3264, #3263, #3262, #3261, #3260, #3259, #3258, #3255, #3253, #3252, #3251, #3250, #3248, #3247, #3246, #3243, #3242, #3241, #3080, #3051, #3013
>
> The following PRs were Skipped and should remain open: #3265 (eslint-plugin-jsdoc), #3257 (@eslint/js), #3249 (eslint), #3245 (globals), #3244 (typescript).

packages/ripple-binary-codec/src/quality.ts
kuan121: If the decimal is 0, decimal.e can be null, which would result in `null - 15`. So it’s safer to keep the || 0.

packages/ripple-binary-codec/src/types/amount.ts
kuan121: Good point. Addressed in https://github.com/XRPLF/xrpl.js/pull/3271/changes/419b60739dfcdc30f8d1a11647777273a81a01f1

packages/ripple-binary-codec/src/quality.ts
kuan121: In my opinion, the current messages are self-explanatory and contain the invalid value. Keeping the original “[BigNumber Error]” text doesn’t add much debugging value and unnecessarily exposes an internal dependency. Let me know if this makes sense.

Patel-Raj11 (dismissed):

packages/ripple-binary-codec/src/types/amount.ts
```diff
@@ -216,9 +221,14 @@ class Amount extends SerializedType {
...
-        `1e${exponent}`,
-      )
+      let value: BigNumber
+      try {
+        value = new BigNumber(`${sign}0x${bytesToHex(mantissa)}`).times(
+          `1e${exponent}`,
+        )
+      } catch (_err) {
+        throw new Error(`${sign}0x${bytesToHex(mantissa)} is an illegal amount`)
+      }
```
cybele-ripple: `mantissa` is always 8 bytes so the error on line 230 would not be thrown unless there is an error with `BigNumber` itself. Would `const value = new BigNumber(`${sign}0x${bytesToHex(mantissa)}`).times(`1e${exponent}`)` be sufficient?

packages/xrpl/src/utils/xrpConversion.ts
```diff
@@ -63,12 +63,12 @@ export function dropsToXrp(dropsToConvert: BigNumber.Value): number {
...
-
-  // check that the value is valid and actually a number
-  if (typeof xrpToConvert === 'string' && xrp === 'NaN') {
+  let xrp: string
+  try {
+    xrp = new BigNumber(xrpToConvert).toString(BASE_TEN)
+  } catch (_err) {
     throw new ValidationError(
-      `xrpToDrops: invalid value '${xrpToConvert}', should be a BigNumber or string-encoded number.`,
+      `xrpToDrops: invalid value '${typeof xrpToConvert === 'string' ? xrpToConvert : JSON.stringify(xrpToConvert)}', should be a BigNumber or string-encoded number.`,
```
cybele-ripple, line 71: Since xrpToConvert is type `BigNumber.Value` can we just use `${String(dropsToConvert)}` instead of `${typeof xrpToConvert === 'string' ? xrpToConvert : JSON.stringify(xrpToConvert)}`

webpack.test.config.js
ckeshava: thanks for the explanation

.claude/skills/batch-deps-upgrade/SKILL.md
ckeshava: ok, thanks for the explanation

packages/xrpl/src/utils/xrpConversion.ts
kuan121: When I remove use `${String(dropsToConvert)}`, I got the following lint error
```
error  'xrpToConvert' may use Object's default stringification format ('[object Object]') when stringified
  @typescript-eslint/no-base-to-string
```
Therefore, I will have to keep the `JSON.stringify` part.

packages/ripple-binary-codec/src/types/amount.ts
kuan121: Good catch. The try/catch was added while addressing this [comment](https://github.com/XRPLF/xrpl.js/pull/3271#discussion_r3125357009). After taking a closer look at the code, it seems unnecessary, so we can remove it.

cybele-ripple (approved):

Patel-Raj11 (approved):

PR #3327, opened 2026-04-21
Redact private key material in errors

cybele-ripple:
Private keys now appear as [redacted] in error messages 

### Context of Change
When key validation failed, keyError() interpolated the full hex key into the thrown Error.message. For private keys this propagated the secret to application logs, error-tracking services, and HTTP responses that echo error messages.

Replace the key with "[redacted]" when type === 'private'. Public keys are still rendered since they are not sensitive. Add a test asserting the private-key hex is absent from the error message and a regression test pinning the redaction contract.

### Type of Change

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation Updates
- [ ] Release

- [ ] Yes

## Test Plan
Added tests to demonstrate security fix


packages/ripple-keypairs/src/utils/getAlgorithmFromKey.ts
ckeshava, line 109: If the length is of a non-standard type, can't we coerce a Dictionary KeyError at this juncture? I have not looked at the entire call stack, however this attack vector is worth an investigation.

A KeyError here could spill the private key in the error logs.

cybele-ripple: In the case of a `KeyError`, the private key would not be returned. I have added a unit test [here](https://github.com/XRPLF/xrpl.js/pull/3327/changes#diff-86890ecd99d39c9d0164529c3d2fb59dee077653673a99b0216df7fc25600528R133-R150) to demonstrate that a non-standard yields `redacted` instead of the key itself

kuan121 (dismissed):

kuan121 (approved):

ckeshava (approved):

PR #3337, opened 2026-05-11
fix(xrpl): infer signing algorithm from seed prefix in Wallet.fromSeed / fromSecret

ckeshava:
### Background

Until now, `Wallet.fromSeed(seed)` (and its alias `Wallet.fromSecret`) silently coerced `opts.algorithm` to ed25519 whenever the caller didn't pass one. For ed25519 family seeds (`sEd…`) this happened to produce the right account. For a secp256k1 family seed (`s…`), it produced a valid keypair on a completely different account — making it look like the user's account "disappeared." This has been one of the most commonly reported migration issues since the 3.x default flip.

### Change

`Wallet.deriveWallet` no longer applies a fallback default when `opts.algorithm` is omitted. The undefined value flows through to `ripple-keypairs.deriveKeypair`, which already infers the algorithm from the seed prefix via `decodeSeed`. Net effect:

- `sEd…` seeds → ed25519 (unchanged result for callers who relied on the previous default)
- `s…` family seeds → secp256k1 (correct curve; previously returned an unrelated ed25519 account)

`Wallet.generate` still passes its algorithm explicitly, so its ed25519 default is preserved. JSDoc updated on `fromSeed`, `fromSecret`, `deriveWallet`, and (in a separate doc-only commit) `walletFromSecretNumbers`.

### Wallet factory defaults after this PR

| Method | Input | Algorithm when `opts.algorithm` is omitted |
|---|---|---|
| `Wallet.generate()` | none (random) | `ed25519` |
| `Wallet.fromSeed(seed)` | seed string | inferred from prefix: `sEd…` → `ed25519`, otherwise `secp256k1` |
| `Wallet.fromSecret(secret)` | seed string (alias of `fromSeed`) | inferred from prefix: `sEd…` → `ed25519`, otherwise `secp256k1` |
| `Wallet.fromEntropy(entropy)` | random bytes | `ed25519` |
| `Wallet.fromMnemonic(mnemonic)` (bip39) | bip39 mnemonic | n/a — HD-derives the keypair directly, no algorithm parameter consulted |
| `Wallet.fromMnemonic(mnemonic, { mnemonicEncoding: 'rfc1751' })` | RFC1751 mnemonic | `ed25519` |
| `walletFromSecretNumbers(numbers)` | Xaman-style secret numbers | `secp256k1` (Xaman compatibility — intentional deviation) |
| `ripple-keypairs.generateSeed()` | none / entropy | `ed25519` |
| `ripple-keypairs.deriveKeypair(seed)` | seed string | inferred from prefix via `decodeSeed` |
| `secret-numbers.Account(numbers)` | secret numbers | `ed25519` |

### Migration

**Only `Wallet.fromSeed` and `Wallet.fromSecret` change behavior. Every other Wallet factory listed above is unaffected by this PR.**

| Call site | Before this PR | After this PR | Required action |
|---|---|---|---|
| `Wallet.fromSeed(seed)` / `Wallet.fromSecret(seed)` where `seed` does **not** start with `sEd` (i.e. `s…` family seed), `opts.algorithm` omitted | returned an ed25519 keypair for an unrelated account | returns the secp256k1 keypair for the account the seed was actually issued for | If you intended secp256k1: **no action** — you now get the correct keypair. If your application has persisted the previously-derived ed25519 keypair and must keep using it: pass `{ algorithm: ECDSA.ed25519 }` explicitly. |
| `Wallet.fromSeed(seed)` / `Wallet.fromSecret(seed)` where `seed` starts with `sEd`, `opts.algorithm` omitted | returned an ed25519 keypair | returns an ed25519 keypair (now via prefix inference instead of the hard-coded default) | **None** — result is identical. Optional: pass `{ algorithm: ECDSA.ed25519 }` explicitly as defense against future default changes. |
| `Wallet.fromSeed` / `Wallet.fromSecret` with an explicit `opts.algorithm` (either curve) | honored the explicit value | honors the explicit value | **None.** |
| `Wallet.generate`, `Wallet.fromEntropy`, `Wallet.fromMnemonic`, `walletFromSecretNumbers`, `ripple-keypairs.generateSeed`, `secret-numbers.Account` | unchanged | unchanged | **None.** |

### Tests

- The two `derives a wallet using default algorithm` cases were locking in the buggy behavior (asserted ed25519 keys for the secp256k1 fixture seed `sh1HiK…`). Renamed to `infers secp256k1 from a non-sEd seed prefix when algorithm is omitted` and updated to assert the correct secp256k1 keys.
- Added sibling cases that ingest `sEdVaw4m9W3H3ou3VnyvDwvPAP5BEz1` and assert the inferred result equals an explicit `{ algorithm: ECDSA.ed25519 }` call.
- Full unit suite: 117 suites / 1040 tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/xrpl/src/Wallet/index.ts
+++ b/packages/xrpl/src/Wallet/index.ts
@@ -146,7 +146,8 @@ export class Wallet {
    *
    * @param seed - A string used to generate a keypair (publicKey/privateKey) to derive a wallet.
    * @param opts - (Optional) Options to derive a Wallet.
-   * @param opts.algorithm - The digital signature algorithm to generate an address for.
+   * @param opts.algorithm - The digital signature algorithm to generate an address for. When omitted,
+   *                         the algorithm is inferred from the seed prefix (`sEd…` → ed25519, otherwise secp256k1).
    * @param opts.masterAddress - Include if a Wallet uses a Regular Key Pair. It must be the master address of the account.
    * @returns A Wallet derived from a seed.
    */
@@ -165,7 +166,8 @@ export class Wallet {
    *
    * @param secret - A string used to generate a keypair (publicKey/privateKey) to derive a wallet.
    * @param opts - (Optional) Options to derive a Wallet.
-   * @param opts.algorithm - The digital signature algorithm to generate an address for.
+   * @param opts.algorithm - The digital signature algorithm to generate an address for. When omitted,
+   *                         the algorithm is inferred from the seed prefix (`sEd…` → ed25519, otherwise secp256k1).
    * @param opts.masterAddress - Include if a Wallet uses a Regular Key Pair. It must be the master address of the account.
    * @returns A Wallet derived from a secret (AKA a seed).
    */
@@ -285,7 +287,8 @@ export class Wallet {
    *
    * @param seed - The seed used to derive the wallet.
    * @param opts - (Optional) Options to derive a Wallet.
-   * @param opts.algorithm - The digital signature algorithm to generate an address for.
+   * @param opts.algorithm - The digital signature algorithm to generate an address for. When omitted,
+   *                         `deriveKeypair` infers it from the seed prefix (`sEd…` → ed25519, otherwise secp256k1).
    * @param opts.masterAddress - Include if a Wallet uses a Regular Key Pair. It must be the master address of the account.
    * @returns A Wallet derived from the seed.
    */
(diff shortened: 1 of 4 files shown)


Patel-Raj11 (approved):
LGMT!

achowdhry-ripple (approved):

ckeshava:
@mvadari  Is this PR not required? there is an equivalent one in the xrpl-py repository as well.

packages/xrpl/src/Wallet/index.ts
```diff
@@ -146,7 +146,8 @@ export class Wallet {
    *
```
kuan121, line 146: Can you ensure that all CI/CD checks pass?

kuan121 (approved):

packages/xrpl/src/Wallet/index.ts
ckeshava: The integration test CI was failing due to an old docker issue. I did not have to make any updates to the library code, a re-run of the CI passed the test.

ckeshava:
thanks for the reviews @kuan121  @Patel-Raj11

PR #3338, opened 2026-05-11
fix(xrpl): use deep amount comparison in handleDeliverMax (#3313)

### LINKED ISSUE #3313: handleDeliverMax uses reference equality for Amount objects
## Description

`tx.Amount !== tx.DeliverMax` uses `!==` (reference equality). For IOU amounts (objects like `{currency, issuer, value}`), two distinct objects with identical values always compare as not-equal, throwing a **spurious validation error**. This mismatch check is broken for all non-XRP payments.

## Location

[packages/xrpl/src/sugar/autofill.ts#L482](https://github.com/XRPLF/xrpl.js/blob/5cf0921b0fa4637f324bcf5b0788e2ef59a83c8a/packages/xrpl/src/sugar/autofill.ts#L482)

## Severity

**Medium** -- Valid IOU Payment transactions with both Amount and DeliverMax erroneously rejected.

## Adversarial Review Notes

The bug only manifests when BOTH Amount and DeliverMax are independently provided as separate objects with identical values. For XRP string amounts, `"1000" !== "1000"` is `false` (string comparison by value). **The bug only affects IOU/MPT object amounts** where two structurally identical objects have different references. **Unit test confirmed: throws** `ValidationError` for identical IOU objects with different references.

Confidence: 92%

## Unit Test Patch

No standalone unit test added to existing files for this bug — `handleDeliverMax` is tested in `autofill.test.ts` which requires WebSocket client setup. The bug was confirmed via a direct import test.

satyakwok:
Closes #3313.

## Summary

`handleDeliverMax` in `src/sugar/autofill.ts` compared the two fields with `!==` (reference equality). For IOU and MPT amounts — objects like `{currency, issuer, value}` — two distinct object literals with identical fields always compare not-equal, so every IOU payment that set both \`Amount\` and \`DeliverMax\` as separate object literals threw a spurious validation error:

```
PaymentTransaction: Amount and DeliverMax fields must be identical when both are provided
```

This silently broke any IOU payment flow that built the two fields independently (e.g. one constructed by the caller, the other autofilled from a higher-level helper).

## Fix

Reuse the existing `amountsEqual` helper from `src/client/partialPayment.ts`. It already handles all three amount shapes correctly:

- **XRP** (string drops): `===`
- **IOU**: `currency` + `issuer` + `BigNumber(value)` (so `"1.0"` and `"1"` compare equal)
- **MPT**: `mpt_issuance_id` + `BigNumber(value)`

Only change to `partialPayment.ts` is making `amountsEqual` exported (was previously file-local). Same implementation, just visible to other modules.

## Tests

Two new tests in `packages/xrpl/test/client/autofill.test.ts`:

1. **Regression** — identical IOU `Amount` and `DeliverMax` as **separate object literals** → must autofill cleanly. Fails on `main` (throws the spurious `ValidationError`), passes after.
2. **Companion** — different `value` with same `currency` / `issuer` → must still throw. Pins that the fix didn't loosen validation.

The pre-existing tests for the `=== string` path and the same-reference IOU case (`paymentTx.DeliverMax = paymentTx.Amount`) still pass.

## Smoke gate

- `npm run build` from repo root — all 6 workspaces build clean.
- `npx jest --config=jest.config.unit.js --testPathPatterns autofill` — 23 passed, 0 failed.

### DIFF
--- a/packages/xrpl/src/client/partialPayment.ts
+++ b/packages/xrpl/src/client/partialPayment.ts
@@ -28,7 +28,16 @@ const WARN_PARTIAL_PAYMENT_CODE = 2001
 
 /* eslint-disable complexity -- check different token types */
 /* eslint-disable @typescript-eslint/consistent-type-assertions -- known currency type */
-function amountsEqual(
+/**
+ * Returns true when two transaction amounts represent the same on-ledger value.
+ *
+ * Handles all three amount shapes (XRP string, IOU object, MPT object) and uses
+ * `BigNumber` for the numeric `value` field so `"1.0"` and `"1"` compare equal.
+ * Exported so callers outside this module (notably `handleDeliverMax` in
+ * `sugar/autofill.ts`) can avoid reference-equality bugs on IOU/MPT objects —
+ * see issue #3313.
+ */
+export function amountsEqual(
   amt1: Amount | MPTAmount,
   amt2: Amount | MPTAmount,
 ): boolean {
--- a/packages/xrpl/src/sugar/autofill.ts
+++ b/packages/xrpl/src/sugar/autofill.ts
@@ -4,6 +4,7 @@ import BigNumber from 'bignumber.js'
 import { xAddressToClassicAddress, isValidXAddress } from 'ripple-address-codec'
 
 import { type Client } from '..'
+import { amountsEqual } from '../client/partialPayment'
 import { ValidationError, XrplError } from '../errors'
 import { LoanBroker } from '../models/ledger'
 import {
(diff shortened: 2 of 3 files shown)


satyakwok:
Thanks @coderabbitai — both fair.

**1. Comment wording (`autofill.test.ts:112`)** — fixed in d1519c0f. You're right, the operand semantics were inverted: `tx.Amount !== tx.DeliverMax` returns `true` for distinct object references (different identity), which is what tripped the validation throw. Updated the comment to match.

**2. MPT regression coverage** — held off intentionally. `Payment.Amount` / `DeliverMax` are typed as `Amount = IssuedCurrencyAmount | string` today (see `models/common/index.ts` — there's a `// TODO: add MPTAmount to Amount once MPTv2 is released` next to it), so an MPT pair would need an `as any` cast to compile. The shared `amountsEqual` helper already has direct MPT coverage in its own call site (`partialPayment.ts`). Happy to add the `Payment` MPT coverage as a follow-up the moment MPT lands in the `Amount` union.

PR #3340, opened 2026-05-18
fix(connection): block reserved internal event types from server messages

### LINKED ISSUE #3318: Server-controlled event injection via unvalidated data.type
## Description

`this.emit(data.type as string, data)` emits whatever event name the server dictates, with **zero validation**. A malicious/compromised server can emit `"connected"`, `"disconnected"`, `"error"`, or `"reconnect"` events, tricking the client into false state transitions, reconnection loops, or believing it's connected when it isn't.

## Location

[packages/xrpl/src/client/connection.ts#L357](https://github.com/XRPLF/xrpl.js/blob/5cf0921b0fa4637f324bcf5b0788e2ef59a83c8a/packages/xrpl/src/client/connection.ts#L357)

## Severity

**Critical** -- Remote event injection from a rogue server.

## Adversarial Review Notes

The EventEmitter `emit()` only fires registered listeners. If no handlers are bound for a spoofed event name, the emit is a no-op. However, the Connection and Client classes DO register handlers for `"connected"`, `"disconnected"`, `"error"`, etc. A rogue server sending `{"type": "error", ...}` could trigger error handlers. The lack of a whitelist is a genuine design issue.

Confidence: 90%

## Unit Test Patch

No standalone unit test for this bug — requires WebSocket mocking. Suggested fix: add a whitelist of allowed event types before `this.emit(data.type, ...)`, e.g. `if (!ALLOWED_EVENTS.has(data.type)) return`.

achowdhry-ripple:
## Summary

Closes #3318

`Connection.onMessage` previously called `this.emit(data.type as string, data)` with zero validation of `data.type`. A malicious or compromised rippled server could therefore send `{"type":"connected"}`, `{"type":"disconnected"}`, `{"type":"error"}`, or `{"type":"reconnect"}` and trigger `Connection`'s internal state-event listeners (registered in `Client`), spoofing the local connection state, faking errors, or driving reconnect logic.

This PR introduces a denylist of reserved internal event names and refuses to forward server messages whose `type` collides with any of them. Such messages are dropped and surfaced via the existing `('error', 'badMessage', ...)` channel so they remain observable.

A denylist (rather than the whitelist suggested in the issue) was chosen deliberately to preserve the existing "unrecognized message type" forward-compatibility contract in `packages/xrpl/test/connection.test.ts`. Unknown future rippled stream types continue to be forwarded as before; only reserved local event names are blocked.

## Changes
- `packages/xrpl/src/client/connection.ts`: add `RESERVED_INTERNAL_EVENTS` set and guard the emit in `onMessage` against it.
- `packages/xrpl/test/connection.test.ts`: add unit tests covering all four reserved event names, including the subtle `type: "error"` case (verifies the raw payload object never reaches `error` listeners).
- `packages/xrpl/HISTORY.md`: changelog entry under Unreleased -> Fixed.

## Test plan
- [ ] `cd packages/xrpl && npm test -- connection.test.ts` - new "drops server message with reserved internal event type" cases pass.
- [ ] Existing "unrecognized message type" test still passes (forward-compat preserved).
- [ ] Existing "emit stream messages" and "propagates error message" tests still pass (legitimate stream/error paths unaffected).
- [ ] `npm run lint` in `packages/xrpl` clean.

### DIFF
--- a/packages/xrpl/src/client/connection.ts
+++ b/packages/xrpl/src/client/connection.ts
@@ -28,13 +28,14 @@ const CONNECTION_TIMEOUT = 5
  * of a server-supplied `data.type`, otherwise a malicious or compromised
  * rippled server could spoof connection state (e.g. claim the client is
  * `connected` when it is not, or inject a fake `error`/`disconnected`/
- * `reconnect` to drive the client into a bad state). See DGE-6740.
+ * `reconnect`/`reconnecting` to drive the client into a bad state).
  */
 const RESERVED_INTERNAL_EVENTS: ReadonlySet<string> = new Set([
   'connected',
   'disconnected',
   'error',
   'reconnect',
+  'reconnecting',
 ])
 
 /**
@@ -373,7 +374,7 @@ export class Connection extends EventEmitter {
       // Refuse to forward server-supplied event names that collide with
       // Connection's internal state events. Otherwise a rogue server could
       // spoof local connection state by sending e.g. `{"type":"connected"}`
-      // or `{"type":"error"}`. See DGE-6740.
+      // or `{"type":"error"}`.
       if (RESERVED_INTERNAL_EVENTS.has(type)) {
         this.emit(
           'error',
--- a/packages/xrpl/src/client/connection.ts
+++ b/packages/xrpl/src/client/connection.ts
@@ -23,6 +23,21 @@ const TIMEOUT = 20
 const CONNECTION_TIMEOUT = 5
 
 /**
+ * Event names that are emitted internally by Connection to signal local
+ * client/socket state transitions. These must never be emitted as a result
+ * of a server-supplied `data.type`, otherwise a malicious or compromised
+ * rippled server could spoof connection state (e.g. claim the client is
+ * `connected` when it is not, or inject a fake `error`/`disconnected`/
+ * `reconnect` to drive the client into a bad state). See DGE-6740.
+ */
+const RESERVED_INTERNAL_EVENTS: ReadonlySet<string> = new Set([
+  'connected',
+  'disconnected',
+  'error',
+  'reconnect',
+])
+
+/**
  * ConnectionOptions is the configuration for the Connection class.
  */
 interface ConnectionOptions {
@@ -354,7 +369,21 @@ export class Connection extends EventEmitter {
     }
     if (data.type) {
       // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Should be true
-      this.emit(data.type as string, data)
+      const type = data.type as string
+      // Refuse to forward server-supplied event names that collide with
+      // Connection's internal state events. Otherwise a rogue server could
+      // spoof local connection state by sending e.g. `{"type":"connected"}`
+      // or `{"type":"error"}`. See DGE-6740.
+      if (RESERVED_INTERNAL_EVENTS.has(type)) {
+        this.emit(
+          'error',
+          'badMessage',
+          `Server message used reserved internal event type "${type}"; dropping.`,
+          message,
+        )
+        return
+      }
+      this.emit(type, data)
     }
     if (data.type === 'response') {
       try {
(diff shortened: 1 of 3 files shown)


packages/xrpl/src/client/connection.ts
```diff
@@ -22,6 +22,21 @@ const SECONDS_PER_MINUTE = 60
...
 
+/**
+ * Event names that are emitted internally by Connection to signal local
+ * client/socket state transitions. These must never be emitted as a result
+ * of a server-supplied `data.type`, otherwise a malicious or compromised
+ * rippled server could spoof connection state (e.g. claim the client is
+ * `connected` when it is not, or inject a fake `error`/`disconnected`/
+ * `reconnect` to drive the client into a bad state). See DGE-6740.
+ */
+const RESERVED_INTERNAL_EVENTS: ReadonlySet<string> = new Set([
```
kuan121, line 33: There is a `reconnecting` event as well -  https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/client/connection.ts#L459

packages/xrpl/src/client/connection.ts
```diff
@@ -354,7 +369,21 @@ export class Connection extends EventEmitter {
     }
     if (data.type) {
       // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Should be true
-      this.emit(data.type as string, data)
+      const type = data.type as string
+      // Refuse to forward server-supplied event names that collide with
+      // Connection's internal state events. Otherwise a rogue server could
+      // spoof local connection state by sending e.g. `{"type":"connected"}`
+      // or `{"type":"error"}`. See DGE-6740.
```
kuan121: nit: Let's remove `See DGE-6740.`

packages/xrpl/test/connection.test.ts
```diff
@@ -903,6 +903,97 @@ describe('Connection', function () {
...
+    TIMEOUT,
+  )
+
+  // DGE-6740: A spoofed `{"type":"error"}` payload must not be forwarded to
+  // 'error' listeners as the raw payload. Instead the fix emits an 'error'
+  // event whose first arg is the string code 'badMessage'. We verify the
+  // 'error' listener never receives the raw spoofed object.
+  it(
+    'drops server message with reserved internal event type "error"',
+    async () => {
```
kuan121: I was wondering why do we need a separate test for the `error` event? Can we combine both tests?

packages/xrpl/test/connection.test.ts
```diff
@@ -903,6 +903,97 @@ describe('Connection', function () {
...
+      clientContext.client.connection.onMessage(
+        JSON.stringify({ type: reservedType }),
+      )
+
+      await Promise.race([errorReceived, spoofTrigger])
+    },
+    TIMEOUT,
+  )
+
+  // DGE-6740: A spoofed `{"type":"error"}` payload must not be forwarded to
```
kuan121: nit: Let's remove `DGE-6740`.

packages/xrpl/test/connection.test.ts
```diff
@@ -903,6 +903,97 @@ describe('Connection', function () {
     TIMEOUT,
   )
 
+  // DGE-6740: Server messages whose `type` collides with a reserved internal
```
kuan121: nit: Let's remove `DGE-6740`.

packages/xrpl/test/connection.test.ts
```diff
@@ -903,6 +903,97 @@ describe('Connection', function () {
     TIMEOUT,
```
kuan121, line 903: browser tests are failing

packages/xrpl/src/client/connection.ts
```diff
@@ -354,7 +370,21 @@ export class Connection extends EventEmitter {
     }
     if (data.type) {
       // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Should be true
-      this.emit(data.type as string, data)
+      const type = data.type as string
```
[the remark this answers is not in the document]
kuan121: @achowdhry-ripple This sounds like a legit comment

ckeshava:
Continuing this work in a different PR: https://github.com/XRPLF/xrpl.js/pull/3378

PR #3341, opened 2026-05-18
fix(xrpl): return string from dropsToXrp to preserve precision (#3316)

### LINKED ISSUE #3316: dropsToXrp() returns number, causing precision loss for large amounts
## Description

The function returns `.toNumber()` (IEEE-754 double). For amounts approaching the total XRP supply (\~10^17 drops), precision loss occurs. `dropsToXrp('9999999999999999')` returns `9999999999.999998` instead of `9999999999.999999`. A round-trip `xrpToDrops(dropsToXrp(x))` **silently loses 1 drop**.

## Location

[packages/xrpl/src/utils/xrpConversion.ts#L18](https://github.com/XRPLF/xrpl.js/blob/5cf0921b0fa4637f324bcf5b0788e2ef59a83c8a/packages/xrpl/src/utils/xrpConversion.ts#L18), [#L53](https://github.com/XRPLF/xrpl.js/blob/5cf0921b0fa4637f324bcf5b0788e2ef59a83c8a/packages/xrpl/src/utils/xrpConversion.ts#L53)

## Severity

**Critical** -- Silent loss of funds on round-trip conversions.

## Adversarial Review Notes

The "Critical" severity is overstated -- it only affects amounts above \~9 billion XRP where sub-XRP precision matters. But the precision loss is real. **Unit test confirmed:** `xrpToDrops(dropsToXrp('9999999999999999'))` returns `"9999999999999998"`, losing 1 drop.

Confidence: 93%

## Unit Test Patch

```diff
--- a/packages/xrpl/test/utils/dropsToXrp.test.ts
+++ b/packages/xrpl/test/utils/dropsToXrp.test.ts
@@ -1,7 +1,7 @@
 import BigNumber from 'bignumber.js'
 import { assert } from 'chai'
 
-import { dropsToXrp } from '../../src/utils'
+import { dropsToXrp, xrpToDrops } from '../../src/utils'
 
 describe('dropsToXrp', function () {
@@ -130,4 +130,11 @@
+
+  it('BUG 5: round-trips large amounts without precision loss', function () {
+    const drops = '9999999999999999'
+    const xrp = dropsToXrp(drops)
+    const roundTripped = xrpToDrops(xrp)
+    assert.strictEqual(roundTripped, drops)
+  })
 })
```

achowdhry-ripple:

Fixes [xrpl.js#3316](https://github.com/XRPLF/xrpl.js/issues/3316).

- `dropsToXrp()` now returns a base-10 decimal `string` instead of a JavaScript `number`. The internal `.toNumber()` call is replaced with `.toString(BASE_TEN)`.
- `Client.getXrpBalance()` (which forwards the value) now returns `Promise<string>` instead of `Promise<number>`.
- `Client.getBalances()` is updated for the new return type (the branch that pushes the XRP balance no longer needs a redundant `.toString()`, and the `!== 0` sentinel check becomes `!== '0'`).
- `getBalanceChanges()` drops a now-redundant `.toString()` on the already-string XRP value.
- Existing `dropsToXrp` tests updated for the new return type; new regression test added per the issue:

  ```ts
  it('round-trips large amounts without precision loss (issue #3316)', function () {
    const drops = '9999999999999999'
    const xrp = dropsToXrp(drops)
    assert.strictEqual(xrp, '9999999999.999999')
    assert.strictEqual(xrpToDrops(xrp), drops)
  })
  ```
- `HISTORY.md` updated under `Unreleased > BREAKING CHANGES`.

### Context of Change

`dropsToXrp` converts a drops integer (≤ ~10^17 for the XRP total supply) into an XRP amount by dividing by 1,000,000. The old implementation called `.toNumber()` on the BigNumber result, coercing to IEEE-754 double. For amounts above ~9 billion XRP, the double cannot represent every drop exactly, so the conversion silently loses precision.

Repro from the issue:

```
dropsToXrp('9999999999999999')              // returned 9999999999.999998 (should be 9999999999.999999)
xrpToDrops(dropsToXrp('9999999999999999'))  // returned '9999999999999998' — 1 drop silently lost on round-trip
```

The fix is to return the BigNumber's exact base-10 string representation. Drops are at most 6 fractional digits of XRP, so the division terminates exactly within BigNumber's default precision and never produces exponential notation.

This is a breaking API change (return type `number` → `string`), which is why it is grouped with the other `Unreleased` breaking changes in `HISTORY.md`. Callers that rely on a JS number can wrap the result in `Number(...)` — this is already done at the two internal call sites in `Wallet/fundWallet.ts`.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

- New unit test `dropsToXrp > round-trips large amounts without precision loss (issue #3316)` directly exercises the regression case from the issue.
- Existing `dropsToXrp` tests were updated to assert against the new string return type (mechanical change — same values, now quoted).
- `getXrpBalance` mock test updated to assert the string return.
- `getBalanceChanges` fixture-driven tests are unchanged (the fixture already expected string values; the new code path produces the same strings).
- Did not re-run the full suite locally — `npm install` is currently hitting the npm "Exit handler never called!" bug on my machine. Relying on CI; happy to push fixups if anything fails.


packages/xrpl/src/client/index.ts
```diff
@@ -936,15 +936,15 @@ class Client extends EventEmitter<EventTypes> {
    * @param [options] - Additional options for fetching the balance (optional).
    * @param [options.ledger_hash] - The hash of the ledger to retrieve the balance from (optional).
    * @param [options.ledger_index] - The index of the ledger to retrieve the balance from (optional).
-   * @returns A promise that resolves with the XRP balance as a number.
+   * @returns A promise that resolves with the XRP balance as a base-10 decimal string.
```
kuan121: Should we return `BigNumber` instead?

We will have the same breaking-change cost (every caller updates either way), but I think it ends up nicer:
- `dropsToXrp` already takes `BigNumber.Value` as input, so returning `BigNumber` keeps things symmetric and doesn't add anything new to the public API surface.
- No string-concat surprise. With `string`, `dropsToXrp(a) + dropsToXrp(b)` silently concatenates. With `BigNumber` you get `.plus()` and TS yells at you for using `+`.

packages/xrpl/src/client/index.ts
```diff
@@ -1043,7 +1046,7 @@ class Client extends EventEmitter<EventTypes> {
         const accountLinesBalance = linesResponses.flatMap((response) =>
           formatBalances(response.result.lines),
         )
-        if (xrpBalance !== 0) {
+        if (!xrpBalance.isZero()) {
```
cybele-ripple, line 1049: We should add test coverage for this change as well to be thorough

packages/xrpl/src/client/index.ts
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable jsdoc/require-jsdoc -- Request has many aliases, but they don't need unique docs */
```
kuan121, line 1: Can you update the PR title and description?

kuan121:
Browser tests are still failing.

packages/xrpl/src/client/index.ts
achowdhry-ripple: Done — switched to `BigNumber` in ceb12d7. Agreed on the symmetry argument, and the type system catching `+` is a nice bonus.

- `dropsToXrp()` returns `BigNumber`; `Client.getXrpBalance()` returns `Promise<BigNumber>`.
- `getBalances()` now uses `.isZero()` for the sentinel and `.toString()` when pushing onto the string-typed `Balance.value`.
- The two internal `fundWallet` call sites that wanted a number switched from `Number(...)` to `.toNumber()`.
- `HISTORY.md` breaking-change entry updated.

Full unit suite (1039 tests) + lint pass locally. The failing browser/integration jobs on this PR look like the same pre-existing CI infra issue affecting other PRs (e.g. #3340) — rippled docker container failing to start with ECONNREFUSED on 6006, unrelated to this change.

PR #3342, opened 2026-05-21
fix(xrpl): infer signing algorithm from seed prefix in Wallet.fromSeed / fromSecret

achowdhry-ripple:
## Summary

Independent fix for the bug also tracked in [#3337](https://github.com/XRPLF/xrpl.js/pull/3337). Filed under [DGE-7133](https://ripplelabs.atlassian.net/browse/DGE-7133).

`Wallet.deriveWallet` no longer applies an `ed25519` fallback when `opts.algorithm` is omitted. The `undefined` value now flows through to `ripple-keypairs.deriveKeypair`, which infers the algorithm from the seed prefix via `decodeSeed`.

### Bug

`Wallet.fromSeed(seed)` (and its alias `Wallet.fromSecret`) silently coerced an omitted `opts.algorithm` to `ed25519`. For `sEd…` family seeds this happened to produce the right account. For an `s…` family seed (secp256k1), it produced a valid keypair on a completely different account — making it look like the caller's account had "disappeared." This has been one of the most commonly reported migration issues since the 3.x default flip.

### After this PR

| Method                              | Algorithm when `opts.algorithm` is omitted                                                                            |
| ----------------------------------- | --------------------------------------------------------------------------------------------------------------------- |
| `Wallet.generate()`                 | `ed25519` (unchanged — passed explicitly)                                                                             |
| `Wallet.fromSeed(seed)`             | inferred from prefix: `sEd…` → `ed25519`, otherwise `secp256k1`                                                       |
| `Wallet.fromSecret(secret)`         | inferred from prefix: `sEd…` → `ed25519`, otherwise `secp256k1` (alias of `fromSeed`)                                 |
| `Wallet.fromEntropy(entropy)`       | `ed25519` (unchanged — passed explicitly)                                                                             |
| `Wallet.fromMnemonic(...)`          | unchanged                                                                                                             |
| `walletFromSecretNumbers(numbers)`  | `secp256k1` (unchanged)                                                                                               |

### Migration

Only `Wallet.fromSeed` and `Wallet.fromSecret` change behavior.

| Call site                                                                                                | Before                                                          | After                                                          | Required action                                                                                                                                                                                  |
| -------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------- | -------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `Wallet.fromSeed(seed)` / `Wallet.fromSecret(seed)`, `seed` does **not** start with `sEd`, no `algorithm` | returned an `ed25519` keypair for an unrelated account          | returns the `secp256k1` keypair for the account the seed encodes | If you intended `secp256k1`: **none** — you now get the correct keypair. If your application persisted the previously-derived `ed25519` keypair and must keep using it: pass `{ algorithm: ECDSA.ed25519 }` explicitly. |
| `Wallet.fromSeed(seed)` / `Wallet.fromSecret(seed)`, `seed` starts with `sEd`, no `algorithm`            | returned an `ed25519` keypair                                   | returns an `ed25519` keypair (via prefix inference)            | **None** — result is identical.                                                                                                                                                                  |
| `Wallet.fromSeed` / `Wallet.fromSecret` with an explicit `opts.algorithm`                                | honored the explicit value                                      | honors the explicit value                                      | **None.**                                                                                                                                                                                        |
| `Wallet.generate`, `Wallet.fromEntropy`, `Wallet.fromMnemonic`, `walletFromSecretNumbers`                | unchanged                                                       | unchanged                                                      | **None.**                                                                                                                                                                                        |

### Tests

- The two `derives a wallet using default algorithm` cases (`fromSeed`, `fromSecret`) were locking in the buggy behavior — they asserted `ed25519` keys against the secp256k1 fixture seed `sh1HiK7SwjS1VxFdXi7qeMHRedrYX`. Renamed to `infers secp256k1 from a non-sEd seed prefix when algorithm is omitted` and updated to assert the correct `secp256k1` keys.
- Added sibling cases that ingest `sEdVaw4m9W3H3ou3VnyvDwvPAP5BEz1` and assert the inferred result equals an explicit `{ algorithm: ECDSA.ed25519 }` call.
- HISTORY.md gets a BREAKING CHANGES entry under `## Unreleased`.

## Test plan

- [x] `packages/xrpl` unit suite: **117 suites / 1040 tests pass** (`jest --config=jest.config.unit.js`)
- [ ] Reviewer sanity check: in a repo of your choice, `Wallet.fromSeed('sh1HiK7SwjS1VxFdXi7qeMHRedrYX')` should return classic address derived from `03BFC2F7AE242C3493187FA0B72BE97B2DF71194FB772E507FF9DEA0AD13CA1625` (secp256k1), not the previous ed25519 result.
- [ ] CI green on the upstream PR.


mvadari:
Note: this needs to be a major version bump

PR #3364, opened 2026-06-05
Support for Confidential MPT (XLS-0096)

kuan121:

This PR adds support for **Confidential Multi-Purpose Tokens ([XLS-0096](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0096-confidential-mpt))**.

Confidential MPT represents an MPT balance on-ledger as EC-ElGamal ciphertexts and uses zero-knowledge proofs, allowing validators to verify transfers (no overdraft, conservation of value) without learning the amounts. This change adds the wire-format definitions, the transaction and ledger models, field validation, fee handling, and a high-level builder API that encapsulates the cryptography. All changes are additive and do not alter existing behavior.

Components:

- **`@xrplf/mpt-crypto` (new package).** A hex-in/hex-out TypeScript API over a vendored WebAssembly build of the reference C crypto library, providing ElGamal encryption, Pedersen commitments, blinding factors, context hashes, and the four proof builders (convert, convert-back, send, clawback). It is a **regular, version-pinned dependency** of `xrpl`, reached only through a lazy dynamic `import()` so bundlers code-split its ~2 MB WASM out of `xrpl`'s main entry. It ships as a first-class published package (with its own `README.md` and `HISTORY.md`); the release pipeline vendors the matching WASM before publishing it to npm.
- **`ripple-binary-codec`.** Definitions for the five new transaction types (`ConfidentialMPTConvert`, `ConfidentialMPTMergeInbox`, `ConfidentialMPTConvertBack`, `ConfidentialMPTSend`, `ConfidentialMPTClawback`), their fields, and the `tecBAD_PROOF` and `temBAD_CIPHERTEXT` result codes.
- **Transaction models (`xrpl`).** The five `ConfidentialMPT*` interfaces and `validate*` functions, the `tfMPTCanHoldConfidentialBalance` flag on `MPTokenIssuanceCreate`, and the `IssuerEncryptionKey` and `AuditorEncryptionKey` fields on `MPTokenIssuanceSet`.
- **Ledger models (`xrpl`).** Confidential fields on `MPToken` (inbox, spending, and version balances; issuer- and auditor-encrypted balances; holder encryption key) and on `MPTokenIssuance` (issuer and auditor keys, outstanding amount).
- **Builder API (`xrpl/confidential`).** The `prepareConfidential*` builders, which read the required ledger state, generate the ciphertexts, commitments, and proofs, and return an unsigned transaction. `prepareConfidentialConvert` derives whether to register the holder encryption key from ledger state (registering on the first convert, omitting it on later ones). `getConfidentialBalance` decrypts a holder's spendable balance.
- **Fees (`autofill`).** Confidential transactions carry a surcharge; the autofilled fee is `base × (10 + signerCount)`, matching rippled. `autofill` fills `Fee` / `LastLedgerSequence` but leaves the proof-pinned `Sequence` untouched.
- **Release pipeline.** `@xrplf/mpt-crypto` is published through the standard `workflow_dispatch` release workflow (`.github/workflows/release.yml`). `packages/mpt-crypto/MPT_CRYPTO_VERSION` pins the upstream `mpt-crypto` release, and `scripts/fetch-wasm.sh` downloads that release's WASM bundle and vendors it into `wasm/` before packing.

### Context of Change

This implements the client side of XLS-0096, tracking rippled PR [#5860](https://github.com/XRPLF/rippled/pull/5860) and the xrpl-py reference implementation (branch `confidential-mpt`).

Key design decisions:

- **Cryptography is isolated in a lazily-loaded package.** The proof system requires a large WebAssembly library, so it lives in `@xrplf/mpt-crypto`, reached only through a dynamic import from the `xrpl/confidential` subpath. It is a **required, version-pinned dependency** — like `ripple-binary-codec`, the cryptography is protocol-coupled, and Confidential MPT is shipping to mainnet — but because it is loaded lazily, bundlers code-split its ~2 MB WASM into a separate chunk, so apps that never assemble a Confidential MPT transaction never pay its load cost.
- **Encryption keys reuse the existing key infrastructure.** The ElGamal encryption key is a distinct secp256k1 keypair, derived with `ripple-keypairs`. No new key type is introduced, the key is recoverable from a seed, and the approach is consistent with the Java SDK (xrpl4j).
- **Builders return sequence-pinned transactions.** Each proof's context hash is bound to the transaction's identity (account, issuance, sequence, destination, balance version). The builders therefore pin `Sequence`, which must not be changed before submission (`autofill` respects it). Because the proof binds to the sequence value, `TicketSequence` is not supported by the builders.
- **The vendored WASM is version-pinned and vendored at release time.** `packages/mpt-crypto/MPT_CRYPTO_VERSION` pins the `mpt-crypto` release; `scripts/fetch-wasm.sh` downloads that release's WASM bundle and vendors `mpt_crypto.{js,mjs,wasm}` into `wasm/`, and the release pipeline runs it before publishing — so the published package always carries the WASM built from the pinned tag. This must match the `mpt-crypto` version rippled pins; a mismatch results in `tecBAD_PROOF`.

Release prerequisites (for maintainers):

- The pinned `mpt-crypto` release must attach the WASM bundle asset (`mpt-crypto-wasm-<tag>.tar.gz`).
- `@xrplf/mpt-crypto` must be published to npm **before** any `xrpl` release that depends on it.

### Type of Change

- [x] Yes

> Added Confidential MPT entries to the `Unreleased` sections of `packages/xrpl/HISTORY.md` and `packages/ripple-binary-codec/HISTORY.md`, and created `packages/mpt-crypto/HISTORY.md` for the new package.

## Test Plan

**Unit, codec, and fee tests** — run in CI with no additional setup (`npm test`):

- `packages/mpt-crypto/test/*.test.ts`: the crypto primitives (ElGamal encrypt/decrypt, Pedersen commitments, blinding factors), context hashes, the proof builders (including the multi-participant send proof), and the hex / WASM-marshalling helpers.
- `packages/xrpl/test/confidential/*.test.ts`: the `prepareConfidential*` builders, `getConfidentialBalance`, key derivation, and the lazy loader — with a mocked client. Includes Convert's ledger-state-derived key registration (register on the first convert, omit on later ones).
- `packages/xrpl/test/models/ConfidentialMPT*.test.ts` and the `MPTokenIssuanceCreate` / `MPTokenIssuanceSet` additions: each `validate*` function accepts well-formed transactions and rejects malformed fields.
- `packages/ripple-binary-codec/test/confidential-mpt.test.ts`: encode/decode round-trips for the new transaction types and fields.
- `packages/xrpl/test/client/autofill.test.ts`: the confidential fee calculation.

**Integration tests** — `packages/xrpl/test/integration/transactions/confidentialMPT*.test.ts`: one file per transaction type plus a four-party lifecycle file, sharing `confidentialMPTUtils.ts`. These require a rippled with the MPTokensV1, Clawback, and ConfidentialTransfer amendments enabled. They run as part of `npm run test:integration` against the `rippleci/xrpld` develop image once #5860 is merged, and were verified against a local standalone rippled built from #5860: all five transaction types and the full four-party lifecycle (convert → merge → send → convert-back → clawback, with auditor disclosure) pass. The Convert file additionally covers subsequent (top-up) converts on an already-registered holder.


packages/xrpl/src/confidential/convert.ts
```diff
@@ -0,0 +1,207 @@
...
+  }
+  if (issuance.AuditorEncryptionKey != null) {
+    tx.AuditorEncryptedAmount = await crypto.encryptAmount(
+      amount,
+      issuance.AuditorEncryptionKey,
+      blindingFactor,
+    )
+  }
+  if (params.registerKey ?? true) {
+    tx.HolderEncryptionKey = holder.publicKey
```
florent-uzio: Should we move the `ZKProof` field here @kuan121 ?

I can read this in the docs:

```
A Schnorr Proof of Knowledge. Required only when HolderEncryptionKey is present.
```

When I do a test locally, my second `ConfidentialMPTConvert` transaction without `HolderEncryptionKey` but with `ZKProof` fails with `temMALFORMED`

According to the [docs](https://opensource.ripple.com/docs/xls-96-confidential-transfers/references/transactions/confidentialmptconvert#error-cases):

```
The transaction is malformed for one of the following reasons:
HolderEncryptionKey is provided but ZKProof is not.
HolderEncryptionKey is not provided but ZKProof is.
```

florent-uzio: I just did a test without `HolderEncryptionKey` and `ZKProof` for a second `ConfidentialMPTConvert` and it worked.

So it should be:

```
const tx: ConfidentialMPTConvert = {
    TransactionType: 'ConfidentialMPTConvert',
    Account: params.account,
    Sequence: sequence,
    MPTokenIssuanceID: params.mptIssuanceID,
    MPTAmount: amount.toString(),
    HolderEncryptedAmount: holderEncryptedAmount,
    IssuerEncryptedAmount: issuerEncryptedAmount,
    BlindingFactor: blindingFactor,
    // removed ZKProof
  }
  if (issuance.AuditorEncryptionKey != null) {
    tx.AuditorEncryptedAmount = await crypto.encryptAmount(
      amount,
      issuance.AuditorEncryptionKey,
      blindingFactor,
    )
  }

if (params.registerKey ?? true) {
    tx.HolderEncryptionKey = holder.publicKey
	tx.ZKProof = zkProof
}
```

kuan121: @florent-uzio Good catch! You are right. I addressed your comment in https://github.com/XRPLF/xrpl.js/pull/3364/changes/14fd7495121012b6b13fc2b56dd15d1a7bbe391d

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
```diff
@@ -38,6 +40,11 @@ export enum MPTokenIssuanceSetFlags {
    * If set, indicates that issuer unlocks the MPT
    */
   tfMPTUnlock = 0x00000002,
+  /**
+   * Enables the lsfMPTCanHoldConfidentialBalance flag on the issuance. Allows
+   * holders to hold confidential (encrypted) balances of this token.
+   */
+  tfMPTSetCanHoldConfidentialBalance = 0x00000100,
```
florent-uzio, line 75: @kuan121 should `tfMPTSetCanHoldConfidentialBalance` be part of the mutable flags instead?
https://xls.xrpl.org/xls/XLS-0096-confidential-mpt.html#12-transaction-mptokenissuanceset 
section 12.2.1

kuan121: @florent-uzio As part of the latest Dynamic MPT update, we no longer have mutable flags. See this [PR](https://github.com/XRPLF/XRPL-Standards/pull/583) to update the XLS spec for Dynamic MPT. I included only `tfMPTSetCanHoldConfidentialBalance` here since it's relevant to Confidential MPT.

packages/mpt-crypto/src/primitives.ts
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * ElGamal-encrypt an amount under a public key.
+ *
+ * @param amount - The integer amount to encrypt.
+ * @param publicKey - The 33-byte hex public key.
+ * @param blindingFactor - The 32-byte hex randomness scalar.
+ * @returns The 66-byte hex ciphertext (C1 || C2).
+ * @throws If inputs are malformed or the WASM call fails.
+ */
+export async function encryptAmount(
```
pdp2121, line 48: `encryptAmount`/`getPedersenCommitment` (and `proofs.ts` amount args) pass the bigint straight to a WASM i64 param — the JS API does `ToBigInt64` (wraps), it doesn't throw. So `encryptAmount(2n**64n, …)` silently encrypts 0. Minor but worth adding a check

packages/mpt-crypto/src/primitives.ts
```diff
@@ -0,0 +1,131 @@
...
+ * @param privateKey - The 32-byte hex private key.
+ * @param rangeHigh - Inclusive upper bound for the search; must be < 2^64 - 1.
+ * @returns The decrypted integer amount.
+ * @throws If inputs are malformed, the amount is outside [0, rangeHigh], or the
+ * WASM call fails.
+ */
+export async function decryptAmount(
+  ciphertext: string,
+  privateKey: string,
+  rangeHigh: bigint,
```
pdp2121, line 91: Should we add a default and bound check for `rangeHigh`?

packages/mpt-crypto/src/constants.ts
```diff
@@ -0,0 +1,52 @@
...
+/** A full ElGamal ciphertext (C1 || C2). */
+export const ELGAMAL_TOTAL_SIZE = 66
+/** A Pedersen commitment point. */
+export const PEDERSEN_COMMIT_SIZE = 33
+/** The 32-byte transaction context hash (challenge) consumed by the proofs. */
+export const CONTEXT_HASH_SIZE = 32
+/** 20-byte XRPL AccountID. */
+export const ACCOUNT_ID_SIZE = 20
+/** 24-byte MPTokenIssuanceID. */
+export const ISSUANCE_ID_SIZE = 24
```
pdp2121, line 22: nit: should we add a length check validations for models using `MPTokenIssuanceID`?

packages/mpt-crypto/src/module.ts
```diff
@@ -0,0 +1,140 @@
...
+ * serves both builds: `require`/Jest get the CJS glue, bundlers/browsers the ESM
+ * glue (whose `new URL(import.meta.url)` lets them emit the wasm as an asset). See
+ * the `package.json` exports and `src/wasm.d.ts`.
+ *
+ * Provenance: `mpt_crypto.{js,mjs,wasm}` come from mpt-crypto's
+ * `.github/scripts/build-wasm.sh` and are vendored by `scripts/fetch-wasm.sh`.
+ *
+ * @returns A promise resolving to the initialized WASM module.
+ */
+export async function loadWasmModule(): Promise<WasmModule> {
```
pdp2121, line 214: nit: should we add a retry mechanism here in case of a transient load failure?

pdp2121:
Looks good to me so far, just some nit comments

packages/mpt-crypto/src/constants.ts
kuan121: I think it's OK since the length will also be checked during serialization on the binary codec side

packages/mpt-crypto/src/module.ts
kuan121: Good point. I added some retry logic in https://github.com/XRPLF/xrpl.js/pull/3364/changes/0b92ab7131beec21be5b92de8bab5d4c31fc5300

packages/mpt-crypto/src/primitives.ts
kuan121: Good point. Addressed in https://github.com/XRPLF/xrpl.js/pull/3364/changes/067443eecb2218d08c47252769b77815c43b0e16

packages/mpt-crypto/src/primitives.ts
kuan121: Good point. Addressed in https://github.com/XRPLF/xrpl.js/pull/3364/changes/067443eecb2218d08c47252769b77815c43b0e16

packages/mpt-crypto/src/proofs.ts
```diff
@@ -0,0 +1,200 @@
...
+        amountCommitmentPtr,
+        balancePtr,
+        outPtr,
+        outLenPtr,
+      ) !== 0
+    ) {
+      throw new Error('mpt_get_confidential_send_proof failed')
+    }
+    return bytesToHex(
+      marshaller.readBytes(outPtr, marshaller.readU32(outLenPtr)),
```
cybele-ripple: Maybe include a check for the `outLenPtr` before `return bytesToHex(...)` like below:
```
const outLen = marshaller.readU32(outLenPtr)
if (outLen > SEND_PROOF_SIZE) {
  throw new Error('send proof output exceeds allocated buffer')
}
```

packages/mpt-crypto/src/proofs.ts
```diff
@@ -0,0 +1,200 @@
...
+    const outPtr = marshaller.alloc(SEND_PROOF_SIZE)
+    const outLenPtr = marshaller.alloc(SIZE_T_BYTES)
+    marshaller.writeU32(outLenPtr, SEND_PROOF_SIZE)
+    if (
+      mod._mpt_get_confidential_send_proof(
+        privPtr,
+        pubPtr,
+        params.amount,
+        participantsPtr,
+        participants.length,
```
cybele-ripple, line 202: Should we check the value of `participants.length` first?

packages/mpt-crypto/src/context.ts
```diff
@@ -0,0 +1,155 @@
...
+  sequence: number,
+): Promise<string> {
+  const acc = hexToBytes(account, 'account', ACCOUNT_ID_SIZE)
+  const iss = hexToBytes(issuance, 'issuance', ISSUANCE_ID_SIZE)
+  return withModule((mod, marshaller) => {
+    const accPtr = marshaller.allocBytes(acc)
+    const issPtr = marshaller.allocBytes(iss)
+    const outPtr = marshaller.alloc(CONTEXT_HASH_SIZE)
+    if (
+      mod._mpt_get_convert_context_hash(accPtr, issPtr, sequence, outPtr) !== 0
```
cybele-ripple, line 33: Should there be some type of check for these values like what is done with account [here](https://github.com/XRPLF/xrpl.js/pull/3364/changes#diff-48b17e5a9122c3eaab945714331add4c1c15c1c30ad06990b6026b2ceb331567R68)?

packages/mpt-crypto/src/primitives.ts
```diff
@@ -0,0 +1,139 @@
...
+): Promise<bigint> {
+  assertUint64(rangeHigh, 'rangeHigh', MAX_DECRYPT_RANGE_HIGH)
+  const ct = hexToBytes(ciphertext, 'ciphertext', ELGAMAL_TOTAL_SIZE)
+  const priv = hexToBytes(privateKey, 'privateKey', PRIVKEY_SIZE)
```
cybele-ripple, line 95: Does this need to be cleared after it's used like you go [here](https://github.com/XRPLF/xrpl.js/pull/3364/changes#diff-70c1531f5263358986b6246b97179a23ad2c3a1b3c89e1260bb6b51eb55471c3R141-R148)?

package-lock.json
```diff
@@ -19271,6 +19283,14 @@
...
       "engines": {
         "node": ">=20.19.0"
+      },
+      "peerDependencies": {
+        "@xrplf/mpt-crypto": "^0.1.0"
+      },
+      "peerDependenciesMeta": {
+        "@xrplf/mpt-crypto": {
+          "optional": true
+        }
```
cybele-ripple: You might have to regenerate these files. I think this entry should not include `"optional": true`

packages/mpt-crypto/scripts/fetch-wasm.sh
```diff
@@ -0,0 +1,130 @@
...
+info "Fetching ${ASSET} from ${REPO}@${TAG} (via ${DOWNLOADER})..."
+if [[ "${DOWNLOADER}" == "gh" ]]; then
+  if ! assets="$(gh release view "${TAG}" --repo "${REPO}" --json assets --jq '.assets[].name' 2>&1)"; then
+    die "release '${TAG}' not found in ${REPO} (or a gh/auth error). gh reported: ${assets}"
+  fi
+  if ! grep -qxF "${ASSET}" <<<"${assets}"; then
+    err "${REPO}@${TAG} exists but does not attach '${ASSET}'."
+    err "The mpt-crypto release must include the WASM bundle (see the mpt-crypto build-wasm workflow)."
+    err "Assets present on ${TAG}:"
+    printf '  - %s\n' ${assets:-"(none)"} >&2
```
cybele-ripple, line 73: Might need to change this to
`printf '  - %s\n' "${assets:-(none)}"`

packages/xrpl/src/confidential/convert.ts
```diff
@@ -0,0 +1,232 @@
...
+  client: Client,
+  params: ConfidentialConvertParams,
+): Promise<ConfidentialMPTConvert> {
+  // Convert may register the holder key with a zero amount, so zero is allowed.
+  assertConfidentialAmount(params.amount, true)
+  const [crypto, issuance, mptoken, sequence] = await Promise.all([
+    loadMptCrypto(),
+    fetchMPTokenIssuance(client, params.mptIssuanceID),
+    fetchMPToken(client, params.account, params.mptIssuanceID),
+    resolveSequence(client, params.account, params.sequence),
```
cybele-ripple, line 53: These are 3 separate, unpinned requests with no `ledger_index`. Would there be an issue if these requests go to different ledgers? Maybe consider doing something like [this](https://github.com/XRPLF/xrpl.js/pull/3364/changes#diff-2cf3f4affc52fb31a782bcff6166266cb2e261b72660da92c36cedc3c5666a33L227)

packages/xrpl/src/confidential/transfer.ts
```diff
@@ -0,0 +1,220 @@
...
+  // would otherwise accept anything up to 2^64 - 1.
+  assertConfidentialAmount(params.amount, true)
+  const [crypto, issuance, senderToken, destToken, sequence] =
+    await Promise.all([
+      loadMptCrypto(),
+      fetchMPTokenIssuance(client, params.mptIssuanceID),
+      fetchMPToken(client, params.account, params.mptIssuanceID),
+      fetchMPToken(client, params.destination, params.mptIssuanceID),
+      resolveSequence(client, params.account, params.sequence),
+    ])
```
cybele-ripple: Same concern as [this comment](https://github.com/XRPLF/xrpl.js/pull/3364/changes#r3779943322)

packages/xrpl/src/models/transactions/common.ts
```diff
@@ -446,6 +480,31 @@ export function validateOptionalField<
...
+ * @param tx - The transaction to validate.
+ * @param allowZero - Whether a zero amount is permitted.
+ * @throws ValidationError if MPTAmount is missing, malformed, or out of range.
+ */
+export function validateConfidentialMPTAmount(
+  tx: Record<string, unknown>,
+  allowZero: boolean,
+): void {
+  validateRequiredField(tx, 'MPTAmount', isString)
+  if (typeof tx.MPTAmount === 'string') {
```
cybele-ripple: nit: I don't think that this line can ever be false due to the line [immediately preceding this](https://github.com/XRPLF/xrpl.js/pull/3364/changes#diff-695e6507bbf3a86833c7318c48da6ec25fe12c7a4f8a90a3a0cfbb8c19f8b3abR496)

packages/xrpl/src/models/transactions/ConfidentialMPTConvert.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+}
+
+/**
+ * Verify the form and type of a ConfidentialMPTConvert at runtime.
+ *
+ * @param tx - A ConfidentialMPTConvert Transaction.
+ * @throws When the ConfidentialMPTConvert is malformed.
+ */
+export function validateConfidentialMPTConvert(
```
pdp2121, line 88: minor and apply to other txn as well: verify `account == issuer` case

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
```diff
@@ -135,6 +137,16 @@ export interface MPTokenIssuanceSet extends BaseTransaction {
    * If omitted, this transaction will apply to all any accounts holding MPTs.
    */
   Holder?: Account
```
pdp2121, line 139: missing the `Holder` + `IssuerEncryptionKey` incompatibility validation, which will be rejected in rippled `temMalformed`

packages/xrpl/HISTORY.md
```diff
@@ -10,6 +10,7 @@ Subscribe to [the **xrpl-announce** mailing list](https://groups.google.com/g/xr
 * Support Dynamic MPTs (XLS-94d) based on latest spec change [XRPL-Standards#583](https://github.com/XRPLF/XRPL-Standards/pull/583).
 * Add `ReferenceHolding` to `MPTokenIssuance` ledger object and `vault_info` response.
 * Add XLS-56 Batch V1_1 support to `signMultiBatch` and `combineBatchSigners` ([XRPLF/rippled#6446](https://github.com/XRPLF/rippled/pull/6446)).
+* Add support for Confidential Transfers for Multi-Purpose Tokens (XLS-96).
```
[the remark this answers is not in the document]
kuan121: Will do.

packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts
kuan121: Good point. Addressed in [43de1bf](https://github.com/XRPLF/xrpl.js/blob/43de1bfc332820ffea25ef9daa0060b779c96c31/packages/xrpl/src/models/transactions/MPTokenIssuanceSet.ts#L278)

packages/xrpl/src/models/transactions/common.ts
kuan121: Good point. I removed it in

packages/mpt-crypto/scripts/fetch-wasm.sh
kuan121: Unquoted, $assets word-splits so `printf` prints one bullet per asset; quoting makes it a single arg, so only the first is bulleted. For comparison:
### current (unquoted):
  \- mpt-crypto-natives-1.0.3.tar.gz
  \- mpt-crypto-wasm-1.0.3.tar.gz

### quoted (suggested):
  \- mpt-crypto-natives-1.0.3.tar.gz
mpt-crypto-wasm-1.0.3.tar.gz

packages/xrpl/src/models/transactions/ConfidentialMPTConvert.ts
kuan121: Good point. It's addressed in [43de1bf](https://github.com/XRPLF/xrpl.js/blob/43de1bfc332820ffea25ef9daa0060b779c96c31/packages/xrpl/src/models/transactions/ConfidentialMPTConvert.ts#L93), and applied across the others — ConvertBack, Send (account + destination), MergeInbox, and the inverse check on Clawback.

packages/mpt-crypto/src/primitives.ts
```diff
@@ -0,0 +1,145 @@
...
+  return withModule((mod, marshaller) => {
+    const ctPtr = marshaller.allocBytes(ct)
+    const privPtr = marshaller.allocBytes(priv)
+    // Wipe the transient JS copy; WASM scratch is zeroed on dispose().
+    priv.fill(0)
```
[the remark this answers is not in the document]
kuan121: Will not address since this is a low severity issue

packages/xrpl/src/confidential/ledger.ts
```diff
@@ -0,0 +1,284 @@
...
+  client: Client,
+  account: string,
+  mptIssuanceID: string,
+  privateKey: string,
+): Promise<bigint> {
+  // Pin both reads to one ledger so the decrypt bound (from the issuance's
+  // outstanding total) can never come from an earlier ledger than the balance
+  // it must cover, which would make the bound too small and the decrypt fail.
+  const ledgerIndex = await client.getLedgerIndex()
+  const mptoken = await fetchMPToken(
```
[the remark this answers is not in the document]
kuan121: Addressed in https://github.com/XRPLF/xrpl.js/pull/3364/commits/b8df6345e82684241d1ad2ef4a73e9eaa2b38698

packages/xrpl/src/confidential/convert.ts
kuan121: Good point. It's addressed in [43de1bf](https://github.com/XRPLF/xrpl.js/blob/43de1bfc332820ffea25ef9daa0060b779c96c31/packages/xrpl/src/confidential/convert.ts#L48) — the reads now pin a single ledgerIndex resolved once up front.

packages/mpt-crypto/src/primitives.ts
kuan121: Good catch. It's addressed in [43de1bf](https://github.com/XRPLF/xrpl.js/blob/43de1bfc332820ffea25ef9daa0060b779c96c31/packages/mpt-crypto/src/primitives.ts#L98) — the transient key copy is zeroed right after it's handed to WASM.

packages/mpt-crypto/src/context.ts
kuan121: What values do you suggest that we perform the check?

pdp2121 (approved):

cybele-ripple (approved):

PR #3378, opened 2026-06-22
fix(connection): block reserved internal event types from server messages

### LINKED ISSUE #3318: Server-controlled event injection via unvalidated data.type

ckeshava:
Blocks server-supplied `data.type` values that collide with `Connection`'s reserved internal events (`connected`, `disconnected`, `error`, `reconnect`, `reconnecting`). Such messages are dropped and surfaced as `badMessage`; unknown types are still forwarded so forward-compatibility is preserved.

Continues the work in #3340.

Fixes #3318

### DIFF
--- a/packages/xrpl/src/client/connection.ts
+++ b/packages/xrpl/src/client/connection.ts
@@ -28,13 +28,14 @@ const CONNECTION_TIMEOUT = 5
  * of a server-supplied `data.type`, otherwise a malicious or compromised
  * rippled server could spoof connection state (e.g. claim the client is
  * `connected` when it is not, or inject a fake `error`/`disconnected`/
- * `reconnect` to drive the client into a bad state). See DGE-6740.
+ * `reconnect`/`reconnecting` to drive the client into a bad state).
  */
 const RESERVED_INTERNAL_EVENTS: ReadonlySet<string> = new Set([
   'connected',
   'disconnected',
   'error',
   'reconnect',
+  'reconnecting',
 ])
 
 /**
@@ -373,7 +374,7 @@ export class Connection extends EventEmitter {
       // Refuse to forward server-supplied event names that collide with
       // Connection's internal state events. Otherwise a rogue server could
       // spoof local connection state by sending e.g. `{"type":"connected"}`
-      // or `{"type":"error"}`. See DGE-6740.
+      // or `{"type":"error"}`.
       if (RESERVED_INTERNAL_EVENTS.has(type)) {
         this.emit(
           'error',
--- a/packages/xrpl/src/client/connection.ts
+++ b/packages/xrpl/src/client/connection.ts
@@ -369,22 +369,7 @@ export class Connection extends EventEmitter {
       return
     }
     if (data.type) {
-      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- Should be true
-      const type = data.type as string
-      // Refuse to forward server-supplied event names that collide with
-      // Connection's internal state events. Otherwise a rogue server could
-      // spoof local connection state by sending e.g. `{"type":"connected"}`
-      // or `{"type":"error"}`.
-      if (RESERVED_INTERNAL_EVENTS.has(type)) {
-        this.emit(
-          'error',
-          'badMessage',
-          `Server message used reserved internal event type "${type}"; dropping.`,
-          message,
-        )
-        return
-      }
-      this.emit(type, data)
+      this.forwardServerMessage(data, message)
     }
     if (data.type === 'response') {
       try {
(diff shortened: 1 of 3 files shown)


packages/xrpl/src/client/connection.ts
```diff
@@ -354,7 +370,21 @@ export class Connection extends EventEmitter {
...
+      if (RESERVED_INTERNAL_EVENTS.has(type)) {
+        this.emit(
+          'error',
+          'badMessage',
+          `Server message used reserved internal event type "${type}"; dropping.`,
+          message,
+        )
+        return
+      }
+      this.emit(type, data)
```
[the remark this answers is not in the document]
ckeshava: Good catch — confirmed and addressed in 5dfb28d42.

`{"type":["error"]}` failed the `RESERVED_INTERNAL_EVENTS.has()` check (an array is never `===` a string in the Set) but still coerced to the string `"error"` when passed to `emit`, re-opening the spoofing hole. `onMessage` now validates `typeof data.type === 'string'` and drops non-string types as `badMessage` *before* the reserved-event check. The forwarding logic was extracted into a `forwardServerMessage` helper to stay within the lint complexity/length limits.

Added parameterized tests covering the coercion vectors (`["error"]`, `["connected"]`, `[["reconnect"]]`, …); each verifies the payload is never forwarded as the coerced internal event and is surfaced as `badMessage` instead.

packages/xrpl/test/connection.test.ts
```diff
@@ -903,6 +903,116 @@ describe('Connection', function () {
...
+              errorCodeOrPayload != null &&
+              typeof errorCodeOrPayload === 'object' &&
+              (errorCodeOrPayload as { type?: unknown }).type === reservedType
+            ) {
+              reject(
+                new XrplError(
+                  'Spoofed server payload was forwarded to error listeners',
+                ),
+              )
+            }
```
cybele-ripple, line 956: Does this need an else statement? If this test fails then we would have to wait for the test to timeout.

packages/xrpl/src/client/connection.ts
```diff
@@ -369,6 +384,52 @@ export class Connection extends EventEmitter {
...
+    // before it is used as an event name. A compile-time `as string` is not
+    // enough: a payload like `{"type":["error"]}` slips past
+    // `RESERVED_INTERNAL_EVENTS.has(...)` (an array never matches a string in
+    // the Set) yet still coerces to the string "error" when passed to `emit`,
+    // re-opening the spoofing hole. Reject any non-string `type` outright.
+    if (typeof data.type !== 'string') {
+      this.emit(
+        'error',
+        'badMessage',
+        `Server message 'type' must be a string; dropping.`,
```
cybele-ripple: Optional change (up to you): maybe include `got: ${typeof data.type}` to the error message to aid in debugging

ckeshava: Done in `bb70c6dde` — the message now reads `Server message 'type' must be a string (got ${typeof data.type}); dropping.`. The raw payload is already forwarded as the 4th `emit` arg, but having the type inline makes the log self-contained.

packages/xrpl/test/connection.test.ts
ckeshava: Good point. The spoof-forwarding regressions that matter already reject immediately — the reserved-event listener for non-`error` types and the payload-object check for `error` — so those fail fast. The gap was an *unexpected* `error` emission shape, which fell through to the 20s `TIMEOUT`. Fixed in `bb70c6dde`: added a `return` after the spoofed-payload reject plus an `else`-style fallback `reject` so any unexpected emission fails immediately. The one case that still relies on the framework timeout is "message silently dropped without emitting `badMessage`" — there's no event to hook there, but Jest reports that as a failure too.

ckeshava: fixed in [bb70c6d](https://github.com/XRPLF/xrpl.js/pull/3378/commits/bb70c6dde7b12bcb212831215da8de448d207a90)

cybele-ripple (approved):

kuan121 (approved):

PR #3429, opened 2026-08-10
fix: use base fee for VaultCreate autofill

### LINKED ISSUE #3427: `autofill` charges the owner reserve for `VaultCreate`, which no longer requires it
### Summary

`calculateFeePerTransactionType` lists `VaultCreate` as an owner-reserve ("special tx cost") type, but rippled removed that fee rule in 3.0.0. On any current network xrpl.js autofills ~200,000 drops where ~12 are required.

### Details

`packages/xrpl/src/sugar/autofill.ts:329`

```ts
const isSpecialTxCost = [
  'AccountDelete',
  'AMMCreate',
  'VaultCreate',
].includes(tx.TransactionType)
```

`VaultCreate.cpp` has no `calculateBaseFee` override in rippled 3.x, the override existed only in 2.6.x and earlier. Without it, `VaultCreate` inherits `Transactor::calculateBaseFee` (`base * (1 + signers)`) and enforces its reserve as an actual reserve via `tecINSUFFICIENT_RESERVE`, not as a fee.

The complete set of transactors calling `calculateOwnerReserveFee` on `develop` is:

- `src/libxrpl/tx/transactors/account/AccountDelete.cpp`
- `src/libxrpl/tx/transactors/dex/AMMCreate.cpp`
### Impact

Mainnet `reserve_inc` is 200,000 drops and `base_fee` is 10, so a `VaultCreate` autofills 200,000 drops instead of ~12, roughly four orders of magnitude.

There's no backstop, because `autofill.ts:385` intentionally bypasses the cap for these types:

```ts
const totalFee = isSpecialTxCost
  ? baseFee
  : BigNumber.min(baseFee, maxFeeDrops)
```

That bypass is correct for `AccountDelete`/`AMMCreate` (clamping below the increment would give `telINSUF_FEE_P`), but it means `maxFeeXRP` cannot catch a type that shouldn't be in the list at all. The overpay is silent regardless of how low the user sets `maxFeeXRP`.

### Suggested fix

Drop `'VaultCreate'` from `isSpecialTxCost`. If 2.6.x support matters, gate on server version rather than leaving it unconditional.

Suede-AI:
Fixes #3427.

## Summary

Remove `VaultCreate` from the transaction types that receive the two-XRP special fee during autofill. `VaultCreate` now uses the network base fee, while `AccountDelete` and `AMMCreate` retain their existing special cost.

## Testing

- `npm run build`
- `npm test -w xrpl -- --runInBand test/client/autofill.test.ts` (22 tests)
- `npm run lint -w xrpl`

The new regression test autofills a valid `VaultCreate` transaction and expects the mocked network fee of 12 drops.

### DIFF
--- a/packages/xrpl/src/sugar/autofill.ts
+++ b/packages/xrpl/src/sugar/autofill.ts
@@ -326,11 +326,9 @@ async function calculateFeePerTransactionType(
   const netFeeDrops = xrpToDrops(netFeeXRP)
   let baseFee = new BigNumber(netFeeDrops)
 
-  const isSpecialTxCost = [
-    'AccountDelete',
-    'AMMCreate',
-    'VaultCreate',
-  ].includes(tx.TransactionType)
+  const isSpecialTxCost = ['AccountDelete', 'AMMCreate'].includes(
+    tx.TransactionType,
+  )
 
   // EscrowFinish Transaction with Fulfillment
   if (tx.TransactionType === 'EscrowFinish' && tx.Fulfillment != null) {
--- a/packages/xrpl/test/client/autofill.test.ts
+++ b/packages/xrpl/test/client/autofill.test.ts
@@ -7,6 +7,8 @@ import {
   Payment,
   Transaction,
   Batch,
+  VaultCreate,
+  VaultWithdrawalPolicy,
   type LoanSet,
 } from '../../src'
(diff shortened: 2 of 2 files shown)


Suede-AI:
Superseded by #3430 from the requested contributor account.

PR #3440, opened 2026-08-13
Expand team required-reviewers into individual member logins for Slack notifications

shichengripple001:
## Summary
- `fetch_reviewers` in the release pipeline now resolves `Team`-type required reviewers on an environment into their individual member logins (via `GET /orgs/{org}/teams/{slug}/members`) instead of only showing the team slug in Slack notifications.
- If member lookup fails (e.g. insufficient token scope), falls back to a clickable team page URL (`https://github.com/orgs/{org}/teams/{slug}`) instead of a bare `@org/slug` reference.

## Test plan
- [x] Ran the Release Pipeline workflow from this branch (with temporary main-only guard relaxed, since reverted) targeting `first-review`/`official-release` environments configured with team reviewers (`ripplex-developer-growth-engineering`, `ripplex-platform`).
- [x] Confirmed `DEV_REVIEWERS`/`SEC_REVIEWERS` output resolved to individual logins when the calling token had org read access.
- [x] Confirmed fallback to team URL when member lookup fails.
- [ ] Note: the default `GITHUB_TOKEN` used by `ask_for_dev_team_review`/`ask_for_sec_team_review` does not have `read:org` scope, so live runs on `main` will currently fall back to the team-URL path rather than expanding to member logins, unless a PAT/fine-grained token with org read access is wired in separately.


.github/workflows/release.yml
```diff
@@ -423,29 +423,60 @@ jobs:
...
+                User:*)
+                  expanded+=("${entry#User:}")
+                  ;;
+                Team:*)
+                  local slug="${entry#Team:}"
+                  logins="$(curl -sSf \
+                    -H "Authorization: Bearer $GH_TOKEN" \
+                    -H "Accept: application/vnd.github+json" \
+                    "https://api.github.com/orgs/$org/teams/$slug/members" \
+                    | jq -r '.[].login' 2>/dev/null || true)"
```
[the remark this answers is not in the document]
shichengripple001: pushed a fix for the pagination issue in 72188b0f: the team-members lookup now uses `gh api --paginate` instead of a single unpaginated `curl` call, so teams larger than one page no longer silently lose members.

On the `GITHUB_TOKEN` org-read permission gap: that's a known, pre-existing limitation, already called out as an open item in the PR description's test plan. The default `GITHUB_TOKEN` can't be granted org-level `Members: read` via the workflow's `permissions:` block — that requires a PAT or fine-grained token with explicit org member-read access, wired in as a repo secret by someone with admin on the org/repo. Tracking that as a follow-up rather than blocking this PR, since the fallback (team page URL) degrades gracefully when that token isn't available.

.github/workflows/release.yml
```diff
@@ -423,29 +423,58 @@ jobs:
```
kuan121, line 423: Where is the team specified?

kuan121 (approved):

.github/workflows/release.yml
shichengripple001: Team is created under xrplf org
https://github.com/orgs/ripple/teams/developer-growth-engineering

kuan121: Where do we specify that the reviewer team should be developer-growth-engineering? Some env variables?

kuan121: Also, which team do we use for DGE?
https://github.com/orgs/XRPLF/teams/ripplex-developer-growth-engineering or https://github.com/orgs/ripple/teams/developer-growth-engineering

pdp2121 (approved):

.github/workflows/release.yml
shichengripple001: we use https://github.com/orgs/XRPLF/teams/ripplex-developer-growth-engineering for repos under xrplf. so for this pipeline it is https://github.com/orgs/XRPLF/teams/ripplex-developer-growth-engineering

Issue #3286, opened 2026-04-17
entropy.slice(0, 16) silently truncates user-provided entropy in generateSeed


ckeshava:
## Description

If user-provided entropy is longer than 16 bytes, it is silently truncated. Only `>= 16` is checked. The original audit claimed that providing 32 bytes gives a false sense of additional security.

## Location

`packages/ripple-keypairs/src/index.ts`, line 39

Source: https://github.com/XRPLF/xrpl.js/blob/5cf0921b0fa4637f324bcf5b0788e2ef59a83c8a/packages/ripple-keypairs/src/index.ts#L39

## Original Severity

Low — Potentially misleading API regarding additional entropy.

## Adversarial Review Outcome: DISPROVED (confidence 90%)

**XRPL seeds are exactly 16 bytes by protocol specification.** The `generateSeed` function slices to 16 bytes because that's the correct seed length. The `>= 16` check ensures sufficient entropy is provided. Providing 32 bytes doesn't add security because the seed format is fixed at 128 bits. The truncation is correct behavior mandated by the protocol. A warning could be nice but the code is correct.

## Scope Note

No code change recommended. A documentation clarification (or an optional warning on oversized entropy) could be considered separately, but the runtime behavior is correct per the XRPL spec.

## Reference

Produced during adversarial review of `SECURITY_AUDIT_REPORT.md` on 2026-04-15.

PR #3469, opened 2026-09-11
fix: throw on invalid entropy in Wallet.fromEntropy and generateSeed

pdp2121:

- `Wallet.fromEntropy` now validates entropy instead of coercing it
- `generateSeed` (ripple-keypairs) now requires exactly 16 bytes instead
  of truncating anything longer

### Context of Change

**Problem**
- `fromEntropy` passed input through `Uint8Array.from()`, which accepts any iterable. Strings were coerced via `Number()` — letters became `NaN`, stored as `0`
- `Wallet.fromEntropy('abcdefghijklmnop')` returned a real, spendable wallet derived from 16 zero bytes, with no error
- `generateSeed` accepted any entropy `>= 16` bytes and sliced to the first 16, so inputs sharing a prefix derived the same wallet (#3286)

**Changes**
- Reject input that is not a `Uint8Array` or an array of byte values
- Require exactly 16 bytes; no truncation, no zero-padding
- Reject sparse arrays — `Array.prototype.every` skips holes, so `new Array(16)` and `Array(16).map(...)` previously passed validation and read back as zero bytes
- Validate the materialized copy rather than re-reading the caller's object, so the validated bytes and the derived bytes cannot differ
- Length error tells callers to hash input down to 16 bytes rather than truncate, so meaningful bytes aren't silently discarded

### Type of Change

- [ ] Documentation Updates
- [ ] Release

- [x] Yes

## Test Plan

Regression tests cover strings, hex strings, sparse arrays, wrong lengths, non-byte values, and that no malformed input derives the known zero-entropy wallet.


packages/ripple-keypairs/src/index.ts
```diff
@@ -36,13 +57,19 @@ function generateSeed(
     !options.algorithm || VALID_ALGORITHMS.includes(options.algorithm),
     `Unsupported algorithm: ${options.algorithm}. Use one of: ${VALID_ALGORITHMS.join(', ')}`,
   )
+  // Entropy must be refused, not resized. Truncating over-length entropy
+  // silently discards the caller's extra bytes, and accepting a non-byte-array
+  // lets a coerced value (a string, say) through as well-formed zero bytes.
+  assert.ok(
+    options.entropy == null || isUint8Array(options.entropy),
```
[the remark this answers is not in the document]
pdp2121: Addressed in [c5f6bc7](https://github.com/XRPLF/xrpl.js/pull/3469/commits/c5f6bc7d86cfa51c6125a1a7d4934d1c022f063b)

packages/xrpl/src/Wallet/utils.ts
```diff
@@ -137,3 +138,107 @@ export function computeSignature(
...
+      `entropy must be a Uint8Array or an array of byte values, received ${typeof entropy}. ` +
+        `If you have a hex string, convert it to bytes first (for example with hexToBytes) ` +
+        `rather than passing the string directly.`,
+    )
+  }
+
+  // Materialize first, then validate what was materialized. The array iterator
+  // yields `undefined` for holes rather than skipping them, so a sparse array
+  // becomes an array of `undefined` that the byte check below rejects.
+  const values = Array.from(entropy)
```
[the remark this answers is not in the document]
pdp2121: Addressed in [c5f6bc7](https://github.com/XRPLF/xrpl.js/pull/3469/commits/c5f6bc7d86cfa51c6125a1a7d4934d1c022f063b)

packages/xrpl/src/Wallet/utils.ts
```diff
@@ -137,3 +138,107 @@ export function computeSignature(
...
+/**
+ * Converts caller-supplied entropy into exactly ENTROPY_LENGTH_BYTES bytes,
+ * rejecting anything that is not already a sequence of bytes.
+ *
+ * The conversion happens up front, and validation runs on the converted copy.
+ * Validating the caller's object and then re-reading it to build the result
+ * lets the two reads disagree, which is how the original bug worked: a bare
+ * `Uint8Array.from()` accepts any iterable, so a string was coerced with
+ * `Number()`, every letter became `NaN`, and it stored as 0 — producing a
+ * spendable wallet from zero bytes with no error, because the bytes reaching
```
kuan121: I don't think we need to mention the original bug.

pdp2121: Removed in [803afa7](https://github.com/XRPLF/xrpl.js/pull/3469/commits/803afa7425fa9e144a1f95833c3bda7643f83f6a)

kuan121 (approved):

packages/xrpl/test/wallet/index.test.ts
```diff
@@ -475,6 +476,157 @@ describe('Wallet', function () {
...
+    it('throws when entropy is short but length-padded', function () {
+      const padded = [1, 2, 3]
+      padded.length = 16
+      assert.throws(() => Wallet.fromEntropy(padded), ValidationError)
+    })
+
+    // Malformed input must be rejected outright, never quietly turned into
+    // all-zero bytes.
+    it('never derives an all-zero wallet from malformed input', function () {
+      const zeroAddress = Wallet.fromEntropy(new Uint8Array(16)).classicAddress
```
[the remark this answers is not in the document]
pdp2121: Fixed in [9aff515](https://github.com/XRPLF/xrpl.js/pull/3469/commits/9aff51564bf7354f9e9102b3e377c1e3425ffb4e)

packages/ripple-keypairs/src/index.ts
```diff
@@ -20,6 +20,27 @@ import {
...
+    value instanceof Uint8Array ||
+    // The tag check alone would also admit other single-byte views, so
+    // require the element size too.
+    (ArrayBuffer.isView(value) &&
+      'BYTES_PER_ELEMENT' in value &&
+      value.BYTES_PER_ELEMENT === 1 &&
+      Object.prototype.toString.call(value) === '[object Uint8Array]')
+  )
+}
+
```
cybele-ripple, line 43: Can this function be shared since it is the exact same as [that in xrpl/src/Wallet?](https://github.com/XRPLF/xrpl.js/pull/3469/changes#diff-7211f097a1f828af1cc5ee4a8613ea28e2f5fa167e66bd1a32743d246f08b2d6R153-R163)

pdp2121: This function is cross-package and not currently export. The natural home for this would be isomorphic but that means we will need another package release. I don't think it's worth the effort

cybele-ripple (approved):

